Issue #64, opened 2020-09-02
Add support for adding multiple files on asset


santteegt:
It might be useful for some use cases to allow users to add more than one file when publishing an asset

mihaisc:
this is part of the new form https://github.com/oceanprotocol/market/issues/862

PR #65, opened 2020-09-02
Add support for adding multiple files on asset

santteegt:
Adds support for adding multiple files when publishing an asset

Solves #64


kremalicious:
Thanks for looking into this! But we're closing this for now since adding multiple files during publish has way more implications which would need to be tackled before this could be rolled out. Both Consume & Compute components would need to be refactored to display those multiple files. Ocean.js and the react hooks would need to be updated to allow a "Download All Files" action. The Compute component would need a way to select individual files to run jobs on.

With all this added complexity it seems a more feasible workaround for now is to tell publishers to use an archive of multiple files, like a zip if they want to provide multiple files.

PR #141, opened 2020-10-27
ga tracking

mihaisc:
- [ ] add valid ga id
- [ ] change gdpr message in gdpr.json

Note: check `do not track`  browser setting, if you have it on the banner will not show, it will reject cookies by default


alexcos20:
remove ga for now

PR #164, opened 2020-10-29
refresh price and pool info

### LINKED ISSUE #155: price & shares do not update after user interaction without browser refresh
the pool still thinks I have the same pool share even after withdrawing.
(from discord user)

mihaisc:
closes #155

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "@loadable/component": "5.13.1",
     "@oceanprotocol/art": "^3.0.0",
     "@oceanprotocol/lib": "^0.9.6",
-    "@oceanprotocol/react": "^0.3.13",
+    "@oceanprotocol/react": "^0.3.14",
     "@oceanprotocol/typographies": "^0.1.0",
     "@sindresorhus/slugify": "^1.0.0",
     "@tippyjs/react": "^4.2.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -75,7 +75,9 @@
     "remove-markdown": "^0.3.0",
     "shortid": "^2.2.16",
     "slugify": "^1.4.5",
+    "styled-components": "^5.2.0",
```
kremalicious, line 78: do we need that? That's a big dependency

mihaisc: noo, forgot it

mihaisc: it's one of the weird dependencies required when i use react hooks locally

kremalicious (approved):
testing on preview with adding liquidity, works perfectly

src/components/organisms/AssetActions/Pool/index.tsx
```diff
@@ -110,18 +112,31 @@ export default function Pool({ ddo }: { ddo: DDO }): ReactElement {
         // swapFee is tricky: to get 0.1% you need to convert from 0.001
         const swapFee = await ocean.pool.getSwapFee(price.address)
         setSwapFee(`${Number(swapFee) * 100}`)
+
+        console.log(
```
kremalicious, line 116: maybe put them behind `Logger` or remove

PR #169, opened 2020-10-30
change weights

### LINKED ISSUE #168: change the % weight of Ocean that initial publishers must put in, from {90% DT, 10% OCEAN} to {50% DT, 50% OCEAN}
1. change the % weight of Ocean that initial publishers must put in, from {90% DT, 10% OCEAN} to {50% DT, 50% OCEAN} or even {10% DT, 90% OCEAN}. 

This better aligns the incentives of the publisher with the rest of the community (that owns OCEAN). Where to change: just Market, not libraries or contracts. Side benefit: keeps prices sane.

mihaisc:
change price weights of datatoken and ocean  to 5-5 
closes #168


kremalicious:
think this also implies we need to change the minimum amount checks in the form validations

kremalicious (approved):

PR #186, opened 2020-10-31
lower min amount for adding token

kremalicious:
Some assets have rather expensive datatoken, making fractions worth a lot. So lower the live-validation amount which can be added back to pool.

Still needs to be verified:
- low OCEAN or DT amounts do not pose a problem along the stack, so this change does not result in failing transactions
- we might need to decide and implement different min amounts for OCEAN and DT


alexcos20:
as for balancer & ocean.js, the only limits are the upper ones, so there is no issue

PR #191, opened 2020-11-02
re-fetch pool numbers every 10 sec.

kremalicious:
Re-fetch price and recalculate all numbers every 10 sec. so users do not have to reload to get new values. Seems ok performance-wise when doing this only on the asset details page within the pool widget.

Screen Shot 2020-11-02 at 02 05 44

### DIFF
--- a/src/components/organisms/AssetActions/Pool/index.tsx
+++ b/src/components/organisms/AssetActions/Pool/index.tsx
@@ -170,11 +170,17 @@ export default function Pool({ ddo }: { ddo: DDO }): ReactElement {
       }
     }
     init()
+
+    // Re-fetch price every 10 sec., triggering re-calculation of everything
+    const interval = setInterval(refreshPrice, 10000)
+    return () => {
+      clearInterval(interval)
+    }
   }, [ocean, accountId, price, ddo.dataToken, refreshPool, ddo.publicKey])
 
   const refreshInfo = async () => {
     setRefreshPool(!refreshPool)
-    await refreshPrice()
+    refreshPrice()
   }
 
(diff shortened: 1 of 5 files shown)


src/components/organisms/AssetActions/Pool/index.tsx
```diff
@@ -170,11 +172,15 @@ export default function Pool({ ddo }: { ddo: DDO }): ReactElement {
...
-  }, [ocean, accountId, price, ddo.dataToken, refreshPool, ddo.publicKey])
+
+    // Re-fetch price periodically, triggering re-calculation of everything
+    const interval = setInterval(() => refreshPrice(), refreshInterval)
+    return () => clearInterval(interval)
+  }, [ocean, accountId, price, ddo, refreshPool])
 
   const refreshInfo = async () => {
     setRefreshPool(!refreshPool)
-    await refreshPrice()
```
mihaisc, line 177: why remove `await`?

kremalicious: editor told me it's useless cause `refreshPrice` is no async method. But could add it back if there was a reason for it?

mihaisc: this is interesting ` async function refreshPrice(): Promise<void>` , could it be because it returns `Promise<void>`, but i use `async` functions in it so it needs to be awaited

kremalicious: not sure, I put it back for now. editor shows me this:
Screen Shot 2020-11-02 at 11 48 30

PR #193, opened 2020-11-02
User pool transactions on asset details

### LINKED ISSUE #185: show user's pool transactions for a specific dataset
The pool transactions list on history page loads painfully slow, and it would be useful to show pool transactions in place on a specific asset details page. Cause when deciding on interacting with a pool it's crucial to know what was put in/removed before.

Could be just simple table below the asset actions widget. And if we can have a method with a reduced query only for specific pool we might get a faster result.

kremalicious:
closes #185

With toggle pattern for less busy interface, and more performant initial pool widget loading.

Default:
Screen Shot 2020-11-02 at 16 20 51

Open:
Screen Shot 2020-11-02 at 16 21 02

### DIFF
--- a/src/components/atoms/SuccessConfetti.tsx
+++ b/src/components/atoms/SuccessConfetti.tsx
@@ -25,10 +25,12 @@ const confettiConfig = {
 
 export default function SuccessConfetti({
   success,
-  action
+  action,
+  className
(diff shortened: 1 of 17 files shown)


mihaisc (approved):
looks good

PR #203, opened 2020-11-03
first aggregated stats

kremalicious:
- output some basic stats in footer
- fetched on start of app, and refetched every 10 sec. when in viewport 

Screen Shot 2020-11-04 at 10 39 51

numbers fetched from new endpoint 
https://github.com/oceanprotocol/market-stats
https://market-stats.oceanprotocol.com

### DIFF
--- a/src/components/molecules/MarketStats.tsx
+++ b/src/components/molecules/MarketStats.tsx
@@ -1,0 +1,63 @@
+import { Logger } from '@oceanprotocol/lib'
+import React, { ReactElement, useEffect, useState } from 'react'
+import PriceUnit from '../atoms/Price/PriceUnit'
+import axios from 'axios'
+import styles from './MarketStats.module.css'
+import { useInView } from 'react-intersection-observer'
+
+interface MarketStatsResponse {
+  datasets: {
+    pools: number
+    exchanges: number
+    none: number
+    total: number
+  }
+  owners: number
+  ocean: number
+  datatoken: number
+}
+
+export default function MarketStats(): ReactElement {
+  const [ref, inView] = useInView()
+  const [stats, setStats] = useState<MarketStatsResponse>()
+
+  useEffect(() => {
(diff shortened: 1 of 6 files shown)


src/components/molecules/MarketStats.tsx
```diff
@@ -0,0 +1,63 @@
...
+  owners: number
+  ocean: number
+  datatoken: number
+}
+
+export default function MarketStats(): ReactElement {
+  const [ref, inView] = useInView()
+  const [stats, setStats] = useState<MarketStatsResponse>()
+
+  useEffect(() => {
```
mihaisc, line 24: we should use a canceling token for axios.
```
useEffect(() => {
    async function getStats() {
      try {
        const response = await axios('https://market-stats.oceanprotocol.com',  { cancelToken: source.token})
        if (!response || response.status !== 200) return
        setStats(response.data)
      } catch (error) {
           if (axios.isCancel(error)) {
              //  cancelled
            } else {
               Logger.error(error.message)
            }
      }
    }

    // Update every 10 sec. when in viewport
    const interval = setInterval(getStats, 10000)

    if (!inView) {
      clearInterval(interval)
    }

    getStats()

    return () => {
          clearInterval(interval)
           source.cancel()
     }
  }, [inView])
```

mihaisc (approved):

mihaisc (commented):
you forgot the error check, without it it will log an error when you cancel the call

kremalicious:
so does it make sense just logging the cancel?

```js
if (axios.isCancel(error)) {
  Logger.log(error.message)
 } else {
  Logger.error(error.message)
}
```

mihaisc:
You can leave it like that, but it doesn't provide much information

kremalicious:
meh, just pushed the change I already committed as that cancel token logging doesn't seem to matter much

PR #204, opened 2020-11-04
Swap tokens

### LINKED ISSUE #110: new asset action: Swap
Add a new tab in the asset actions widget allowing users to swap OCEAN ←→ Datatoken.

This is so we do not end up handling this on the _Use_ screen, which should stay focused on the usage of dataset, so only 1 Datatoken can be bought from there.

In there, handle those rules:

```
2. Swap:
        a) buyDT
                       i) Make sure that DT_AMOUNT_REQIRED < getMaxBuyQuantity(pool, dtaddress)
                      ii) Calculate how many ocean you need:  getOceanNeeded(pool, DT_AMOUNT_REQUIRED)
        b) sellDT
                      i) Calculate how many ocean you will receive:  getOceanReceived(pool,DT_AMOUNT)
                     ii) Make sure that ocean_amount < getMaxBuyQuantity(pool, pool.oceanAddress)
```

Motivation: people can't sell datatokens. Yet Ocean Market is meant to be an *exchange*; so selling is a critical part; it's also critical to truly establish datatokens as assets. There's also a sub-problem: people who buy to speculate may end up inadvertently consuming, and losing. 

It's for these reasons that this issue is a bug (vs enhancement), and high priority.

----

Defer until later:
- Slow slippage. [Github issue](https://app.zenhub.com/workspaces/ocean-main-5e5e5a97e91ce03f380fb9b7/issues/oceanprotocol/market/197)
- Show "max/min expected amount received" (like uniswap or Balancer do)

### LINKED ISSUE #197: In swap, show slippage
It's a safety feature in Uniswap and Balancer.

### LINKED ISSUE #198: In swap page, show "max/min expected amount received" (like Uniswap & Balancer do)
This will help set user expectations.

mihaisc:
Adds a new asset action, Trade, where users can swap between OCEAN and the respective datatoken.

Screen Shot 2020-11-16 at 15 44 09

- closes #110
- closes #197 
- closes #198 

Further changes:

New message under Buy button for `Use`:

Screen Shot 2020-11-09 at 21 15 14


kremalicious:
Pretty cool so far, I'll do some light styling updates and probably unify that `userLiqudity` styling thing.

As for data flow, would be cool if you could look into making most of UI show up by default. I found following this pattern in every component works well and has the benefit that most UI is present for non-wallet users:

1. get as much info needed for UI directly from the `ddo`, so for the swap thing we should be able to show pretty much everything based on that, like both symbols
2. after that data is there, enhance it and add more data like we do now

kremalicious:
also running into that 0 price for whatever reason on re-fetch even with current interval. That problem does not show on the pool tab I think

content/price.json
```diff
@@ -47,5 +47,10 @@
       },
       "action": "Approve & Remove"
     }
+  },
+
+  "trade": {
```
kremalicious, line 52: and I forgot the reasoning again behind using `Trade`, seems if we title it with `Swap tokens` we should also call the whole thing `Swap` down to the component names

kremalicious:
so far, so good

Screen Shot 2020-11-05 at 19 55 49

content/price.json
mihaisc: Ah, yes, I changed the names a few times. I took the trade from uniswap, the button that takes you to the swap interface is called "Trade"

mihaisc:
I'll test some more the balance stuff, something is not right

kremalicious:
and some quick thoughts on interaction design after playing around with what we wanted to build. While this UI is nicely flexible, where users can type in both fields, this also makes it harder to grasp in the beginning. But I feel we actually do not need that full flexibility in this context here, since we only deal with 2 tokens.

So we could switch to only showing one input, where the prefix is a dropdown for both tokens. Depending on what users select, the bottom part shows a more "You will receive' style output instead of another input field.

mihaisc:
Well I copied the UX from uniswap. Your suggestion has one disadvantage ( at least from my point of view) , when i want to input how much i want to buy : for example i want to buy 3 dt, it's easier to input the value in the second field, or I want to sell dt and get 500 ocean (because that how much i need, and i don't want to sell more dt)  
image

kremalicious:
ok, got it, the fallback view in non-web3 browser shows pretty much everything now:

Screen Shot 2020-11-06 at 12 07 42

kremalicious:
after some testing, we have to solve #197 and #198 as part of the first rollout. The differences in what has been received are just too big and we can expect users to be highly confused. So a new output section will be added in this PR already:

Screen Shot 2020-11-06 at 14 53 44

kremalicious:
What exactly is the pool balance at the bottom now?

Pool statistics:
Screen Shot 2020-11-09 at 21 17 54

Pool balance display:
Screen Shot 2020-11-09 at 21 18 06

Should pool balance not be the same to what's in pool statistics for the datatoken?

And this is what shows up when switching to OCEAN at bottom:
Screen Shot 2020-11-09 at 21 20 48

If this pool balance at bottom is now some value based on some pool maximum then we should change the label to be more precise

mihaisc:
No, pool balance is the maximum number of tokens that can be bought (confirmed with Alex).  And when you think about it, `Pool balance` is confusing

kremalicious:
changed labels and made the minimum received amount work. I think. Not sure it is actually the minimum account.

Screen Shot 2020-11-09 at 23 12 20

Now, how do we go about calculating the slippage? Is this even feasible in current architecture state?

src/components/organisms/AssetActions/Trade/Output.tsx
```diff
@@ -0,0 +1,70 @@
...
+  // Get output values
+  useEffect(() => {
+    if (!ocean || !poolAddress) return
+
+    async function getOutput() {
+      // Minimum received
+      // TODO: check ity is actually the minimum received after transaction.
+      const minOutput =
+        values.type === 'buy'
+          ? await ocean.pool.getDTReceived(poolAddress, `${values.ocean}`)
```
kremalicious, line 43: might be redundant fetching this here since the output from this seems to be always the same as whatever is put into the bottom input field.

mihaisc: Yes, we should put something like 'Maximum price'  since both buy and sell have a fixed amount and a variable price. The slippage is hardcoded to 10% for now, we can't actually calculate it

kremalicious:
add:
- slippage tolerance setting UI, with default set to ...%
- price impact display based on spot price

src/components/organisms/AssetActions/Trade/Output.tsx
```diff
@@ -0,0 +1,67 @@
...
+  useEffect(() => {
+    if (!ocean || !poolAddress) return
+
+    async function getOutput() {
+      // Minimum received
+      // TODO: check ity is actually the minimum received after transaction.
+      const maxPrice =
+        values.type === 'buy'
+          ? await ocean.pool.getOceanNeeded(poolAddress, `${values.ocean}`)
+          : await ocean.pool.getDTNeeded(poolAddress, `${values.datatoken}`)
```
kremalicious, line 44: the output values for this are kinda weird, had to remove that `1` param in `ocean.pool.getDTNeeded` cause the typings want something else.

With this change I'm getting this output, which looks wrong:

Screen Shot 2020-11-10 at 16 03 47

Screen Shot 2020-11-10 at 16 00 57

mihaisc: Oh, sorry I totally ignored those values because we said that we remove that section. Are getting it back? It was unclear to me

kremalicious:
Quick slippage tolerance UI, collects everything in form data:

Screen Shot 2020-11-10 at 23 02 52

kremalicious:
modified slippage UI, opting for native `select` element instead of some custom component

Screen Shot 2020-11-11 at 14 20 07

src/components/organisms/AssetActions/Trade/Output.tsx
```diff
@@ -0,0 +1,66 @@
...
+      <div>
+        <p>Maxiumum Paid</p>
+        <Token
+          symbol={values.type === 'buy' ? dtSymbol : 'OCEAN'}
+          balance={maxOutput}
+        />
+      </div>
+      <div>
+        <p>&nbsp;</p>
+        <Token symbol="% swap fee" balance={swapFee} />
```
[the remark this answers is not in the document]
mihaisc: I have no idea how to calculate that, as far as I know the fee is a % of datatokens. But is it the same when you buy and when you sell? The fee is calculated from what amount?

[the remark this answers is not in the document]
kremalicious: The swap fee is taken from input value, so if I spend 100 OCEAN and the swap fee is 1%, then we can display 1 OCEAN as swap fee if I get this right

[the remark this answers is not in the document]
mihaisc: But we've been only guessing until now, hence so many refactors.

mihaisc:
Refactored price impact, left only %.
`Minimum Received` is calculated properly

image

kremalicious (approved):

PR #214, opened 2020-11-05
Data Partners

kremalicious:
Many UI tweaks across the board favoring our data partners.

New front-page with data partner assets in prominent carousel:

Screen Shot 2020-11-05 at 12 40 20

New badges for those partner assets across all asset lists:

Screen Shot 2020-11-05 at 12 40 54

New enhanced account display for partners:

Screen Shot 2020-11-05 at 12 41 35

Also output in the list of full metadata:

Screen Shot 2020-11-05 at 12 42 11

### Mechanism

This is all based on the moderated list https://github.com/oceanprotocol/list-datapartners, used as npm module. There is a `useDataPartner(accountId)` hook which gives back the respective `partner` data when found.


package.json
```diff
@@ -62,6 +63,7 @@
     "lodash.omit": "^4.5.0",
     "query-string": "^6.13.6",
     "react": "^16.14.0",
+    "react-alice-carousel": "^2.0.2",
```
kremalicious, line 66: not fully happy about this but damn, carousel components out there are bad and this had the least amount of issues. What I actually wanted is on large screens a 4x2 grid on each carousel pane, but 2x1 on small screens, and more in-between. This was not easily possible with this library either, so until we have 100 data partner assets it should suffice.

trentmc (approved):

mihaisc (approved):

PR #222, opened 2020-11-06
Remove slider quick wins

### LINKED ISSUE #196: Remove slider UX is terrible
Native input element so while accessible, how we connect it results in a sub-par experience. While the constant conversion jumps are not nice they are kinda expected right now, but there is too much going on when putting the slider to 100%.

Investigate and decide if native `input type="range"` should be replaced by a component with more control.

kremalicious:
Without switching to new component some quick wins addressing #196 to make the slider a bit more bearable:

- Address main pain point: turn the maximum % display into a max button:

  Screen Shot 2020-11-06 at 23 24 14

- Make custom slider styles appear in Chrome too
- Majority of weirdness on simple screen comes from the multiple calls happening when slider is moved, and then when multiple values are updated in component, effects fire too often while the calls are still running, all leading to this number sluggishness when moving the slider. 
- Addressed this a bit by splitting up and reducing the effect usage and doing more directly in the `onChange` handlers. 
- And finally split up that method with the 4 pool calls, this was blocking getting `amountOcean`, the displayed value in UI. All not perfect but a bit better before we switch to some other component.

closes #196

### DIFF
--- a/src/components/atoms/Input/InputElement.module.css
+++ b/src/components/atoms/Input/InputElement.module.css
@@ -211,16 +211,28 @@
 }
 
 input[type='range'] {
-  background: none;
+  background: transparent;
+  appearance: none;
 }
 
 input[type='range']:focus {
   outline: none;
 }
 
-input[type='range']::-webkit-slider-thumb,
+/* Selectors need to be split up to work in both engines */
+input[type='range']::-webkit-slider-thumb {
+  -webkit-appearance: none;
+  margin-top: -0.5rem;
+  background: var(--brand-gradient);
+  border: 2px solid var(--border-color);
+  width: var(--font-size-h4);
+  height: var(--font-size-h4);
+  border-radius: 50%;
+  cursor: pointer;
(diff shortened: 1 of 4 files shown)


kremalicious:
alright, finished like 20 tests with this in different `amountMaxPercent` combinations and I'm mostly sure everything works. All a bit more snappy and the max button alone feels like a huge improvement for that user flow, so we can get this out

mihaisc:
Something is wrong with the pool, same asset on this branch and on main
image
image

mihaisc:
Debounce was not actually working, it worked only for the first call. Did multiple tests with useCallback, throttle and debounce and finally I found [this](https://stackoverflow.com/questions/54666401/how-to-use-throttle-or-debounce-with-react-hook/54666498). He is actually right, I did the tests.
A downside is that you have to pass as params the values that change, I added only `amountPoolShares` and `isAdvanced` since the rest don't change

mihaisc (approved):

kremalicious:
wow, what a weird trick, good find!

PR #223, opened 2020-11-07
package updates

kremalicious:
Bump many packages, and upgrade to [React 17](https://reactjs.org/blog/2020/10/20/react-v17.html)


mihaisc (approved):
Looking at the blog there shouldn't be any problems. Didn't do any extensive testing but browsing through all the pages all seems to be ok

Issue #174, opened 2020-10-30
Tweet published dataset


alexcos20:
from within Ocean Market, make it easy for people to tweet about the dataset that they just published. 

Have a pre-defined tweet including #oceanmarketdata and $DT-SYMBOL.

Helps rug pulls, because Twitter is a proxy for identity / reputation.

Bonus: helps virality further.

PR #230, opened 2020-11-08
share feature

mihaisc:
A quick look at a share feature.

- [ ] Proper styling - use our own icons
- [ ] What should be the text,tags?
- [ ] Who can share? If all, will there be a difference in the text of the publisher and the rest of the users?
- [ ] Do we care about other platforms except twitter? Telegram looks interesting
- [ ] Use web share api when possible and have a fallback

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23595,6 +23595,14 @@
       "resolved": "https://registry.npmjs.org/jsonify/-/jsonify-0.0.0.tgz",
       "integrity": "sha1-LHS27kHZPKUbe1qu6PUDYx0lKnM="
     },
+    "jsonp": {
+      "version": "0.2.1",
+      "resolved": "https://registry.npmjs.org/jsonp/-/jsonp-0.2.1.tgz",
+      "integrity": "sha1-pltPoPEL2nGaBUQep7lMVfPhW64=",
+      "requires": {
+        "debug": "^2.1.3"
+      }
+    },
     "jsprim": {
       "version": "1.4.1",
       "resolved": "https://registry.npmjs.org/jsprim/-/jsprim-1.4.1.tgz",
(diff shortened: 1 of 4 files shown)


kremalicious:
to me, https://github.com/oceanprotocol/market/issues/174 implies that there should be a prominent CTA for publisher to tweet the dataset, not just a share button at the bottom. In general users know how to share a URL to their favorite social media site, even more so on mobile. Based on the issue description, the main thing we want is to control the text when they do so, also as a workaround for missing titles because of missing SSR for the asset pages. 

As for the share buttons itself, we have our own icons so they should be used https://www.figma.com/file/UXKjbq8bNSGWHkHh2UPIsM/Assets?node-id=0%3A1178 Design-wise it could make sense only using the glyphs and not the box depending how prominent we want them to be, and where we place them. 

So might be ok with just manually constructing the share links without further dependencies, like, for twitter `https://twitter.com/intent/tweet?text=Hello%20world`: https://developer.twitter.com/en/docs/twitter-for-websites/tweet-button/guides/web-intent

Also worth noting this is all mostly for desktop, enforcing people to use Twitter website. Half of our users are on mobile so far, and majority of mobile users use their OS' native share sheet using their native Twitter app, whatever it might be. If we want to go full circle and also control the text in there we have to work with the Web Share API https://css-tricks.com/how-to-use-the-web-share-api/

mihaisc:
This was a 10 min pr, because I stumbled the web-intent feature for twitter, and then I found this repo and because it has a lot of apps integrated so easy and I was curious if some would be usefull.  
Regarding the web share api, it's kinda useless on desktop so we need a fallback (no necessarily this, if only twitter we can use web intent)
image

And what do you mean we don't control the text? You can set the text, hashtags, url.

mihaisc:
This is no longer of interest

PR #238, opened 2020-11-10
move analytics id to environment variable

kremalicious:
Forked locally running visits started to show up in analytics, so move the id to environment variables, setup on Vercel


mihaisc (approved):
I guess it's ok, can't really test it

PR #248, opened 2020-11-13
Liquidity & price history graph

### LINKED ISSUE #247: minimum viable pool price graph
Make use of Aquarius' shiny new data endpoints to ship the most minimal but useful implementation of some price graph in the pool widget. Only with basic interaction for showing historic pool price values on hover

kremalicious:
Start with an interactive OCEAN liquidity & price history graph for each data set. Makes use of Aquarius' new `/pools/history/` endpoint which live-fetches the data. 

Because of that, the more transactions were happening on a data set, the longer the graph loading times. So new pools load the data in under 1s, while older pools (like, ancient 3 weeks) with many transactions can take some time. The Swash pool was measured at 16s...

Screen Shot 2020-11-16 at 14 31 57

closes #247


mariozidov:
If you open multiple tabs (Chrome) and connect wallet (Metamask), values on chart starts don't give correct values anymore.
image - example where it is ok.
image  - same pool with connected wallet (on multiple tabs)
I'm not a DEV so, sorry if this is not the place for reviews.

mihaisc:
This happens to me too, will try to investigate more.

kremalicious:
@mariozidov this is exactly the right place and time for a review. Thanks, good find!

kremalicious:
So we cannot fix this from front-end with a live-updating graph. In Aquarius, @ssallam seems to calculate the timestamps between the first and last block. This means on every fetch call the timestamps are different. Only way to solve this on front-end is to disable the auto-update of graph, which breaks the mental model we tell users in the pool widget, that all in there refreshes every 10 sec.

kremalicious:
Double values solved with updated Aquarius endpoint delivering already cumulative sum calculated values.

Next up: plot price

mihaisc (approved):

PR #250, opened 2020-11-14
Put a "ratings" tab onto the datasets

SarahKay99:
Included the library react-rating-stars-component

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3570,6 +3570,9 @@
       "version": "1.0.2",
       "resolved": "https://registry.npmjs.org/@oceanprotocol/list-datapartners/-/list-datapartners-1.0.2.tgz",
       "integrity": "sha512-7WCQyiaNuUMhXh0x3cS/l4Fv34mjZTrXueOzVTfirE+2DD/8Y/BrtBbFV6tYZ0ALjUEulExEAW6ELqn4qSDQHg=="
+      "version": "1.0.3",
+      "resolved": "https://registry.npmjs.org/@oceanprotocol/list-datapartners/-/list-datapartners-1.0.3.tgz",
+      "integrity": "sha512-MMyy81FvnRGwl2cQ4+cucq/YWjUTGzStHyAUVM6P2pFA8zMc3jouuWN2WSAjmvhxeKZU7jvJRwZCoi+miEYKjw=="
     },
     "@oceanprotocol/react": {
       "version": "0.3.16",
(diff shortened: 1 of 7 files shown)


kremalicious:
Hey there, thanks for that. Unfortunately we do not have any system in place to actually do the ratings. As described in https://github.com/oceanprotocol/multi-repo-issue/issues/32, doing the UI is the easy part but how the ratings actually work in a decentralized way is still part of research.

SarahKay99:
I've got nothing in mind yet, though I'll play around and try incorporate the ratings system with the chain. 

To rate in a decentralised manner, I imagine we would only let people with datatokens rate, and we would weight their rating based on how many datatokens they have.

We would need a mechanism to prevent people from: 

- trading tokens in/out/in/out so they can keep rating. 

- quickly buying a lot of tokens so they can rate and then instantly selling them back. 

[please suggest any other problems you foresee]

I suggest a few proposals: 

- Once an address rates, they can only rate again if they buy more datatokens than what they initially rated with. If they sell datatokens they cannot rate the dataset. 

- Addresses cannot rate until they've held datatokens for a certain period of time. This prevents people buying datatokens with the sole purpose of manipulating ratings. 

- Addresses cannot rate until they have consumed the data to satisfactory extent. It would be easier to tell the lengths to which they've consumed data with a C2D environment. Personally I don't know how we'd implement this but it might be possible. 

As I understand, you don't want a comments / reviews system?

SarahKay99:
Perhaps there can be two rating systems: one from existing investors and one from previous investors. Current investors in a dataset will likely rate it high so they can increase the value, whilst previous investors may rate it low because they are mad about losses / so they can decrease the value. 

As I ponder it more, I think having one 5-star rating system is too subjective. Even having two is. 

Instead of having one rating for users to encompass all factors of the dataset, we can have multiple ratings, based on... 

- How consistent the dataset is with its promises. Imo, we need to have stricter guidelines for uploaders describing data (when they enter the dataset in the upload page) but that is thought for later. 
- How well the data trains certain neural networks. Datatoken holders can enter the neural network(s) they used to train their model + the results they received. 

Unrelated suggestion: To monitor the marketplace in a decentralised way, Ocean token holders (and datatoken holders) can vote periodically, like they do in the Polkadot system. They can vote on a team of admins committed to regulating the market as a decentralised community. Voting for the admins occurs once every few months or so. So it's like a democracy.

kremalicious:
Those are all wonderful ideas and I would suggest you add them on the actual issue https://github.com/oceanprotocol/multi-repo-issue/issues/32

For now, I'll close this PR since this feature is only in ideation phase. It does not make sense implementing anything in the UI before we know what exact decentralized system(s) are being used and how, and until they are actually implemented. Once we have that, we can then define where exactly a rating action should be in the UI, and how exactly the interaction will be. 

Likewise, the implementation in this PR would need more work too, like loading a big icon font file to get only 1 icon seems a bit ineffective. And using icon fonts comes [with a lot of problems too](https://cloudfour.com/thinks/seriously-dont-use-icon-fonts/), which is why all our icons are inlined SVGs so any rating implementation would need to follow this

PR #264, opened 2020-11-19
3Box publisher profiles

### LINKED ISSUE #234: Retire datapartners list or make it community-maintained

### LINKED ISSUE #240: Link Eth address to other online profiles
Problems: (1) [IP rights violations](https://app.zenhub.com/workspaces/ocean-main-5e5e5a97e91ce03f380fb9b7/issues/oceanprotocol/multi-repo-issue/34) and (2) [rug pulls](https://app.zenhub.com/workspaces/ocean-main-5e5e5a97e91ce03f380fb9b7/issues/oceanprotocol/multi-repo-issue/30), typically both by scam publishers.

What can help to address this: help the legit publishers establish their credibility better in Ocean Market. 

Specific goal here: 
- [ ] Alice can link her Eth address to other online profiles (ENS, twitter, etc), which each are signals towards credibility (or lack thereof)
- [ ] Ocean Market lets Bob see Alice's links 

This will likely need a (short) tech spike.

Tools that may help (more info below):
- [3Box User Profiles](https://3box.io/products/profiles) - links multiple identities; Ethereum-first
- [Keybase](https://www.keybase.io) - links multiple identities; blockchain-friendly but probably more friction than 3Box
- [ENS domains](https://ens.domains/) - human-readable names for Eth addresses. Doesn't directly tie to other identities.
- (More generally, this can be viewed as an identity problem, which is a big rabbit hole. The top Ethereum-based solution is 3Box.)

**About 3Box User Profiles**

"Profiles API makes it easy to add social user profiles and basic reputation to your Ethereum application. With profiles, your users no longer need to interact with robotic hexadecimal addresses, but can begin to enjoy a more usable, familiar, and human web3 experience.

3Box profile data is stored in a decentralized IPFS datastore that is controlled by the user, so data is secure and can be shared across various apps and services.

- Humanize your decentralized application
- Access shared social identity and basic reputation
- Support for shared or application specific profiles
- Compatible with Ethereum accounts

From [here](https://3box.io/products/profiles).

**About Keybase**

"Keybase allows users to prove a link between certain online identities (such as a Twitter or Reddit account) and their encryption keys. Instead of using a system such as OAuth, identities are proven by posting a signed statement as the account a user wishes to prove ownership of." [Ref](https://en.wikipedia.org/wiki/Keybase)

**About ENS domains**

In the web3 world we have .eth domains and we should make use of them as a basic decentralized verification system. Publishers can increase their credibility by attaching their account address to a .eth domain. 

Wherever in UI we display an account address, we check if that address is attached to an .eth domain, and if it is, output this domain for display. So instead of `0x324...3143` we display e.g. `coolpublisher.eth`.

This would also mean that in the user wallet widget the ENS name would show up instead of address if user has one attached, probably will use same component.

Note: there's a ticket specifically for ENS, [here](https://github.com/oceanprotocol/market/issues/188).

mihaisc:
Enhance publisher line with profile data from [3Box Hub](https://www.3box.io/hub), and deprecate `list-datapartners`.

Default view as before:
Screen Shot 2020-11-24 at 18 21 03

The owner of the dataset is shown an affordance to set up a profile if none is found yet:
Screen Shot 2020-11-26 at 10 38 23

New 3Box-enhanced view, when 3Box profile is found for account:
Screen Shot 2020-11-26 at 11 29 09

Hovering the _Profile_ link shows all the data from respective 3Box profile we allow to display:

- image (from IPFS, loaded via our IPFS gateway) 
- name 
- “spirit emoji” 
- description 
- website 
- verified twitter & github accounts

Screen Shot 2020-11-26 at 12 01 59

Also output on asset teasers:
Screen Shot 2020-11-24 at 18 23 58

Screen Shot 2020-11-26 at 11 31 04

### Implementation

This implementation uses the [3Box REST API](https://docs.3box.io/api/rest-api) to check and fetch profile info. To get silent errors without spilling the browser console with 404 errors for accounts without a profile, a [API proxy](https://github.com/oceanprotocol/market-stats/blob/main/api/profile.ts) is used.

This proxy caches requests for 5 min., together with [`stale-while-revalidate`](https://web.dev/stale-while-revalidate/), so updates on 3Box might take some minutes to populate into `market` UI.

Using [`3box-js`](https://github.com/3box/3box-js) was deemed not feasible for simple read-only data, since it introduced too many new dependencies and security vulnerabilities, and with that doubled the app's build time.

- closes #240
- closes #234
- closes https://github.com/oceanprotocol/multi-repo-issue/issues/34, as the last item on the list
- further addressing https://github.com/oceanprotocol/multi-repo-issue/issues/30


kremalicious:
Some build findings for 3box:
- because of lots of Node.js C bindings the dependency install time went up like 1-2 min.
- the build time went up by 3 min., until the error
- package introduces 71 security vulnerabilities marked as high

As for build error, looks like we need to prevent stuff in the Profile provider from happening when we are not in the browser. All providers are currently wrapped around browser and SSR build, so for SSR build we can leave the provider, but then need to add e.g. `typeof window !== 'undefined'` before initializing things in there, so it only fires in browser

mihaisc:
Yes, the lib is pretty big, we can optimize using https://github.com/3box/3box-js#build  and lazy loading the full library only when the user wants to update the profile.

kremalicious:
and main production app.js delivered first to user went up by 2MB, definitely needs to be lazy loaded.

Screen Shot 2020-11-19 at 11 24 14

trentmc:
Note: this PR seems to be related to [Market-240](https://github.com/oceanprotocol/market/issues/240)

src/utils/profile.ts
```diff
@@ -0,0 +1,121 @@
...
+    const response: AxiosResponse<ResponseData3Box> = await axios(
+      `${apiUri}/profile?address=${accountId}`,
+      {
+        cancelToken
+        // validateStatus: () => true
+      }
+    )
+    // axios.interceptors.request.eject(interceptor)
+
+    // TODO: prevent 404 from showing up in console somehow.
```
kremalicious, line 102: as expected, this is tricky. Cause the browser is doing that, not the library, and showing an error in console upon any `404` response is the default in most browsers. And we can't ship production code where by default `404` errors will be thrown.

I had 3 strategies so far:
1. silent error `catch`
2. make axios trick into resolving promise with any status code, and go silent from there
3. intercept the response with axios, and if `404`, go silent

But all of them can't get to the lower levels required to prevent a 404 being shown in browser console.

As far as I see, the only possibility would be to create our own quick API proxy which always returns a `200`. That's fine cause 3Box response sends `{ "status": "error" }` when no profile is found. So we just need another response status code which does not trigger browser errors. 

Any other ideas?

mihaisc: The API proxy would be the best approach, 1 or 3 look interesting, maybe less effort

kremalicious: What I meant was that I already tried option 1-3, and they are not working, and option 4 is too obtrusive. So we have this proxy now: https://github.com/oceanprotocol/market-stats/blob/main/api/profile.ts

mihaisc: Oh, i misunderstood then, my bad.

mihaisc:
When you click the 3box link ideally it should open in a new tab

mihaisc:
lgtm

w1kke:
Tested it with https://www.3box.io/0x655efe6eb2021b8cefe22794d90293aec37bb325
Works!

kaimeinke:
Works wonderful.

image

kremalicious:
have one small remaining issue on small viewports, but it's no blocker. Somehow tippy.js creates a `body` overflow although the popover fits perfectly within the viewport. This leads to some weird interaction where the body behind the popover scrolls a bit while the popover stays correctly placed. None of our css is causing that and it seems an issue with tippy.js

kremalicious (approved):

PR #265, opened 2020-11-19
Handle price fetching better

mihaisc:
Price is not longer fetched from the chain in the `useMetadata` hook, it is manually fetched and set
Moved `AssetProvider` from react hooks and implemented asset purgatory in market.
`AssetProvider` is still WIP (in a later pr, not urgent now), we need to move most of the ddo logic there and then use it in all the components.


src/components/organisms/AssetActions/Pool/index.tsx
```diff
@@ -196,17 +204,19 @@ export default function Pool({ ddo }: { ddo: DDO }): ReactElement {
     }
     init()
 
-    // Re-fetch price periodically, triggering re-calculation of everything
-    const interval = setInterval(() => refreshPrice(), refreshInterval)
-    return () => clearInterval(interval)
+    // TODO why is this again here ?!?!?!
```
kremalicious, line 207: use case was so that people could stay with opened Pool widget and everything in there updates in the shown interval. So all subcomponents react to the price change, and then recalculate

mihaisc: removed

mihaisc:
Found some issues in pool, WIP

kremalicious:
when going to Trade tab, and accepting the banner:

Screen Shot 2020-11-20 at 11 54 48

That `_callee` in Swap.tsx points to:
Screen Shot 2020-11-20 at 11 55 31

kremalicious:
found a pool refresh problem, interval refreshing only works when user arrives on asset details page on client-side from another page, like front-page. If you arrive at an asset detail page, or reload the asset details page, no refreshing happens

kremalicious (approved):

PR #266, opened 2020-11-19
refactor transaction titles

### LINKED ISSUE #261: swap transactions have wrong symbols
In the pool transaction tables, all swap transactions show up the the same symbol:

Screen Shot 2020-11-18 at 13 20 55

kremalicious:
closes #261

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "@coingecko/cryptoformat": "^0.4.2",
     "@loadable/component": "^5.14.1",
     "@oceanprotocol/art": "^3.0.0",
-    "@oceanprotocol/lib": "^0.9.16",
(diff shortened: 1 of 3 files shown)


mihaisc:
is it supposed to look like this?

image

kremalicious:
well, it's like before. Still fighting with that react data table, where the title column is supposed to grow when space is available, which works quite well on history page transactions table but for whatever reason in this small space it shrinks correctly but the table is still pushed in overflow for whatever reason

kremalicious:
table does lots of auto-calculation based on actual content so not sure how much I can influence this with CSS-only, or the possible table settings. Hacked something in so scrollbar shouldn't be there, which is fine as long as the time text is not super long

mihaisc (approved):

PR #269, opened 2020-11-20
use Gatsby’s new File System Route API for asset routes

kremalicious:
Remove all the needed hacks to get our client-only asset routes to work just like in Next.js.

https://www.gatsbyjs.com/docs/file-system-route-api


kremalicious:
after tests: on production versions the 404 page is rendered for a brief second before asset details kick in. This looks like it needs fixing on the Gatsby side

kremalicious:
that's just not working for our case without the 404 page coming up for a brief moment. Can be revisited later but since it doesn't provide any new improvement gonna close this for now

Issue #202, opened 2020-11-03
Set and edit consumption timeout


trentmc:
Issues:
1. In 'publish' form, there is no way to specify length of access control. It defaults to timeout = 0, which is perpetual. I'm not sure if many people realize this. Lack of visibility could lead to contract disputes.
2. The buyer doesn't know how long they're buying for either, unless they look into the metadata and see timeout = 0. (And know that this interprets to 'perpetual'). Lack of visibility could lead to contract disputes.
3. The default is timeout = 0, which means that the publisher is on the hook to provide the data perpetually. This is too onerous of a default.
4. The terms don't have language on timeout at all. And, do they distinguish between "can I use the data I've downloaded for _x_ length of time or forever" versus "here's how long I promise to make the data (updated or not) available for you to access"?
5. (Maybe an issue? I can't tell.) In consume, if timeout is not perpetual, can the user tell if they've timed out or not?

The solution needs to address each of these issues.

Towards a solution, the starting point is addressing (1). The GUI should have a means to make it easy to specify the time span for the license. There could be a few good defaults, e.g. 1 hour, 1 day, 1 week, 1 month, maybe 1 year. And perhaps a "custom" field. I would steer clear of giving a 'perpetual' license in terms of ability to access (though it's ok to give perpetual in terms of the right to use, once you've downloaded it).

Marking this a bug because not doing it could lead to contract disputes and onerous conditions to the publisher.

### UI Spec

- add new field to publish form, show in preview too: `timeout`
  - label: `Download Allowance` (draft)
  - help: `Define how long buyers should be able to download the data set for 1 buy` (draft)
  - make it a `select` with options: `Forever`, `1 week`, `1 month`, `1 year`
- add timeout to editable fields in edit asset screen
- output set timeout as part of text under Buy button
- logic for buy button needs to be adapted to decide when to show `Buy`, and when `Download`

alexcos20:
blocked by https://github.com/oceanprotocol/ocean.js/pull/544

claudiaHash:
What about this approach of displaying the timeout?

![Screenshot (456)](https://user-images.githubusercontent.com/49017601/104747339-115dd500-5759-11eb-93ff-76f27104ab91.png)
![Screenshot (457)](https://user-images.githubusercontent.com/49017601/104747343-11f66b80-5759-11eb-83fd-2e73475fed19.png)

kremalicious:
[...]

PR #272, opened 2020-11-24
add provider uri and timeouts to publish form

azenyem:
addressing #202


src/components/molecules/FormFields/DurationInput/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+      ...field.value,
+      [name]: value.length ? Number(value) : ''
+    })
+  }
+  useEffect(() => {
+    setSeconds(computeTimeoutInSecons(field.value))
+  }, [field])
+  return (
+    <>
+      <InputGroup customClass="duration-input-group">
```
kremalicious, line 95:
```suggestion
      <InputGroup className={styles.duration-input-group}>
```

kremalicious:
Since this is already rather old, I would suggest closing this and addressing only #202 as a separate new PR

PR #274, opened 2020-11-26
price optimizations and bug fixes

### LINKED ISSUE #147: revert hasNoPrice check
as soon as Aquarius is fixed, revert parts of #146 for checking if an asset has a price.

mihaisc:
Give proper message on use tab when there is not enough liquidity in the pool
image (6)
image (5)

Fixed some bugs regarding various scenarios where you couldn't consume the data set even tough you had datatokens

closes #147


src/components/organisms/AssetContent/index.tsx
```diff
@@ -28,9 +29,13 @@ export default function AssetContent({
...
   const { owner } = useMetadata(ddo)
   const { dtSymbol, dtName } = usePricing(ddo)
-  const isOwner = accountId === owner
-  const hasNoPrice = ddo.price.datatoken === 0 && ddo.price.value === 0
-  const showPricing = isOwner && hasNoPrice
+  const [showPricing, setShowPricing] = useState(false)
+  const { price } = useAsset()
+
+  useEffect(() => {
+    console.log(price)
```
kremalicious, line 36: the last remaining console.log :-)

src/components/organisms/AssetActions/Consume.tsx
```diff
@@ -118,7 +130,13 @@ export default function Consume({
           <File file={file} />
         </div>
         <div className={styles.pricewrapper}>
-          <Price ddo={ddo} conversion />
+          {isConsumable ? (
+            <Price ddo={ddo} conversion />
+          ) : (
+            <div className={styles.help}>
+              There is not enough liquidity in the pool to consume the data set.
```
kremalicious, line 137: We don't use 'consume' in the UI so far, and if we use 'buy' instead it also makes it less confusing in the case of low liquidity/has datatokens

```suggestion
              There is not enough liquidity in the pool to buy this data set.
```

kremalicious:
and some UX testing: depending on connection speed, users see the low liquidity message first, then price kicks in. Think we should either show price by default, or nothing.

And I get weird prices in Pool tab on pretty much any asset:

Screen Shot 2020-11-26 at 13 32 16
Screen Shot 2020-11-26 at 13 32 20

mihaisc:
Yes, apparently i made a mistake in the hooks

mihaisc:
Regarding the message order, this happens because we don't have this info in aquarius. I'll look if i can do some hacks

kremalicious:
and as a workaround, throwing in a loading spinner first might turn out to be best compromise between showing nothing and showing error by default. But would also be fine to show price from Aquarius by default first, then switch for error when applicable. Cause the edge case should not influence the majority of use cases

mihaisc:
I'll show the price from aquarius first. I can't really show the spinner because i'm hooked to the automatic price fetching from the provider, so i'll have to refactor a bit for just this temporary hack.

kremalicious:
do you have a test asset on Rinkeby by any chance where the condition low liquidity is met? Want to check if Pool tab shows up properly for those assets, and the asset teaser for it

mihaisc:
Check this one, https://market-n4t5p1b97.vercel.app/asset/did:op:89eEe7002f0Af4c40326A40E5c8f83B112952F63 . Price will be 0 because it's like 0.000009

kremalicious:
cool, that all seems to be working well now

kremalicious (approved):

PR #276, opened 2020-11-27
fix cropped graph on mobile

### LINKED ISSUE #273: A bug discovered in charts while using mobile version.
Ive discovered a bug in the marketplace, while using mobile. The charts for each given dataset, isn't shown entirely. It is cropped of 10% from both ends on timescale. ONLY way to see the whole chart in its total time span, is to tilt the mobile horizontal, otherwise you miss out on about 20% of the chart. It would be great if it could be fixed as soon as possible.

kremalicious:
closes #273


alexcos20 (approved):
lgtm

PR #285, opened 2020-12-02
fix buying 

### LINKED ISSUE #283: Buying an asset takes a long time
The time from button click to first approval popup from MetaMask can take minutes for whatever reason

mihaisc:
Closes #283


kremalicious (approved):
confirmed on main net, approve comes up right away now

PR #288, opened 2020-12-02
Pool statistics from the graph

### LINKED ISSUE #284: query subgraph for all pool statistics (part of TheGraph work)
Refactor to get all pool statistics from our subgraph, with Apollo's `useQuery` hook https://www.apollographql.com/docs/react/data/queries/

mihaisc:
Get pool statistics directly from the graph without connecting a wallet.

closes #284


src/utils/getAddressesForMain.ts
```diff
@@ -0,0 +1,6 @@
+import { ConfigHelper } from '@oceanprotocol/lib'
```
kremalicious, line 1: do we need this utility?

kremalicious:
This works amazing! 

Only have this one issue with the price graph where on first render it's fine but when clicking on another tab, and then back to Pool tab they render cut off for me:

Screen Shot 2020-12-10 at 20 06 20

Something definitely changed in rendering those, we also had a subtle animation when they were initially rendered.

kremalicious:
But all data seems to be alright, also for the graphs when comparing live with this PR, the shapes very closely match with bigger differences only in the very beginning as it seems

Live:
Screen Shot 2020-12-10 at 20 16 31

The Graph:
Screen Shot 2020-12-10 at 20 16 24

kremalicious:
more findings: we need to put the `ApolloProvider` and its `client` in the highest component possible. Otherwise we always have to create a new client in different places. Cause I also will need it for metadata history. And we need it for pool transactions. So we should maybe have some `getApolloClient()` helper passing `subgraphUri` to it

Could make sense adding the `ApolloProvider` the standard gatsby way with `wrapRootElement` but only in gatsby-browser.js but then again we have this problem with switching config

src/utils/getAddressesForMain.ts
mihaisc: no, i needed it before the last update in the graph, i had to filter the tokens, now the ocean reserve is on the transaction. Will remove it

mihaisc:
You are right, i will move it somewhere under OceanProvider so i can catch config changes
I removed the animations because i with them the update looks bad, because i get chunks of 1k items
I'll check the bug when you switch tabs.

kremalicious:
all the graph (the visual one) rendering issues seem to be gone, at least in Chrome and Safari. Might be some chart.js update which fixed it

kremalicious (approved):
working great! Can be released once Rinkeby & Ropsten subgraphs are up again

kremalicious:
all confirmed working in all supported networks, wohoo

PR #291, opened 2020-12-03
Add Portis wallet connection

### LINKED ISSUE #289: Integrate Portis wallet for fiat on/off ramp, more
The main motivation to integrate is traction. Portis has a fiat on/off ramp [[ref]](https://twitter.com/portis_io/status/1334151882655358979), an app store, and good traction. Also, we're in touch with the team; doing this would be a stepping stone to deeper 'datatokens' integration. 

Action for this ticket:
- [x] Make Portis wallet an option, side-by-side with Metamask

Future actions (not part of this ticket):
- Ocean in Portis app store
- Native datatokens support in Portis
- And more

### LINKED ISSUE #332: wallet popover always says MetaMask
When connected with WalletConnect, it still says `MetaMask`:

Screen Shot 2021-01-25 at 14 41 57

It seems web3modal's `getInjectedProviderName()` always returns `MetaMask`

kremalicious:
- add portis, closes #289 
- fix wallet name display, closes #332

**Caveats**

- support only Mainnet accounts for now
- initial connection upon tile click has a slight delay we cannot detect, which might make users click it multiple times (Web3Modal/web3modal#42)

**UI**

Screen Shot 2021-01-25 at 13 13 08

Screen Shot 2021-01-29 at 11 05 29

Workaround for auto-reconnection and the huge delay this has:

Screen Shot 2021-01-29 at 11 30 36

Along the way, all wallets get their icon shown:

Screen Shot 2021-01-29 at 13 39 40


kremalicious:
As for ports, lots of things going wrong, this is as far as I can get:

Screen Shot 2020-12-04 at 12 30 37

This package also does lots of connections by default, for all users, regardless if users interact with it or not and loads a bunch of remote js from remote resources:

- widget.portis.io
- auth.portis.io
- loading some fonts from Typekit
- loading like 40 individual js files(!)

Screen Shot 2020-12-04 at 12 34 46

tomteman:
Hi, we fixed that behaviour in our latest SDK. Please upgrade to 3.0.2, you should be very pleased :)

kremalicious:
@tomteman thanks for dropping by. Can confirm, all the issues we found are gone with `@portis/web3` v3.0.2, beautiful!

mihaisc:
Successfully created an account and logged in without leaving the market. 
Only issue i can find is that in the wallet widget it says `MetaMask`
image

mihaisc:
Can you change the network in portis, like you do in metamask?

tomteman:
@mihaisc yep (we had it first 😏): https://docs.portis.io/#/methods?id=changenetwork

kremalicious:
Portis documentation implies network switching, allows to pass a network string when creating an instance: https://docs.portis.io/#/quick-start

But web3modal does not provide any more options: https://github.com/Web3Modal/web3modal/blob/master/docs/providers/portis.md 

So either that's not supported with web3modal, or no config is needed and it just works. But not sure how to switch networks in Portis wallet, there's a toggle to "show test networks" but not sure where they would show up

mihaisc:
@tomteman thanks 👍 
@kremalicious we need to implement the network change in our wallet widget. We get the portis provider from web3modal and use it like in the documentation https://codesandbox.io/s/1j6j0ry4q

kremalicious:
We also need to add a button to invoke the widget after user has connected, and there then might be the network switch: https://docs.portis.io/#/methods?id=showportis

mihaisc:
@kremalicious we don't need a button for that, that just logs you in, we already do that with web3modal. We just need to create a network switcher (a select control ) in our wallet widget ( maybe were we show the network) and use the code provided in the example
https://codesandbox.io/s/1j6j0ry4q

image

kremalicious:
UX find on the web3modal side: When clicking the Portis option in web3modal, there is a noticeable delay before anything happens. This might make users click the button multiple times. If we can't find and remove reason for that delay, only option left is some loading feedback like spinner. But not sure we can inject that there

mihaisc:
https://github.com/Web3Modal/web3modal/issues/42

kremalicious:
and another web3modal issue, which is present on live: https://github.com/oceanprotocol/market/issues/332

kremalicious:
So far so good:

Screen Shot 2021-01-25 at 15 47 19

While visually there's space we're kinda at the limit of the interactions in that space, all this popup on top of popup shows we should think about a different pattern for our wallet interactions if we add more there.

src/components/molecules/Wallet/Details.tsx
```diff
@@ -1,16 +1,35 @@
...
+
+    const providerInfo = getProviderInfo(web3Provider)
+    setProviderInfo(providerInfo)
+  }, [web3Provider])
+
+  // Handle network change for Portis
+  async function handlePortisNetworkChange(e: ChangeEvent<HTMLSelectElement>) {
+    setPortisNetwork(e.target.value)
+    web3Provider._portis.changeNetwork(e.target.value.toLowerCase())
+    await connect()
```
kremalicious, line 31: @mihaisc something is happening in Portis with this implementation already but our `connect()` does not seem to work. Did you have some other idea here? Feel free to hack it in here

mihaisc: ah, i think we need to refactor more

kremalicious:
We can launch this supporting only Mainnet supporting majority of use cases. We are a bit at a dead end with our initial web3/ocean/wallet connection and switching implementation which highly depends on some exposed `networkChanged` event but technically switching network in Portis only switches accounts. So a huge refactor would be required in our `OceanProvider` in react hooks, which we are not tackling right now.

mihaisc:
Did you try publish/consume?

tomteman:
@kremalicious is there something we could do to help on our end?

kremalicious:
@tomteman thanks, but that's totally on us :-) Even if portis library would expose some event on network change with the new chain ID I'm not sure our current implementation would work. It was created as a plug-and-play solution with external devs in mind where web3modal/web3/ocean are highly coupled but it's overdue uncoupling this for the market to get more flexible

kremalicious:
Consume is a success on mainnet for me but damn, those fees...:

Tested with this account: https://etherscan.io/address/0x0D685BE3419724394aC58C14Ee3507ca31A316Ad

Screen Shot 2021-01-29 at 13 14 46

Screen Shot 2021-01-29 at 13 15 20

Screen Shot 2021-01-29 at 13 15 43

Screen Shot 2021-01-29 at 13 16 00

mihaisc (approved):

Issue #282, opened 2020-12-01
Asset edit history


alexcos20:
- Use TheGraph to get history
- Display history in asset details

kremalicious:
a new metadata history component is included in #292, but will be not shipped initially:

<img width="732" alt="Screen Shot 2020-12-09 at 13 13 02" src="https://user-images.githubusercontent.com/90316/101628765-96e6ba80-3a20-11eb-8707-288e7a3b30c3.png">

It runs on fake data so once we have The Graph in place, we can ship it

kremalicious:
this can be re-evaluated to see if current subgraph allows us to ship the existing component seen in screenshot above. Or to figure out what we need in the subgraph to fill that component with real data

kremalicious:
unblocked. Subgraph now has `updates` on `Datatoken`:
- https://github.com/oceanprotocol/ocean-subgraph/blob/625389490e0c83d07d0d8866f7d462d1d61e6e1a/schema.graphql#L119-L154
- [some example query](https://subgraph.mainnet.oceanprotocol.com/subgraphs/name/oceanprotocol/ocean-subgraph/graphql?query=%7B%0A%20%20datatokens%20%7B%0A%20%20%20%20id%0A%20%20%20%20metadataUpdateCount%0A%20%20%20%20updates%20%7B%0A%20%20%20%20%20%20id%0A%20%20%20%20%20%20tx%0A%20%20%20%20%20%20timestamp%0A%20%20%20%20%20%20userAddress%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%0A)

PR #292, opened 2020-12-04
Update metadata, the proper way

### LINKED ISSUE #237: Edit metadata in UI
The smart contracts and libraries support editing metadata. The Ocean Market UI does not.

### UI

- edit link action, visible only for `owner`
- invokes new edit screen making asset details disappear, same layout as in publish screen
- edit screen has a page title and a description, something like `Edit Metadata`
- edit screen with new form starting with fields for title & description, on left side
- edit screen with title & description preview on the right side
- loading state during submit (can be same UI pattern as in publish)
- after success, redirect to (and possibly reload) respective asset detail with updated metadata
- UI pattern for showing asset edit history, might not be possible from tech side yet (see #282)
- possible bonus: delete action on same screen, see #209

### Code

- new folder `/Edit` in `src/components/organisms/AssetActions`
- new Formic form `FormEdit`, with its own validation scheme
- all fields are optional, so we need to make sure if e.g. only title is filled out, only the title gets edited
- form can be submitted if:
  - account connected
  - correct network
  - one of the fields is filled out
- call `ocean.assets.editMetadata` in ocean.js, see https://github.com/oceanprotocol/market/pull/242/files#diff-ca7956d6692b9236c359b455fb8b0df1de430909ea165ade400a4cf16db65742R31-R35
- double-check if we need the artificial delay when waiting for Aquarius to pick up changes like in publish flow
- reuse existing copy & code:
  - use field definitions from `content/pages/publish.json` if possible
  - refactor, move and use `src/components/pages/Publish/Preview.tsx` & `src/components/pages/Publish/Feedback.tsx` also in edit screen
- edit history only possible if we solve #284 before releasing edit metadata

kremalicious:
- closes #237 
- replaces interim solution, closes #242
- touching #282
- new owner actions UI, and edit flow on the respective asset details view
- new components shared among publish and edit flows: `DebugOutput`, `MetadataPreview`, `MetadataFeedback`
- switch to `ddo.created` for showing publish date, and `ddo.updated` for update date, both of which are set based on block timestamp together with https://github.com/oceanprotocol/aquarius/pull/330
- reduce passing of `ddo` and other props to components, replace with `useAsset()` hook calls
- remove lots of `useMetadata()` calls made redundant by `useAsset()`
- metadata preview tweaks: deprioritize author, show datatoken values
- visual tweaks to metadata output

### UI changes

New owner actions row only visible to data set owner:

Screen Shot 2020-12-07 at 18 10 23

Clicking it opens the data set edit view:

Screen Shot 2020-12-08 at 22 51 24

Loading view:

Screen Shot 2020-12-07 at 17 06 10

Success View:

Screen Shot 2020-12-09 at 13 11 33

When data set was updated, will be shown in the moved date output, and all dates show the exact time when hovered:

Screen Shot 2020-12-07 at 17 03 51

Also includes the new metadata history component, but it can't be shipped until #288 and https://github.com/oceanprotocol/aquarius/issues/324 are finalized. Tracked in #282:

Screen Shot 2020-12-09 at 13 13 02


mihaisc:
Title seems to not have max limit, might be an issue in publish too. 
https://market-2pcgzh971.vercel.app/asset/did:op:C81EAEab66cDD2edb1d19607d6A902F576a5eF5C ( on rinkeby)

mihaisc:
Same blank space works for description too
image

mihaisc:
Still an issue with the file validation
image

mihaisc:
Published a data set with all blanks, including tags,
https://market-2pcgzh971.vercel.app/asset/did:op:25Caf45fE360f8b9133fc8341b2142838d3fbdc1

image

kremalicious:
solved most empty space issues by introducing minimum character lengths for title & description. All other validation issues are caused by Formik/Yup not validating on each `onChange` event, despite this being the default: https://formik.org/docs/api/formik#validateonchange-boolean

So I think somewhere in our chain of `Input` component calls these handlers get removed somewhere

kremalicious:
hacked in semi-manual live-validation to fix most weirdnesses around that. The file input behaves so weird cause the blur event of the `input` fires just before the mousedown event of the `button` fires when user clicks it. Which is why I only could make the `Required` disappear after like 300ms, but it's not shown anymore for the full loading time.

Definitely all needs to be revisited in case we add more forms but for now should be good enough.

kremalicious:
As for limiting the title, I would not go there cause it can be problematic. Not technically, but from user perspective where there can be many reasons to have a longer title and the problem then only lies in the UI. So we should deal with long titles somehow, like cutting them off on the asset teasers, but on asset details show them in full, ideally with some variable font size the longer it gets or something

mihaisc (approved):

PR #297, opened 2020-12-10
minimum viable search

### LINKED ISSUE #38: minimum viable search
Quick wins with current technology where we use Aquarius `nativeSearch` in the `/search` route. This also means all sorting & filtering actions fire a server search to Aquarius:

### Sorting

- by default, sort results by date created, newest first
- sort results by liquidity, from highest to lowest (`ddo.price.ocean`)
- sort results by price, from highest to lowest (`ddo.price.value`)
- needs UI pattern for sorting actions

### Filtering

- filter to show only pools/only exchanges
- needs UI pattern for filter actions

azenyem:
closes #38


kremalicious:
Cool, it works already 🎉 

My first thought for the filtering was to use that UI pattern we have for the price graphs, something like that as a quick mockup, placed on desktop at the top right of the search results:

Screen Shot 2020-12-10 at 12 05 14

As for filters, thought about a filter sidebar but would not really make sense yet if we only have one filter. So we could maybe go with the same UI pattern as above, but placed to the left. Not sure about exact naming, think we can name it `Price` where the options could be `All` (selected by default), `Fixed`, `Dynamic`

src/components/templates/Search/index.tsx
```diff
@@ -7,6 +7,43 @@ import queryString from 'query-string'
 import { getResults } from './utils'
 import Loader from '../../atoms/Loader'
 import { useOcean } from '@oceanprotocol/react'
+import { DragDropContext, Droppable, Draggable } from 'react-beautiful-dnd'
```
kremalicious, line 10: what's the drag & drop list here? What's the use case?

src/components/templates/Search/index.tsx
```diff
@@ -7,6 +7,43 @@ import queryString from 'query-string'
...
+const defaultSearchItems = () =>
+  ['created', 'price.ocean', 'price.value'].map((e) => ({
+    id: e,
+    direction: -1
+  }))
+
+// a little function to help us with reordering the result
```
kremalicious, line 17: Sorting the results on client could lead to many problems, but more importantly we need to reorder across all search result pages, not just for the current page. So every change in sorting simply needs to fire a modified search query to Aquarius

src/components/templates/Search/index.tsx
```diff
@@ -7,6 +7,43 @@ import queryString from 'query-string'
...
+  result.splice(endIndex, 0, removed)
+
+  return result
+}
+
+const grid = 8
+
+const getItemStyle = (isDragging, draggableStyle) => ({
+  // some basic styles to make the items look a bit nicer
+  userSelect: 'none',
```
kremalicious, line 30: all styling should happen in CSS module files, with switching classes based in js variables in here, like `isDragging`. But don't think we should have any drag & drop thingy here

src/components/templates/Search/utils.ts
```diff
@@ -4,23 +4,45 @@ import {
...
     : tags
     ? // eslint-disable-next-line no-useless-escape
       `(service.attributes.additionalInformation.tags:\"${tags}\")`
     : categories
     ? // eslint-disable-next-line no-useless-escape
       `(service.attributes.additionalInformation.categories:\"${categories}\")`
-    : text || ''
+    : text || '';
+
+  searchTerm = (priceType) ? searchTerm+' AND price.type:'+priceType : searchTerm
```
kremalicious, line 45: we usually use template literals to combine strings, so:

```suggestion
  searchTerm = priceType ? `${searchTerm} AND price.type:${priceType}` : searchTerm
```

bogdanfazakas: yeah sure, I'll modify using your suggestion.

kremalicious:
Looks a bit broken to me:

Screen Shot 2021-01-12 at 09 28 56

I see the drag & drop stuff is still in there as mentioned in https://github.com/oceanprotocol/market/pull/297#discussion_r540945958. Let's remove that, it just makes the search more confusing and I don't see a use case for it

src/components/templates/Search/filterPrice.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react'
+import generalStyles from './index.module.css'
+import styles from './filter.module.css'
```
kremalicious, line 3: if the tax is named `filterPrice` the css module file should have the same name

src/components/templates/Search/index.module.css
```diff
@@ -2,6 +2,31 @@
...
 }
 
+.row {
+  display: flex;
+  flex-direction: column;
+  align-items: stretch;
+  justify-content: space-between;
+  margin-bottom: calc(var(--spacer) / 2);
+}
+@media (min-width: 930px) {
```
kremalicious, line 12: no fixed units please, for media queries we always use `rem`

src/components/templates/Search/index.tsx
```diff
@@ -3,10 +3,13 @@ import { QueryResult } from '@oceanprotocol/lib/dist/node/metadatacache/Metadata
 import SearchBar from '../../molecules/SearchBar'
 import AssetQueryList from '../../organisms/AssetQueryList'
 import styles from './index.module.css'
+import PriceFilter from './filterPrice'
```
kremalicious, line 6: consistenty, decide on one name and then stick to it in file, component, and import naming

src/components/templates/Search/index.tsx
```diff
@@ -15,28 +18,71 @@ export default function SearchPage({
...
+            priceType,
+            items
+          })
+      )
+      const queryResult = await getResults(
+        { ...parsed, sort: items, priceType, text: text2, page: page2 },
+        config.metadataCacheUri
+      )
       setQueryResult(queryResult)
+      console.log(queryResult)
```
kremalicious, line 57: console.log should be removed

src/components/templates/Search/sort.tsx
```diff
@@ -0,0 +1,169 @@
+import React from 'react'
```
kremalicious, line 1: there is a lot going on in here but if I see this correctly it tries to reorder the existing results on screen. This will not work because we only show 1 page of the search results and when resorting we want to resort across all pages, not just the one displayed. So changing the sorting needs to trigger a complete new search

kremalicious:
superseeded by #322

PR #307, opened 2021-01-05
Pool tx history

### LINKED ISSUE #279: [BUG]-Data-set names are not changing/scrolling under history/poolshares section
**How to replicate:**

**Go-to:** 
Market - History
Leave it for a few minutes for the pool shares to load up.

Once loaded if you scroll through you will see DT/YOUR POOLSHARES and TOTAL will all change correctly while the actual name of the dataset remains static not changing.

pool1
pool2
pool3
pool4

This has been replicated on both Brave browser and Metamask browser on Android.

Thanks Lee

mihaisc:
- Fixes #279 
- Addressing #326

ToDo:

- [ ] Transactions control keeps refreshing with the price in the pool section ( no idea why it does that, i removed all relations between them)
- [x] The pagination is done locally, need to switch to the graph

Because it is impossible to get the total items returned by a query it is hard to implement server pagination with the current react table (it requires to know the total pages).  We are going to stick with local pagination since  these queries don't return a very large number of items and we fetch only what information we need so there is not a big impact on the client.


mihaisc:
Except for the refresh issue, this is done

kremalicious:
so so fast, amazing!

Just checking all the displayed data and I'm not sure about those transactions where it shows me those double removals:

Screen Shot 2021-01-22 at 12 23 23

Cause I think those are actually the initial liquidity providing transactions, see how my liquidity numbers match the remove transaction numbers, I never removed anything in this pool:

Screen Shot 2021-01-22 at 12 21 59

(testing on Rinkeby right now)

src/components/molecules/PoolTransactions.tsx
```diff
@@ -52,52 +71,55 @@ async function getSymbol(
...
+      locale
+    )}${inTokenSymbol} for ${formatPrice(
+      Math.abs(outToken.value).toString(),
+      locale
+    )}${outTokenSymbol}`
+  } else {
+    for (let i = 0; i < row.tokens.length; i++) {
+      const tokenSymbol = await getSymbol(ocean, row.tokens[i].tokenAddress)
+      if (i > 0) title += '\n'
+      title += `${row.event === 'join' ? 'Add' : 'Remove'} ${formatPrice(
```
kremalicious, line 95: just checked, the mentioned transactions are of event type `setup` so that's another event to account for

mihaisc: Ah, i see, i'll add `setup` too. In the older implementation setup wasn't considered so that's why i had no ideea it actually existed

mihaisc: Should the text  look like `join` ? Meaning `Add 21 Ocean / Add 9 DEFPEL-72` or create a different message?

kremalicious: That 2nd one seems more clear

mihaisc:
`setup` event added, but there is still an issue where when you change the page the `Data set` column doesn't update

src/components/molecules/AssetListTitle.tsx
```diff
@@ -19,8 +19,7 @@ export default function AssetListTitle({
   const [assetTitle, setAssetTitle] = useState<string>(title)
 
   useEffect(() => {
-    if (assetTitle || !config?.metadataCacheUri) return
-
+    if (title || !config?.metadataCacheUri) return
```
mihaisc, line 22: the update issue was here, for some reason the table did not refresh the components and they kept their state, so `assetTitle` was always set with the values of the first page. Will test a bit more so see if this impacts the rest of the site

mihaisc: seems to not break anything

kremalicious:
that rerendering of tx table in pool section we probably can't fix, that might be because we are passing `ddo.price?.address` and we get `ddo` via `useAsset()` and in there the `ddo` might always change. So the problem lies deeper but also not worth solving this since overhead seems to be minimal

kremalicious (approved):
Great stuff!

Now that this pool transactions table is so fast, the slow loading of pool shares becomes even more annoying where it pushes everything down once loaded. So we should refactor the pool shares with The Graph next

PR #308, opened 2021-01-07
Added create date for each asset card

### LINKED ISSUE #133: Add publish/create date on asset card
I was looking in "all data sets" to figure out how old were the assets and the only way to do that was to click on each one to figure out when they were created. There should be enough room to add the date.

claudiaHash:
The create and update date are shown for each asset card, above the price.

closes #133

### DIFF
--- a/src/components/molecules/AssetTeaser.module.css
+++ b/src/components/molecules/AssetTeaser.module.css
@@ -61,3 +61,8 @@
 .symbol {
   display: block;
 }
+
+.date {
+  font-size: var(--font-size-mini);
+  margin-top: calc(var(--spacer) / 2);
+}
--- a/src/components/molecules/AssetTeaser.module.css
+++ b/src/components/molecules/AssetTeaser.module.css
@@ -40,6 +40,9 @@
 
 .foot {
   margin-top: calc(var(--spacer) / 4);
+  display: flex;
+  justify-content: space-between;
+  align-items: flex-end;
 }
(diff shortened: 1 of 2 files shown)


trentmc:
Please link this to a github issue.

kremalicious:
Somewhere the flexbox flow of the cards create weird spacing making the date kinda its own thing: 

Screen Shot 2021-01-12 at 08 45 32

We should decide if date belongs to card footer, header, or description and place it accordingly. Stylistically, having a 4th font size within the card definitely makes them all look more busy. From a visual hierarchy standpoint it is the least important info so if we don't want another font size we might want to try placing it below or beside the price.

And we should stay consistent between date presentation on cards and asset details where we use relative dates. So either we use relative dates on card too, or we switch to standard date on asset details too.

kremalicious:
Something like this would prevent the cards getting so much longer because of the date but space might become a problem. We could work around that by making the price itself a bit smaller since it already is made to look more important through the font color:

Screen Shot 2021-01-12 at 08 59 36

claudiaHash:
Thanks! on it :)


În mar., 12 ian. 2021 la 10:01, Matthias Kretschmann <

kremalicious (approved):
perfect, works great

claudiaHash:
Thanks!

În mie., 13 ian. 2021 la 11:23, Matthias Kretschmann <

PR #309, opened 2021-01-08
cut asset teasers title length

### LINKED ISSUE #295: handle longer titles
As for limiting the title, I would not go there cause it can be problematic. Not technically, but from user perspective where there can be many reasons to have a longer title and the problem then only lies in the UI. So we should deal with long titles somehow, like cutting them off on the asset teasers, but on asset details show them in full, ideally with some variable font size the longer it gets or something

_Originally posted by @kremalicious in https://github.com/oceanprotocol/market/issues/292#issuecomment-742374903_

KatunaNorbert:
- I cut the title length inside asset teasers to 2 lines

closes #295

### DIFF
--- a/src/components/molecules/PageHeader.tsx
+++ b/src/components/molecules/PageHeader.tsx
@@ -1,7 +1,6 @@
-import React, { ReactElement, useEffect, useState } from 'react'
+import React, { ReactElement } from 'react'
 import classNames from 'classnames/bind'
 import styles from './PageHeader.module.css'
-import { ca } from 'date-fns/esm/locale'
 
 const cx = classNames.bind(styles)
 
--- a/src/components/molecules/PageHeader.tsx
+++ b/src/components/molecules/PageHeader.tsx
@@ -1,8 +1,7 @@
 import React, { ReactElement, useEffect, useState } from 'react'
-import ReactDOM from "react-dom";
 import classNames from 'classnames/bind'
 import styles from './PageHeader.module.css'
-import { ca } from 'date-fns/esm/locale';
+import { ca } from 'date-fns/esm/locale'
 
 const cx = classNames.bind(styles)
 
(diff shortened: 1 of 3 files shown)


trentmc:
Please link this to a github issue.

src/components/molecules/PageHeader.tsx
```diff
@@ -18,9 +19,59 @@ export default function PageHeader({
...
+        .getComputedStyle(element, null)
+        .getPropertyValue('line-height')
+    )
+    const lines: number = elementHeight / lineHeight
+    return { linesNumber: lines, lineHeight: lineHeight }
+  }
+
+  useEffect(() => {
+    const titleElement = document.getElementsByClassName(
+      'PageHeader-module--title--2n3-e'
```
kremalicious, line 52: this could go wrong since those classnames are auto-generated, we can only trust that `PageHeader-module--title` won't change (unless we rename the CSS module files or class names). Would suggest rather using CSS attribute selector via `querySelector` so you can do `class*='PageHeader-module--title'`

src/components/molecules/PageHeader.tsx
```diff
@@ -18,9 +19,59 @@ export default function PageHeader({
...
+      lines.lineHeight
+    )
+  }, [])
+
   return (
     <header className={styleClasses}>
-      <h1 className={styles.title}>{title}</h1>
+      <h1
+        className={styles.title}
+        style={{ fontSize: fontSize, lineHeight: lineHeight }}
```
kremalicious, line 71: we like short stuff :-) 

```suggestion
        style={{ fontSize, lineHeight }}
```

kremalicious:
Asset teasers: think 1 line is too aggressive, we should at least allow 2 lines there. Which is then hard to do with CSS only (we do not want to set any fixed widths or heights), which is why in some places we use [react-dotdotdot](https://github.com/CezaryDanielNowak/React-dotdotdot) for this

Asset details: cool approach with the variable font size although [CSS-only true variable font treatment](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Fonts/Variable_Fonts_Guide) with a js fallback would be better. Cause the js solution now detaches the styling from the CSS, which already created different font sizes for the titles. With this change all titles now appear too small on mobile viewports. So ideally js only takes over the styling when a long title is detected, guess something like 3 lines would be a good threshold but not sure how feasible that is

kremalicious (approved):

PR #312, opened 2021-01-12
displayed search results count

### LINKED ISSUE #299: output search results count in all searches
Find a UI pattern to output the total count of DDOs in all search results we get from the search response. One way could be to simply replace the page titles in searches to something like:

```
271 results for SEARCHTERM
No results for SEARCHTERM
```

Then for the all assets search we could have as title:

```
271 datasets
```

Once we output this for the `All Datasets` search, we can remove those numbers from the footer stats.

KatunaNorbert:
closes #299

### DIFF
--- a/src/pages/search.tsx
+++ b/src/pages/search.tsx
@@ -19,9 +19,14 @@ export default function PageGatsbySearch(props: PageProps): ReactElement {
   const title = owner
     ? `Published by ${accountTruncate(owner as string)}`
     : `${
-        totalResults
+        totalResults !== undefined
           ? searchValue
-            ? totalResults + ' results for ' + searchValue
+            ? totalResults === 0
+              ? 'No results'
+              : totalResults +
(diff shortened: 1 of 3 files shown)


kremalicious (approved):
Works like a charm

src/pages/search.tsx
```diff
@@ -17,11 +18,20 @@ export default function PageGatsbySearch(props: PageProps): ReactElement {
     categories
   const title = owner
     ? `Published by ${accountTruncate(owner as string)}`
-    : `Search for ${searchValue || 'all data sets'}`
+    : `${
+        totalResults
+          ? searchValue
+            ? totalResults + ' results for ' + searchValue
```
kremalicious, line 24: small change could be to also account for singular & plural, so the `s` needs to be removed if there's only 1 result. And if there's no result it could say `No results`

KatunaNorbert: sure, i'll do that

PR #317, opened 2021-01-13
Download timeout

claudiaHash:
Compute and render asset availability

requires https://github.com/oceanprotocol/ocean.js/pull/544
addressing #202


src/components/organisms/AssetContent/MetaFull.tsx
```diff
@@ -7,6 +7,23 @@ import { useAsset } from '../../../providers/Asset'
...
+  const [assetTimeout, setAssetTimeout] = useState('')
+
+  const renderedTimeout = (timeout: number) => {
+    switch (timeout) {
+      case 0:
+        return 'Forever'
+      case 3600:
+        return '1 hour'
+      default:
+        return ''
```
alexcos20, line 19: what if timeout is not 0 or 3600?   should we have a function that is doing some nice formating ? like 4 hours or 2 weeks?

after all the timeout is set by the publisher according to his business plan

kremalicious: good point. We probably enforce some predefined values during publish, but when displaying we could just default to conversion from milliseconds with something like https://github.com/vercel/ms

PR #322, opened 2021-01-14
add sort filter features to search

### LINKED ISSUE #38: minimum viable search

bogdanfazakas:
Closes #38

### DIFF
--- a/src/components/templates/Search/filterPrice.module.css
+++ b/src/components/templates/Search/filterPrice.module.css
@@ -3,17 +3,31 @@
   float: left;
 } */
 
-.filter {
-  padding: calc(var(--spacer) / 6);
+.filter,
+button.filter,
+.filter:hover,
+.filter:active,
+.filter:focus {
   border: 1px solid var(--border-color);
   text-transform: uppercase;
   border-radius: var(--border-radius);
   margin-right: calc(var(--spacer) / 6);
-  font-weight: var(--font-weight-bold);
   color: var(--color-secondary);
   background: var(--background-body);
-  cursor: pointer;
+
+  /* the only thing not possible to overwrite button style="text" with more specifity of selectors, so sledgehammer */
+  padding: calc(var(--spacer) / 5) !important;
 }
-.selected {
+
+.filter:hover,
+.filter:focus {
   color: var(--font-color-text);
+  background: inherit;
+  transform: none;
+}
+
+.filter.selected {
+  color: var(--background-body);
+  background: var(--brand-white);
+  border-color: var(--background-body);
 }
--- a/src/components/templates/Search/filterPrice.module.css
+++ b/src/components/templates/Search/filterPrice.module.css
@@ -1,0 +1,14 @@
+.filter {
+    padding: calc(var(--spacer) / 4);
+    border: 1px solid var(--border-color);
+    text-transform: uppercase;
+    border-radius: var(--border-radius);
+    margin-right: calc(var(--spacer) / 4);
+    font-weight: var(--font-weight-bold);
+    color: var(--color-secondary);
+    background: var(--background-body);
+    cursor: pointer;
+  }
+  .selected {
+    color: var(--font-color-text);
+  }
--- a/src/components/templates/Search/filterPrice.module.css
+++ b/src/components/templates/Search/filterPrice.module.css
@@ -1,14 +1,14 @@
-.filterList {
+/* .filterList {
   display: inline-flex;
   float: left;
-}
+} */
 
 .filter {
-  padding: calc(var(--spacer) / 4);
+  padding: calc(var(--spacer) / 6);
   border: 1px solid var(--border-color);
   text-transform: uppercase;
   border-radius: var(--border-radius);
-  margin-right: calc(var(--spacer) / 4);
+  margin-right: calc(var(--spacer) / 6);
   font-weight: var(--font-weight-bold);
   color: var(--color-secondary);
   background: var(--background-body);
--- a/src/components/templates/Search/filterPrice.module.css
+++ b/src/components/templates/Search/filterPrice.module.css
@@ -1,3 +1,8 @@
+.filterList {
+    display: inline-flex;
+    float: left;
+}
+
 .filter {
     padding: calc(var(--spacer) / 4);
     border: 1px solid var(--border-color);
--- a/src/components/templates/Search/filterPrice.module.css
+++ b/src/components/templates/Search/filterPrice.module.css
@@ -28,6 +28,6 @@ button.filter,
 
 .filter.selected {
   color: var(--background-body);
-  background: var(--brand-white);
+  background: var(--font-color-text);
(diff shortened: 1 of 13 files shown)


mihaisc:
Sorting by price is not working

image

mihaisc:
Ideally switching sort direction should be done by clicking anywhere on the button not only the arrow like :
- I navigate to search page, filter is by default all and sort is published
- click on `Liquidity` , sort by liquidity with default direction is selected
- click again on `Liquidity`, sort direction is inverted

bogdanfazakas:
> Ideally switching sort direction should be done by clicking anywhere on the button not only the arrow like :
> 
> * I navigate to search page, filter is by default all and sort is published
> * click on `Liquidity` , sort by liquidity with default direction is selected
> * click again on `Liquidity`, sort direction is inverted

Sure I'll take care of this

kremalicious:
great stuff! Some semantic and visual findings:

- semantically, all interactive elements should be `<button>`. You could use our own `<Button style="text" size="small" />` which outputs `<button>` by default if it does not have a `href`
- price filter: let's reduce this so we do not need a label for the buttons, use `--font-size-small` (which should happen automatically when using above `Button` snippet, and reduce top/bottom inside padding, like so:
  Screen Shot 2021-01-18 at 17 26 38
- the sorting elements should then have kinda the same unified style, but since sorting is different from filtering in a user's mind, it justifies a slightly different style for the sorting. 
  - Maybe turning around the styling for label and elements and visually grouping them, all quickly mocked in browser:
  Screen Shot 2021-01-18 at 17 40 30

  - so we highlight the current choice, and clicking on selected item again reverses sort order, so we do not need a dedicated button on the arrow, the whole item is the button

src/components/templates/Search/filterPrice.module.css
```diff
@@ -0,0 +1,19 @@
+.filterList {
+  display: inline-flex;
+  float: left;
```
kremalicious, line 3: we usually try to avoid using `float` for any layout, but also the isolated component should not define its placement within its parent, to keep it flexible and isolated.

This can be achieved by defining the layout on the parent of sorting component & filtering component with e.g. `display: flex; justify-content: space-between`

src/components/templates/Search/index.module.css
```diff
@@ -2,6 +2,34 @@
   display: grid;
 }
 
+row {
```
kremalicious, line 5: missing a dot so these styles were not kicking in:

```suggestion
.row {
```

src/components/templates/Search/sort.module.css
```diff
@@ -0,0 +1,31 @@
...
+
+.selected {
+  color: var(--font-color-text);
+}
+
+.direction {
+  background: transparent;
+  border: none;
+  color: var(--font-color-text);
+  font-size: 1.3em;
```
kremalicious, line 27: I used `font-size: .75rem;` to get that size seen in my screenshot. Fine making it smaller when the whole element is  a button and not just the arrow

mihaisc:
The sort buttons are not aligned
image

src/components/templates/Search/sort.tsx
```diff
@@ -0,0 +1,88 @@
...
+          [styles.selected]: e.value === sortType,
+          [styles.sorted]: true
+        })
+        return (
+          <Button
+            key={index}
+            className={sorted}
+            size="small"
+            onClick={() => {
+              if (e.value === sortType) {
```
kremalicious, line 57: a bit long so the justifies moving this into it's own `handleButtonClick` handler. Even more optimization could be to remove that double `return`, and have a `<SortItem />` component in same file abstracting out the single component from the `<Sort />` component

src/components/templates/Search/sort.tsx
```diff
@@ -0,0 +1,88 @@
...
+                }
+              } else {
+                setSortType(e.value)
+                sortResults(e.value, null)
+              }
+            }}
+          >
+            {e.display}
+            {e.value === sortType ? (
+              sortDirection === SortValueOptions.Descending ? (
```
kremalicious, line 73: this can be further simplified so markup does not get duplicated:

```
<span className={styles.direction}>
  {String.fromCharCode(sortDirection === SortValueOptions.Descending ? 9660 : 9650)}
</span>
```

and I think then we could move the whole `String.fromCharCode(sortDirection === SortValueOptions.Descending ? 9660 : 9650)` as a `const` outside of the component to the top of file so we end up in component markup with:

```
<span className={styles.direction}>
  {directionArrow}
</span>
```

bogdanfazakas: Sure I'll take care of that

src/components/templates/Search/sort.tsx
bogdanfazakas: You mean merging both Filter and Sort in the same file and returning everything within the same component ?

bogdanfazakas: ok, got it 😃

kremalicious: uuuh, hang on, just looking into this and that might be actually too much refactor. So let's keep the double return but move out the `onClick()` stuff, then it's a good compromise

kremalicious:
alright, styling should be fine now, also on mobile. I would be totally fine with shipping this without that `with price` thingy, once Aquarius is fixed we can revisit. So after above mentioned small markup simplification & onClick handler move this is good to go

bogdanfazakas:
> alright, styling should be fine now, also on mobile. I would be totally fine with shipping this without that `with price` thingy, once Aquarius is fixed we can revisit. So after above mentioned small markup simplification & onClick handler move this is good to go

I was about to commit almost same changes regarding the mobile styling, good that I've checked the origin before 😃 . I'll do the mentioned simplifications and push a new commit soon

mihaisc:
I agree to shipping without `with price`

mihaisc:
Browsing not working, this is the [url](https://market-mquqmgi0n.vercel.app/search?sort=created&sortOrder=desc), it seems that the query is not ok
image
The sort component looks good enough on windows too, mobile is better 
image

bogdanfazakas:
> Browsing not working, this is the [url](https://market-mquqmgi0n.vercel.app/search?sort=created&sortOrder=desc), it seems that the query is not ok
> image
> The sort component looks good enough on windows too, mobile is better
> image

I'll fix the browsing issue along with the mentioned simplifications

kremalicious (approved):
awesome, great work fighting with our somewhat tricky search

PR #324, opened 2021-01-15
Set, edit, and display timeout

### LINKED ISSUE #202: Set and edit consumption timeout

KatunaNorbert:
Fixes #202 .

Changes proposed in this PR:

- added timeout to publish asset form
- added timeout to edit metadata form
- output timeout under buy button
- fix various checks around buy/download button
- improve help text messaging for buy/download

### States for consumers

Timeout set as forever (default), user didn't buy yet and has no datatoken:
Screen Shot 2021-01-22 at 14 14 03

Timeout is set to other than forever, user didn't buy yet and has no datatoken:
Screen Shot 2021-01-22 at 14 19 21

User bought this dataset already and dataset either has forever timeout, or user is still within the timeout:
Screen Shot 2021-01-22 at 14 15 57

Timeout is set, but user has the datatoken, and has not bought dataset before or is outside the timeout:
Screen Shot 2021-01-22 at 14 17 25


src/components/organisms/AssetActions/Edit/index.tsx
```diff
@@ -67,22 +68,43 @@ export default function Edit({
...
+        Logger.error(content.form.error)
+        return
+      }
 
-      // Edit failed
-      if (!ddo) {
+      const service = ddo.findServiceByType('access')
+      const timeout = mapTimeoutStringToSeconds(values.timeout)
+      const ddoEditedTimeout = await ocean.assets.editServiceTimeout(
+        ddo,
```
mihaisc, line 84: Shouldn't it be `ddoEditedMetdata` instead of `ddo`, you ignore the edits to title and description this way. Or am i misunderstanding

KatunaNorbert: Yes it should, I'll change that, strangely it works with `ddo` to

mihaisc: Ah, i see, the return value is kind of useless, ocean.js uses the reference so it will change `ddo`, and thus working as you said

KatunaNorbert: Yas that's it

mihaisc:
Publishing works fine, when i try to edit i get 
image

There is also an issue with sorting the options, should be week, month, year, forever. I'm not sure how to fix this though. The input component has a property `sortOptions`, but don't know how to pass it in the form.

image

mihaisc:
Fixed the autosorting issue, order is as intended now

image

KatunaNorbert:
> Fixed the autosorting issue, order is as intended now
> 
> image



> Publishing works fine, when i try to edit i get
> image
> 
> There is also an issue with sorting the options, should be week, month, year, forever. I'm not sure how to fix this though. The input component has a property `sortOptions`, but don't know how to pass it in the form.
> 
> image

I think that is because the changes in the ocean.js are not up yet. I tested it by adding the dependencies locally. I just reinstalled the dependencies getting them from oceanprotocol and I get an error too.

kaimeinke:
Publish seems to be sorted correctly, but Edit does not.

Publish:
image

Edit:
image

Thanks for the great work on this feature, can't await it.

kaimeinke:
Just one questions. Would you include "1 day" as an option? This is missing and would be a valuable option for daily subscriptions.

KatunaNorbert:
Fixed autosorting issue in edit metadata form
Screenshot 2021-01-18 at 14 38 32

mihaisc:
Yes, we can add `1 day` too.

kaimeinke:
Thank you very much, the daily subscription model is relevant for Evotegra and might be for other publishers that provide daily updates on their data. Thank you @mihaisc

KatunaNorbert:
Added "1 day" option to timeout options
Screenshot 2021-01-18 at 14 57 28

src/models/FormEditMetadata.ts
```diff
@@ -5,14 +5,24 @@ export const validationSchema = Yup.object().shape({
...
-  metadata: MetadataMarket
+  metadata: MetadataMarket,
+  timeout: number
 ): Partial<MetadataPublishForm> {
   return {
     name: metadata.main.name,
-    description: metadata.additionalInformation?.description
+    description: metadata.additionalInformation.description,
+    timeout:
+      timeout === 0
```
kremalicious, line 20: while it's fine hardcoding our own mapping in the publish form, we can't do that for any display of existing timeout in the DDO. Background is that users can push into market without using the UI, with libraries or other interfaces and this always needs to be considered. Most of the time we will encounter `timeout` values we have defined in publish form, but if not, the value display can't be broken.

So I think we do not get around using some library for this, maybe only overwriting its behavior for mapping `0` to `Forever`. This one was mentioned somewhere else https://github.com/vercel/ms but it seem to only deal with milliseconds. Unifying this mapping will also be important in combination with https://github.com/oceanprotocol/market/pull/317 where we display timeout in more places.

mihaisc: I'm not sure we can have a custom option, to actually show the exact value. Maybe we can calculate and add it somewhere in the form and alter `initialValues` and `content.form`. While I agree with you I'm not sure we need to focus that much at this scenario at this point.

Or the easiest way is to put a simple input in edit, not select. And then the user just adds the number of seconds. This way it becomes a bad UX because the user needs to calculate how many seconds are in 1 week or 1 month etc.

Ideally we have a control that adds predefined values in the input, but it implies more dev time. Similar to a date picker, you can manually add the number of seconds but you can also select predetermined values

kremalicious:
This is coming along great, nice work!

Form label naming: now seeing that in context, think `Access Allowance` makes sense? Has the advantage of covering download now, and also compute later.

@claudiaHash if your changes in https://github.com/oceanprotocol/market/pull/317 are not that big yet, it might make sense opening that PR against this `feature/add-timeout-to-publish-form` branch PR here instead, and we merge it into here. Then we would have all timeout-related changes together for testing and we make sure not to double efforts regarding display mapping. As far as I see both branches don't have overlapping changes yet so should just work opening your existing branch against this one without any merge/rebase.

Travis checks should go green once a new ocean.js with https://github.com/oceanprotocol/ocean.js/pull/544 is released and updated in here. CodeClimate's `Cognitive Complexity` complain can be fine, our stuff is complex :-) Worth trying if switch/case statements instead of ternary operator chains makes it more happy, at least with many conditions switch/case is easier to read.

src/models/FormEditMetadata.ts
kremalicious: manually entering seconds indeed is a bit weird. Think then the compromise is simply having some fallback for the case of a timeout not within our presets, this change here would always output `1 year` in that case. It should then just output the existing value from DDO, but feels like we can simply do the conversion for user in this case too. How about simply converting from seconds to hours for the fallback as a compromise? So we don't introduce new library for like 0.5% of use cases, and it's more human-friendly than seconds

mihaisc:
Switch is also faster than ternary operator chains  ( at least chains this big )

claudiaHash:
Sure, 'Access Allowance' sounds good to me.

Before making the PR as suggested, I have a question: for converting the
timeout number to a readable string I use (the already installed library)
'moment' which gives us an output like: 'an hour', 'a day', '7 days' and so
one, which is different from what the user selects in the create/edit form.
Should we install vercel/ms which gives us the output like '1 hour', '1
day', '1 week' to make sure everything is consistent?

În lun., 18 ian. 2021 la 15:27, Matthias Kretschmann <

kremalicious:
we were discussing that timeout display stuff a bit above: https://github.com/oceanprotocol/market/pull/324#discussion_r559569392

So we try to prevent adding any new library, and only have a fallback display for non-preset case. Interesting find with moment.js, always tried to stay away from introducing that particular library since it is know to be rather huge in file size. Guess some dependency of ours installs this

KatunaNorbert:
so i'm changing now the ternary operator to switch, what should be the returned value when displaying the timeout inside the edit form if none of the cases are matched?

KatunaNorbert:
Changed ternary operators to switch

kremalicious:
the fallback should be a value pass-through then so we then output whatever is the DDO. Only suggested to at least convert that fallback to hours, cause it might be easy to do without any new library

KatunaNorbert:
so if there is no match I should add a new option to select containing the timeout value converted to hours, right?

kremalicious:
quick summary:

- any timeout display runs through one method converting seconds to human readable, with `0` being mapped to `forever`. Probably [this method](https://github.com/oceanprotocol/market/pull/324/files#diff-2ec51fbdbf76b95773fe87b47bbac25114e58f14be9da6f2ef5772fb6299b9c8R30) with a modified `default` to that code snippet discussed in Slack, or we use this snippet for everything instead of hardcoding our options
- edit `select`: populated with our default options, but if existing `timeout` does not match any of them, the custom timeout needs to be shown, again ran through above method

KatunaNorbert:
- added method to map seconds to string
- show timeout option inside the edit metadata form for value that doesn't match the default values

mihaisc:
Display and edit seem to be working ok 
image

src/components/organisms/AssetActions/Edit/FormEditMetadata.tsx
```diff
@@ -31,6 +36,30 @@ export default function FormEditMetadata({
...
+      data[timeoutInputIndex].options.pop()
+    } else if (
+      data[timeoutInputIndex].options.length === 6 &&
+      data[timeoutInputIndex].options[5] !== values.timeout
+    ) {
+      data[timeoutInputIndex].options[5] = values.timeout
+    }
+  }
+
+  handleTimeoutCustomOption()
```
mihaisc, line 61: this `handleTimeoutCustomOption()` shouldn't be directly in the component, use an effect

KatunaNorbert: `useEffect()` is not working too well. The option doesn't appear right away, I need to click the select then click elsewhere to make it render my option.

kremalicious: problem is that in current implementation that method runs on every user keystroke so we need to prevent that somehow. Usually helps just testing a bunch of things in the `useEffect` dependency array to figure out upon which value change this should run. But yeah, can be lots of edge cases where UI does not update, pretty sure the whole Formic stuff adds to problems here too. In that case using some Formik method to update its values sometime helped

mihaisc: it should work ok with `useEffect(()=>{handleTimeoutCustomOption()},[data])`

KatunaNorbert: I tried that but it doesn't, is the same problem that I described above

kremalicious:
going through all the places we display timeout, had this idea on consume/compute action which might be better cause it makes the timeout way more clear. Otherwise it feels like we need a tooltip to explain `Access Allowance` in that place:

Forever timeout, keep as is:
Screen Shot 2021-01-20 at 15 42 52

But when a specific timeout is set:
Screen Shot 2021-01-20 at 15 43 27


Does this make it more clear?

claudiaHash:
Seems like a good idea to me, is more explicit. I'll change it right now.

În mie., 20 ian. 2021 la 16:45, Matthias Kretschmann <

src/components/organisms/AssetActions/Edit/FormEditMetadata.tsx
mihaisc: I've done some testing and it will never work in the effect in the current format. When you update the array of the data object you will never trigger an event for react so `data` will need to be moved into a state and change all the things after. Fortunately because this component is handled by formik it's not rendered like a "normal" react component, `handleTimeoutCustomOption` is called only once so i think we don't need to bother with this anymore

mihaisc:
We kind of derailed a bit but if we are polishing the consume button

image

If the user already owns dt shouldn't we hide the second text? the first one is very clear

kremalicious:
Seemed like a good time to clear up all the confusions around the downlod/buy button since we're introducing another factor now. Sorry @KatunaNorbert & @claudiaHash for the side-tracking, will push some changes for this in here because of that.

I think hiding the 2nd text in that case was always the intention but the `!hasDatatoken` logic seems buggy. Cause in my case 2nd text is properly hidden:
Screen Shot 2021-01-21 at 14 56 01

src/utils/metadata.ts
```diff
@@ -10,6 +10,66 @@ export function transformTags(value: string): string[] {
...
+    default:
+      return 0
+  }
+}
+
+function numberEnding(number: number): string {
+  return number > 1 ? 's' : ''
+}
+
+export function millisecondsToStr(milliseconds: number): string {
```
kremalicious, line 34: while trying to make this method a bit shorter, realized we should create a method which does exactly what we need. And we need seconds → string. So would suggest making this into `secondsToString()` so we do not need that other `mapSecondsToTimeoutString()` method too. If milliseconds are then required for internal calculation, this can happen within the method

src/utils/metadata.ts
```diff
@@ -10,6 +10,66 @@ export function transformTags(value: string): string[] {
...
+    ? `${hours} hour${numberEnding(hours)}`
+    : minutes
+    ? `${minutes} minute${numberEnding(minutes)}`
+    : seconds
+    ? `${seconds} second${numberEnding(seconds)}`
+    : 'less than a second'
+}
+
+export function checkIfTimeoutInPredefinedValues(timeout: string): boolean {
+  const predefinedValues = ['Forever', '1 day', '1 week', '1 month', '1 year']
```
kremalicious, line 61: I feel we are having this list in too many places by now and we should be able to keep it down to 2 content files, for publish & edit form (ideally we would only have 1 content file for this but this was tricky to do). So the list in the content file should be the source of truth for this list instead of repeating it within the code, which opens up possible maintainability mistakes.

So in this case we could either import the content json directly in here (with downside that whole file is then included in js bundle), or accept `predefinedValues` as a method parameter and pass them from the component where we get the content with Gatsby's GraphQL

mihaisc:
In your case did you already buy

> Seemed like a good time to clear up all the confusions around the downlod/buy button since we're introducing another factor now. Sorry @KatunaNorbert & @claudiaHash for the side-tracking, will push some changes for this in here because of that.
> 
> I think hiding the 2nd text in that case was always the intention but the `!hasDatatoken` logic seems buggy. Cause in my case 2nd text is properly hidden:
> Screen Shot 2021-01-21 at 14 56 01

Weird things happen, some because of the bug in aquarius.
image
image

Plus in the second image, it says there is no liquidity in the pool but it's a fixed price asset

src/components/organisms/AssetActions/Consume.tsx
```diff
@@ -104,12 +111,17 @@ export default function Consume({
       ) : (
         <>
           <Button style="primary" onClick={handleConsume} disabled={isDisabled}>
-            {hasDatatoken || hasPreviousOrder ? 'Download' : 'Buy'}
+            {hasPreviousOrder
```
kremalicious, line 114: now that we have this slightly revised logic, next step is to incorporate the timeout here right? Something with comparing time of previous order with the timeout in DDO?

mihaisc: the timeout logic is already handled in `checkPreviousOrder`. So if timeout is expired ( you understand what i mean ), `checkPreviousOrder` will not return anything so you `hasPreviousOrder` will be `false`

mihaisc:
for the second help text, the condition should be `(!hasDatatoken &&  !hasPreviousOrder)` instead of  ` || `

src/utils/metadata.ts
claudiaHash: I can take care of that.

src/components/organisms/AssetActions/Consume.tsx
kremalicious: oooh, amazing info! Considered all that and tweaked the help text logic instead

src/components/organisms/AssetActions/Consume.tsx
```diff
@@ -11,6 +11,24 @@ import { useOcean, useConsume, usePricing } from '@oceanprotocol/react'
 import { useSiteMetadata } from '../../../hooks/useSiteMetadata'
 import checkPreviousOrder from '../../../utils/checkPreviousOrder'
 import { useAsset } from '../../../providers/Asset'
+import { mapSecondsToTimeoutString } from '../../../utils/metadata'
+
+function getHelpText(
```
kremalicious, line 16: ok, now we are here. Revised that whole logic we are using to display help text and introduced another sentence which considers the timeout too. Totally needs testing but think then it all makes sense with all the combinations of button + help text

mihaisc: went through all 3 scenarios and it seems ok

src/components/organisms/AssetActions/Edit/FormEditMetadata.tsx
kremalicious: great investigation! So let's leave it as is and be done with it

src/utils/metadata.ts
kremalicious: tinkered around with this and realized we already have the actual content, with the input field definition and the timeout options, passed down to component already so I quickly refactored it to make use of that. So I was able to remove this hardcoded list in https://github.com/oceanprotocol/market/pull/324/commits/1391d2663dd0b5185dd49393eb642d2fc93c86eb

Now we are down to 3 places (where 2 are purely content files) we would have to update, ideally we only have to update the 2 content files and the app does the rest. But not sure we can incorporate that timeout options list into that `mapTimeoutStringToSeconds()` method https://github.com/oceanprotocol/market/pull/324/files#diff-2ec51fbdbf76b95773fe87b47bbac25114e58f14be9da6f2ef5772fb6299b9c8R13

If anybody has an idea to accomplish this please go ahead, otherwise I think it's good enough and we keep that in mind if we update the list

kremalicious:
Agree, all functionality is in place now and couldn't find any bugs with it when testing it through, and the remaining CodeClimate issues are fine to introduce. Left those 2 [bonus](https://github.com/oceanprotocol/market/commit/1391d2663dd0b5185dd49393eb642d2fc93c86eb#r46236391) [points](https://github.com/oceanprotocol/market/pull/324#discussion_r562556146) which is only about reducing possibilities of future dev errors so no need to stress about it if not done.

Remaining display thing is just copy, we need to unify a bit between form & preview:

Screen Shot 2021-01-22 at 13 59 50

Still not fully sure if `Allowance` is more clear than `Timeout`. The help text makes it more clear now I think, so tending towards calling the label in the form `Download Timeout` again, which has advantage that we can just keep as it is in preview. Any other ideas?

mihaisc:
I'm ok with timeout. Honestly with either of them you still need to read the helper text to properly understand the meaning. 
And we should remove `Download` , keep it just `Timeout` , this way it includes compute as well. When we actually implement compute we just modify the helper text, this way it's less of a visual impact on the structure of the form

kremalicious (approved):
Great job @claudiaHash & @KatunaNorbert dealing with our combined product design/building process here, one of those features we had to build and iterate to figure out what works best.

PR #325, opened 2021-01-18
Feature/asset timeout

claudiaHash:
Added the access allowance timeout for each asset. 

closes #317


kremalicious (approved):
awesome, thanks. You can just go ahead and merge it so we can all continue on the big PR with all changes included. We won't get Travis green in here until we update ocean.js anyway

Next thing to look out for is unifying together with @KatunaNorbert whatever is being used to map seconds to string, like the other PR also has a method but it's not exported from there yet https://github.com/oceanprotocol/market/pull/324/files#diff-41b2dae7c1c0602645ae4a49717155df609cea7bb7a335fb711b1b05509eb532R12

claudiaHash:
Oh, thanks, I was wondering about the failing Travis stage! :)

Maybe we should include that function in the /utils/metadata.ts file and
take it from there every time we need it. What do you think?

În lun., 18 ian. 2021 la 19:12, Matthias Kretschmann <

kremalicious:
yup, that seems like a good place for that method

PR #327, opened 2021-01-19
bump ocean.js & react hooks

kremalicious:
bump to ocean.js `v0.10.0` with breaking change to `ocean.assets.editMetadata`, and react hooks to `v0.5.0`. Requirement to make https://github.com/oceanprotocol/market/pull/324 work.


mihaisc:
Edit seems to not be working, were we expecting this?
image

kremalicious:
yup, forgot to add the new `ocean.assets.updateMetadata()`, just committed that and tested locally against Rinkeby

mihaisc (approved):
tested on rinkeby, working

kremalicious:
Mainnet testing is ... interesting... So no matter if the actual spent gas fee is lower, users need to have around 120€ worth of ETH in their wallet

Screen Shot 2021-01-19 at 14 29 49

PR #346, opened 2021-01-27
fix search navigation on localhost

### LINKED ISSUE #347: On localhost search redirect url is broken

mihaisc:
On localhost when you search the url is broken, we used the wrong `navigate`
Fixes #347


kremalicious (approved):
good find!

PR #352, opened 2021-01-28
Fileinfo replacement

### LINKED ISSUE #318: URL validation during publish through provider
Do not use market's microservice to check the published URL, but use ocean.js instead

Market should use ocean.provider.fileinfo (https://github.com/oceanprotocol/ocean.js/blob/main/src/provider/Provider.ts#L81) instead of using it's own endpoint
(https://github.com/oceanprotocol/market/blob/main/src/utils/index.ts#L45)

> Once this is done we should remove this endpoint https://fileinfo.oceanprotocol.com

bogdanfazakas:
Fixes #318.

Changes proposed in this PR:

- replaces market fileinfo with provider fileinfo


src/components/molecules/FormFields/FilesInput/index.tsx
```diff
@@ -20,8 +22,9 @@ export default function FilesInput(props: InputProps): ReactElement {
 
     try {
       setIsLoading(true)
-      const newFileInfo = await getFileInfo(url)
-      newFileInfo && helpers.setValue([newFileInfo])
+      const filesArray: FileMetadata[] = await ocean.provider.fileinfo(url)
```
kremalicious, line 25: the downside now is that adding a file can't be tested without a wallet, users get this, making them assume their file is not working:

Screen Shot 2021-01-29 at 15 45 09

That is always the problem when using the `ocean` object which is why in the `market` we often work around that by hitting some API directly. Haven't checked but maybe we can hit provider API directly? Or is a web3 account really required for this?

If we want to keep using `ocean` here, then we need to adapt the UI for it, like:

- disabling the input (but then we have to tell users why it's disabled)
- adapt the toast error message

bogdanfazakas: I believe it better to try to hit the provider api firstly, if we won't be able to use the api without a web3 account i guess it would be better to adapt to toast message depending on the error thrown

mihaisc:
For some reason assets published on this branch are not processed by aquarius. Publish is successful but you can't view the asset afterwards. I tried with 3 assets, with different file url

src/components/molecules/FormFields/FilesInput/index.tsx
kremalicious: Well, tried to find new provider endpoints but [they are undocumented](https://github.com/oceanprotocol/provider/issues/76). There's a `/fileinfo` route we could hit directly https://github.com/oceanprotocol/provider/blob/master/ocean_provider/routes.py#L223, so I guess full path would be `/api/v1/services/fileinfo`

mihaisc: While I agree with what you said, what is the point of the form without a wallet? If a user is not initiated in the world of web3 and he starts filling the form and at the end realizes that he doesn't have a wallet are we not wasting his time?

kremalicious:
The use case here would be to allow users exploring, testing their files. In general it's about a UX principle we try to follow in market overall: every functionality we can make work without Web3 and a wallet we should.

src/components/molecules/FormFields/FilesInput/index.tsx
kremalicious: The use case here would be to allow users exploring, testing their files. In general it's about a UX principle we try to follow in market overall: every functionality we can make work without Web3 and a wallet we should.

bogdanfazakas: I agree, in fact we would somehow treat differently the input fields and the user will be able to fill all required fields except file and sample file inputs, so filling name, description is also a waste of time for the user if no wallet is connected.
I will try to hit the provider api directly (initially I was trying to use the provider class from OceanJS )  if we decide to stick with the principle of allowing the users to explore and test their files.

mihaisc: Then call `https://provider.mainnet.oceanprotocol.com/api/v1/services/fileinfo` like we directly call other endpoints

mihaisc:
The issue is the `valid?: boolean` field, this is not in the validation on aquarius but was added in ocean-lib.  We can just delete the property for now before adding the file to the ddo

src/components/molecules/FormFields/FilesInput/index.tsx
kremalicious: yup, ocean.js does nothing else so we can mimic that, with the bonus of having our axios cancel token although not super useful in this case https://github.com/oceanprotocol/ocean.js/blob/main/src/provider/Provider.ts#L77-L100

bogdanfazakas:
> The issue is the `valid?: boolean` field, this is not in the validation on aquarius but was added in ocean-lib. We can just delete the property for now before adding the file to the ddo

Yes saw on slack, I will remove the property before adding the file to the ddo

src/components/molecules/FormFields/FilesInput/index.tsx
mihaisc: but if the wallet is connected we should get the url from the config. Even though it does the same thing we should call the correct provider (based on network ), maybe we are testing something on test network.

kremalicious: yes, for the Aquarius replacement methods we always pass the URL for that reason from the component where we get that from `useOcean()`: https://github.com/oceanprotocol/market/blob/main/src/utils/aquarius.ts#L40

mihaisc: actually get `config.providerUri` like we get `metadataCacheUri` https://github.com/oceanprotocol/market/blob/ab609b93c0aa41b1dc567bba2163c95b71434828/src/components/pages/Home.tsx#L60 since this works without a wallet.

src/utils/index.ts
```diff
@@ -42,19 +42,27 @@ export function toStringNoMS(date: Date): string {
   return date.toISOString().replace(/\.[0-9]{3}Z/, 'Z')
 }
 
-export async function getFileInfo(url: string): Promise<FileMetadata> {
+export async function getFileInfo(
```
kremalicious, line 45: perfect. Probably good time to move this out of this generic file. For the ocean.js-aquarius replacement methods we put them all into its own file so we could do that here too and put it into a new file `utils/provider.ts`. And then name the method exactly like ocean.js does to make the mimicking more obvious in code, so `fileinfo()`

src/components/molecules/FormFields/FilesInput/index.tsx
```diff
@@ -1,14 +1,44 @@
...
+  useEffect(() => {
+    const source = axios.CancelToken.source()
+
+    async function validateUrl() {
+      try {
+        setIsLoading(true)
+        const checkedFile = await fileinfo(
+          fileUrl,
+          config.providerUri,
+          undefined
```
kremalicious, line 26: this should be the actual cancel token, right? So `source.token`?

kremalicious:
tested publishing + downloading on Rinkeby with this change, all working perfectly 
https://market-git-fix-fileinfo-replacement.oceanprotocol.vercel.app/asset/did:op:7a50BB7ad0363A4565207f5f24bD178B7F3e32Ec

Not really noticeable by end users so we can then release this without a tweet

src/components/molecules/FormFields/FilesInput/index.tsx
bogdanfazakas: yes, my bad 😃  run some tests with it set to undefined and forgot to set it back. I'll update this right away

kremalicious (approved):
awesome! Thanks also for keeping the CodeClimate warnings down

PR #353, opened 2021-01-28
Display TVL

### LINKED ISSUE #286: "Number of datatokens" displayed in the market footer is meaningless/confusing; TVL would be super-useful
The Ocean Market footer currently shows e.g. "1,298,989.836 OCEAN≈ €502,197.26  and 40,107.252 datatokens in 167 data set pools."

The part "40,107.252 datatokens" is meaningless. It's like giving a stat for "# BTC tokens" + "# ETH tokens" + "# XRP tokens" + ... . Each token has a different value / meaning. This does not help users, and can actually confuse them as they try to sort out what it might mean.

This is precious real estate. There are plenty of stats that are more useful. E.g. min/avg/max # consumes per dataset, min/avg/max OCEAN price per dataset, min/avg/max OCEAN liquidity per dataset. And more. 

[Originally proposed] TODO for this ticket: simply remove the part that gives "40,107.252 datatokens". 

[Updated proposed] TODO for this ticket: Update the footer to give information that accounts for the value of the datatokens (and doesn't show "number of datatokens"). Full proposal is a comment in this PR:  https://github.com/oceanprotocol/market/pull/311#issuecomment-758425502

### Spec

Replace current footer statistics line with:

```
€1,209,894.27 TVL (total value locked) across 156 dataset pools that contain 1,095,651.439 OCEAN and the datatokens for each pool. 
```

claudiaHash:
Fixes #286 .

Changes proposed in this PR:
- tvl value fetched from graph
- converted tvl (ocean) to selected currency
-  tvl displayed in footer stats

--- 

~~blocked by oceanprotocol/ocean-subgraph#16~~


kremalicious:
#354 should help so we do not have to deal with formatting errors in the future

src/components/App.tsx
```diff
@@ -65,28 +65,28 @@ export default function App({
...
-          <Alert
-            title={purgatory.title}
-            badge={`Reason: ${accountPurgatory?.reason}`}
-            text={purgatory.description}
-            state="error"
-          />
-        )}
-        {client ? (
-          <ApolloProvider client={client}>
+      {client ? (
```
kremalicious, line 68: not sure about the whole change in this file, what was the intention? This logic would prevent anything from being rendered unless the Apollo `client` is present?

src/components/atoms/Price/Conversion.tsx
```diff
@@ -59,7 +61,8 @@ export default function Conversion({
       className={styleClasses}
       title="Approximation based on current OCEAN spot price on Coingecko"
     >
-      ≈ <strong dangerouslySetInnerHTML={{ __html: priceConverted }} />{' '}
+      {hideApproximationSign ? <> </> : <> ≈ </>}
```
kremalicious, line 64: I think this would also work, right? Just a bit shorter and no need for those fragments:

```jsx
{!hideApproximationSign && '≈ ' }
```

src/components/molecules/MarketStats.tsx
```diff
@@ -19,9 +22,30 @@ interface MarketStatsResponse {
...
+      poolCount
+      finalizedPoolCount
+    }
+  }
+`
+
 export default function MarketStats(): ReactElement {
   const [ref, inView] = useInView()
   const [stats, setStats] = useState<MarketStatsResponse>()
+  const [totalLockedValue, setTotalLockedValue] = useState()
```
kremalicious, line 41: when no initial state is set, the state should be explicitly typed depending on what `totalLockedValue` should be. Think in this case it would be `useState<string>()`

src/components/App.tsx
claudiaHash: done that because i needed the Apollo client above the footer so i could use it in footer which includes MarketStats. i'm thinking about this...is there another way of doing this better?

src/components/molecules/MarketStats.tsx
```diff
@@ -62,20 +86,9 @@ export default function MarketStats(): ReactElement {
...
-      />{' '}
-      and{' '}
-      <PriceUnit
-        price={`${stats?.datatoken}`}
-        symbol="datatokens"
-        small
-        className={styles.total}
-      />{' '}
-      in <strong>{stats?.datasets.pools}</strong> data set pools.
+      <Conversion price={`${totalLockedValue}`} hideApproximationSign /> TVL
```
kremalicious, line 89: we can make use of HTML as semantically intended for [abbreviations](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/abbr) to get this sentence shorter:

```html
<abbr title="Total Value Locked">TVL</abbr>
```

Not sure we have any default `abbr` styles though so that might need a tweak

src/components/atoms/Price/Conversion.tsx
claudiaHash: yup, it's a much better approach. thanks!

src/components/molecules/MarketStats.tsx
```diff
@@ -62,20 +86,9 @@ export default function MarketStats(): ReactElement {
...
+      (total value locked) across <strong>{stats?.datasets.pools}</strong> data
```
kremalicious, line 90: good point here actually, we need to decide if we should get this from `market-stats` endpoint, or from the The Graph query. In this case, since it's about the pools, it should come from The Graph I think.

Definitely would be most effective to get all data within this component from one source only but not sure that's feasible for all numbers, like number of datasets and accounts.

src/components/App.tsx
kremalicious: ah got it, didn't think of that. Current live implementation tries to make sure that header & footer are not re-rendered upon page changes, slo so that not every page change initializes new API requests happening in footer. So that would be the goal somehow.

Alternatively, since it's Gatsby, the correct place for providers wrapping the whole app would be in `helpers/wrapRootElement` so maybe we want to move this there. But think @mihaisc ran into issues with that back then for initial implementation

src/components/molecules/MarketStats.tsx
mihaisc: well from the original issue i understand that we settled on just `€1,209,894.27 TVL (total value locked) across 156 dataset pools that contain 1,095,651.439 OCEAN and the datatokens for each pool. ` and drop `Total of 278 data sets & unique datatokens published by 172 accounts.` since that doesn't have that much value and we keep it simple. We have all that info from the graph  (or we will have it at least once it is updated) . So basically this pr is blocked until then

kremalicious:
Blocked by https://github.com/oceanprotocol/ocean-subgraph/issues/16

Any other blockers for this PR from the data source?

claudiaHash:
> Blocked by [oceanprotocol/ocean-subgraph#16](https://github.com/oceanprotocol/ocean-subgraph/issues/16)
> 
> Any other blockers for this PR from the data source?

Nothing, as far as I know.

kremalicious:
think we can get this already implemented with suggested changes, and where we assume correct numbers from the subgraph so we end up with only this sentence as [mentioned here](https://github.com/oceanprotocol/market/pull/353#discussion_r566878681):

```
€1,209,894.27 TVL across 156 dataset pools that contain 1,095,651.439 OCEAN and the datatokens for each pool.
```

So that as soon as https://github.com/oceanprotocol/ocean-subgraph/issues/16 is fixed we can release this here.

src/components/molecules/MarketStats.tsx
```diff
@@ -19,9 +22,30 @@ interface MarketStatsResponse {
 
 const refreshInterval = 60000 // 60 sec.
 
+const getTotalPoolsValues = gql`
+  query PoolsData {
+    poolFactories {
+      totalLockedValue
```
trentmc, line 28: This should be `totalValueLocked` (TVL) not `totalLockedValue`. That's the industry standard. If we don't follow the standard we will confuse people. Example: see "TVL" in https://defipulse.com/

mihaisc: This a sub-graph issue, not here

src/components/molecules/MarketStats.tsx
```diff
@@ -59,23 +82,11 @@ export default function MarketStats(): ReactElement {
...
-      <PriceUnit
-        price={`${stats?.datatoken}`}
-        symbol="datatokens"
-        small
-        className={styles.total}
-      />{' '}
-      in <strong>{stats?.datasets.pools}</strong> data set pools.
+      <Conversion price={`${totalLockedValue}`} hideApproximationSign />{' '}
+      <abbr title="Total Value Locked">TVL</abbr> (total value locked) across{' '}
+      <strong>{stats?.datasets.pools}</strong> data set pools that contain{' '}
```
kremalicious, line 87: we should get the pools number also from a subgraph query, then we can completely remove fetching from `market-stats.oceanprotocol.com`

And it's fine removing `(total value locked)` from the text display since it's behind the `abbr`

src/helpers/wrapRootElement.tsx
```diff
@@ -10,6 +10,7 @@ import {
 } from '@oceanprotocol/lib/dist/node/utils/ConfigHelper'
 import { UserPreferencesProvider } from '../providers/UserPreferences'
 import PricesProvider from '../providers/Prices'
+import ApolloClentProvider from '../providers/ApolloClientProvider'
```
kremalicious, line 13: small typo: `ApolloClentProvider` should be `ApolloClientProvider`

src/components/molecules/MarketStats.tsx
```diff
@@ -59,23 +82,11 @@ export default function MarketStats(): ReactElement {
...
+      <PriceUnit price={totalOceanLiquidity} small className={styles.total} />{' '}
+      and datatokens for each pool.
       <br />
       <strong>{stats?.datasets.none}</strong> data sets have no price set yet.
```
kremalicious, line 91: would also kill this line completely since it doesn't really have any value, and we remove another need to fetch `market-stats` endpoint

src/components/molecules/MarketStats.tsx
```diff
@@ -19,9 +22,29 @@ interface MarketStatsResponse {
 
 const refreshInterval = 60000 // 60 sec.
 
+const getTotalPoolsValues = gql`
+  query PoolsData {
+    poolFactories {
+      totalLockedValue
```
kremalicious, line 28: just a heads up, over in `ocean-subgraph` multiple fixes were added today, among them fixes for TVL value but also fixing the naming of the keys to be correct and consistent. So `totalLockedValue` is now correctly `totalValueLocked`: https://github.com/oceanprotocol/ocean-subgraph/blob/main/src/mappings/factory.ts#L15

So once that's deployed (always a bit unclear if `main` is deployed or not), we need to switch key in the query here and also make sure that within this PR we always use `totalValueLocked` e.g for variable names

src/helpers/wrapRootElement.tsx
kremalicious: also, now the app has 2 `ApolloProvider`, one added in `App.tsx` and one here. So should be removed again from `App.tsx` if all works when having the provider here

src/providers/ApolloClientProvider.tsx
```diff
@@ -0,0 +1,36 @@
...
+export default function ApolloClentProvider({
+  children
+}: {
+  children: ReactNode
+}): ReactElement {
+  const [client, setClient] = useState<ApolloClient<NormalizedCacheObject>>()
+  const { config } = useOcean()
+
+  useEffect(() => {
+    const newClient = new ApolloClient({
```
kremalicious, line 21: good solution, putting it behind `useEffect` should make Gatsby not fail during its SSR build where it probably can't handle this client. To be safe, the effect could start with: 

```ts
useEffect(() => {
  if (!config?.subgraphUri) return
 
  const newClient = ...
}, [config]
```

Then we make sure it only ever fires once `config` and `config. subgraphUri` are available. This is a general pattern we have to follow whenever using values from `useOcean` and reacting to it. Because by default most of its exposed values are undefined and are filled during the lifecycle of the `OceanProvider`. So we always need to make sure we only do stuff once a value is not undefined anymore

src/providers/ApolloClientProvider.tsx
```diff
@@ -27,10 +29,11 @@ export default function ApolloClentProvider({
       }),
       cache: new InMemoryCache()
     })
+
     setClient(newClient)
-  }, [config])
+  }, [config, client])
```
kremalicious, line 34: I think this creates a loop. By putting `client` in the dependency array of the effect, it tells it to run whenever `client` changes. But it then sets `client` which will trigger it doing the same thing again and again

mihaisc:
I'm looking into this and the `ApolloClientProvider` is acting weird

mihaisc:
So, unfortunately we need this logic  (i know you don't like it @kremalicious ). Basically you can't have a `ApolloProvider` without a valid `client`. I think it's an ok compromise since we have a url from the start. 
```
 return client ? (
    <ApolloProvider client={client}>{children}</ApolloProvider>
  ) : (
    <></>
  )
``` 
And @claudiaHash please remove everything related to Apollo from `App.tsx` having a `ApolloProvider` in another `ApolloProvider` is messing stuff up

mihaisc:
Should be ok

kremalicious:
Yeah, think everything seems correct! Thanks for fixing that Apollo setup, seems to not make the typical problems with SSR 

Screen Shot 2021-02-10 at 17 16 32

mihaisc:
Looking at [this](https://subgraph.mainnet.oceanprotocol.com/subgraphs/name/oceanprotocol/ocean-subgraph/graphql?query=%7B%0A%20%20pools(first%3A1000)%7B%0A%20%20%20%20id%0A%20%20%7D%0A%7D%0A) query there actually are.

kremalicious:
new reference with `finalizedPoolCount`, much more realistic considering the purgatory assets counted in it

Screen Shot 2021-02-10 at 18 13 29

Is this the correct place for using an Oxford comma? Or is it clear that number of datatokens is not the same number as OCEAN?

> ...that contain 1,119,869.616 OCEAN, and datatokens for each pool.

Albeit technically correct, from usability standpoint it stays a bit confusing when comparing data set numbers from searches. Might be good place to have some tooltip explaining this number slightly

mihaisc:
You lost me with the comma. Before i never had a doubt when reading it, now i am wondering...
Good idea with the tooltip, always good to be transparent with the data

claudiaHash:
Screenshot from 2021-02-10 19-45-28
should the icon be smaller?

kremalicious:
could be, but could also look alright if we put the icon at the end of sentence, since tooltip info concerns all numbers

kremalicious:
network switching works super well too. But did not test against barge yet, will do that tomorrow. Cause then we'll have a `subgraphUri : null`, so want to make sure this doesn't break whole app https://github.com/oceanprotocol/ocean.js/blob/main/src/utils/ConfigHelper.ts#L43

mihaisc:
Another option would be to specifically add the provider where we need it ( AssetActions, Footer, History )

mihaisc:
Update : I blocked subgraph.mainnet.oceanprotocol.com and tested the website. Works fine, just the stuff from the graph doesn't load so I think we are ok
image
image
image
image
Maybe we create a component that monitors the status of our services ( not related to this PR )

kremalicious:
phew, got it, local barge development was broken cause the `handleNetworkChange` in network monitor never fired, so subgraphUri was never overwritten, resulting in ApolloClient never getting initiated, which then broke things. Pushing fix in a second

src/providers/ApolloClientProvider.tsx
```diff
@@ -0,0 +1,47 @@
...
+    if (!(config as ConfigHelperConfig)?.subgraphUri) return
+
+    const newClient = createClient((config as ConfigHelperConfig).subgraphUri)
+    setClient(newClient)
+  }, [config])
+
+  return client ? (
+    <ApolloProvider client={client}>{children}</ApolloProvider>
+  ) : (
+    <></>
```
kremalicious, line 43: while in 5703ff5 I made sure `client` is present in all networks it seems like if this condition happens and there's no ApolloProvider the whole app breaks. We forgot to pass `children`

kremalicious: so tricky, think we have to leave it like that as the best compromise. Basically whenever app encounters an undefined `config.subgraphUri` we prevent all children on `ApolloProvider` from rendering.

If I pass `children` in that case all the subcomponents using Apollo make app break. Since we're using hooks to access `ApolloProvider` there is no way in React to conditionally use them, unless we prevent every component from rendering whenever `ApolloProvider` is not present.

So best pragmatic way seems to make sure that `config.subgraphUri` is always defined which should happen successfully now in this PR. I'll log an error to console for all other cases

kremalicious:
> Another option would be to specifically add the provider where we need it ( AssetActions, Footer, History )

also though about that but think it's worth figuring it out globally:
- we do not have to deal with the Apollo setup again in all future The Graph work
- we do not constantly unmount and mount the provider on route changes

Putting it in Gatsby's `wrapRootComponent` should make sure of both, but it also requires us to always test the deployed version because locally there's no SSR build. But seems to be all working well now

src/providers/ApolloClientProvider.tsx
mihaisc: check my comment above regarding how the app behaves when you don't have subgraph. The check is more or less to make ssr happy. After deployment even when the graph endpoint is broken the app works ok

kremalicious:
final before/after reference snapshots, think that's shippable if everybody agrees on the validity of all numbers:

Screen Shot 2021-02-11 at 12 39 50

Screen Shot 2021-02-11 at 12 39 38

trentmc:
> think that's shippable if everybody agrees on the validity of all numbers:

+1 from my side

trentmc:
As commented in slack: Rather than "OCEAN and datatokens for each", I'd say "OCEAN, plus datatokens for each"

PR #354, opened 2021-01-28
Dev experience tweaks

kremalicious:
Trying to get those formatting errors out of the way we all run into from time to time. We shouldn't even think about it so let's automate some of that in this PR:

- add VS Code workspace settings:
  * add everything required for auto-formatting on filesave
  * add list of suggested extensions
- add pre-commit hook for `prettier` with [Husky](https://typicode.github.io/husky) + [pretty-quick](https://github.com/azz/pretty-quick)
  * will run `prettier` on staged files before a commit is made
  * after merge of this PR everybody needs to run `npm install`, which will install `husky` in your environment along the way
- update code style docs

So if you use VS Code, then you can accept the suggested extensions and formatting happens automatically when hitting _Save_. For everything else the pre-commit hook will deal with that.


mihaisc:
.vscode files are not fetched, rest seems ok

mihaisc (approved):

Issue #351, opened 2021-01-28
"Published" section in History needs to be paginated


mihaisc:
The "Published" section currently uses asset tiles and it takes too much space. 
We have two options: 
- paginate it and have only  6 items per page ( maybe less on mobile)
- move it to a table format like the other sections on the page

PR #356, opened 2021-01-29
Feature/asset list isolated pagination

### LINKED ISSUE #91: AssetList needs to have its own isolated pagination
As of right now, our `Pagination` component within `AssetList` only works on the search page because we use it to manipulate the url query which then fires a new search query to Aquarius.

This mechanism needs to change so that every `AssetList` can be placed wherever and the pagination still works, without updating any URLs. Might be tricky because the component itself does not fetch the assets but just receives the results of a search query. 

A solution might be to have a `SearchQuery` passed to `AssetList` which then fetches itself, and manipulates the `SearchQuery` when pagination is clicked.

KatunaNorbert:
Fixes #91 .

Changes proposed in this PR:

- isolated pagination inside `AssetQueryList` so it doesn't depend on `Search` component


src/components/molecules/Pagination.tsx
```diff
@@ -40,7 +44,9 @@ export default function Pagination({
       marginPagesDisplayed={smallViewport ? 0 : 1}
       pageRangeDisplayed={smallViewport ? 3 : 6}
       onPageChange={(data) => onPageChange(data.selected)}
-      hrefBuilder={(pageIndex) => hrefBuilder(pageIndex)}
+      hrefBuilder={(pageIndex) => {
```
mihaisc, line 47: There shouldn't be anything related to url here, this is a simple pagination component, it's should work with server side pagination and client side. `hrefBuilder` should be called in the parent component in `onPageChange`

mihaisc: For example `AssetQueryList` is also used in `PublishedList` and now it's not working. If i think more about it `AssetQueryList` should be called just `AssetList` and accept a list of `DDO` and `Pagination` should do just that. The rest of the logic is done in the parent component , for example in search : you do the query, pagination logic and the url stuff in the search component and then pass the query results  (the `DDO` list) to the `AssetList` and what params you need to the pagination

KatunaNorbert: the function will be called inside `AssetQueryList`, I just tried to make it optional inside `Pagination`. But I can move that check inside `AssetQueryList`

src/components/molecules/PoolTransactions.tsx
```diff
@@ -11,7 +11,7 @@ import { formatPrice } from '../atoms/Price/PriceUnit'
 import { gql, useQuery } from '@apollo/client'
 import {
   TransactionHistory,
-  TransactionHistoryPoolTransactions
+  TransactionHistory_poolTransactions as TransactionHistoryPoolTransactions
```
kremalicious, line 14: ran into this weirdness too, the Apollo typings generated locally and the the ones generated in CI differ for whatever reason. This is also why Travis fails https://travis-ci.com/github/oceanprotocol/market/builds/215211030#L579

I ended up leaving this as is (`TransactionHistoryPoolTransactions `) and learned to ignore the error in my editor 🙃

KatunaNorbert: well locally this change fixed the error, but in github Travis test seems to fail either way

mihaisc: if you get that locally it means something after `find` isn't working for you locally
https://github.com/oceanprotocol/market/blob/a76093556495444b0f6735e0fe52211e5fcc8ec9/package.json#L22

src/components/molecules/Pagination.tsx
kremalicious: so I called it `AssetQueryList` back then to make it clear this always accepts OUR search query. And that always has pagination info.

We just need to decide what use case this should support, it seems the work here is to ultimately tackle #351. Based on that it seems we should figure out how `AssetQueryList` can always have a pagination inside the component, and not just on the search page

mihaisc: We can leave the pagination in there, and the `AssetList` component can have `assets: DDO[]` , `page`, `totalPages`, `onPageChange` . This way we can use it in multiple ways, we can use it on homepage as well, instead of showing two rows each subsection can be paginated and the use can just browse with preset filters, for the algos, in the history, and others. We need to decouple this component from the data source

KatunaNorbert: this sounds good for me. Maybe we should add a property like `showPagination` or `hidePagination` in case we don't want the pagination to be displayed

mihaisc: yes, and except for `assets` and maybe `showPagination` all the other are optional, maybe you just want to display 4 items and that's it.

KatunaNorbert: ok, then i'll also rename the `AssetQueryList` to `AssetList`

src/components/molecules/PoolTransactions.tsx
kremalicious: gonna investigate this cause with the new GiHub Actions setup I am able to actually capture this, this fails on macOS https://github.com/oceanprotocol/market/runs/1806022553?check_suite_focus=true

KatunaNorbert: something strange its happening when I try to resolve the conflicts locally. I have conflicts in 24 files, and none of them was changed by me

src/components/templates/Search/index.tsx
```diff
@@ -1,12 +1,14 @@
...
 import SearchBar from '../../molecules/SearchBar'
-import AssetQueryList from '../../organisms/AssetQueryList'
+import AssetList from '../../organisms/AssetList'
 import styles from './index.module.css'
 import queryString from 'query-string'
 import PriceFilter from './filterPrice'
 import Sort from './sort'
-import { getResults, addExistingParamsToUrl } from './utils'
+import { getResults } from './utils'
+import { useNavigate } from '@reach/router'
```
mihaisc, line 10: Use `import { navigate } from 'gatsby'`

kremalicious (approved):
perfect, makes it more flexible. Quickly tested and couldn't find anything breaking wherever `AssetList` was used, so good to go

PR #357, opened 2021-01-29
Fix random null balance 

### LINKED ISSUE #145: OCEAN user balance is null randomly
Happens a lot on Rinkeby. Ran into that on the Cobra Capital asset during the demo:

Screen Shot 2020-10-27 at 16 22 27

mihaisc:
This is not exactly a fix but it was needed anyway. We need to periodically re-fetch the balance of the chain. 
The main issue is when you have the network on rinkeby and navigate to the market. By default the market connects to main so basically we start two balance fetch operations the first for main and immediately one for rinkeby. The fetch with main ocean address is actually executed on rinkeby so it fails but it fails after the fetch from rinkeby is executed so the returned ocean balance is null. Until we completely refactor the hooks to separate web3 from ocean we can't properly fix this.
Closes #145


kremalicious (approved):
nice! Confirmed working on Rinkeby/MetaMask, OCEAN was shown some seconds after connecting wallet

PR #360, opened 2021-02-01
Get pool shares using The Graph

### LINKED ISSUE #326: Pool Shares and Pool Transactions at History page take a long time to load
**To Reproduce:**
1) Go to[ market.oceanprotocol.com/history](https://market.oceanprotocol.com/history)
2) Scroll down to below as shown in screenshot.
3) Sometimes it takes up to a minute or more to load the pool shares and pool transactions data.

image

KatunaNorbert:
Fixes #326 .

Changes proposed in this PR:

- used the graph to get the poolShares list

final piece to unblock https://github.com/oceanprotocol/aquarius/issues/329

Screen Shot 2021-02-10 at 12 52 41

---

- ~~blocked by https://github.com/oceanprotocol/ocean-subgraph/issues/15~~
- ~~nice-to-have https://github.com/oceanprotocol/ocean-subgraph/issues/14~~
- nice-to-have, but we can work around it https://github.com/oceanprotocol/ocean-subgraph/issues/13


src/components/pages/History/PoolShares.tsx
```diff
@@ -56,20 +79,33 @@ export default function PoolShares(): ReactElement {
...
 
       try {
-        const pools = await ocean.pool.getPoolSharesByAddress(accountId)
         const metadataCache = new MetadataCache(config.metadataCacheUri, Logger)
         const result: Asset[] = []
-
         for (const pool of pools) {
-          const ddo = await metadataCache.retrieveDDO(pool.did)
-          ddo && result.push({ ddo, shares: pool.shares })
+          const did = web3.utils
```
mihaisc, line 104: minor performance hit here because you fetch all the ddos before and not on row but considering there are few items and we will change the way we fetch the name it's ok

mihaisc (dismissed):

kremalicious:
what is the order of the pool shares now? They should be ascending based on when user added pool share, so newest at top

src/components/pages/History/PoolShares.tsx
kremalicious: not sure this is that minor. When comparing the 2 tables on history page here https://market-git-feature-get-pool-shares-using-graphql.oceanprotocol.now.sh/history the pool shares table is significantly slower for me, although I have more items in the pool transactions than in the pool shares. Pool transactions come in in under 1 sec., pool shares take 10 sec. for me

kremalicious:
@KatunaNorbert and for dev flow, please always complete this line in your issue description: `Fixes # .`

KatunaNorbert:
It is descending, so the newest is at the bottom

src/components/pages/History/PoolShares.tsx
KatunaNorbert: they happen at the same time for me, but I have just a few items, 2 pool shares and 3 pool transaction

kremalicious:
haha, always so confused about ascending/descending. We want newest to old, just like the pool transactions table, which defines `orderDirection: desc`. So is the ordering tuned around for the pool shares query?

KatunaNorbert:
Wait i was wrong, I forgot to add the `orderBy` field, so it isn't based on timestamp like the transactions. But pool shares doesn't have timestamp field so I can't order them by that.

src/components/pages/History/PoolShares.tsx
mihaisc: Then i underestimated the impact so it needs to be as it is on the tx table

mihaisc:
Pool share has only
```
id: ID!
userAddress: User!
poolId: Pool!
balance: BigDecimal!
```

We can't really filter them by anything at this point. To add a timestamp we will need some custom logic on the graph because what is the timestamp of the share? The first tx? the last tx? I think we should order by balance desc. I think you want to see your highest liquidity on top

kremalicious:
good points. I like how we can finally think about how and what to display here cause we don't wait for 10 min. to see them :-) 

I like the order by user liquidity idea. Think that is more valuable than seeing the total pool liquidity. So I would suggest switch that total column for the user liquidity, and then order by that. Maybe name the column header just `Value`, or keep it blank.

But while we are at it (we can also do this in another issue/PR) and we probably can do this now with The Graph, the pool shares number is not that useful, it would be great to display what this represents in a `Your Liquidity` column, like Uniswap does:

Screen Shot 2021-02-01 at 15 42 50

mihaisc:
Agree, we add `datatokenReserve` and `oceanReserve` to `poolId` in the query and we have that informatio. Or we can use `lockedValue` ( as well in `poolId` , value is in ocean) and convert it in the currency that the user selected in the preferences (top right)

src/components/pages/History/PoolShares.tsx
KatunaNorbert: as it is now the table doesn't get just the data set name from the ddo but the total pool liquidity too, so if I implement it as in transactions component there would be 2 requests, one for the name column and one for the total pool liquidity. But if the total pool liquidity will be replaced with something else that we can take from pool shares than we can get the asset title only

KatunaNorbert:
Ok, we can replace the total pool liquidity with user liquidity but the graph doesn't support an ordering by that, so it should be implemented on the graph side

mihaisc:
Order by balance , here is the [query](https://subgraph.rinkeby.oceanprotocol.com/subgraphs/name/oceanprotocol/ocean-subgraph/graphql?query=%7B%0A%20%20poolShares(orderBy%3Abalance%2CorderDirection%3A%20desc%2C%20first%3A%201000%2C%20where%3A%20%7BuserAddress%3A%20%220x4d156a2ef69ffddc55838176c6712c90f60a2285%22%7D)%20%7B%0A%20%20%20%20id%0A%20%20%20%20balance%0A%20%20%20%20userAddress%20%7B%0A%20%20%20%20%20%20id%0A%20%20%20%20%7D%0A%20%20%20%20poolId%20%7B%0A%20%20%20%20%20%20id%0A%20%20%20%20%20%20datatokenReserve%0A%20%20%20%20%20%20oceanReserve%0A%20%20%20%20%20%20lockedValue%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%0A)

src/components/pages/History/PoolShares.tsx
```diff
@@ -1,86 +1,103 @@
...
   return (
-    <Conversion
-      price={`${totalLiquidityInOcean}`}
-      className={styles.totalLiquidity}
-    />
+    <Conversion price={`${lockedValue}`} className={styles.totalLiquidity} />
   )
 }
 
+function Symbol({ tokenAddress }: { tokenAddress: string }) {
```
kremalicious, line 43: we might not need this, how about just getting the symbol within the GraphQL query, [like so](https://subgraph.rinkeby.oceanprotocol.com/subgraphs/name/oceanprotocol/ocean-subgraph/graphql?query=%7B%0A%20%20poolShares(orderBy%3Abalance%2CorderDirection%3A%20desc%2C%20first%3A%201000%2C%20where%3A%20%7BuserAddress%3A%20%220x4d156a2ef69ffddc55838176c6712c90f60a2285%22%7D)%20%7B%0A%20%20%20%20id%0A%20%20%20%20balance%0A%20%20%20%20userAddress%20%7B%0A%20%20%20%20%20%20id%0A%20%20%20%20%7D%0A%20%20%20%20poolId%20%7B%0A%20%20%20%20%20%20id%0A%20%20%20%20%20%20symbol%0A%20%20%20%20%20%20datatokenReserve%0A%20%20%20%20%20%20oceanReserve%0A%20%20%20%20%20%20lockedValue%0A%20%20%20%20%20%20tokens%20%7B%0A%20%20%20%20%20%20%20%20tokenId%20%7B%0A%20%20%20%20%20%20%20%20%20%20symbol%0A%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%0A)?

The `null` for where `OCEAN` should be is a bug though https://github.com/oceanprotocol/ocean-subgraph/issues/13

kremalicious:
woah, I just tested latest changes and I thought I didn't reload the page yet, the table is now coming in before I can finish blinking, amazing job!

giphy

kremalicious:
And @KatunaNorbert is right, ordering by the actual locked value is not possible, but it should be. In cases like that please always open an issue on the subgraph. We control it so no need to accept its current state https://github.com/oceanprotocol/ocean-subgraph/issues/14

mihaisc:
But not a blocker

kremalicious:
And the data for last column is not right, they all show the total locked value, but should show the user's liquidity value:

Screen Shot 2021-02-02 at 16 44 25

mihaisc:
I think there are some inconsistencies in the data again
image
Global Station : 
image
Polar caps: 
image

KatunaNorbert:
> Agree, we add `datatokenReserve` and `oceanReserve` to `poolId` in the query and we have that information. Or we can use `lockedValue` ( as well in `poolId` , value is in ocean) and convert it in the currency that the user selected in the preferences (top right)

Ok, I might used the wrong field for that. From here I understood that I should use the `lockedValue`, so I used that and converted it to the currency selected by the user using the `conversion` component. What field should I use for user liquidity?

kremalicious:
good question. The data source really is flaky, remember stumbling upon the code for that over in https://github.com/oceanprotocol/ocean-subgraph/pull/6#pullrequestreview-576367167

Regardless it seems `poolId.lockedValue` is always the total locked value

kremalicious:
And I have a lot of those balances. Think this was all because of our weird liquidity removal slider back then:

Screen Shot 2021-02-02 at 18 15 45


Might be safe to add `balance_gt: 0.001` within the `where` filter on `poolShares`

src/components/pages/History/PoolShares.tsx
KatunaNorbert: yep, good idea, I tried to find the symbol in the graph but it wasn't looking that deep. About the problem with the null, using the graph there are always 2 tokens returned and one of them has a `null` `tokenId`

KatunaNorbert:
Yes, that can be done but I'm not sure about that. I mean even if the share is bellow `0.001`, It is still my share and want to see it. I think we should find a way of showing it, maybe `<0`

kremalicious:
good point but it's kinda our tech's fault. I agree in general, if there is anything left in wallet we should show it but not in this case. People's intention is to remove everything, but they can't, we always leave something if I understood this correctly. So showing all those shares users intended to get rid off would be confusing, and we would need to do some work on not using our number component in this place

mihaisc:
The reality is that those shares are not usable, they can't remove them (don't remember if it's our fault or if it's just how the pools work). Showing them might be more confusing, people will ask why can't they remove them.

mihaisc:
Should we block this issue? We can release it in it's original form (show pool liquidity, not user). We don't really have a estimation on the subgraph fixes so in the meantime this can be useful for people.

mihaisc:
The changes on the graph can't be done, so the solution would be the following :
We can do something similar to the [ pool liquidity](https://github.com/oceanprotocol/market/blob/a951f9bf3bf15cda0c7df338819e842315b65c57/src/components/organisms/AssetActions/Pool/index.tsx#L195). We add in the query `oceanReserve`, `datatokenReserve`, `totalShares`, `spotPrice`  to `poolId` and then we just calculate the liquidity. We can use the weight as well but what i suggested is easier to fetch 
Pagination is done client-side anyway , we can do sorting as well on the table headers (not sure if this component has this feature out of the box)
```
ocean = ( balance / totalShares ) *oceanReserve
datatokens = ( balance / totalShares ) *datatokenReserve
totalLiquidity = ocean + datatoken * spotPrice
```

KatunaNorbert:
Screenshot 2021-02-04 at 15 04 50
Screenshot 2021-02-04 at 15 06 50
Still not the most accurate but definitely better

mihaisc:
What do you mean by not accurate?

KatunaNorbert:
Yes, in the first image I calculated it with your formula. I checked the values.
This is inside the poolShares
Screenshot 2021-02-04 at 16 13 09

And this inside the pool liquidity
Screenshot 2021-02-04 at 16 12 31

The differences are in `oceanReserve` and the more semnificative one in `spotPrice`.

KatunaNorbert:
I was looking into sorting the table by `userLiquidity` and it is quite tricky. The table we use looks into the row attributes and can be sorted based on those. The problem is we calculate the `useLiquidity` on each row, so we don't have it as an attribute therefore it doesn't know how to sort by that. One option would be to calculate the `userLiquidity` before and add it as an attribute, but It will take more time to load the table.

mihaisc:
OceanReserve is not the same as creatorOceanBalance. The difference in price is big, we need to see why

KatunaNorbert:
Screenshot 2021-02-05 at 12 48 47
Screenshot 2021-02-05 at 12 48 56
I think I found the problem. It should be `consumePrice` on query instead of `spotPrice`. Now it is accurate.

kremalicious:
amazing progress in here, good detective work! On quick look calculating user liquidity before rendering table should be fast™ since it calculates on existing values and doesn't fetch again? But guess we only know once we test it, so feel free to push whatever is there then so we can then decide if it's too much overhead with that intended sorting.

Also merging `main` in here again would be handy, cause there's a new table on History page too so we should test this all together

KatunaNorbert:
-Fixed userLiquidity value
-Sorted table descending based on userLiquidity
For me the sorting it's fast but I don't have too much pool shares.

kremalicious:
quick testing: it's as fast as before so think that sorting is all good!

Amazing improvement as is. But ideally, we only leave a column `Your Liquidity`, where in that same column we show a bit smaller the OCEAN & datatoken this represents, so the user's ocean & datatoken reserve in the pool [as Uniswap does](https://github.com/oceanprotocol/market/pull/360#issuecomment-770909527). But only if we do not have to do more crazy calculations

KatunaNorbert:
So we combine `DATATOKEN` and 'VALUE ' columns into 'YOUR LIQUIDITY', and show the value in the currency chosen by the user and bellow that the value in oceans and value in datatoken?

kremalicious:
Sorry wasn't clear, what I meant was combining what is currently `YOUR POOL SHARES` & `VALUE` into one `YOUR LIQUIDITY` column. And then like you said, so we end up with something like this (that's from Uniswap) for this `YOUR LIQUIDITY` column:

Screen Shot 2021-02-01 at 15 42 50

src/components/pages/History/PoolShares.tsx
kremalicious: found that too and reported here oceanprotocol/ocean-subgraph#13 but not blocking cause we work around it in here

kremalicious:
small visual thing I found affecting all our data tables:

when hovering the column headers, they turn black, which is weird especially in dark theme:

Screen Shot 2021-02-05 at 14 40 43

Guess some built-in affordance in the table component we use which also allows changing sort order when clicking the column headers but we're not using this functionality here. So some style needs to be overwritten, column headers should never do anything on hover unless we add sorting option

mihaisc:
> Sorry wasn't clear, what I meant was combining what is currently `YOUR POOL SHARES` & `VALUE` into one `YOUR LIQUIDITY` column. And then like you said, so we end up with something like this (that's from Uniswap) for this `YOUR LIQUIDITY` column:
> 
> Screen Shot 2021-02-01 at 15 42 50

Well, we already have all this info to calculate the total liquidity of the user so i think it's just a matter of displaying it in a pretty way in the collumn

mihaisc:
> Screenshot 2021-02-05 at 12 48 47
> Screenshot 2021-02-05 at 12 48 56
> I think I found the problem. It should be `consumePrice` on query instead of `spotPrice`. Now it is accurate. Just have to find a way of sorting the table

that is great!!!! I was really worried that we messed something up with the pricing (again 😢 )

KatunaNorbert:
About the problem with the table header hovering, It seems like both simple and sortable headers are connected by class so if I remove the hover It wold be removed on both. If I want to keep the hover on sortable headers I have to use some JS to set the style, but I think an easier approach is just to overwrite the hover color, maybe we could use `--color-primary`

kremalicious:
Ah interesting. If we're currently not using sortable columns we could overwrite hover color with `--color-secondary` which should be the color used on column headers by default I think. Then there's no affordance on hover visually

mihaisc:
But ideally we would like to use sortable columns (right?)

KatunaNorbert:
Yes I think that's the goal. Meanwhile I tried to display the values in Your Liquidity column, not sure if this is what wanted
Screenshot 2021-02-08 at 16 21 29

kremalicious:
yeah, just like that, great! Some visual tweaks in table: center aligning the cells looks weird with multiple lines so might look better if we top align the cells content in this table.

Also we can remove that approximate symbol in this context, think we recently added some prop to conversion component to easily control that.

mihaisc:
Since we have only 3 columns i think we can add some more info. We can get back in total pool liquidity

KatunaNorbert:
It wasn't any prop for controlling the symbol display inside conversion component, but I added one. So I made the changes you suggested, also brought the pool shares column back. How does it look like?
Screenshot 2021-02-09 at 09 39 50

kremalicious:
giphy-1

This is great! Think we arrived at feature completion. Needs some small padding and spacing adjustments and then it can be shipped.

Hard to describe exactly, quickly mocked up like so:

Screen Shot 2021-02-09 at 08 58 47

- cell top & bottom padding: slightly more spacing, and then needs to be the same measured from top of text to line at top, and bottom of text to line at bottom
- in `YOUR LIQUIDITY` column all text lines should have same spacing between them, they belong together so visually they should be together

kremalicious:
we should also do a whale check, making sure cells also work with like `$20 000 000.00`

KatunaNorbert:
How about now, should there be more spacing?
Screenshot 2021-02-09 at 10 13 49

I think it looks fine with big values too
Screenshot 2021-02-09 at 10 18 22

kremalicious:
cool, already more consistent. Exact spacing depends a bit on context so it stays consistent to other tables. We could use the space between bottom of column header text and line as a guideline. So then the cell padding in this table should be a bit more like that. If that makes this table too spacious then, alternative would be to reduce the column header padding everywhere.

sidenote: forgot that the approximation symbol switch was in https://github.com/oceanprotocol/market/pull/353

mihaisc:
Instead of your pools shares i think it would have a lot more value to have `Pool liquidity` , in the same format as `Your liquidity`. `x pool shares` has no value for you.

KatunaNorbert:
Do you mean to show the same fields as in the `Your liquidity` column but for the entire pool?

mihaisc:
Yep

KatunaNorbert:
> cool, already more consistent. Exact spacing depends a bit on context so it stays consistent to other tables. We could use the space between bottom of column header text and line as a guideline. So then the cell padding in this table should be a bit more like that. If that makes this table too spacious then, alternative would be to reduce the column header padding everywhere.
> 
> sidenote: forgot that the approximation symbol switch was in #353

The space between the header and the line bellow it is bigger in all the tables, it is done by using a `min-height` which is bigger for the header cell, but that can be changed so both will have the same height.

I think we can use my changes for the approximation symbol if this will be merged first.

kremalicious:
> Do you mean to show the same fields as in the Your liquidity column but for the entire pool?

> Yes

oh, also didn't get that. I mean can't hurt showing total pool liquidity although it might make table look rather busy. If we do it, then `YOUR LIQUIDITY` should come first cause that's the most important info for user.

As for spacing, feel free to push what you have then, would rather try to play around in browser than trying to describe it. Cause pretty sure we're running with some defaults in some tables and this was never adapted so hard to describe a measurable baseline.

Approx symbol: yup, fine having that change in here too. Almost inclined to remove that altogether so we save some logic but we don't need to do that in here

KatunaNorbert:
Do you know if I can get the  creator pool tokens inside  the `poolShares` query, or a more efficient way than getting the `ddo` compute the `shareId` and get It from another query using that?

mihaisc:
You don't need pool creator shares, you have  `totalShares`, `oceanReserve`, `datatokenReserve` on `poolId` in the query

KatunaNorbert:
Right, I misunderstood something

kremalicious:
just playing around and yes, it seems not that easy getting everything in one query. So if that all turns out to be too ineffective, it's still an option to leave total liquidity out. Like, `oceanReserve` on `poolId` is the user's reserve but in this context couldn't figure out a way to also get total ocean reserve

.husky/_/husky.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
```
kremalicious, line 1: this should not be here. In `main` this file is ignored via `.gitignore`

mihaisc:
Shouldn't `DATA SET` and `DATATOKEN` be aligned? The data set name is a few px lower, it's wierd.
image

kremalicious:
haha yeah, ignore me for The Graph queries, I just started learning them :-) Good explanation, thanks!

Also having that lower title in Chrome on macOS:

Screen Shot 2021-02-09 at 17 15 23

Removing the `line-height` from `.AssetListTitle-module--title` fixed it for me in browser

kremalicious:
And after some playing, it looks most balanced and less busy if we use `--color-secondary` like this, so only the big fiat values stay prominent. This also visually aligns everything better where the dark/white colors are all in one line:

Screen Shot 2021-02-09 at 17 21 18

Screen Shot 2021-02-09 at 17 21 46

Using `!important` for this somewhere would be totally fine here

KatunaNorbert:
That's what I was trying to do, I'm also using macOS but it was fine for me, how about now?

kremalicious:
Still lower title for me with 6e0953c change, those selectors either need to be more specific as it looks like or `!important` I guess. We don't want to figure out right now why exactly the asset list title has a line-height defined and if we could remove it in general from there.

The difference in rendering might be based on if any of us has Sharp Sans installed. The desktop font version behaves a bit different than the webfont version which is often a pain.

And after testing with more pool shares: let's limit them to `5` per page, they make page quite long with `10`

src/components/pages/History/PoolShares.tsx
```diff
@@ -1,86 +1,188 @@
...
 }
 
-function TotalLiquidity({ ddo }: { ddo: DDO }): ReactElement {
-  const { price } = useMetadata(ddo)
-  const totalLiquidityInOcean = price?.ocean + price?.datatoken * price?.value
+function Symbol({ tokens }: { tokens: PoolSharePoolIdTokens[] }) {
+  return <>{findValidToken(tokens)}</>
+}
 
+function YourLiquidity({ row }: { row: Asset }) {
```
kremalicious, line 68: just a small architectural thing we don't necessarily need to bother with right now to not add more scope but would remove duplication and is more of a React-y-way: `YourLiquidity` and `PoolLiquidity` components should be just one component e.g. `Liquidity` with a `type="user" and `type="pool"` prop. Based on that inside of components we switch values based on that. Cause both have the exact same markup, and only the value change

kremalicious:
All is perfect now, good job!

But: I have problems getting pool shares from Rinkeby now on https://market-git-feature-get-pool-shares-using-graphql.oceanprotocol.vercel.app/history

I think this already has a newer subgraph deployment which is not on mainnet yet. Which might imply new subgraph breaks our PR here, checking...

src/components/pages/History/PoolShares.tsx
```diff
@@ -1,86 +1,176 @@
...
+    poolShares(where: { userAddress: $user, balance_gt: 0.001 }, first: 1000) {
+      id
+      balance
+      userAddress {
+        id
+      }
+      poolId {
+        id
+        datatokenAddress
+        lockedValue
```
kremalicious, line 27: ok, this was changed into `valueLocked`. This right now will only work against Rinkeby where latest subgraph is deployed

KatunaNorbert: right, this fixed it, thanks :)

KatunaNorbert: Also sometimes I don't get the Published assets on my history page

kremalicious: 99.5% sure this was the only change affecting us here. The deployment of latest subgraph to mainnet will be started now, and once that's finished this PR can be shipped

kremalicious: aaaand it's deployed :-) So should now also work against mainnet

KatunaNorbert: I really see the need of a pagination for the published assets now that I have more of that

mihaisc: it takes some time to sync, wait for it

kremalicious:
tests and deployment failed because they ran before mainnet subgraph was deployed. All restarted and should be good then

kremalicious:
and we need to wait for the new subgraph to sync. Apparently this takes about 2 hours... Until then the pool share numbers will not make sense against mainnet, they should become correct once the node is synced

kremalicious:
All pool shares data now makes sense for me on mainnet, so looks like the graph node finished syncing.


giphy-3

.husky/_/husky.sh
kremalicious: wait, forgot about this, this should not be part of the PR. So deleting file and commit the deletion

kremalicious (approved):
awesome, final thing is faster than any other pool shares lists out there.

PR #361, opened 2021-02-02
End-to-end integration testing

### LINKED ISSUE #128: add pragmatic testing
We still have the test setup leftovers in the app so we should make them work again. So components can either have a Storybook story, or a dedicated unit test with `react-testing-library`.

So the current `npm test` _just_ needs to succeed.

Ideally we get full unit testing coverage, and ave additional integration test runs.

After this is all setup, the branch protections should be updated to not allow any regression of coverage.

Better alternative: add integration testing with Cypress as we had in Commons for the very basic user flows:

- publish
- search
- consume

This is way more valuable to catch errors

kremalicious:
_Highly WIP..._

* use [Cypress](https://www.cypress.io) for easy scripting of end-to-end testing
* setup and isolate under `tests/integration`
* GitHub Action setup: run barge, run integration test
* new npm scripts

closes #128


DimitarSD:
The custom provider is working (resolve some of the previous errors) only when we use `@walletconnect/web3-provider`. I think it's because we use web3modal for connecting. In the other cases (using different providers), the provider is created and attached to `window.ethereum` successfully. However, it fails during `web3modal.connect()`. Still, in all cases (even that there is a working provider attached to `window.ethereum`) the reconnecting wallet option never passes, it keeps spinning.

kremalicious:
Superseded by #799

PR #363, opened 2021-02-03
Feature/consume history

### LINKED ISSUE #126: add Consume history to History page
We need a list of all the consumed assets of a user. We can get the order list from the graph

claudiaHash:
Fixes #126 

Changes proposed in this PR:

-  user's orders history table created and displayed in History


mihaisc:
Looks good, remove the title column. 
We removed some auto renaming of the classes generated by apollo so you need to fix this ` has no exported member named 'OrdersDataTokenOrders'. Did you mean 'OrdersData_tokenOrders'`
The big question now is should we leave this as a table with 2 columns or change it to a tile grid like publish (both will be paginated when the other pr is done)

claudiaHash:
#126 should be linked already, isn't it showing?
What about adding a 'Datatoken' column instead of title? 
Screenshot from 2021-02-03 13-50-08

mihaisc:
I'm sorry, didn't see the issue in the description and i presumed it's not linked, my bad.
Sure, `Datatoken` may have some value

kremalicious:
This works great! 

I would at least switch the Compute Jobs & Orders tables since almost nobody will have anything listed in Compute Jobs.

And for now the table title could be `Downloads` which makes it more clear what this table is.

mihaisc:
You can't show the price, you don't know when the dt was bought and when the asset was consumed. Basically i can buy from the pool 10 dt and after a month i use 1 dt to consume the asset. What is the price? The price at the moment of consumption or when i actually bought the dt? Furthermore you can't actually link the event in the background, only lets say by time proximity.

kremalicious:
good points. So then I think this PR is ready-to-go after we remove the Orders.module.css file. If we don't need any further styling we can always leave out css files for a component

claudiaHash:
done! sorry i forgot about that

kremalicious (approved):
perfect! Tweet is prepared so can be merged once checks are green

PR #364, opened 2021-02-03
added entry for provider inside publish asset form

### LINKED ISSUE #75: Add custom provider endpoint during publish
In Ocean Market, OPF has to run Provider and therefore holds the data service decryption key. This means that OPF has custody of these keys.

What to do for this issue: 
- [ ]  In Ocean Market, give an option for the user to provide an endpoint of the Provider. Therefore the user can run their own Provider, and retain decentralization. We can expect enterprises and others to do this. Handled in https://github.com/oceanprotocol/market/issues/702
- [ ]  In Ocean Market, if the user chooses to use the OPF-run Provider, have a GUI affordance to set the expectation that this Provider will only be run until some specific cut-off date: Dec 31, 2021. This ensures that OPF is not bound to some weird long-term commitment that it didn't mean to make. Handled in https://github.com/oceanprotocol/market/issues/703
- [x] Allow external provider in ocean-lib-js [Issue](https://app.zenhub.com/workspaces/ocean-v30-5e5e5a97e91ce03f380fb9b7/issues/oceanprotocol/ocean-lib-js/278)
- [x] Allow external provider in ocean-lib-py [Issue](https://app.zenhub.com/workspaces/ocean-v30-5e5e5a97e91ce03f380fb9b7/issues/oceanprotocol/ocean-lib-py/49)

Note: in the future we will want more decentralized storage of private keys. See [related tech spike](https://app.zenhub.com/workspaces/ocean-v30-5e5e5a97e91ce03f380fb9b7/issues/oceanprotocol/tech-spikes/99).

Note: moved from [this issue](https://app.zenhub.com/workspaces/ocean-v30-5e5e5a97e91ce03f380fb9b7/issues/oceanprotocol/multi-repo-issue/18)

KatunaNorbert:
Fixes #75 .

Changes proposed in this PR:

- add entry for custom provider endpoint when publishing asset

### DIFF
--- a/content/pages/publish.json
+++ b/content/pages/publish.json
@@ -35,6 +35,13 @@
         "type": "files"
       },
       {
+        "name": "providerUri",
+        "label": "Provider",
+        "placeholder": "e.g. https://myproviders.com/provider",
+        "help": "You can provide an endpoint to your own provider",
+        "type": "url"
+      },
+      {
         "name": "access",
         "label": "Access Type",
         "help": "Choose how you want your files to be accessible for the specified price.",
--- a/src/@types/MetaData.d.ts
+++ b/src/@types/MetaData.d.ts
@@ -31,6 +31,7 @@ export interface MetadataPublishForm {
   files: string | File[]
   author: string
   timeout: string
+  providerUri: string
   dataTokenOptions: DataTokenOptions
   access: 'Download' | 'Compute' | string
   termsAndConditions: boolean
--- a/src/components/pages/Publish/index.tsx
+++ b/src/components/pages/Publish/index.tsx
@@ -42,6 +42,8 @@ export default function PublishPage({
     const metadata = transformPublishFormToMetadata(values)
     const timeout = mapTimeoutStringToSeconds(values.timeout)
 
+    console.log(values)
+
     const serviceType = values.access === 'Download' ? 'access' : 'compute'
 
     try {
@@ -49,7 +51,9 @@ export default function PublishPage({
         'Publish with ',
         metadata,
         serviceType,
-        values.dataTokenOptions
+        values.dataTokenOptions,
+        timeout,
+        values.providerUri
       )
 
       const ddo = await publish(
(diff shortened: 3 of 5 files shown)


KatunaNorbert:
I'm not sure what the second checkbox from the epic #75 description means. If the user sticks with OPF-run Provider there should be a message to inform him how long can he use this provider?

kremalicious:
This is more tricky than the issue suggests, and of course the issue does not describe all the tricky parts, sorry. 

1. It is an advanced use case and the majority of our users will be confused by this. Which implies we should put it behind some progressive disclosure pattern, like an `Advanced` section at end of form closed by default.
2. Honestly, right now, I do not want to make users believe they have to run their own provider. There are no instructions whatsoever how to do that. The provider repo still says it is in beta https://github.com/oceanprotocol/provider
3. When using a custom provider it has implications on https://github.com/oceanprotocol/market/pull/352 which then needs to be made to use that custom provider. This would then imply we need to ask for the custom provider right at the start of the form before the file input. But again, it is an advanced use case and putting it prominently for all users is not really justifiable

KatunaNorbert:
Ok, this made it more clear, thanks. We can add it as you suggested inside an `Advanced` section which will be closed by default. Then if the user gives his own provider we can remove the value from fileInfo field and ask him to add it again using his own provider. With this we can also check that the provider is working before submitting the form.

kremalicious:
All great ideas! Just not sure the use cases we want to have it for justify all the work. Maybe a more pragmatic approach would be to just put that field as added in this PR behind some feature flag, which we could trigger based on some url param and be done with it. This would cover the enterprise use case where we work with them together closely and can provide them with a custom link, something like:

`https://market.oceanprotocol.com/publish&customProvider=true`

This would save us dealing with all those form & UI checks and we do not confuse majority of our users

mihaisc:
Don't forget that we didn't even consider very well what implications this will have in the rest of the ui, and if the current workflows will work ( in theory yes, but...)

kremalicious:
closing for now until an actual need arises for majority of our users or we decide to have some more advanced publish fields

PR #368, opened 2021-02-08
Fix/url encoding categories tags

### LINKED ISSUE #231: URL Encoding issue in categories and tags
If a tag or category for an asset contains an ampersand (&), when creating links in the market, characters needs to be url encoded.

For example, in Rinkeby Test Network:
https://market.oceanprotocol.com/asset/did:op:46D82feFC506eDD1831c11EA91fA3149179c4465

The category is "Agriculture & Bio Engineering", but if you click on the category near the top, you are redirected to:
https://market.oceanprotocol.com/search?categories=Agriculture%20&%20Bio%20Engineering
instead of
https://market.oceanprotocol.com/search?categories=Agriculture%20%26%20Bio%20Engineering
leading to different results.

Same thing with the tags field.

Other characters may need to be encoded as well.

claudiaHash:
Fixes #231  .

Changes proposed in this PR:

- fixed url encoding for tag
- fixed url encoding for search value

Did not find the category, as mentioned in issue description. The only place where I can see them is in DDO JSON, activating the "Debug Mode"

### DIFF
--- a/src/components/atoms/Tags.tsx
+++ b/src/components/atoms/Tags.tsx
@@ -12,12 +12,12 @@ declare type TagsProps = {
 }
 
 const Tag = ({ tag, noLinks }: { tag: string; noLinks?: boolean }) => {
-  const encodedTag = encodeURIComponent(tag)
+  const urlEncodedTag = encodeURIComponent(tag)
   return noLinks ? (
     <span className={styles.tag}>{tag}</span>
   ) : (
     <Link
-      to={`/search?tags=${encodedTag}&sort=created&sortOrder=desc`}
+      to={`/search?tags=${urlEncodedTag}&sort=created&sortOrder=desc`}
       className={styles.tag}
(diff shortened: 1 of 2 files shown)


kremalicious:
Categories do not matter in our market UI anymore, we never allowed them to be set in the UI but were still displaying them when users were pushing data sets from outside of the market. But by now we also removed that to keep it simple with only tags.

kremalicious:
not sure how to fully test this but just following the flow of the `tags=` value, I'm wondering if we need to URL decode that in some place. It then hits this line and is run through `query-string` package https://github.com/oceanprotocol/market/blob/main/src/components/templates/Search/index.tsx#L23

kremalicious:
all good, should work. Package sets a `decode` option by default https://github.com/sindresorhus/query-string#api

kremalicious (approved):

PR #369, opened 2021-02-09
Fix development with Barge

kremalicious:
prerequisite for #361

- new revised instructions which actually work
- new copy contracts script adapted from https://github.com/oceanprotocol/commons/blob/master/scripts/keeper.sh
- read out and inject contract addresses from locally generated `address.json`
- also pass `oceanTokenAddress` from that `address.json`
- workaround for `chainId` mismatch between MetaMask and barge, which was the main issue breaking the app
- fix providerUri switching in file input


mihaisc:
I get this (you mentioned this on slack) 
image

Other than this, seems good. Can't really test because this breaks the site 😄

mihaisc (approved):

kremalicious:
yup, I even created hacks and workarounds for this but in the end was caused by older elasticsearch database in stale Docker container. So just added the docker pruning note in readme and called it a day

PR #375, opened 2021-02-11
Added link to highest liquidity pool search

### LINKED ISSUE #370: Link to highest liquidity pool search on front page
Just like we link to `ALL DATA SETS` at the end of the front page, we should have a similar link below the `Highest Liquidity Pools` grid, linking to [`/search?sortOrder=desc&sort=liquidity&priceType=pool`](https://market.oceanprotocol.com/search?sortOrder=desc&sort=liquidity&priceType=pool), labeled `All Pools` or `All Data Set Pools`

Screen Shot 2021-02-09 at 21 39 47

KatunaNorbert:
Fixes #370  .

Changes proposed in this PR:

- Link to highest liquidity pool search on front page

### DIFF
--- a/src/components/pages/Home.tsx
+++ b/src/components/pages/Home.tsx
@@ -111,7 +111,7 @@ export default function HomePage(): ReactElement {
             style="text"
             to="/search?priceType=pool&sort=liquidity&sortOrder=desc"
           >
-            Highest Liquidity pools →
+            All data set pools →
           </Button>
         }
       />
--- a/src/components/pages/Home.tsx
+++ b/src/components/pages/Home.tsx
@@ -106,6 +106,14 @@ export default function HomePage(): ReactElement {
       <SectionQueryResult
         title="Highest Liquidity Pools"
         query={queryHighest}
+        action={
+          <Button
+            style="text"
+            to="/search?priceType=pool&sort=liquidity&sortOrder=desc"
+          >
+            Highest Liquidity pools →
+          </Button>
+        }
       />
 
       <SectionQueryResult
--- a/src/components/pages/Home.tsx
+++ b/src/components/pages/Home.tsx
@@ -104,20 +104,20 @@ export default function HomePage(): ReactElement {
       </section>
 
       <SectionQueryResult
-        title="Highest Liquidity Pools"
+        title="Highest Liquidity"
         query={queryHighest}
         action={
           <Button
             style="text"
             to="/search?priceType=pool&sort=liquidity&sortOrder=desc"
           >
-            All data set pools →
+            All data sets with pool →
           </Button>
         }
       />
 
       <SectionQueryResult
-        title="New Data Sets"
+        title="Recently Published"
         query={queryLatest}
         action={
           <Button style="text" to="/search?sort=created&sortOrder=desc">


KatunaNorbert:
I'm thinking of the right name for the link, and I came up with `SEARCH HIGHEST LIQUIDITY POOLS` , but this sound a bit confusing since the section title says `HIGHEST LIQUIDITY POOLS` , why would search on that if there are already assets shown. So maybe something like `ALL ASSETS  ORDERED BY HEIGHEST LIQUIDITY POOLS` would be more clear, but It seems too long. Do you have any suggestions?

kremalicious:
also wasn't sure, `All Data Set Pools` made most sense to me considering the context with the other link on the page for `All Data Sets`. There can be data sets (at least in UI we never use _asset_ but always _data set_) with and without pool so this seems most precise, since the link also filters data sets to only those with a pool

KatunaNorbert:
yeah, I think that's better

mihaisc:
`All Data Set Pools` implies you display  all the pools, not the data sets. Maybe `All Data Sets with Pool`

KatunaNorbert:
I think the section title Is confusing too, it says `Highest Liquidity Pools` it impose that you know there will be data sets displayed, on the other hand in the section bellow it, there is specified that there are data sets displayed `New Data Sets`. I think there should be consistent, we either put `Data Sets` in both titles or we remove it.

kremalicious:
So far in our UI browsing pools and browsing data sets is kinda the same. But we have the space, so being more precise is fine with `All Data Sets with Pool`. Or `All Pool Data Sets`? Any other variations of 4 words we can come up with? :-)

kremalicious:
we could remove `Pools` from `Highest Liquidity Pools`. Think just `Highest Liquidity` as title still makes it clear, and then the new link describes it more

KatunaNorbert:
and then how about changing also the `New Data Sets` to `Newest` or `Recently added`, to remove Data Sets from that title?

kremalicious:
yes! `Recently Added` or `Recently Published` should be good. Removing data set from its title nicely makes it consistent across whole page

mihaisc:
And the link can be a simple `All` , or `All items`

KatunaNorbert:
For me `All Data Sets` it's more consistent since we have `All Data Sets with Pool` above. Also as @kremalicious said the link describe it more if we keep the title simple

mihaisc:
I meant for all of them, the link is in the component context, so it's implied what items you will get

KatunaNorbert:
Ok, I see. That would make the page more clean, but In this case I think the title should be more suggestive. With the titles discussed for me it's not clear that if I click the `All` link bellow `Highest Liquidity` I would get only the data sets which has pools, I would expect to get all the data sets.

kremalicious:
good points @KatunaNorbert. While it's hard to argue against Amazon cause they probably tested this like crazy, the difference in our case is that title & link can be far apart like on small viewports. So having more precise link titles as [those suggested ones](https://github.com/oceanprotocol/market/pull/375#issuecomment-778096406) would help with that.

kremalicious (approved):
Perfect. Naming is always the hardest part :-)

Issue #374, opened 2021-02-10
Reorganize History page


kremalicious:
At the moment the history page is a place to put all kinds of lists creating a very busy page, visually, interactively, and performance-wise. Some quick fixes like #351 we can do but we should think about a reorganization of the whole page.

A first goal is to come up with a good way to make it more focused before starting any implementation, some quick suggestions could be, more welcome!:

- split up different lists into different screens
- via sub-menu, or tabs
- reorganize lists across whole app to maybe put them more in context, e.g. does it make sense moving the pool shares to the front page?
- have some lists in some "global" UI, like a button in main menu 

What should be considered:
- performance: adding stuff to front-page has implication on Time to Interaction, which is really important on front-page
- layout shifts: empty/loading tables vs. filled tables create a lot of layout shifts which should mostly be either prevented or made sure they don't affect too much on one screen

PR #377, opened 2021-02-11
added pagination to Published list, used aquarius to get assets

### LINKED ISSUE #351: "Published" section in History needs to be paginated

KatunaNorbert:
Fixes #351  .

Changes proposed in this PR:

- add pagination to published list


KatunaNorbert:
I used aquarius to get the published assets because ocean doesn't support pagination

kremalicious:
functionality-wise this is perfect, nice that in-page pagination on that component now works. Opens up way more possibilities, like on the front-page.

But UX/UI-wise we now have some new problems:

1. Pagination UI inconsistencies 

This becomes most obvious when using AssetList and our Table component on one page. I know they are 2 different components but to a user, elements doing the same thing should look and behave the same.

Screen Shot 2021-02-12 at 10 28 57

So in this case we should visually & interactively unify both pagination components as much as we can and find a compromise between the functionality of the 2, so we can visually unify them.

2. Layout collapse

When clicking a new page number whole list collapses and comes back. That's especially a problem on mobile. Not sure we can do something about this since we do a server fetch, and setting fixed heights is probably not an option. There might be ways where we keep current list for as long as we have the new assets after the fetch, and only then replace the list.

Another quick fix is to reduce the list from displaying 9 to 6 data sets.

KatunaNorbert:
Yes, I agree. But I saw that there is an issue #374 about reorganizing the history page, so maybe the assets would be separated from the tables, and then it would be more consistent.  
About Layout collapse, I'll try to find something, maybe loading only if there is no data at all which is when opening the page

mihaisc:
The data-table can get a `paginationComponent` so maybe we can push the same pagination component or a component that looks the same but has the requirements : 
```
a component that overrides the default pagination component. It must satisfy the following API: 
{ rowsPerPage: PropTypes.number.isRequired,
rowCount: PropTypes.number.isRequired,
 onChangePage: PropTypes.func.isRequired,
 onChangeRowsPerPage: PropTypes.func.isRequired,
 currentPage: PropTypes.number.isRequired };
```

kremalicious:
Would be fine using a unified minimal pagination like UniSwap has in some places:

Screen Shot 2021-02-15 at 06 11 39

But main problem in making that pagination not only look but also work consistently is our data table component counting items, and not pages of items as our `AssetList` does. The [pagination options](https://github.com/jbetancur/react-data-table-component#pagination) do not mention any total pages or something so think we're stuck with this.

We can at least unify some styling with what we have so we're not blocked by this, like using the same arrows, making AssetList pagination slightly smaller, and without the signal color. On data-table pagination removing the last/first page icons. I'll also tinker around a bit

kremalicious:
This is as far as I can get without creating a custom pagination component for data table:

Screen Shot 2021-02-15 at 06 54 43

Centering both kinda only works if we figure out how to move the text between the arrows in the data table pagination. But [rebuilding this whole thing](https://github.com/jbetancur/react-data-table-component/blob/master/src/DataTable/Pagination.js) only for that seems overkill.

So tending towards just leaving it like this. Since we can't unify both pagination's functionality, it's justify they look a bit different but now look less out-of-place when used together. I'll push what I have right now:

Screen Shot 2021-02-15 at 07 09 37

KatunaNorbert:
Nice. I was trying to create a custom pagination.
Screenshot 2021-02-15 at 09 19 55
I'm just struggling to make the `onChangePage` function to work

KatunaNorbert:
Ok, so the custom pagination works. Should I use that or we keep the changes made by @kremalicious ?

kremalicious:
oh nice! Then I would say we need a combination, where the styling for our `Pagination` is from my changes (makes sense toning the pagination down a bit everywhere so it can be used in more contexts), but functionality is from your changes with the custom page number pagination on data tables too.

Is the shown data table pagination our reused `Pagination` or a rebuild? Just checking cause our `Pagination` has some heavy truncating going on depending on amount of page numbers

KatunaNorbert:
Well It is the same component but rebuild to match the table needs. But I think they can be combined in 1 component, I'm just working on that and I'm almost done

KatunaNorbert:
This is how it turned out with both modifications combined
Screenshot 2021-02-15 at 13 29 24

kremalicious:
yay, almost there, final style change would be to add this border back for the last table row so we end up like this:

Screen Shot 2021-02-16 at 10 55 00

So that then the space between top of text of pagination to border, and top of text of pagination to beginning of AssetList grid is kinda the same

src/components/molecules/PaginationTable.tsx
```diff
@@ -0,0 +1,61 @@
+import React, { useState, useEffect, ReactElement } from 'react'
```
kremalicious, line 1: is this new component in use anywhere? Couldn't quickly see it being used

src/components/pages/History/PublishedList.tsx
```diff
@@ -4,33 +4,60 @@ import { useOcean } from '@oceanprotocol/react'
...
+  const [page, setPage] = useState<number>(1)
 
   useEffect(() => {
     async function getPublished() {
-      if (!accountId || !ocean) return
-
+      if (!accountId) return
+      const queryPublishedAssets = {
+        page: page,
+        offset: 9,
```
kremalicious, line 23: good thinking migrating `ocean.assets.ownerAssets` to Aquarius query. If we can't easily deal with the layout collapse on page change, I would then limit the displayed items to `6` here so only 2 rows on desktop are shown at once. Just slightly improves the layout collapse

src/components/molecules/PaginationTable.tsx
KatunaNorbert: It isn't just leaved it there until we sure to use the same component for both. But I'll remove it

src/components/pages/History/PublishedList.tsx
KatunaNorbert: I found a solution for that, basically I don't show the loading icon between page changes so It doesn't collapse. Only show the loading on page mount. Not sure if trading the loading icon for that is a good idea, what do you think?

kremalicious: ah, sorry, should have checked out preview, this is great as is!

src/components/pages/History/PublishedList.tsx
```diff
@@ -4,33 +4,60 @@ import { useOcean } from '@oceanprotocol/react'
...
+        setQueryResult(result)
       } catch (error) {
         Logger.error(error.message)
       } finally {
         setIsLoading(false)
       }
     }
     getPublished()
-  }, [ocean, status, accountId])
+  }, [accountId, page])
```
kremalicious, line 47: this effect should fire when `config.metadataCacheUri` changes too, which usually happens when users switch networks in their wallet. So `config.metadataCacheUri` should be part of this dependency array

KatunaNorbert: right, I'll add that

kremalicious: perfect, switching networks works now for showing different assets. Somewhere we also need to trigger an update to the pagination calculation, that didn't update for me on network change

kremalicious (approved):
great work on solving this in a super DRY way. Feels so nice having just one pagination UI we have to maintain now, but used in all the places

PR #378, opened 2021-02-11
docs: mention Barge + The Graph setup

kremalicious:
Describe how The Graph can be used all locally together with Barge.


mihaisc (approved):
good enough for now. With the changes Alex made to the subgraph I started the node and deployed the subgraph without an issue and without help. After that you just have to change the url in the market config.

PR #384, opened 2021-02-16
Compute: publish algorithm

### LINKED ISSUE #341: Publish algorithm
On the Publish page:

- introduce tab interface: Data Set, Algorithm. Data set is selected by default
- new form for algorithm publishing
   - title
   - description
   - file URL
   - pre-defined Docker image selection: Node.js, Python
   - tags
   - optional checkbox "Keep my algorithm private"
- needs to be synced with editing form
- algorithm is public by default

bogdanfazakas:
Fixes #341  .

Changes proposed in this PR:

- added new form for algorithm publish on publish page


src/components/pages/Publish/index.tsx
```diff
@@ -24,20 +37,29 @@ import Button from '../../atoms/Button'
...
   const [success, setSuccess] = useState<string>()
   const [error, setError] = useState<string>()
+  const [title, setTitle] = useState<string>()
   const [did, setDid] = useState<string>()
   const [publishType, setPublishType] = useState<string>()
 
   const hasFeedback = isLoading || error || success
 
+  useEffect(() => {
+    publishType === 'data'
```
kremalicious, line 58: just small thing, we should keep value names consistent. So in this case this refers to the `type` set in the DDO and for those it is either `dataset` or `algorithm` so we can use that naming here too

src/components/pages/Publish/index.tsx
```diff
@@ -81,16 +103,52 @@ export default function PublishPage({
...
+  async function handleAlgorithmSubmit(
+    values: Partial<AlgorithmPublishForm>,
+    resetForm: () => void
+  ): Promise<void> {
+    const metadata = transformPublishAlgorithmFormToMetadata(values)
+
+    try {
+      Logger.log('Publish Algorithm with ', metadata)
+
+      const ddo = await publish((metadata as unknown) as Metadata, 'access')
```
kremalicious, line 115: this line is totally the tricky part cause we map something else in the UI to something different in our tech stack. This concerns the `optional checkbox "Keep my algorithm private"` part from #341 which implies:

- user sets it to public → we set service type === 'access'
- user sets it to private → we set service type === 'compute'

bogdanfazakas: ok got it, wanted to ask after the planning what 'keep my algorithm private implies', but it all clear now

mihaisc: For now we can have a simple condition  `await publish((metadata as unknown) as Metadata, algoPrivacy ==='public'? 'access': 'compute')`

content/pages/publishAlgo.json
```diff
@@ -0,0 +1,69 @@
+{
+  "title": "Publish",
+  "description": "Highlight the important features of your algorith to make it more discoverable and catch the interest of data consumers.",
```
kremalicious, line 3: might be cleaner to split this up into fixed page content and forms content since this is content which should not change on tab change, and we can minimize maintenance duplication. So a `pages/publish.json` where we only have the page title, description, and warning (warning should go above tabs anyway).

Then either put both forms with different keys in this file (like `formDataset` & `formAlgorithm`) or create like so we end up with files:

- `pages/publish.json`
- `pages/publish-forms.json`

Or it might be already time to create a whole content folder for the publish page since this is constantly extended, where we could use:

- `pages/publish/index.json`
- `pages/publish/forms.json`, or `pages/publish/form-dataset.json` & `pages/publish/form-algorithm.json`

Then the page component just queries for the index file content, and each form can query the exact content they need in their components instead of passing all content as prop.

content/pages/publishAlgo.json
```diff
@@ -0,0 +1,69 @@
...
+  "warning": "Given the beta status, publishing on Ropsten or Rinkeby first is strongly recommended. Please familiarize yourself with [the market](https://oceanprotocol.com/technology/marketplaces), [the risks](https://blog.oceanprotocol.com/on-staking-on-data-in-ocean-market-3d8e09eb0a13), and the [Terms of Use](/terms).",
+  "form": {
+    "title": "Publish",
```
kremalicious, line 6: I think we do not output that form title right now but with 2 forms to switch between, we should do that. So that the form title switches between "Publish a data set" & "Publish an algorithm". And we might want to have a form description under that title too so we could already prepare that. Cause then we can keep the page description more generic, and use those form's title & description to explain the difference between them individually

content/pages/publishAlgo.json
bogdanfazakas: Yes good point, also thought about that, I'll do it as soon as the Custom Image selection is ready.

content/pages/publishAlgo.json
```diff
@@ -0,0 +1,69 @@
...
+        "type": "files",
+        "required": true
+      },
+      {
+        "name": "dockerImage",
+        "label": "Docker Image",
+        "placeholder": "e.g. python3.7",
+        "help": "Please select a predefined image to run your algorithm.",
+        "type": "select",
+        "options": ["NodeJS", "Python 3.7"],
```
kremalicious, line 36: since it's Docker images might be more clear if we reference them in the UI by actual image name in that Docker image syntax, like `node:latest` & `python:latest` or whatever specific images we use

content/pages/publishAlgo.json
```diff
@@ -0,0 +1,69 @@
...
+        "label": "Algorithm Privacy",
+        "type": "checkbox",
+        "options": ["Keep my algorithm private"],
+        "required": false
+      },
+      {
+        "name": "author",
+        "label": "Author",
+        "placeholder": "e.g. Jelly McJellyfish",
+        "help": "Give proper attribution for your algorith.",
```
kremalicious, line 50: small typo:

```suggestion
        "help": "Give proper attribution for your algorithm.",
```

src/@types/MetaData.d.ts
```diff
@@ -39,6 +39,19 @@ export interface MetadataPublishForm {
   links?: string | File[]
 }
 
+export interface AlgorithmPublishForm {
```
kremalicious, line 42: can't think of any naming for both interfaces here making the 2 types consistent without making it super long, like:

- `MetadataPublishFormDataset`
- `MetadataPublishFormAlgorithm`

Or is not too long? On other hand kinda clear that publish form holds metadata so maybe we can get a way with:

- `PublishFormDataset`
- `PublishFormAlgorithm`

but that might collide with our publish forms component naming

bogdanfazakas: I believe using `MetadataPublishFormDataset` and `MetadataPublishFormAlgorithm` will be more consistent and more clear what that form holds, even if it's a bit too long

src/components/pages/Publish/FormAlgoPublish.tsx
```diff
@@ -0,0 +1,88 @@
...
+    initialValues,
+    validateField,
+    setFieldValue
+  }: FormikContextType<AlgorithmPublishForm> = useFormikContext()
+
+  // reset form validation on every mount
+  useEffect(() => {
+    setErrors({})
+    setTouched({})
+    resetForm({ values: initialValues, status: 'empty' })
```
kremalicious, line 32: this might actually lead to problems now with 2 forms behind tabs cause it will clear all entered (and saved in `localStorage`) form values when user switches tabs. And this saving form data and restore on mount is broken in this PR. We haven't fully defined yet how this feature works UX-wise with 2 forms but first guess would be that both forms save their state, and only successfully submitting one of them clears their `localStorage`

mihaisc: why don't we have 2 separate keys for the localState `ocean-publish-form-data` and `ocean-publish-form-algo`

kremalicious: right, that might be the reason it breaks right now, would be great if it's as simple as that to restore functionality

bogdanfazakas:
@kremalicious and @mihaisc, saw that the algorithm object in the asset's service.main has language property, should we split the algorithm file url and get file extension and set it here or from where we could get that so we don't hardcode it ?

see https://github.com/oceanprotocol/ocean.js/blob/main/test/integration/ComputeFlow.test.ts#L265

content/pages/publishAlgo.json
bogdanfazakas: shouldn't we use options like `node:pre-defined` & `python:pre-defined`, or are we going to use the `latest` version of those two every time as pre-defined images?

kremalicious: right, totally, we should use specific versions. Probably whatever is their latest LTS release or however that is called in the Python world

kremalicious:
to keep it pragmatic, we could just switch that `language` based on selected Docker image which at least works for the pre-set images cause we know them. Tricky of course for custom images, but just when this option is selected we could show a `language` selector input.

But of course many assumptions in between but not sure we can solve any of that better without controlling input & output, e.g. with a textarea editor users enter their code so we can detect and transform when needed, or have the /fileinfo endpoint give us back more info

mihaisc:
That is not used as far as i know so ignore it. @alexcos20 am i wrong?

mihaisc:
Not sure how easy it is, but i shouldn't be able to edit image/version/entrypoint for default images.

bogdanfazakas:
Related to the custom docker image input fields (`image`, `version`, `entrypoint`) 
We have two options I guess:

- to hide those fields when a pre-defined image is selected
- or to disable those fields and display them all the time and make them editable when custom image is selected from dropdown  (like in the picture bellow) 
Screenshot 2021-02-18 at 17 34 45

mihaisc:
Considering we can see the image/version/entrypoint in the preview on the right we can also hide them so at this point whatever is the easiest

kremalicious:
would be in favor of progressive disclosure, so for pre-set images we hide the relevant form fields, and show when custom image is selected. Since it seems that input values for pre-set images are reflected in their option name anyway. But no clue how hard (or easy?) that is with current form setup, don't wanna create forms within forms or something just for that

The _Version_ field is a bit confusing from user perspective. Does it reference a language version number or a Docker tag (which happens to be a version number sometimes)? If it's the latter, we should label it accordingly like _Docker Image Tag_. But isn't this then referenced by the DockerHub image url anyway?

bogdanfazakas:
> would be in favor of progressive disclosure, so for pre-set images we hide the relevant form fields, and show when custom image is selected. Since it seems that input values for pre-set images are reflected in their option name anyway. But no clue how hard (or easy?) that is with current form setup, don't wanna create forms within forms or something just for that
> 
> The _Version_ field is a bit confusing from user perspective. Does it reference a language version number or a Docker tag (which happens to be a version number sometimes)? If it's the latter, we should label it accordingly like _Docker Image Tag_. But isn't this then referenced by the DockerHub image url anyway?

Yeah I couldn't decide on that either, but thought that form would be more consistent and we would not have fields flying around when a select input is changed, thats why I went for disabling those fields.

Regarding the `version` field I was a bit confused to 😃  but i believe it references the docker image tag not the algorithm version and we should rename that accordingly

mihaisc:
It's from docker so for oceanprotocol/aquarius:latest , image = oceanprotocol/aquarius , version = latest

mihaisc:
Regarding the validation, we need to validate if the docker image is valid but what about 3rd party repos? Do we just do a http head or options on the url and see if it's 200? We can't do much in that scenario.

kremalicious:
Some UI notes for Docker image selection, which seems to be the major thing to still solve in here: 

- should default to one of our pre-defined images
- let's try to hide _Image URL_, _Version_, _Entrypoint_ fields by default, and show them when _Custom_ is selected
- rename _Version_ to _Docker Image Tag_, and for pre-defined images it should reflect what their labels say
- _Image URL_ should be clearer, if either docker image name (+ tag ?) or DockerHub URLs are allowed it should be labeled like _Docker Image_. But then again if we allow image name + tag in there, the _Version_/_Docker Image Tag_ field becomes redundant. We should be able to get that automatically from image name + tag, or DockerHub URL maybe?

We should then look into #397 since it interferes a bit when using the form but fine doing this individually

bogdanfazakas:
we kind of need the fix for the persistence issue and set corect initial values for the form on each tab because it's a feature break issues for publish algorithm. The default values for the docker image and private algorithm are set using the initial values set in FormAlgoPublish model that is not passed when switching tabs. 
I've created a fix for this issue in this PR #402

bogdanfazakas:
Made a small update on metadata preview and set fixed width for  metadata item content also set word wrap to break word to fix this
Before
Screenshot 2021-03-04 at 18 42 54
After
Screenshot 2021-03-04 at 18 42 32

Another option would have been to update the tabs width but i considered this approach better

With this i think publish is ready to be merged

src/components/organisms/AssetContent/MetaItem.module.css
```diff
@@ -10,3 +10,9 @@
   color: var(--color-secondary);
   text-transform: uppercase;
 }
+
+.content {
+  width: 11.5rem;
```
kremalicious, line 15: double this is wider than most mobile viewports, resulting in scrollbar and generally misaligned things on whole screen:

Screen Shot 2021-03-05 at 11 28 40

Setting fixed widths by default with our mobile first styles can become problematic later on. In this case I guess easiest is to put this width behind a `min-width` media query

kremalicious: for reference: `grid-template-columns: repeat(auto-fit, minmax(10rem, 1fr));` on `.metaFull` and removing that `width` seems to solve all our problems

kremalicious (approved):
amazing work going through all those known and unknown hurdles. Pretty sure we'll iterate on some details further and we now have a good base for the overall publish flow

PR #388, opened 2021-02-17
view algorithms on homepage

### LINKED ISSUE #336: View new algos on homepage
Create a new section on homepage to view new algos and have a link to 'All algorithms'. This is basically the search page with a filter set `service.main.type == algorithm`

### LINKED ISSUE #337: Add algorithms to search page
Add 'data' and 'algorithms' to the filters. This query is based on `service.main.type` from the ddo

claudiaHash:
Fixes #336
Fixes #337

Changes proposed in this PR:

- view new algorithms on homepage
- global color system tweak to create distinction between page background & content
- asset teaser distinction between dataset & algorithm


src/components/pages/Home.tsx
```diff
@@ -35,6 +35,18 @@ const queryLatest = {
   sort: { created: -1 }
 }
 
+const queryAlgorithms = {
+  page: 1,
+  offset: 9,
+  query: {
+    nativeSearch: 1,
+    query_string: {
+      query: `service.main.type == algorithm`
```
kremalicious, line 44: this query seems to not work completely. Works fo me when connected to Rinkeby where algorithms are already there. But  when connected to mainnet, I get assets with `dataset` type shown in this section when it should just return nothing right now. 

And for this query we should exclude purgatory assets with `isInPurgatory` too, like we do for other queries https://github.com/oceanprotocol/market/pull/388/files#diff-1df2f2876c20c90c294da5abb9fcf0812eec3448de2aea5eb20e5e5f4a7d6cc7R20

claudiaHash: I understand, I'm working right now to correct this.

src/components/pages/Home.tsx
```diff
@@ -117,6 +129,16 @@ export default function HomePage(): ReactElement {
           </Button>
         }
       />
+
+      <SectionQueryResult
+        title="New Algorithms"
```
kremalicious, line 134: we should get in latest changes from `feature/compute` which itself includes latest changes from `main` where we have some changed labeling on front-page so however we name it here needs to be considered in that context.

Then, if we have a section with new algorithms only, it implies that the section above only shows latest data sets, but no algorithms.

Not sure about the Highest Liquidity section yet if we want to mix them there, or rather separate. Likewise, I'm thinking about some visual separation cause right now we visually separate between download/compute datasets but not between datasets/algorithms. On top of that for algorithms, pretty sure we do not want to have a `Compute` label show up on their `AssetTeaser` despite them having a compute service

KatunaNorbert: maybe we can keep only the sections "Highest Liquidity" and "Recently Published" which contains both algorithms and data sets and after sorting by service type is implemented on the search page we can change the links from sections by replacing `DATA SETS` with `ASSETS`. Then the user can do the sorting o search page

claudiaHash: I just have merged the changes from `feature/compute` in this branch so now it's up to date .

On the datasets section I  added a filter, so now we can have a clear separation between datasets and algorithms sections.

Maybe separating Highest Liquidity Datasets / Algorithms in two sections would result in way too many sections on market page. A visual separation between algorithms and datasets would make things more clear for user, but I'm not sure how we could achieve that visually, without crowding the design.

kremalicious: yup, keeps front-page more simple. But requires some visual distinction, currently trying to come up with some quick visual system accounting for download/compute/dataset/algorithm. If we have that we can show them all mixed everywhere by default

kremalicious:
First tinkering for some simple visual system to start from, to differentiate between download/compute/dataset/algorithm:

Screen Shot 2021-02-18 at 15 49 18

data set: where the circle icon placeholder signals either download or compute. Cause this distinction concerns for what the price is paid so belongs around the price. Also does not visually prioritize either one any more, which is better than currently used `Compute` label.

And then with that space free we can use it to add a text label on algorithms, and just very slightly dim their backgrounds (I used `#f6f6f6`). Dim because despite having a label making it more prominent, data sets should be visually most important. Although probably debatable given that on tech level they are almost the same

claudiaHash:
This looks nice and clean, and resolves all the problems mentioned above. I think we should find really expressive icons for the `Download` / `Compute` actions so the user understands those purpose.

mihaisc:
Oh boy, please re-think the UI, we don't need more clutter in that teaser. That information is irrelevant. We can add a help message in the detailed view somewhere to add more context to what that purchase is, but you can't even "buy" an algo directly. In the teaser it's enough to know that it's an algorithm.  The price of the algo will be usefull in the "start job" interface where we calculate how much the user has to pay ( compute data price + algo price ), and there we can explain more.

kremalicious:
even better if we can skip download/compute as unneeded info completely and leave price line as is.

Was also thinking about leaving price for algorithms out but then realized it's not covering all user groups. If we allow pools to be created on algos, the price becomes kinda important. Not for the usage part, I agree, but for the trade part. If we hide prices on algorithm pool teasers, it communicates that they are less important than data set pools. If we deprioritize them anyway, then why have algorithm pools in the first place? So if we want those pools, we should show the price consistent to how we show it on dataset pools

KatunaNorbert:
That looks good, now I can easily differentiate the asset type. 

About differentiating download/compute I think that could be useful. If there would be multiple algorithms published that does the same thing and I want to find one that I can download I don't have to open multiple assets just to see if I can download them. But we should find something suggestive, I think a text would do the best.

mihaisc:
I didn't say to hide the price, i just said to leave the teaser as it is + the Algo label. I said that the extra circle/icon was too much

mihaisc:
> Not sure about the Highest Liquidity section yet if we want to mix them there, or rather separate. 

Show all assets that have a pool ( dataset  + algo ) people in this section don't care about the content that much they are here for the pools.

For distinction : 
image

image

image

image

kremalicious:
yeah, was my first try using existing style and just add more labels, but when seen in context with many teasers in `AssetList`, using that existing label becomes super busy and suddenly those labels are visually the most important thing on screen.

Distinction between download/compute is fine to leave out as we discussed above. Then for dataset/algorithm we use that same place but with the more subtle label (`--color-secondary`), and adding label to both of them as in 2 last images above. At least makes it really clear:

Screen Shot 2021-02-19 at 10 50 23

The rest depends on if we want to communicate them on same level, or prioritize one type more. Thinking behind that white/dimmed background as another visual hierarchy level was to put focus on datasets. The assumption was that we want to encourage usage of data sets more than with algorithms which would justify making them somehow more prominent. 

Screen Shot 2021-02-19 at 10 50 59

Screen Shot 2021-02-19 at 10 57 12

mihaisc:
Cool ideea, i like the slightly distinct background. And for compute we slap the label that we have now or how would it look?

kremalicious:
At least download is easy in minimal to add this distinction back without adding too much to the UI:
Screen Shot 2021-02-19 at 11 24 10

Compute not so much, first thought: boxes cause jobs running in containers:
Screen Shot 2021-02-19 at 11 25 43

Or just one box so just an outlined square, compute is so abstract anyway:

Screen Shot 2021-02-19 at 11 34 31

In any case, principle could work to only have small icons for download/compute on the dataset labels, although hard to find clear icon to communicate compute

mihaisc:
I think the 3 squares look better. Don't think we need to stress too much. Once people understand it it's just a visual queue. And for people who are not familiar yet it's not relevant if it's an icon or it says "compute" they will not understand either one

src/components/pages/Home.tsx
kremalicious: Aaaaand we arrived back at this after some global color tinkering in 2427c8e. We can now: 

- mix both types in all sections on front page
- remove this _New Algorithms_ section. But might be handy to leave it in for now, but remove it before final release of everything

kremalicious:
While we added this functionality in here and it works, think we do not get around rethinking that filter bar a bit:

Screen Shot 2021-02-19 at 13 36 16

But let's do that in another PR on its own

mihaisc:
That filter works, already tested it on bb fronted

kremalicious:
After playing with having another tooltip on the card: I would be in favor of removing all of them. They create a bunch of interaction problems and we do not use them here how we use tooltips in other places. We always have a visible trigger or other affordance for them, one reason being support for either mouse or touch input. And with our usage on the AssetTeaser cards we alienate touch users already since touching any place on the card goes to the details page. For mouse users the tooltip is then a rather surprising element and can get in the way by simply scrolling.

Tooltip on the datatoken symbol was never thought through and more a way to quickly solve the novelty of it. But think we can get away to have other ways in UI to teach users what symbol/icon on cards mean, e.g. by making sure that datatoken symbol/name & some sort of representation of download/consume along with same icons is shown at top of AssetDetails

src/components/atoms/Tooltip.module.css
```diff
@@ -5,8 +5,6 @@
 .content {
   composes: box from './Box.module.css';
   padding: calc(var(--spacer) / 4);
-  width: calc(100% - var(--spacer) / 3);
-  max-width: 25rem;
```
kremalicious, line 9: that custom `width` calculation removal nicely solves a bunch of issues across whole app since tippy.js seems to figure everything out correctly by now. But the `max-width: 25rem;` we still need to leave in cause tippy.js tries to use full viewport width otherwise:

Screen Shot 2021-02-22 at 11 17 47

src/components/molecules/AssetTeaser.tsx
```diff
@@ -37,7 +39,17 @@ const AssetTeaser: React.FC<AssetTeaserProps> = ({ ddo }: AssetTeaserProps) => {
           </Dotdotdot>
           <Publisher account={owner} minimal className={styles.publisher} />
         </header>
-        {isCompute && <div className={styles.accessLabel}>Compute</div>}
+
+        <div className={styles.typeDetails}>
+          <Tooltip content={accessType === 'access' ? 'Download' : 'Compute'}>
+            {accessType === 'access' ? (
+              <Download className={styles.icon} />
```
kremalicious, line 46: since these icons have no accompanying text but still convey information we should add some ARIA attributes for accessibility reasons, in this case: `role="img" aria-label="Download"`

mihaisc:
Looks good

kremalicious:
pushing final visual tweak in a bit with some small spacing tweaks, and to make sure text & icon are not seen as the same thing:

Screen Shot 2021-02-22 at 12 29 33

mihaisc:
The filter is just UI right, it's not meant to work in this pr, correct?
image

claudiaHash:
Yes, it's just the view. We're not sure about its functionality yet. Some options are presented in [#392](https://github.com/oceanprotocol/market/issues/392)

kremalicious:
there seemed to be more things to figure out in #392 until we can finalize that filter bar. But you're right, it stopped working somehow. So I would say when clicking _Algorithms_ it should show me algorithms on Rinkeby as the baseline within this PR.

Everything else should be good now and we can just merge this PR

mihaisc:
i'll fix it in a moment

mihaisc:
Shouldn't it be `DATASET` instead of `DATA` ?

claudiaHash:
Think `Dataset` is more correct, initially I inspired from #337 description

mihaisc:
`DATA SETS` to be more accurate since this is how we use it on the home page

mihaisc (approved):
Looks really good to me. @kremalicious i see you are still tweaking the ui a bit, merge it when you're done

mihaisc:
I just saw that in the tag we have `DATASET` 
image

claudiaHash:
Yes, that's set on the DDO attributes `const { name, type } = attributes.main`. Try to fix that before merging.

src/components/pages/Home.tsx
```diff
@@ -29,7 +29,7 @@ const queryLatest = {
   query: {
     nativeSearch: 1,
     query_string: {
-      query: `-isInPurgatory:true`
+      query: `(service.attributes.main.type:dataset) -isInPurgatory:true`
```
kremalicious, line 32: this slipped through, the _Recently Published_ section should show both so this should be enough:

```
query: `-isInPurgatory:true`
```

mihaisc: good catch

claudiaHash: Totally forgot about the query. Thanks!

Issue #343, opened 2021-01-27
Start compute job on data set


kremalicious:
On the asset details page under the _Use_ tab there's a _Select Algorithm_ action which opens a modal with a list of existing algorithms. Alternatively only populate a `select` with pre-defined default Algorithms we publish.

Then there's a _Start Compute Job_ button which starts compute job with selected algorithm.

On dedicated screen, or modal show:

- list of algorithms published in market, where they show sufficient infos for users to know hat they are (title, linked to algorithm detail)
- one needs to be selectable
- with selected algorithm start compute job
- show status → ideally use new to be created global transaction provider around the app, and show further compute job status there as defined in #385

kremalicious:
quick mockup, this was my first thought apart form a `select`, to allow us to show mire metadata in place. Still lots of room for getting more space out of the _Use_ screen, like the file is not that important anymore so we could make it way smaller giving us more space.

<img width="516" alt="Screen Shot 2021-03-09 at 11 51 29" src="https://user-images.githubusercontent.com/90316/110459881-ca8fba00-80cd-11eb-8425-4d417e5c78eb.png">

Just to check: did I get this right that if algorithms have price set, then that price is added to the data set price upon ordering on the data set, right? If so then we need to show both prices. Ideally we can also show a sum under the action button

mihaisc:
The file icon has no meaning here, we can remove it completely.
Yes we need a sum ( data set + algo ) somewhere at the bottom

mihaisc:
> show status → ideally use new to be created global transaction provider around the app, and show further compute job status there as defined in #385

With the suggested implementation this status is not that prio now. You can just check the compute history.

kremalicious:
I think those are all the elements we want on there:

<img width="454" alt="Screen Shot 2021-03-09 at 16 03 31" src="https://user-images.githubusercontent.com/90316/110490869-012afc00-80f1-11eb-8046-eb8978c6f082.png">

The algorithm titles should be linked to their details pages, get pink on hover, and possibly best to open them in new tab. Not sure how useful the image type is but we can add it as a smaller label instead of _| Node.js_. The `+` before the algorithm price implies that we could have some `prefix` param on our `PriceUnit` component where either non, or our existing `≈` is the default.

Then we should see if we can have titles at maximum of 2 lines per row and then truncate, or if 1 line is also enough. In any case their `title` attribute should include the full non-truncated title.

[...]

PR #389, opened 2021-02-18
View Algorithm details page

### LINKED ISSUE #342: Algorithm detail view
Main question: what should be removed compared to data sets?

Suggestions:
- remove all actions, except for editing

KatunaNorbert:
Fixes #342  .

Changes proposed in this PR:

- Create page for displaying algorithm details


kremalicious:
Going all in on the details refactoring :-) Separating dataset & algorithm views that much might not be necessary, so far it looks like only some stuff in the _Use_ tab is different between the two. For all the rest we should get away with just one detail view component, and switch some elements and maybe some copy based on `metadata.main.type` only.

That was also the thinking behind using in code `Asset` term (which we never expose to users) so we have a collection term for both, `dataset` or `algorithm`. Cause technically they are almost the same, both are just the same DDO structure with minimal differences.

As a start we can first make sure that all new metadata set in #384 is also output as is. Then we could figure out if for some elements it would make sense styling them more, like the Docker time type

KatunaNorbert:
Ok, It seems a lot of work with separating the 2, wasn't sure how much differences there are between them because there are quite a few components involved, so if there would be more changes could be tricky to tack them.

KatunaNorbert:
The info seems right, I think the `use` action should be changed for compute to allow adding data to run the compute job on

mihaisc:
I suggest you ignore the use section for now until we properly define how we start a job. Regarding the details i agree with @kremalicious it's the same data + some fields so just an extra section based on `type`. 
Suggestions : 
- for access (download) : in the use tab we show something like "this asset can't be downloaded, it can only be used in a compute job "
- for compute : we just show the compute component  (however will that look)

KatunaNorbert:
Alright, here are the changes:
Screenshot 2021-02-19 at 12 18 14
Screenshot 2021-02-19 at 12 18 37

Changed the select from compute because it was confusing for algorithms

mihaisc:
The compute part is not relevant now because we need to change that anyway so it's just a placeholder for now. 
Other suggestions for the text in use are welcomed, not sure if it's the best way to express this idea.

kremalicious:
not even sure how much of the technical fact that algorithms have download or compute service we want to expose to users. I guess download, meaning user has set algorithm to be public, would imply having the same interface within _Use_ to download the algorithm for a price. When use has set algorithm to be private we add a compute service, but I feel exposing compute on algorithms is not that useful in the end. So currently tending towards having a label in this case with no actions and saying something like "This algorithm is private and can't be downloaded". Then it could make sense to have an interface to show a list of datasets a given algorithm is allowed to be run on.

But this all does not have to be solved in this PR and addressed individually as some _Use_ issue later on. As for this PR, there are already some things coming in other PRs together which imply:

- output the new metadata we get with #384 somewhere. Always easy to output in `MetaFull`, but worth looking into more prominent/meaningful display for some, like Docker image type which could be as important as file info icon
- to make it more clear we could pick up the `AssetTeaser` labels from #388, and output them also somewhere at the top

KatunaNorbert:
Than I think that I should wait with this issue for the other PR to be done

mihaisc:
I'm afraid that we might needlessly complicate things with the 'public' and 'private' tags on the algorithm. Starting a compute job from an algorithm with a compute service is a valid use case that we will want to support in the future.  You add the compute algo to the job and the add a dataset with a consume service (or multiple datasets) and you have a valid compute job.

kremalicious:
I like that simple solution of repeating the component from AssetTeaser which might be all we need for this. Placement looks right, but think it creates an interaction problem since it might be perceived as actionable. So would move it as the first item in `MetaMain` without absolute positioning:

Screen Shot 2021-02-24 at 11 56 42

Or to save some space we could use that line where we only have the date now:

Screen Shot 2021-02-24 at 12 06 55

While testing, also looked good moving that whole line then to the very top of `MetaMain`, couldn't decide 😄 

Would imply making the `AssetTypeDetails` (we could name it just `AssetType`? Does it conflict with other names then?) more isolated to styling only, not positioning, and then only on `AssetTeaser` use the absolute positioning. Also `aside` then is not the best generic markup for that component anymore.

Another idea which would help in learning the icons is to somehow write out as label what they stand for. Problem is that as soon as the arrow icon and _Download_ are put together it looks like a download action.

kremalicious:
As for metadata output, we should at least output metadata we know all algorithms have (they are already published with that in Rinkeby so that metadata is not only introduced with #384) in `MetaFull`. That would be what is under `attributes.main.algorithm.container`. 

We can have one item _Docker Image_ which then combines `container.image` & `container.tag` as value: `${container.image}:${container.tag}`. This covers already most use cases, we can figure out later what to do with custom images

src/components/organisms/AssetActions/Consume.tsx
```diff
@@ -157,16 +157,23 @@ export default function Consume({
...
-          {isConsumable ? (
-            <Price ddo={ddo} conversion />
-          ) : (
-            <div className={styles.help}>
-              There is not enough liquidity in the pool to buy this data set.
-            </div>
-          )}
-          {!isInPurgatory && <PurchaseButton />}
-        </div>
+        {type === 'algorithm' ? (
```
kremalicious, line 160: I think we can remove this condition again since we want to allow compute on an algorithm. So we do not specifically have to state that, it is implied by switching from `Download` to `Compute` component, just like with data sets.

src/components/organisms/AssetContent/index.module.css
```diff
@@ -42,3 +42,7 @@
   margin-left: calc(var(--spacer) / 4);
   margin-right: calc(var(--spacer) / 4);
 }
+
+.content [class*='typeDetails'] {
```
kremalicious, line 46: just by the way, using attribute selectors to select things should only be a last resort for cases where we do not control markup and can't add a CSS module class. This can pile up to specifity and maintenance problems later on.

So for everything where we control markup a better pattern is to have all components accept an optional `className` so we can set CSS module classes on them wherever we need to style them individually

src/components/organisms/AssetContent/index.tsx
```diff
@@ -46,8 +47,9 @@ export default function AssetContent(props: AssetContentProps): ReactElement {
   const { owner, isInPurgatory, purgatoryData } = useAsset()
   const [showPricing, setShowPricing] = useState(false)
   const [showEdit, setShowEdit] = useState<boolean>()
-  const { ddo, price, metadata } = useAsset()
+  const { ddo, price, metadata, type } = useAsset()
   const isOwner = accountId === owner
+  const accessType = ddo.service[1].type
```
kremalicious, line 52: right, was seeing that we already have this in `feature/compute` but worth tweaking that in here. While this works right now with assets published through our UI, we should not rely on the `service` array order. And to make it more fun, in there, there could be both, `access` & `compute` service but we can almost always assume that if there's a compute service there should be no download possible.

Not sure about best logic here:
- get the value of the `type` key in one of the `service` objects, where the value is either `access` or `compute`, that seems complicated
- just rely on checking for compute service as a boolean, [as we do in `main`](https://github.com/oceanprotocol/market/blob/main/src/components/organisms/AssetActions/index.tsx#L21)
- any other ideas?

src/providers/Asset.tsx
```diff
@@ -25,6 +25,7 @@ interface AssetProviderValue {
   title: string | undefined
   owner: string | undefined
   price: BestPrice | undefined
+  type: string | undefined
```
kremalicious, line 28: We should be more specific here since we require a specific string. We can get this from ocean.js typings with `MetadataMain['type']`. Then it's also more future proof in case another `type` is added in ocean.js

mihaisc:
I like the second image where algo is under "published by... ", looks cleaner

src/components/organisms/AssetContent/index.tsx
mihaisc: I think that we can run with this assumption , but just for now. This is a compromise. Ideally in some future we actually show all the services.
> if there's a compute service there should be no download possible.

src/providers/Asset.tsx
```diff
@@ -52,6 +53,7 @@ function AssetProvider({
   const [price, setPrice] = useState<BestPrice>()
   const [owner, setOwner] = useState<string>()
   const [error, setError] = useState<string>()
+  const [type, setType] = useState<string>()
```
kremalicious, line 56: now that we are more specific in the type typing, this state should use the same typings then, hence my suggestion of using `MetadataMain['type']` from ocean.js also in the above `AssetProviderValue` so we do not have to repeat those hardcoded strings

kremalicious:
all cool so far, not sure we can add more here at this point in time. Think we can just merge it and then look at the exact details of compute actions as part of #343 for both, algorithms & data sets

kremalicious (approved):

PR #394, opened 2021-02-22
refactor publish page for tab usage

kremalicious:
Making use of our existing tab component which turned out to require a bit more refactor so added as additional PR against #384.

- use existing `Tabs` component 
- style that component better to accommodate new usage
- move publish page content files into `content/pages/publish/`
- split up form names for `Persist` component (don't think this makes it actually work yet)
- remove redundant `TypeOfPublish`, and rely on default `MetadataMain` typings

Screen Shot 2021-02-22 at 18 35 18


content/pages/publish/form-algorithm.json
```diff
@@ -1,5 +1,5 @@
 {
-  "title": "Publish",
+  "title": "Publish an Alorithm",
```
mihaisc, line 2: algorithm

mihaisc (approved):
Looks awesome. Can merge after you update 'Alorithm'

src/components/pages/Publish/index.tsx
```diff
@@ -137,73 +164,78 @@ export default function PublishPage({
...
+          }
+        ]
+
+        return (
+          <>
+            <Persist
+              name={
+                publishType === 'dataset'
+                  ? formNameDatasets
+                  : formNameAlgorithms
```
bogdanfazakas, line 200: Unfortunately this is not enough, I tried it also today, you should also remove the resetForm call from the FormAlgoPublish.ts (https://github.com/oceanprotocol/market/blob/e157278e37a82305852581a89f28b00c96cf953f/src/components/pages/Publish/FormAlgoPublish.tsx#L60)

It's something strange i guess it's related to formik values, somehow the values from localStorage is override when switching tabs and the other form mounts. Also strange why there are stored in the initial values for the algorithm form the same values as for the dataset even though we pass other values there.
Screenshot 2021-02-22 at 22 30 17

src/pages/publish.tsx
```diff
@@ -4,7 +4,7 @@ import Page from '../components/templates/Page'
 import { graphql, PageProps } from 'gatsby'
 
 export default function PageGatsbyPublish(props: PageProps): ReactElement {
-  const content = (props.data as any).content.edges[0].node.childPagesJson
+  const content = (props.data as any).content.edges[0].node.childPublishJson
```
bogdanfazakas, line 7: Nice 😃 , I also initially tried to move them into publish folder but the query didn't really work, but I see now why 👍

kremalicious: yes, this is super confusing that Gatsby picks up the folder names to construct the query structure

mihaisc:
this is supposed to be a pr just to handle the tab usage, not fix other form bugs/issues. So i think we are good on this one.

src/components/pages/Publish/index.tsx
kremalicious: good findings, that's worth looking at individually in publish PR or as another PR to fully restore this behavior. But we should not get blocked by this, would be also fine moving this into a new issue to deal with it later

kremalicious:
yup, agree, just unified both forms and do not intend to look at that localStorage bug in here further

bogdanfazakas:
> this is supposed to be a pr just to handle the tab usage, not fix other form bugs/issues. So i think we are good on this one.

yup agree, just wanted to let @kremalicious know that that small update didn't fix the storage issue 😃

PR #399, opened 2021-02-23
Feature/allow changing of fixed rate price

### LINKED ISSUE #330: Allow changing of fixed-rate price
**Is your feature request related to a problem? Please describe.**

If the publisher has set the price in a fixed fashion (FRE = fixed rate exchange), they can't change it afterwards. Yet there are circumstances where they would want to. (FIXME add why)

Managing a FRE:
  - display the current allowance & rate
  - increase/decrease (edit) the allowance
  - set new rate

KatunaNorbert:
Fixes #330.

Changes proposed in this PR:

- Allow changing price inside edit form

Screen Shot 2021-04-21 at 15 15 41


KatunaNorbert:
I managed to get the `exchangeId` and get the `setRate` function working but the price does not update.
I'm not sure if I'm looking at the right price, but I think is the price from ddo:
Screenshot 2021-02-24 at 13 36 22

src/components/organisms/AssetActions/Edit/index.tsx
```diff
@@ -76,6 +82,26 @@ export default function Edit({
         description: values.description
       })
 
+      console.log('herrrreeee')
+
+      const exchandeId = await ocean.fixedRateExchange.generateExchangeId(
```
kremalicious, line 87: this is generating a new exchange ID which ocean.js uses internally. In here, we need to get the exchange ID based on the given datatoken address somehow.

Bummer that [`getExchange()`](https://github.com/oceanprotocol/ocean.js/blob/main/src/exchange/FixedRateExchange.ts#L319) requires an exchange ID 😄 

But I'm almost sure `ddo.price.address` is the exchange ID, it's the address of that specific fixed rate exchange contract

kremalicious: but good exercise in error reporting. If it is the case that you tried to set the price of a randomly generated `exchangeId` (which does not exist) the transaction should have failed.

kremalicious:
This might be the first field where we need some activate/deactivate UI pattern. Because right now, we always save all fields when user hits submit. This was fine so far because we populate with existing values, and metadata updating is one transaction anyway.

Price editing introduces another transaction but we can't have 2 transactions when user did not intend to change the price. Same the other way around, user only wants to change price so there should be no metadata editing transaction.

So once we find a UI for selecting fields to edit, we can use it on all fields so users will be able to selectively change what they want, and based on selections we figure out which exact transactions to send.

mihaisc:
If you do a quick price fetch before you will get the exchangeId in address `address: tokenExchange.exchangeID || '',`
Another option would be to fetch the exchange directly `await ocean.fixedRateExchange.searchforDT(dataTokenAddress, '1' )`  and get the exchangeId from the first item in the returned array

KatunaNorbert:
I see, I used `ocean.fixedRateExchange.searchforDT()` and then on set price It says that the owner address I give is invalid
Screenshot 2021-02-24 at 17 14 50
Screenshot 2021-02-24 at 17 12 19

KatunaNorbert:
> This might be the first field where we need some activate/deactivate UI pattern. Because right now, we always save all fields when user hits submit. This was fine so far because we populate with existing values, and metadata updating is one transaction anyway.
> 
> Price editing introduces another transaction but we can't have 2 transactions when user did not intend to change the price. Same the other way around, user only wants to change price so there should be no metadata editing transaction.
> 
> So once we find a UI for selecting fields to edit, we can use it on all fields so users will be able to selectively change what they want, and based on selections we figure out which exact transactions to send.

Can't we do those checks in the edit method so we don't have to change the UI?

kremalicious:
`ocean.fixedRateExchange.searchforDT()` is probably way slower than using `price.address` from `useAsset()` which the asset provider already has when user lands on asset details.

Then for the last argument in `setRate()` the current wallet account from the currently connected user has to be used. So `accountId`  from `useOcean()`.

src/components/organisms/AssetActions/Edit/index.tsx
KatunaNorbert: well It wasn't really on purpose, I mean I wasn't sure if that is right or not :)

kremalicious:
> Can't we do those checks in the edit method so we don't have to change the UI?

That could also work, would imply checking every input value against existing DDO values, and then go from there to only update what has been changed by user

KatunaNorbert:
In my case `accountId` and `exchangeOwner` are the same so I get the same error

kremalicious:
then we might have found a bug somewhere across the stack but requires more testing, I mean it's probably the first time that ocean.js method is used. Feel free to push any code change you have already so everybody has same testing base

mihaisc:
so the error in the image (invalid exchange owner) you get after calling setRate?

KatunaNorbert:
Yes

mihaisc:
It seems that something doesn't work in ocean.js. Suggest you drop it, i'll take a look

mihaisc:
One issue is with the new price component. If we have `const { price } = useMetadata(ddo)` it works on the asset teaser but it will not update in asset details when price is updated. If we have `const { price } = useAsset()` it will properly update in asset details but it will not work on the asset teaser.

mihaisc:
The price is being updated , the issue is that aqua is not updating it. Issue created https://github.com/oceanprotocol/aquarius/issues/403
image

KatunaNorbert:
Was testing this and run into the following error:
Screenshot 2021-03-30 at 15 27 03

Caused by `ocean.fixedRateExchange.searchforDT` returning an empty array

KatunaNorbert:
Oh, my bad, was trying to update the price on dataset with dynamic price.
Retested on fixed price and works. Should make some changes to disabled the price update on datasets with pools

mihaisc:
Yes, please add that

KatunaNorbert:
Disabled the price input for `pool` type prices

kremalicious:
Quick UI testing:

- the edit price field should not even show up on dynamic priced assets, should only show up on fixed price assets
  Screen Shot 2021-04-15 at 17 08 03

- the edit price field should follow our pattern we use whenever we edit prices in the create pricing flow, otherwise hard to get what the number represents. So should mimic this, either in full showing the datatoken relationship, or just with the input + prefix:
   Screen Shot 2021-04-15 at 17 10 44

Other than that, actual functionality of changing the fixed price works like charm!

src/@types/MetaData.d.ts
```diff
@@ -39,6 +40,7 @@ export interface MetadataPublishForm {
   files: string | File[]
   author: string
   timeout: string
+  price?: number
```
kremalicious, line 43: don't think we need `price` on here, since in our publish form we do not capture the price so should be removed from here

content/pages/edit.json
```diff
@@ -20,6 +20,15 @@
         "rows": 10,
         "required": true
       },
+      {
+        "name": "price",
+        "label": "New Price",
+        "type": "number",
+        "min": "1",
+        "placeholder": "e.g. Shapes of Desert Plants",
```
kremalicious, line 28: we can just use `"0"` as placeholder for the price input

src/components/organisms/AssetActions/Edit/index.tsx
```diff
@@ -79,6 +96,8 @@ export default function Edit({
         links: typeof values.links !== 'string' ? values.links : []
       })
 
+      ddo.price.type === 'pool' || (await updateFixedPrice(values.price))
```
kremalicious, line 99: This breaks actual editing, what was the intention of this line?

Was the intention to only fire method when FRE asset? Then more like so maybe, fire when FRE & only when price has been changed:

```js
ddo.price.type === 'exchange' && values.price !== ddo.price.value && await updateFixedPrice(values.price)
```

src/components/organisms/AssetActions/Edit/index.tsx
```diff
@@ -124,7 +143,8 @@ export default function Edit({
     <Formik
       initialValues={getInitialValues(
         metadata,
-        ddo.findServiceByType('access').attributes.main.timeout
+        ddo.findServiceByType('access').attributes.main.timeout,
+        ddo.price.value
```
kremalicious, line 147: we are within the `AssetProvider` so we should be able to not rely on the DDO price value for future reasons. So rather use the `price.value` from `useAsset()` instead

content/pages/edit.json
KatunaNorbert: right, missed that since we won't have a case when that would pe displayed

kremalicious:
- updating of FRE price works now for me, although I somehow get 3 transactions for the whole metadata editing now
- metadata editing screen for non-FRE assets is broken now

src/components/organisms/AssetActions/Edit/index.tsx
```diff
@@ -60,13 +61,29 @@ export default function Edit({
...
-  const { metadata, ddo, refreshDdo } = useAsset()
+  const { metadata, ddo, refreshDdo, price } = useAsset()
   const [success, setSuccess] = useState<string>()
   const [error, setError] = useState<string>()
   const [timeoutStringValue, setTimeoutStringValue] = useState<string>()
 
   const hasFeedback = error || success
 
+  async function updateFixedPrice(newPrice: number) {
+    const exchanges = await ocean.fixedRateExchange.searchforDT(
```
kremalicious, line 72: do we need to do this expensive search? Cause it looks like we get the exchange ID from `price.address` for FRE assets https://github.com/oceanprotocol/market/blob/main/src/providers/Asset.tsx#L117

KatunaNorbert:
Hmm, I only have to confirm 2 transactions when changing the price and 1 when not.

src/components/organisms/AssetActions/Edit/index.tsx
KatunaNorbert: yeah, good point, we don't

src/models/FormEditMetadata.ts
```diff
@@ -8,17 +8,20 @@ export const validationSchema = Yup.object().shape({
...
   links: Yup.array<EditableMetadataLinks[]>().nullable(),
   timeout: Yup.string().required('Required')
 })
 
 export function getInitialValues(
   metadata: MetadataMarket,
-  timeout: number
+  timeout: number,
+  price: number
 ): Partial<MetadataPublishForm> {
```
kremalicious, line 20: almost sure this typing here is wrong and should be `MetadataEditForm`. This should also make all those errors go away

kremalicious:
ok, 2 transactions it is indeed. Either I can't count or this switching between 10 different market versions in my browser leads to unexpected results. Should be all perfect now.

Immediately had this question regarding user transparency and since `setRate` is another transaction, it might be feasible to show some price history, just like the metadata history. But that's for another new feature.

kremalicious (approved):
awesome, all good. Pending decision to merge in Slack

PR #400, opened 2021-02-24
Search filter bar

### LINKED ISSUE #392: Search filter bar reorganization
With #388 a new filter group is introduced to switch between algorithms & datasets in addition to existing filters:

108505305-73a97880-72b7-11eb-9f6d-17bbc3fd19c5

Some possible options:
- put existing filters on one line on big viewports
- add label to each filter group, following the style of sorting bar 
- combine price & type filter bars into one. So that _All_ refers to all of both, and all other items in that one group can be clicked and preferably combined, so multi-selection

claudiaHash:
Fixes #392  .

Changes proposed in this PR:

- added  service type filters 
- combined price filters with service type filters
- possibility of selecting and deselecting filters


kremalicious:
this is great visually!

As for interactions:
- remove _All_ item
- with this multi-selection pattern we should make it possible that all of them can be multi-selected, not just 1 of each group
- not fully sure, but the deselection on click of whole item might be a bit surprising, so we could make it deselect only when that `x` is hit. So we would set like `pointer-events: none` on the selected item, except on the `x` which we make slightly bigger too

Since we're at the search header in this PR we should also do something with the search title because _datasets_ is hardcoded there:

Screen Shot 2021-02-24 at 11 37 22

To reflect the new asset types and possible more coming filters in future, we could simply call it _results_.

src/components/templates/Search/filterPrice.tsx
```diff
@@ -57,41 +58,96 @@ export default function FilterPrice({
...
               key={index}
-              className={filter}
-              onClick={async () => {
-                await applyTypeFilter(e.value)
-              }}
+              className={selectFilter}
+              onClick={
+                // only works for two price options
+                isSelected
+                  ? async () => {
```
kremalicious, line 77: good for initial prototype but we should try moving all this logic into ideally just one `handleFilterSelection()` method to which we can pass the `isSelected` as a parameter. So we can then do all logic within that handler method instead of repeating them

claudiaHash: I'm refactoring that right now. I'll push the new version when ready

mihaisc:
If you play around with the filters it sometimes bugs. Can't seem to replicate consistently 

image

mihaisc:
If you selected in the filter dynamic and fixed should we show assets without price?
image

mihaisc:
If you mess around with the sorting as well as with the filtering, sometimes the sorting cancels the filtering. In the image i should have 16 items not 60
image

claudiaHash:
As discussed yesterday, selecting both filters from a category works as selecting none, because in the url we can't have ...&priceType=exchanged&priceType=pool

mihaisc:
> If you selected in the filter dynamic and fixed should we show assets without price?
wasn't a bug report, it was a question. Should we try to hide assets without a price if you select both fixed and dynamic? Assets without a price are basically useless.

kremalicious:
right, sounds most logical to hide no price assets when either one of the price filters is selected. In all other searches we should continue showing them

mihaisc:
This query works , matches assets that have `price.type`  `exchange` or `pool`
```
{
    "page": 1,
    "offset": 21,
    "query": {
        "nativeSearch": 1,
        "query_string": {
            "query": "price.type:pool exchange  -isInPurgatory:true"
        }
    },
    "sort": {
        "created": -1
    }
}
```

kremalicious:
damn conflicts, I'll solve them locally. Other than that I think all is working as expected now after some quick testing

kremalicious:
> If you selected in the filter dynamic and fixed should we show assets without price?

Coming back to that we found actually some issues, since we detect if `price.value: 0` to show "No Price Found", and that could have multiple reasons:

- price has not been set (so also no `price.address` which would be better check if we want those filtered out)
- low liquidity in a pool asset (for which we also have `price.isConsumable`)

So sounds like we want `price.isConsumable` here when price filter are selected. On the other hand, it can make sense to show pools despite them having low liquidity, how else would they regain liquidity if they can't be discovered?

mihaisc:
Well, i agree we need to take another look at the `Price` component, but we need to handle this here too. using `isConsumable` in this case is not ok because of the reason you stated. I think it should be ok to use `price.address`

mihaisc:
Should we add a clear filters button?

claudiaHash:
Not sure if it's necessary since the clear filters option is achieved  selecting none of them

mihaisc:
Yes, but if i select 3 of them, then i have to click 3 times to deselect instead of a `clear` button

mihaisc (approved):

kremalicious:
that `Price` component thing was worth moving into a PR against `main`: https://github.com/oceanprotocol/market/pull/416

src/components/templates/Search/filterPrice.tsx
```diff
@@ -36,7 +36,11 @@ export default function FilterPrice({
 }): ReactElement {
   const navigate = useNavigate()
 
-  async function applyFilter(filterBy: string) {
+  const [priceSelections, setPriceSelections] = useState<string[]>([])
+  const [serviceSelections, setServiceSelections] = useState<string[]>([])
+  const [clearSelected, setClearSelected] = useState<boolean>(false)
```
kremalicious, line 41: not sure I get why the clear action gets its own state. So we want a clear button to clear all user selection. Usually a clear button is only visible after user has selected something. Then when that button/small text link is clicked all selections should be turned off, meaning we can simply reset `priceSelections` & `serviceSelections` states so I don't see a reason to have this clear button have its own state here which is then turned on/off multiple times further down. Think all we need is that `applyClearFilter` method resetting states.

UI-layout-wise I would try like this:

Screen Shot 2021-03-02 at 12 05 27

src/components/templates/Search/filterPrice.tsx
```diff
@@ -45,52 +49,156 @@ export default function FilterPrice({
...
+    setPriceSelections([])
+
+    setClearSelected(true)
+    setPriceType(undefined)
+    setServiceType(undefined)
+    navigate(urlLocation)
+  }
+
   return (
     <div>
```
kremalicious, line 143: for whatever reason, this double `div` was added back in 8cce1898fdd6e0a163c441825d3e8401cf3232c8. Does not look like this is needed

src/components/templates/Search/filterPrice.tsx
mihaisc: Yep, i agree, it doesn't make sense to be the same as the other filters. It needs to be a separate action

claudiaHash: Almost done that. I'm only having an issue related to the new solution: after `onClick` on the `Clear` button, all elements behave correctly except the button doesn't hide unless an action is done ( a page reload, click anywhere). Once that is fixed, I'll push the solution.

mihaisc: Add another state `isClearVisible` and set that `onClick` on in a `useEffect`

kremalicious (approved):
perfect, all working great now

PR #402, opened 2021-02-24
save form values for both publish forms

### LINKED ISSUE #397: publish: saved form values for both forms
Introducing 2 different forms on the publish page broke the form values saving in user's `localStorage`, done by our `Persist` component.

**To Reproduce**

1. Go to `/publish`
2. Enter some text into _Title_
3. Navigate to front page, come back to page
4. See that _Title_ is still there, all good so far
5. Click on _Algorithm_ tab, has now same title
6. Enter some text in Algorithm _Title_ & _Description_, switch back to _Dataset_ tab
7. Entered values for _Algorithm_ are now used in _Dataset_

**Expected behavior**
- clicking on _Algorithm_ tab on first time should always start with empty form
- entered values in both forms are isolated and not shared between them
- switching tabs with values entered in both forms should retain their individual saved values
- sending either form should clear stored values for both forms

bogdanfazakas:
Fixes #397  .

Changes proposed in this PR:

- enable formik reinitialize when initial values change and use existing values

### DIFF
--- a/src/components/pages/Publish/FormAlgoPublish.tsx
+++ b/src/components/pages/Publish/FormAlgoPublish.tsx
@@ -7,6 +7,8 @@ import Input from '../../atoms/Input'
 import Button from '../../atoms/Button'
 import { FormContent, FormFieldProps } from '../../../@types/Form'
 import { MetadataPublishFormAlgorithm } from '../../../@types/MetaData'
+import { initialValues as initialValuesAlgorithm } from '../../../models/FormAlgoPublish'
+
 import stylesIndex from './index.module.css'
 
 const query = graphql`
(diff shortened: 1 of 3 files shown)


kremalicious:
yeah, much better, that works!

Only user flow I could find where algorithm form is cleared but shouldn't:

1. go to `/publish`, type in some title for dataset
2. switch to algorithm, type in some title
3. leave the page, then come back
4. dataset form still has the data, but algorithm does not

I guess if we would make the type part of each form's data we could also retain tab selection but that's too much optimization for now

kremalicious (approved):
perfect, did not find any user flows breaking things so should cover everything now

PR #405, opened 2021-02-26
Validate custom Docker images

### LINKED ISSUE #398: publish: set and validate custom Docker images
By default, one of our predefined images is selected for an algorithm but users can also select a custom one. In that case we provide a text input which could allow for:

- Docker image name + tag name, e.g. `node:10`
- DockerHub URL, e.g. `https://hub.docker.com/repository/docker/oceanprotocol/aquarius`
- Custom Docker image URL, e.g. `https://mydockerimages.com/oceanprotocol/aquarius`

I guess since those DockerHub URLs hav no tag name reference we do not get around also showing a _Docker Image Tag_ field for user to fill out, but ideally we have as less input fields as possible to get image name & tag name.

We should then validate as much as possible if a given Docker image can be retrieved, like with the DockerHub API. If it can't an error on the respective field should be shown to user.

bogdanfazakas:
Fixes #398 .

Changes proposed in this PR:

- validate docker hub images ( the get req to the docker hub is blocked by CORS when running locally)
- validate 3rd party images (just a options request to the provided url and checked response status)

### DIFF
--- a/src/utils/metadata.ts
+++ b/src/utils/metadata.ts
@@ -1,5 +1,6 @@
 import axios, { CancelToken, AxiosResponse } from 'axios'
 import { toast } from 'react-toastify'
+import isUrl from 'is-url-superb'
 import {
   MetadataMarket,
   MetadataPublishFormDataset,
@@ -157,7 +158,7 @@ async function isDockerHubImageValid(
 
 async function is3rdPartyImageValid(imageURL: string): Promise<boolean> {
   try {
-    const response = await axios.options(imageURL)
+    const response = await axios.head(imageURL)
     if (!response || response.status !== 200) {
       toast.error(
         'Could not fetch docker image info. Please check URL and try again'
@@ -172,19 +173,6 @@ async function is3rdPartyImageValid(imageURL: string): Promise<boolean> {
     )
     return false
   }
-}
-
-function isUrl(image: string): boolean {
-  const pattern = new RegExp(
-    '^(https?:\\/\\/)?' +
-      '((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|' +
-      '((\\d{1,3}\\.){3}\\d{1,3}))' +
-      '(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*' +
-      '(\\?[;&a-z\\d%_.~+=-]*)?' +
-      '(\\#[-a-z\\d_]*)?$',
-    'i'
-  )
-  return !!pattern.test(image)
 }
 
 export async function validateDockerImage(
--- a/src/utils/metadata.ts
+++ b/src/utils/metadata.ts
@@ -148,6 +148,9 @@ async function isDockerHubImageValid(
     return true
   } catch (error) {
     Logger.error(error.message)
+    toast.error(
+      'Could not fetch docker hub image info. Please check image name and tag and try again'
+    )
     return false
   }
 }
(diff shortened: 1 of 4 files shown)


bogdanfazakas:
Regarding the docker image validation there are at least 2 thing that we need to decide on:

1. what text imput we provide to the end user. As it is now we have separate fields for the image url and container tag. 

- we can opt for image and tag combined but it will get more trickier to handle the the docker hub images (we will have to parse the input and then validate and set algorithm object values) and harder for the user to figure out what to fill in in that input field. The input should probably look like this oceanprotocol/aquarius:v2.1.12 (this is how docker hub displays them within their website)
- if we decide to stick with both input fields the tag is somehow useless for the 3'rd party images because it's probably already in the url. On the other hand it will be much easier and maybe more clear for the user that is using public docker hub images to fill in the publish form.

2. the other thing would be the validation step. As it is implemented in this PR the validation takes place after hitting publish

- we could stick with this approach
- or we can have a validate button similar to the file input field, but if we stick with having separate field for image url and image tag this button should cover both input fields

kremalicious:
as for 1.: while it would be nicer for user to just have one field `Docker Image` which accepts image name + tag or image URL and we figure out the required values in background, this could get tricky and I think for image URLs the tag is usually not part of URL. So think having image name/url & tag individually is good enough. We could try to detect tag name if it is added when user adds imagename:tag to input, and auto-populate the _Docker Image Tag_ field. But then again, we expect this not be used by majority of our users so we do not need to bother too much with it for now.

as for 2.: also think it's fine for now validating on hitting publish, showing the toast as the warning is not ideal (validation error as part of error messages shown on respective input field would be), but also good enough

src/utils/metadata.ts
```diff
@@ -128,6 +130,73 @@ export function transformPublishFormToMetadata(
...
+  } catch (error) {
+    Logger.error(error.message)
+    toast.error(
+      'Could not fetch docker image info. Please check URL and try again'
+    )
+    return false
+  }
+}
+
+function isUrl(image: string): boolean {
```
kremalicious, line 177: we have a package [`is-url-superb`](https://github.com/sindresorhus/is-url-superb) in our dependencies, so we might be able to re-use that instead of maintaining our own patterns here

bogdanfazakas:
> as for 1.: while it would be nicer for user to just have one field `Docker Image` which accepts image name + tag or image URL and we figure out the required values in background, this could get tricky and I think for image URLs the tag is usually not part of URL. So think having image name/url & tag individually is good enough. We could try to detect tag name if it is added when user adds imagename:tag to input, and auto-populate the _Docker Image Tag_ field. But then again, we expect this not be used by majority of our users so we do not need to bother too much with it for now.
> 
> as for 2.: also think it's fine for now validating on hitting publish, showing the toast as the warning is not ideal (validation error as part of error messages shown on respective input field would be), but also good enough

I agree that showing the toast is not ideal, I took a look over manually setting a validate error to a specific formik input, that's not really easy with our implementation but also in our case it a little bit trickier because we cannot know for sure witch input is wrong or causing the error (image url or image tag) and on witch to place the validation error. So for the moment I think we can display the error as a toast and it will be similar with the validation of the algorithm file or we can display the error using setError method from the publish page index. I would personally opt for the the toast 😄

kremalicious (approved):

kremalicious:
All good, let's merge. Looks like that failing Windows test happens pretty often but can be ignored for now

bogdanfazakas:
> All good, let's merge. Looks like that failing Windows test happens pretty often but can be ignored for now

Yeah that Windows test is becoming annoying.

PR #406, opened 2021-02-26
Add polygon network

### LINKED ISSUE #413: Add polygon network/explorer to transaction links component
After adding liquidity to a pool the transaction link is incorrect 

https://polygon.etherscan.io//tx/0x8d9e5bc7e7f8e049d62c252bc22dca531a040acfa73b626a7d9fdf6f6529a207

Should be 

https://explorer-mainnet.maticvigil.com/tx/0x8d9e5bc7e7f8e049d62c252bc22dca531a040acfa73b626a7d9fdf6f6529a207/token-transfers

The same is for all the tx links (pool, datatoken, etc..)

### LINKED ISSUE #418: Incorrect token displayed in wallet
On polygon network in the wallet we show eth instead of matic.

image

mihaisc:
Related to: https://github.com/oceanprotocol/multi-repo-issue/issues/82

- addresses https://github.com/oceanprotocol/multi-repo-issue/issues/82
- fixes #418
- fixes #413

There are still some issue with the rpc node randomly failing. Needs some more testing.
More help :
https://deploy-preview-577--docs-oceanprotocol.netlify.app/tutorials/polygon-bridge/
https://deploy-preview-577--docs-oceanprotocol.netlify.app/concepts/networks/#polygon-ex-matic-mainnet


mihaisc:
Showing proper currency is more or less a quick fix, it has it's dedicated issue.

src/components/molecules/Wallet/Details.tsx
```diff
@@ -23,6 +44,14 @@ export default function Details(): ReactElement {
     setProviderInfo(providerInfo)
   }, [web3Provider])
 
+  useEffect(() => {
+    if (!networkId) return
+    console.log('networks', networksList, networkId)
```
kremalicious, line 49: we can remove this `console.log`, or put it behind `Logger`

kremalicious:
that currency "quick fix" looks like a good solution already, not sure there's much left to do for #418 then 🥳

src/components/molecules/Wallet/Details.tsx
mihaisc: busted

mihaisc:
> that currency "quick fix" looks like a good solution already, not sure there's much left to do for #418 then 🥳

Well, I was thinking at least renaming `eth` from `Balance` to `main` or something, and ... not sure actually. Guess you are right, no point in sinking more time into it.

kremalicious:
I went through the linked docs pages and the connection did just work™ so also kudos to @DimitarSD for the instructions. Paid about $22 for moving 100 OCEAN via the Matic Bridge, way less than I would have expected so that's great

Screen Shot 2021-03-05 at 13 28 30

Would be great to also have `mOCEAN` displayed, not sure we can fetch this from somewhere but guess we could just hardcode it for Matic for now

kremalicious:
Only errors I get are a lot of these on initial loading but could not find anything in the UI being affected by this, all user balances are coming in correctly for me regardless:

Screen Shot 2021-03-05 at 13 37 53

kremalicious:
Other than that, all actions going great and all transactions finish super quickly.

Screen Shot 2021-03-05 at 13 59 19

Since we're not in a hurry to merge this PR, we could leave it open and make #413 against it, that was the only annoyance I found. After that and possibly figuring out that `mOCEAN` display we then really support Matic/Polygon in all the places.

Final merging might depend on the communication strategy, I also tend towards shipping and announcing everything together, this PR + docs + tweets

mihaisc:
> Final merging might depend on the communication strategy, I also tend towards shipping and announcing everything together, this PR + docs + tweets

yes, at this point i was looking for feedback and another person to test.

The errors in console are random and are probably because of the rpc node, so you can ignore them.

Regarding OCEAN vs mOcean, i was thinking about that and i think it should be ok if we leave just OCEAN on all networks. The button near it to add the token to metamask would be nice, but it's not a must have, we can add it later.

#413 I'll at this in this PR, it's a must have, it's weird without. I'll add it in `ConfigHelper` 

Another thing that I was thinking to add in `ConfigHelper` would be `releaseState : alpha/beta` so that we can have something like this for matic (maybe another color)
image

kremalicious:
ah sorry, only arriving at comment now, did the `ConfigHelper` change already over in https://github.com/oceanprotocol/ocean.js/pull/643.

Still feels a bit weird to me going down to an _alpha_ label because it was always not fully clear what it refers to. Contracts? UI? Chain? Network? Now we want it to refer to the state of a chain. And the user expectation is that it can only go up, into final/production/no label. But I can see how it makes sense in this context to have something different there for not fully stable chains.

Maybe rather use something like `preview` instead of `alpha`? I get that we want something below beta but following greek lettering does not really convey what we mean by that, and what the difference to `beta` is. And pretty sure we won't use `gamma` for `releaseState`

DimitarSD (approved):

DimitarSD:
Having a `releaseState` is a good idea. Regarding labels like `alpha`, `beta` - what if we have an info pop-up on hover with a short description of why is in alpha or beta state?

mihaisc:
It doesn't have to be like I suggested, but we need to do something to express to the user that the market on matic is not stable, or may encounter errors from time to time. If we don't do that, the user will blame the market.

mihaisc:
Any suggestions for new text to replace "View on Etherscan" and 
image

DimitarSD:
> Any suggestions for new text to replace "View on Etherscan" and

A few suggestions:

- View transaction
- View transaction details
- Transaction details
- Check details
- View on the Block Explorer (or just "Block Explorer")

mihaisc:
image

mihaisc:
Updated the text 
image
Didn't like the `alpha` party, this this is more appropriate.  I put the text in the same box because we don't want to get too much attention to the `early stages` part, but we still want people to be aware of it while not discouraging them from using it

mihaisc:
moved polygon banner in form
image

mihaisc:
this pr was merged in https://github.com/oceanprotocol/market/pull/433

PR #409, opened 2021-03-01
optimize did fetching in tables

### LINKED ISSUE #395: optimize did fetching in tables
Make use of the endpoint provided by aquarius in https://github.com/oceanprotocol/aquarius/pull/376. This update can be done in   the tables in the "History" page : pool shares, pool transactions, compute jobs and downloads.
Basically instead of fetching the whole ddo for each item just fetch the title.

KatunaNorbert:
Fixes #395 .

Changes proposed in this PR:

- Use endpoint to get only the asset name

### DIFF
--- a/src/components/molecules/AssetListTitle.tsx
+++ b/src/components/molecules/AssetListTitle.tsx
@@ -29,11 +29,12 @@ export default function AssetListTitle({
     const source = axios.CancelToken.source()
 
     async function getAssetName() {
-      getAssetsNames([did], config.metadataCacheUri, source.token).then(
-        (resp) => {
-          setAssetTitle(resp[did])
-        }
+      const title = await getAssetsNames(
+        [did],
+        config.metadataCacheUri,
+        source.token
       )
+      setAssetTitle(title[did])
     }
 
     !ddo && did && getAssetName()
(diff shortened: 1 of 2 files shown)


src/components/molecules/AssetListTitle.tsx
```diff
@@ -28,15 +28,15 @@ export default function AssetListTitle({
...
     const source = axios.CancelToken.source()
 
-    async function getDDO() {
-      const ddo = await retrieveDDO(did, config.metadataCacheUri, source.token)
-
-      if (!ddo) return
-      const { attributes } = ddo.findServiceByType('metadata')
-      setAssetTitle(attributes.main.name)
+    async function getAssetName() {
+      getAssetsNames([did], config.metadataCacheUri, source.token).then(
```
kremalicious, line 32: just for consistency, we usually use async/await for resolving promises so something like this would also end up shorter:

```js
const title = await getAssetsNames()
setAssetTitle(title)
```

kremalicious:
this can be marked as ready for review, right? Already tested a bit and could not find any issues

KatunaNorbert:
Yes, I was also looking into the refreshing transactions table issue, thought it would be easier to find the problem but I didn't managed to do it yet. I saw that there was an issue opened related to this, so that bug can be done over there and we can close this.

kremalicious:
yup, I also secretly hoped by doing this PR we solve #414 but its cause remains a mystery. Now we know it's not caused by the DDO fetching I guess

kremalicious (approved):

PR #417, opened 2021-03-02
Edit compute dataset

### LINKED ISSUE #344: Edit compute dataset: the way to approve algorithms
From https://github.com/oceanprotocol/multi-repo-issue/issues/53

For data assets with compute services the publisher needs to edit the compute settings. Ideally we don't mix these with metadata editing because in the future there will be more options/ flows in this section.
Fields on this page :
~~a) Allow raw algorithms - checkbox~~
b) Allow all published algorithms - checkbox
c) Publisher trusted algorithms - a list of trusted published algorithms

KatunaNorbert:
Fixes #344 .

Changes proposed in this PR:

- Provide a page where user can change compute settings of data assets

Screen Shot 2021-03-22 at 14 26 26

Screen Shot 2021-03-22 at 14 30 35


KatunaNorbert:
Any thoughts on how that publisher trusted algorithms list should work?

mihaisc:
Basically you need a component to show all the published algorithms and the publisher can select a few of them as trusted

src/components/organisms/AssetContent/index.tsx
```diff
@@ -91,6 +104,16 @@ export default function AssetContent(props: AssetContentProps): ReactElement {
                   <Button style="text" size="small" onClick={handleEditButton}>
                     Edit Metadata
                   </Button>
+                  <span className={styles.separator}>|</span>
```
kremalicious, line 107: separator should be part of the condition in next line, no need to show it when only 1 button exists

src/components/organisms/AssetActions/Edit/EditComputeDataset.tsx
```diff
@@ -0,0 +1,249 @@
...
+  const source = axios.CancelToken.source()
+
+  const hasFeedback = error || success
+
+  interface AlgorithmOption {
+    did: string
+    name: string
+  }
+
+  async function getAlgorithms() {
```
kremalicious, line 94: would suggest moving this method out of this component, either above the component function in this file or as some util method, maybe in `utils/aquarius`. Seems to just need `metadataCacheUri` as param then.

Same goes for the TypeScript interface, this should never be within a component function but always outside.

src/components/organisms/AssetActions/Edit/EditComputeDataset.tsx
```diff
@@ -0,0 +1,249 @@
...
+      })
+    })
+    setAlgorithms(algorithmList)
+  }
+
+  useEffect(() => {
+    getAlgorithms()
+  }, [])
+
+  async function createTrustedAlgorithmList(
```
kremalicious, line 126: same for this method, this does not seem to rely on any changes within this component so we can just move it out

src/components/organisms/AssetActions/Edit/EditComputeDataset.tsx
KatunaNorbert: right, I'll move it to `utils/aquarius`, than it could also be used in #343

KatunaNorbert:
Not exactly sure what you were thinking when you said to center, just center everything like this?

Screenshot 2021-03-10 at 13 49 54

kremalicious:
yeah, this would need some overall design session to unify that pattern of "secondary screens" we are getting into more and more now. We already have edit metadata & publish feedback as such screens and so far we are not consistent between them. Just like with this screen they could use some consistent UI elements in all of them for going back, cancel and such. But all that is out of scope for moving forward here.

For this case, yeah, why not centering everything. I was just thinking about centering the content box but I get that then the page header group looks a bit weird left-aligned.

kremalicious:
this is now in desperate need of a rebase/merge against `feature/compute`. We should then be able to replace the `select` used in here with the new `AssetSelection` component, it accepts a `multiple` prop for the use case in here

src/utils/aquarius.ts
```diff
@@ -97,3 +99,47 @@ export async function getAssetsNames(
     }
   }
 }
+
+export async function getAlgorithmsOptions(
```
kremalicious, line 103: nice, will also help getting the `Compute/FormComputeDataset.tsx` smaller in https://github.com/oceanprotocol/market/pull/439

2 small things:
1. naming: always hard, and I can see how `getAlgorithms()` is too generic since they are returned in a specific way which `...Options` tries to convey but it's also very generic. Maybe we should be precise with `getAlgorithmsForAssetSelection()` or `getAlgorithmsForSelection()` or something else to signal what is returned, or the use case
2. config typing: we can type the `config` parameter, usually `ConfigHelperConfig` | `Config` works

src/components/molecules/FormFields/AssetSelection.tsx
```diff
@@ -26,9 +26,9 @@ export default function AssetSelection({
           <input
             id={slugify(asset.name)}
```
kremalicious, line 27: this is leading to a bug I had put in, when asset names are the same in the list. It's fixed in start compute branch and we should also add the fix here so we do not run into that bug when testing this branch. So just switch here `asset.name` to `asset.did`

src/components/organisms/AssetActions/Edit/FormEditComputeDataset.tsx
```diff
@@ -0,0 +1,111 @@
...
+        break
+      default:
+        value = JSON.parse(e.target.value)
+    }
+    setFieldValue(field.name, value)
+    validateField(field.name)
+  }
+
+  useEffect(() => {
+    algorithmList &&
```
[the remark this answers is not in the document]
KatunaNorbert: cool

kremalicious:
Updated all the copy, and hardcoded `allowRawAlgorithm: false` in the background cause couldn't quickly figure out how to not pass it because of the `ServiceComputePrivacy` typing from ocean.js.

This is what we have now:

Screen Shot 2021-03-18 at 15 32 19

From UI side, _All_ & _Selected_ options are now exclusive so UI should reflect that. So we should disable the whole _Selected Algorithms_ part when user selects _Allow any published algorithm_. For this simple dim seems enough:

Screen Shot 2021-03-18 at 15 36 42

that's `opacity: 0.4; pointer-events: none;` on the `.selection` component.

And while testing found a bug I might have introduced with copy/field changes: 
1. open compute settings
2. click _Allow any published algorithm_
3. Submit button becomes disabled

KatunaNorbert:
Nice, just tested the search and works well, really a game changer. A bug I see here is that the selected algorithms are unchecked after the search(enter a value in search and then delete that value), but I'll work on that.
About the bug you mentioned I couldn't reproduce it.

kremalicious:
awesome, thanks for looking into it. Just pushed last change from me, some usability tweaks for selecting assets, and also outputting DID, seemed useful

Screen Shot 2021-03-18 at 16 29 50

KatunaNorbert:
Looks good, I was also thinking that it would be useful to sort the list so that the selected algorithms appear first, it would help when editing.

kremalicious:
yeah, so much room for micro-interactions! Would be useful to have selected items at top of list when they are selected in the DDO already, so on initial mount of screen. If that's simple to do then we should do it, but solving the bugs has priority.

Also getting this in browser console whenever I select an algorithm:
Screen Shot 2021-03-18 at 17 14 06

I am not fully sure but we might have a problem here with form data flow, somewhere we probably modify the Formik form `values` without going through Formik. When modifying the form `values` we usually have to use that Formik field helper `setValue` https://formik.org/docs/api/useField#fieldhelperprops, in code base we do this for datatoken form field https://github.com/oceanprotocol/market/blob/main/src/components/molecules/FormFields/Datatoken/index.tsx#L11-L16

What usually helps to figure out data flow is to use our debug mode and just like on metadata creation/editing we live-output the formik form state under the form

src/components/organisms/AssetActions/Edit/EditComputeDataset.tsx
```diff
@@ -0,0 +1,189 @@
...
+        ) : (
+          <>
+            <p className={styles.description}>{content.description}</p>
+            <article className={styles.grid}>
+              <>
+                <FormEditComputeDataset
+                  title={content.form.title}
+                  data={content.form.data}
+                  setShowEdit={setShowEdit}
+                  values={initialValues}
```
kremalicious, line 179: this is the trick I think, we are modifying in the `FormEditComputeDataset` subcomponent those `values` but these are not the `values` we are looking for. The actual form `values` should be on here https://github.com/oceanprotocol/market/pull/417/files#diff-12dc7146bf292ba49087391d56cba42ed37a4b4339410e318535a430235225ffR155

We then pass those as `values`. And we pass `initialValues` as a new prop `initialValues`

src/components/organisms/AssetActions/Edit/FormEditComputeDataset.tsx
```diff
@@ -0,0 +1,114 @@
...
+  values: ServiceComputePrivacy
+  algorithmList: AssetSelectionAsset[]
+}): ReactElement {
+  const { accountId } = useWeb3()
+  const { ocean } = useOcean()
+  const {
+    isValid,
+    validateField,
+    setFieldValue
+  }: FormikContextType<ServiceComputePrivacy> = useFormikContext()
```
kremalicious, line 33: or even simpler, you can get `values` & `setValues` & `initialValues` exported from `useFormikContext()`. So we do not need to pass them all as props and just work in this component. Then in this component only work with `setValues()` instead of `values.push`

kremalicious:
and switch to UI side. Just double checking this flow:

- user has previously allowed selected algorithms 
- edits compute setting again, but now checks `Allow any published algorithm`

What has precedence then in the background? The previously selected algorithms, or the allow all setting? Do we need to remove existing trusted algorithms for this flow?

KatunaNorbert:
I see, the error doesn't appear inside my console, but I avoided changing values directly and used `set` method, hope this fixes it
Also did some changes, added a `checked` field to `AssetSelectionAsset` so we don't have to set the checked value every time an input is rendered but rather store it on `AssetSelectionAsset` object, also fixed the error I mentioned above
And sorted the list so the selected algorithm are now first in list

kremalicious:
Cool, was also hacking a bit around, think we still do too much and could leave capturing and setting values to Formik, it seems we do both, modifying form values, and the transformation of those into `ServiceComputePrivacy` all manually.

Regardless, I do not get to compute edit screen now, the first `forEach` error already appears when going to asset details page. Upon clicking edit, I get blank screen with next error. Seems code does not account for case where user does not have anything edited before, and therefore no `selectedAlgorithms` or `assets` to iterate or filter through:

Screen Shot 2021-03-19 at 11 09 14

KatunaNorbert:
That's strange, I see that there is a message `No account connected` on asset detail page. As soon as that disappears it works

kremalicious:
still wrapping my head around that data flow, what I meant with we might do too much is that Formik and its `Field` will have all sorts of states for us to use. Like, we might not need to manually keep track of checked algorithms, it's all there in the React state of the `Field`, and would be also available in the form Formik state:

Screen Shot 2021-03-19 at 12 26 29

Still have this issue where I can't select algorithms, but now with different error. To help with that, added a debug output under the form

kremalicious:
the other data flow problem is the mix between `ServiceComputePrivacy` values in the DDO vs. in the form. Current implementation where we keep track of selected algorithms in the state of parent `AssetContent` will lead to them getting out of sync:
- user goes to edit compute settings
- selects some algorithms, but then cancels the screen
- state in `AssetContent` does not reflect the reality of the set algorithms in the DDO anymore 

And somehow I was able to select an algorithm (but now not anymore). We should remove the re-sort on algorithm selection, that was too surprising and felt like my algorithm selection disappeared. Fine to only have selected at top on initial mount

kremalicious:
more findings from a call:

- `allowNetworkAccess` is not what we think it is, we need to make sure to never reset it and always have it reflect what's in the DDO
- the _Allow any published algorithm_ setting is implied from the `publisherTrustedAlgorithms` array. If it's empty, all algorithms are allowed

That last part is a mistake which will be fixed in ocean.js, so we will get a new value like `allowAllAlgorithms` we read and set in here

kremalicious:
honestly, after lots of tinkering think it's most simple to go with the discussed option 2, where we split form & DDO data structure. So I end up with this basically in the debug output:

Screen Shot 2021-03-19 at 15 40 10

Did lots of refactor for this and will try to rebase them against your latest changes and see how that goes. Many advantages, was able to completely ditch our custom field change/add algorithms/remove algo

KatunaNorbert:
Alright, I won't push other changes

kremalicious:
Form data flow and transformation seems under control now with #448.

The final step is to figure out what exact data we need to send because we do not capture everything what's required in `ServiceComputePrivacy`:

- `allowAllAlgorithms` (working title) → Required to be added to ocean.js (this currently makes the tests fail)
- `publisherTrustedAlgorithms` → all good, populated from `AssetSelection`
- `allowRawAlgorithm` → untouched, always pass through and send whatever is in DDO
- `allowNetworkAccess` → untouched, always pass through and send whatever is in DDO

Ideally we never send `allowRawAlgorithm` & `allowNetworkAccess` but think we do not get around this with ocean.js.

In our form we then only capture `allowAllAlgorithms` & `publisherTrustedAlgorithms` and need to have a relationship between them:

- when `allowAllAlgorithms` is checked:
  - disable/dim `AssetSelection` UI
  - upon submit, existing `publisherTrustedAlgorithms` need to be removed, so we send an empty array 
- when `allowAllAlgorithms` is unchecked:
  - enable `AssetSelection` UI
  - upon submit, set to `false`

kremalicious:
nice, all seems in place. Totally bonus points for solving the disabled handling in the perfectly proper accessible way. I would have just hacked in `pointer-events: none`, yolo 🥳 

I think then the only thing left is that ocean.js gives us `allowAllAlgorithms`. And maybe some form validation checks although that could be alright as is right now

src/utils/compute.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+  return trustedAlgorithms
+}
+
+export async function transformComputeFormToServiceComputePrivacy(
+  values: ComputePrivacyForm,
+  ddo: DDO,
+  ocean: Ocean
+): Promise<ServiceComputePrivacy> {
+  // pass through allowRawAlgorithm & allowNetworkAccess from DDO
```
kremalicious, line 29: I'm tending towards ditching that and setting both, `allowRawAlgorithm` & `allowNetworkAccess` further down in the final `privacy` return to `false` by default. While those will be the defaults for every asset published through our market, we would then need to make sure to switch them depending on if `allowAllPublishedAlgorithms` is set or not. Since we do not expose these 2 options right now most simple way should be to just set them always to `false`

kremalicious (approved):
did final test run, all good as it looks like! Only thing I could find is that metadata history does not get refetched when DDO changes but minor for now and existing issue on live too

we can do the empty state for `AssetSelection` over in #439 since there we are going to need it and can do it in context

src/hooks/usePublish.ts
```diff
@@ -115,6 +115,7 @@ function usePublish(): UsePublish {
           const origComputePrivacy: ServiceComputePrivacy = {
             allowRawAlgorithm: true,
             allowNetworkAccess: false,
+            allowAllPublishedAlgorithms: true,
```
kremalicious, line 118: wait, upon publish we actually want this to be `false`

KatunaNorbert: oh ok, my bad, thanks for fixing

Issue #401, opened 2021-02-24
When connected to unsupported network the market freezes


mihaisc:
If you have metamask connected to the market and you switch to an unsupported network the market UI freezes and you need to close the tab.

Quick suggestion : ignore the network switch in `NetworkMonitor` if the network id is not supported , not sure this is enough

PR #421, opened 2021-03-02
refactor network name output

### LINKED ISSUE #412: Better representation of the selected chain/network
Currently if you are connected to the ethereum mainnet we don't show anything in the wallet component, if you are connected to rinkeby/ropsten we show a warning sign and show the name of the network.
Once we go multi chain we need to show the chain as well so we need to show that.

One suggestion would be to show the chain name near the wallet so that the user knows all the time on what chain he is.
image
We can keep the same rules for the network: if on mainnet  (either eth mainnet or polygon mainnet, or whatever other chain in the future) don't show anything, and if on testnet show warning + network name

kremalicious:
Fixes #412.

Changes proposed in this PR:

- move network detection and warning out of generic Web3 feedback component, so it's not displayed in popover anymore
- output beside account button, based on app config or user network
- output nothing when connected to Mainnet
- fetch chain metadata from [ethereum-lists/chains](https://github.com/ethereum-lists/chains) on build time, for mapping `networkId` to display name among other things
- detect supported networks based on ocean.js `ConfigHelper` configs, and error icon output based on it
- cleanup `utils/wallet.ts`, remove unused or redundant methods

Resulting in:

Supported & test network:

Screen Shot 2021-03-03 at 03 12 34

Unsupported & test network:

Screen Shot 2021-03-03 at 03 38 18

### DIFF
--- a/app.config.js
+++ b/app.config.js
@@ -4,11 +4,17 @@ module.exports = {
   // networks in their wallet.
   // Ocean Protocol contracts are deployed for: 'mainnet', 'rinkeby', 'development'
   network: process.env.GATSBY_NETWORK || 'mainnet',
+
+  // List of supported networkId the market works with. Used for showing UI hints.
+  supportedNetworks: [1, 3, 4, 137, 80001],
+
   infuraProjectId: process.env.GATSBY_INFURA_PROJECT_ID || 'xxx',
+
   // The ETH address the marketplace fee will be sent to.
   marketFeeAddress:
     process.env.GATSBY_MARKET_FEE_ADDRESS ||
     '0x903322C7E45A60d7c8C3EA236c5beA9Af86310c7',
+
   // Used for conversion display, can be whatever coingecko API supports
   // see: https://api.coingecko.com/api/v3/simple/supported_vs_currencies
   currencies: [
--- a/app.config.js
+++ b/app.config.js
@@ -5,9 +5,6 @@ module.exports = {
   // Ocean Protocol contracts are deployed for: 'mainnet', 'rinkeby', 'development'
   network: process.env.GATSBY_NETWORK || 'mainnet',
 
-  // List of supported networkId the market works with. Used for showing UI hints.
-  supportedNetworks: [1, 3, 4, 137, 80001],
-
   infuraProjectId: process.env.GATSBY_INFURA_PROJECT_ID || 'xxx',
 
   // The ETH address the marketplace fee will be sent to.
--- a/.gitignore
+++ b/.gitignore
@@ -12,4 +12,5 @@ public/storybook
 .artifacts
 .vercel
 repo-metadata.json
+networks-metadata.json
 src/@types/apollo
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -11,6 +11,14 @@ execSync(`node ./scripts/write-repo-metadata > repo-metadata.json`, {
 // Generate Apollo typings
 execSync(`npm run apollo:codegen`, { stdio: 'inherit' })
 
+// Fetch EVM networks metadata
+execSync(
+  `node ./scripts/write-networks-metadata > content/networks-metadata.json`,
+  {
+    stdio: 'inherit'
+  }
+)
+
 exports.onCreateNode = ({ node, actions, getNode }) => {
   createFields(node, actions, getNode)
 }
--- a/scripts/write-networks-metadata.js
+++ b/scripts/write-networks-metadata.js
@@ -1,0 +1,11 @@
+#!/usr/bin/env node
+'use strict'
+
+const axios = require('axios')
+
+// https://github.com/ethereum-lists/chains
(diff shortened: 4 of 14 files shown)


mihaisc:
Not really solves the issue, or maybe i am misunderstanding something.
The main idea is that in the future we will have multiple "mainnet" for each chain that we implement. My suggestion was something across these lines:
- eth mainnet
image
- eth test net (any of them) and continue to show network in popup like we used to do
image
- polygon mainnet ( or any other chains) 
image

We can add the chain name in `ConfigHelper` with any other stuff that we need

kremalicious:
right, each chain can have multiple networks so to speak, and PR only accounts for current status quo.

Warning icon:
Logic as in your mockups makes sense then, it's basically warning when stuff is not "real", as in tokens representing real money value being transacted. And that would be then true for other chain's main networks too.

Another possibility would be to redefine the warning condition. Was also thinking of not using warning icon at all for all supported networks. But use it to signal when user is connected to an unsupported network, for which we usually can still output the network name.

As for network name label:
- would still want to hide that for ETH mainnet, it's our default, and also reflects how other dapps handle that. User expectation in dapps is that it works against ETH mainnet.
- for ETH test networks think it's fine showing the respective network name. They are usually well known and attached to ETH in user's minds.
- for every other networks should be fine only showing the chain name, like _Polygon_ as in the mock. That is unless they also have and encourage to use their own test networks. In that case we should also add e.g. _Testnet_ after the network name

kremalicious:
After some researching, tending towards outsourcing the network naming standardization and just fetch all metadata from https://github.com/ethereum-lists/chains

They capture _chain_ & _network_ which we could combine, so e.g. _ETH Rinkeby_ or _Matic Mainnet_ based on their data structure: https://chainid.network/chains.json. This should then cover all future chains/networks as long as they are in that list, which they probably are.

There's also _nativeCurrency_ which might help with #418

mihaisc:
I think i agree with everything you said. Summary :
- don't show chain for eth main, show only for other chains. for now we don't really want to add other test networks. The purpose of rinkeby and ropsten is to test the market not the chain itself. So we already have 2 testnets no need for more, unless some specific scenario
image
- change warning icon for testnets, something like (obviously make it look proper)
image

- integrate with https://chainid.network/

kremalicious:
alrighty, with chainid.network we get those network naming combinations if we want to keep it flexible:

Screen Shot 2021-03-03 at 01 05 21

As per https://chainid.network/chains/#matic-mainnet--, for Matic it would be _Matic Mainnet_ then. 

The goal I had in mind was to make this network name output completely network agnostic, meaning we simply output whatever user is connected to, regardless if we support it or not. So the source of truth for all network metadata becomes that community-maintained repo.

Like that idea of just using _test_ label but would keep it around network name, that account element is now exclusively for the account info. The condition based on chainid.network response data, `if (network. network) !== 'mainnet'`, should be safe for that.

With that, maybe it makes sense to use the warning icon then for networks we do not support. Although I guess everything crashes anyway right now when user is connected to unsupported network but ideally user should be able to connect to whatever network and our UI should give hints if it works or not for which warning icon seems a better use.

mihaisc:
Not crashing can be patched easily. It's not a proper fix, but it will not crash.
Agree with everything.

mihaisc:
So... just slapping a `if (!initialNewConfig) return` in `NetworkMonitor` will make the app not crash but then you don't detect the new network  so it's kinda useless

mihaisc:
Not sure why but sometimes it gets stuck, even when changing through supported networks 
image

app.config.js
```diff
@@ -4,11 +4,17 @@ module.exports = {
   // networks in their wallet.
   // Ocean Protocol contracts are deployed for: 'mainnet', 'rinkeby', 'development'
   network: process.env.GATSBY_NETWORK || 'mainnet',
+
+  // List of supported networkId the market works with. Used for showing UI hints.
+  supportedNetworks: [1, 3, 4, 137, 80001],
```
mihaisc, line 9: do we need this? we already have a list of supported networks [here](https://github.com/oceanprotocol/ocean.js/blob/main/src/utils/ConfigHelper.ts) so something like `ConfigHelper().getConfig()` returning `null` means we don't support it. I want to prevent having multiple lists with the same meaning.

kremalicious: great point, will remove this and rather cross-reference with `ConfigHelper`. Btw, `80001` is the Matic Mumbai test network but I'm pretty sure we did not deploy contracts to it. But we maybe should before release and then add it to `ConfigHelper`

kremalicious:
one remaining thing: response from https://chainid.network can be slow, it's just Jekyll site on GitHub Pages. So seems better to fetch the list on build time into Gatsby data structure, and query that on client side so it's instant

kremalicious:
works pretty well by now in my testing, getting name is now instant. Will not touch any other network init stuff in here and leave that to #401.

mihaisc (approved):

PR #422, opened 2021-03-04
Check previous order from the graph

### LINKED ISSUE #420: Get order id from the graph instead of web3
When consuming an asset we check if it was bought using `checkPreviousOrder`. We need to do this check on the graph. It will be faster and it currently can't be done using web3 on other chains except eth because of a cap in searching previous blocks

mihaisc:
Closes #420


kremalicious (approved):
all perfect, works as before in my testing

Issue #432, opened 2021-03-09
UI to switch between chains


kremalicious:
The more chains we run on like #406, the more important it becomes for users to switch between them to see the assets for each chain and their networks. Without a wallet connected, there is currently no way to see assets in another chain/network. 

At the moment we only react to the user wallet's connected network, and if not present connect to default network configured in app on build time (ETH Mainnet). 

### Proposal

As part of user settings popover, include switches for all chains we are deployed to, switching between their Mainnets. As a secondary visual UI, allow switching between the networks of each chain we are deployed to.

Bancor does it plain and simple:

<img width="401" alt="110461231-784f9880-80cf-11eb-8e6e-8156a1e465c7" src="https://user-images.githubusercontent.com/90316/110516609-16f8eb00-810a-11eb-80b8-4c1ae1f3c2f4.png">

- this requires #433, prototype is in there in https://github.com/oceanprotocol/market/commit/fa9fabf256e17eb87aba8e5261f3dbc2e779c3e9
- in our case those buttons switch the data source, but not the user wallet. To reduce confusion probably best to only show those switching buttons when users are not connected with a wallet. When they are connected with their wallet we can assume they switch in their wallet

PR #433, opened 2021-03-10
Web3/Ocean splitup + full Polygon/Matic support

### LINKED ISSUE #145: OCEAN user balance is null randomly

### LINKED ISSUE #401: When connected to unsupported network the market freezes

### LINKED ISSUE #413: Add polygon network/explorer to transaction links component

### LINKED ISSUE #418: Incorrect token displayed in wallet

kremalicious:
In essence: new `useWeb3()` & `useOcean()` maintained in `market` + all the changes from #406 for Polygon/Matic support.

### Polygon/Matic support

- closes https://github.com/oceanprotocol/multi-repo-issue/issues/82
- fixes #418
- fixes #413

Screen Shot 2021-03-16 at 16 07 59

Screen Shot 2021-03-16 at 19 26 01

### Web3/Ocean split-up

* new `Web3Provider`/`useWeb3()` wrapping the app
  * handling web3 connection with web3modal as before
  * handles user account/network changes
  * exposes network metadata from `https://chainid.network`, and constructs a network display name
* move over `OceanProvider`/`useOcean()` from  `react`, highly reduce what it does: 
  * central place for setting, holding, and changing endpoint configs based on `networkId`
  * hold the Ocean-specific `Account` object needed in some places
  * hold the Ocean-specific user balance (this one could be moved to `Web3Provider` but getting ERC-20 token balances without the helper in ocean.js is cumbersome)
  * change all the data logic flow, everything simply changes based on `accountId` & `networkId` from `Web3Provider`
* between both providers, highly optimized data flow to make sure only necessary RPC and Ocean component endpoint calls are made
  * makes network switching crazy fast and more reliable, with way less memory usage
  * fixes many issues randomly popping up caused by too many RPC calls, takes load from all Ocean component endpoints
  * in some cases, especially in test networks, endpoint calls should be down by 66%, app was repeating many of them 3 times
  * verifiable by looking at the highly reduced amount of browser console messages in debug mode
  * fixes #145 for real without workarounds
  * seems to also fix #401
  * truly decouples user wallet network from ocean endpoints, allowing new UI functionality
  * prototype buttons for endpoint switching independent of user wallet network, for verifying #432
* split-up should make finding a solution for unblocking #361 easier
* new `useAccountPurgatory()` hook
* port over all existing hooks as is, including `useCompute()`, which we should then use for #343 to keep pattern consistent
* remove `useMetadata()` hook, replace with direct DDO value access
* kick out `NetworkMonitor`, functionality is integrated in new `Web3Provider`
* completely remove `@oceanprotocol/react` dependency
* because of code removal and refactoring fixes a bunch of CodeClimate issues along the way

includes learnings from https://github.com/oceanprotocol/react/pull/255 and the [redacted] frontend

Touches almost everything so terrible to review, new magic is happening mainly in:

- `Web3Provider`/`useWeb3()`: https://github.com/oceanprotocol/market/pull/433/files#diff-e44ed1e725b9c26ecad28498d80d998baff95ce5a5527ba88fd6bdb8759057ed
- `OceanProvider`/`useOcean()`: https://github.com/oceanprotocol/market/pull/433/files#diff-63869f26cd5f7eb06fcbaf70892ac8e5e0c018adebaa7b7ebd30650d069b7332

Resulting in nice clean logging in debug mode:

Screen Shot 2021-03-12 at 12 22 23


kremalicious:
Some findings to figure out how to deal with the `react` repo

- moving over `OceanProvider` right now implies moving over all hooks, because they itself [call `useOcean`](https://github.com/oceanprotocol/react/blob/main/src/hooks/usePublish/usePublish.ts#L28), leading to drastic code base changes for all functionality
- if we do that then there's not much need for the `react` repo at all
- alternative compromise would be to make all hooks over in `react` accept the values exposed by `useOcean` as props

But if we want to minimize the refactor in this repo and for forks:
- we could go back to original idea and leave the `OceanProvider` over there in its new minimal form:
  - exposing `ocean`, `balance` and purgatory info, reacting to web3 account  changes 
  - accepting `web3Provider`, `accountId`, `networkId` as a prop
- would allow easier migration path for users of `react` and `market` forks

In both cases, leave the web3 handling to the client apps. As for this PR this means continuing on the consolidation of all web3-related stuff into one `Web3Provider/useWeb3`

kremalicious:
I'm just not sure deprecating `react` repo is best way because it increases entry barrier to build an app with ocean. Cause this is what moving all the hooks implies to me, not sure we can handle maintenance of `react` repo if we do not use it ourselves.

Fully agree our own maintenance would be greatly reduced by moving everything.  But the hooks do a bunch of more stuff than being simple helper methods, like the feedback & loading states which we would have to rebuild in each component if we limit them to pure helper methods. So would probably move the hooks as is anyway. This would then mean we will direct people to copy out the respective hooks from `market` into their apps which is of course not as convenient as getting it from npm package.

Removed `useMetadata` already, no use in it. Will first try to get `Web3Provider/OceanProvider` to play nice in this repo before moving on to deal with rest of hooks so still plenty of time to decide what to do with the react repo

mihaisc:
Well, i always said to keep the react hooks. They don't need a lot of maintenance and we can keep them as they are now. They represent the basic usage of our tech. In the market we need more customization, because the market is our interpretation of the tech, it's not the only way.
So :
- move what we need to market  (indeed it's a bit more that i suggested but not that much, we can work on this branch in parallel,  like we do with compute)
- we keep the react hooks as they are now and we fix the example. There is no need to change them after, only update ocean js lib (we can configure dependabot for that) 

Considering the overhead that the react hooks generate now, decoupling the repos and supporting them separately will take less time overall.

kremalicious:
and just like that we are finally able to do stuff like this, quick prototype for verifying changes work for solving #432 in the end:

https://user-images.githubusercontent.com/90316/110710611-77695480-81fe-11eb-99d3-056dc2c2ee5c.mov

kremalicious:
nice, also verified our new recently optimized `Wallet` component was created flexible enough so that this combination now just works through UI button network switching:

Screen Shot 2021-03-11 at 17 30 57

mihaisc (approved):

PR #376, opened 2021-02-11
Compute-to-data

bogdanfazakas:
Fixes https://github.com/oceanprotocol/multi-repo-issue/issues/53

Main development branch where everything from the [Compute-to-Data milestone](https://github.com/oceanprotocol/market/milestone/1) comes together.

Changes proposed in this PR:

- new publish form for Algorithms (#384)
- interface to switch between publishing data set and algorithm (#394)
- separation between data sets & algorithms throughout the app (#388)
- algorithms detail view (#389)
- new search filter bar (#400)
- validate custom Docker images on algorithm publish (#405)
- Asset selection form component (#441)
- Edit compute dataset to allow algorithms (#417)
- Start compute job (#439)
- Help messages & price display for starting compute job (#478)
- Compute job history (#462)
- Subgraph helpers for assetSelection list and previous orders (#488)
- Refactor compute methods calls (#502)
- Job details (#528)

#### Components

| ETH main                | Rinkeby                 | Polygon main            |
| ----------------------- | ----------------------- | ----------------------- |
| <ul><li>- [x] deployed Aquarius `v2.2.8`</li></ul> | <ul><li>- [x] deployed Aquarius `v2.2.8`</li></ul> | <ul><li>- [x] deployed Aquarius `v2.2.8`</li></ul> |
| <ul><li>- [x] deployed Provider `v0.4.9`</li></ul> | <ul><li>- [x] deployed Provider `v0.4.9`</li></ul> | <ul><li>- [x] deployed Provider `v0.4.9`</li></ul> |
| <ul><li>- [x] deployed Subgraph `v1.1.1`</li></ul> | <ul><li>- [x] deployed Subgraph `v1.1.1`</li></ul> | <ul><li>- [x] deployed Subgraph `v1.1.1`</li></ul> |
| <ul><li>- [x] ocean.js `v0.14.x`</li></ul> | <ul><li>- [x] ocean.js `v0.14.x`</li></ul> | <ul><li>- [x] ocean.js `v0.14.x`</li></ul> |
| <ul><li>- [ ] tested with `feature/compute`</li></ul> | <ul><li>- [x] tested with `feature/compute`</li></ul> | <ul><li>- [x] tested with `feature/compute`</li></ul> |

### New UI

Screen Shot 2021-04-08 at 17 18 48

Screen Shot 2021-04-08 at 17 24 11

Screen Shot 2021-04-08 at 17 19 26

Screen Shot 2021-04-14 at 16 25 06

[...]

PR #434, opened 2021-03-10
Reorganize history page

### LINKED ISSUE #374: Reorganize History page

KatunaNorbert:
Fixes #374 .

Changes proposed in this PR:

- reorganize history page

### DIFF
--- a/src/components/pages/History/index.module.css
+++ b/src/components/pages/History/index.module.css
@@ -31,6 +31,9 @@
   border-bottom: 0;
   border-top-left-radius: var(--border-radius);
   border-top-right-radius: var(--border-radius);
+  white-space: nowrap;
+  overflow-y: hidden;
+  overflow-x: auto;
 }
 
 .tabs div[class*='tabContent'] {
--- a/src/components/pages/History/index.module.css
+++ b/src/components/pages/History/index.module.css
@@ -18,24 +18,21 @@
 
 .content {
   margin-top: var(--spacer);
+  background-color: var(--background-content) !important;
 }
 
-.historyTabs [class*='tabList'] {
-  text-align: start;
+.tabs div[class*='tabs'] {
+  background-color: var(--background-content);
+}
+
+.tabs ul[class*='tabList'] {
+  background-color: var(--background-content);
+  border: 1px solid var(--border-color);
   border-bottom: 0;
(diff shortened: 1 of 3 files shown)


mihaisc:
Looks/works nice but the tables need some tweaks. The waves are making part of them hard to read

image

mihaisc:
Dark theme looks better than white theme but i think it's good enough. Only one more issue, if there are no items returned you get this text that blends in the waves

image

kremalicious:
we usually try to put main content on a screen into content boxes, our `Box` style component. For this here it seems we should use the UI pattern already in #376 for the publish form where tab bar and content are visually connected:

Screen Shot 2021-03-11 at 12 42 32

There's also a global background colors change in #376 to separate content & background better but that we do not need to bring over since that requires some drastic changes.

As for mobile, I would suggest a simple overflow pattern for the tabs, like our main menu.

Another thing we need to have is being able to link to respective tabs. But we do not want tab clicking be part of browser history. So not sure we can find a compromise here where tab selection is part of URL but user hitting back button in browser should not make them go through tabs, but really to last page.

kremalicious:
Woah, it looks perfect now and is already a huge usability win for this page 🤩  I would say we can ship that, and figure out improvements when we need them like the linking.

Only interaction improvements I can think of would be to find a way so that not every tab click reloads the data so the box collapse is minimized, e.g. by centralizing data fetching in some parent component but that sounds like a bit too much optimization at this point in time

kremalicious (approved):

PR #435, opened 2021-03-11
fix edit history

mihaisc:
Market crashes on rinkeby
https://market.oceanprotocol.com/asset/did:op:147A0313B3546129b4F8C69b28CaAB1045fcBf25

image


kremalicious:
interesting! Verified the fix works.

Do we know why the subgraph would not have update datatoken info? Is it because asset was only published, but never updated?

kremalicious (approved):

bogdanfazakas (approved):

PR #437, opened 2021-03-12
Add custom token from the wallet

### LINKED ISSUE #411: Add custom token from the wallet
We need to push custom tokens from the market. Maybe from the wallet component. 
This feature is required mostly for other chains. For example on Polygon we need to push to metamask the mOcean token so the user doesn't need to add it manually in metamask.

There is an example here https://github.com/MetaMask/Add-Token

KatunaNorbert:
Fixes #411 .

Changes proposed in this PR:

- user should be able to add custom token to metamask from the wallet


KatunaNorbert:
Should there be a form inside the wallet that allows to enter a custom token informations(address, symbol, decimal) and then add it to metamask?

kremalicious:
ah, good thinking but from UI it can be simple: it's _just_ a button e.g. _Add mOCEAN to MetaMask_, upon click a method is called with predefined values, we should know all token infos. Button would call this method: https://github.com/MetaMask/Add-Token/blob/master/src/AddTokenPanel.jsx#L126-L137

We then most likely have to figure out how to map detected token to required token info to make it super generic, there are various token list packages out there we could utilize. Then we could use that component for any token we like. 

But we can also start small in here, and only solve it for the Matic/mOCEAN use case first, so we could hardcode that token metadata in the component.

KatunaNorbert:
I see, I didn't know how much token types we need to cover. In the case of adding only mOCEAN definitely make more sense to mock the data and show only a button

kremalicious:
As for placement, this might work since it should only appear with MetaMask anyway:

Screen Shot 2021-03-12 at 13 25 06

Or maybe like this, we could give the wallet popover a bit more space so everything fits: 

Screen Shot 2021-03-12 at 13 27 25

kremalicious:
For starting small we could have the condition metamask + Matic to show the button, and only deal with that.

But instead of super generic we could make it generic for OCEAN only, which would increase the usefulness of the feature. Then people could add all OCEAN in all the networks to their wallet, cause they would have to do that for each network. The token addresses for OCEAN are simple to get, they are always exposed under `config.oceanTokenAddress`, and I think also mOCEAN then under same config key for Matic. If we do that, we would cover the condition metamask + all networks to show the button.

KatunaNorbert:
The first one looks better for me, that's also how I imagined it. The only thing that might create confusion is the token name, having OCEAN above and MOCEAN on the button.
Second one is more intuitive, but the design looks a bit overloaded.

kremalicious:
yup, that OCEAN/mOCEAN labeling was a known issue but only applies to Matic. Not fully sure but #406 might solve it, otherwise I agree we need to make the token name consistent in UI

src/components/molecules/Wallet/Details.tsx
```diff
@@ -23,6 +25,33 @@ export default function Details(): ReactElement {
...
+    provider.sendAsync(
+      {
+        method: 'metamask_watchAsset',
+        params: {
+          type: 'ERC20',
+          options: {
+            address: config.oceanTokenAddress,
+            symbol: 'OCEAN',
+            decimals: 18,
+            image: Logo
```
kremalicious, line 40: would be great if this works but not sure it will. Found the docs: https://docs.metamask.io/guide/registering-your-token.html#code-free-example

They say "A string url of the token logo" so I suppose they will always fetch it remotely. For which we have an asset in our `art` repo: https://github.com/oceanprotocol/art/blob/main/logo/token.png

So we should be able to use the raw URL for it: `https://raw.githubusercontent.com/oceanprotocol/art/main/logo/token.png`

KatunaNorbert: Right, I'll change that

src/components/molecules/Wallet/Details.tsx
```diff
@@ -9,9 +9,10 @@ import { formatCurrency } from '@coingecko/cryptoformat'
 import { useUserPreferences } from '../../../providers/UserPreferences'
 
 export default function Details(): ReactElement {
-  const { balance, connect, logout, web3Provider } = useOcean()
+  const { balance, config, connect, logout, web3Provider } = useOcean()
   const { locale } = useUserPreferences()
   const [providerInfo, setProviderInfo] = useState<IProviderInfo>()
+  const provider = window.web3.currentProvider
```
kremalicious, line 15: we need to always use the initialized `web3` & `web3provider` from `useOcean()` from within components. Otherwise this code here will break app in browsers without MetaMask installed. Also, `window.web3` is deprecated, see the browser console:
Screen Shot 2021-03-15 at 16 23 51

kremalicious:
And just like that, it just works, great!. Tested on ETH Mainnet & Rinkeby & Matic Mainnet. For Matic I would suggest we change the pushed `symbol` to `mOCEAN`.

As for what exact code to Usenet sure about differences between their example code and example app. The example code here seems more clean because of async/await usage https://docs.metamask.io/guide/registering-your-token.html#example but in the end it doesn't matter since we do not have any UI feedback for this

src/components/molecules/Wallet/Details.tsx
```diff
@@ -23,6 +27,28 @@ export default function Details(): ReactElement {
...
+          symbol: networkName.localeCompare('Matic') === 1 ? 'mOCEAN' : 'OCEAN',
+          decimals: 18,
+          image:
+            'https://raw.githubusercontent.com/oceanprotocol/art/main/logo/token.png'
+        }
+      },
+      id: Math.round(Math.random() * 100000)
+    })
+    if (wasAdded) {
+      console.log('Added')
```
kremalicious, line 46: that's nice, to make this useful for debugging we should put it behind our ocean.js `Logger.log()` and then maybe output like so:

```js
Logger.log(`Added ${tokenMetadata.params.symbol} (${tokenMetadata.params.address})`)
```

Should work easily if we move the data passed to `sendAsync()` into a const `tokenMetadata` at beginning of `addOceanToWallet()` method

content/pages/cutomToken.json
```diff
@@ -0,0 +1,28 @@
+{
```
kremalicious, line 1: we can remove this file then now that we have the more simple flow

src/components/molecules/Wallet/Details.tsx
```diff
@@ -23,6 +27,28 @@ export default function Details(): ReactElement {
...
   }, [web3Provider])
 
+  async function addOceanToWallet() {
+    const wasAdded = await web3Provider.sendAsync({
+      method: 'metamask_watchAsset',
+      params: {
+        type: 'ERC20',
+        options: {
+          address: config.oceanTokenAddress,
+          symbol: networkName.localeCompare('Matic') === 1 ? 'mOCEAN' : 'OCEAN',
```
kremalicious, line 37: Think it's safer to rely on the `networkId` here. Because we get this name from data in https://chainid.network, and I expect the Polygon people to rename their network over there in next weeks. If we base our condition on `networkId`, then we do not need to change anything when network names change. 

The `networkId` for Matic Mainnet is `137`

src/components/molecules/Wallet/Details.tsx
```diff
@@ -26,6 +27,39 @@ export default function Details(): ReactElement {
     setProviderInfo(providerInfo)
   }, [web3Provider])
 
+  async function addOceanToWallet() {
+    const tokenMetadata = {
+      type: 'ERC20',
+      options: {
+        address: config.oceanTokenAddress,
+        symbol: networkId === 137 ? 'mOCEAN' : 'OCEAN',
```
mihaisc, line 35: this should be in config as well `config.oceanTokenSymbol`, we must work under the premise that we have multiple chains

src/components/molecules/Wallet/Details.tsx
```diff
@@ -26,6 +27,39 @@ export default function Details(): ReactElement {
...
+    }
+    web3Provider.sendAsync(
+      {
+        method: 'wallet_watchAsset',
+        params: tokenMetadata,
+        id: Math.round(Math.random() * 100000)
+      },
+      (err: string, added: any) => {
+        if (err || 'error' in added) {
+          Logger.log(
```
kremalicious, line 49: nice, I like the more user-focused error messaging, has small typo `Counld't` at beginning. For errors we can then really use `Logger.error()` instead of `.log()`

src/components/molecules/Wallet/Details.tsx
KatunaNorbert: yes this would be nice but I can't access it from `config`

kremalicious: fully agree, but not sure we will solve this as part of this PR so the dedicated detection here should be fine for launching it. I also hacked in a bunch of temporary `networkId === 137` checks in web3 branch, also for symbol. So we should first add `config.oceanTokenSymbol` over in ocean.js and after that change all the dedicated network checks. Prepared that PR over on ocean.js https://github.com/oceanprotocol/ocean.js/pull/670

src/components/molecules/Wallet/Details.tsx
```diff
@@ -62,9 +96,11 @@ export default function Details(): ReactElement {
         ))}
 
         <li className={styles.actions}>
-          <span title="Connected provider">
-            <img className={styles.walletLogo} src={providerInfo?.logo} />
-            {providerInfo?.name}
+          <span title="Connected provider" className={styles.walletInfo}>
+            <div>
```
kremalicious, line 100: a block-level HTML tag within an inline-level tag is invalid HTML. Ideally, remove the `div` and see if it works with less markup, otherwise switch the `span` & `div` tags

src/components/molecules/Wallet/Details.tsx
```diff
@@ -75,7 +111,18 @@ export default function Details(): ReactElement {
...
-          </span>
+            {providerInfo?.name === 'MetaMask' && (
+              <Button
+                style="text"
+                size="small"
+                onClick={() => {
+                  addOceanToWallet()
+                }}
+              >
+                Add Ocean
```
kremalicious, line 122: aaaand right, another symbol :-) We could use same check as above, but shouldn't be any blockers to getting that PR on ocean.js released quickly, so we can make use of `oceanTokenSymbol`. Only the CI over there is slow

mihaisc: pr+release at best 30-40 minutes

KatunaNorbert: right, I'll change that after release. I see that with the changes from main the tokenSymbol from wallet `balance` is also changed to `mOCEAN`. Didn't changed what you spotted to keep it consistent to what the value was before changes.

kremalicious: yup, those symbol hacks come in from multiple PRs now. If anybody can approve my PR I can handle the release https://github.com/oceanprotocol/ocean.js/pull/670

KatunaNorbert:
Just updated the oceanJS version and now I can acces the `oceanTokenSymbol` from the `config`, but you missed to add the field to [`Config` class](https://github.com/oceanprotocol/ocean.js/blob/main/src/models/Config.ts), so it doesn't show up on config in the way it is declared right now `config: ConfigHelperConfig | Config`. I can fix it if I remove the config and declare it like this `config: ConfigHelperConfig`, but I think it should be modified in `oceanJS`.

kremalicious:
yes, those OR in the typings in ocean.js are not working as we think it should. In market, I often have to do `(config as ConfigHelperConfig).oceanTokenSymbol` to get it right when accessing properties which only exist on the ConfigHelper. Cause the `oceanTokenSymbol` should not be on the `Config` you pass to ocean.js, it really only is present for the `ConfigHelper`

mihaisc:
What if we drop the OR and instead of `ConfigHelperConfig | Config` we keep just `ConfigHelperConfig ` in OceanProvider.

kremalicious:
I thought so too cause we totally do not ever use the actual `Config` in the market. But when doing the mega-web3-rewrite found this then all was conflicting with `new ConfigHelper().getConfig()` returning a `Config` in ocean.js. So I just left the OR typing without looking what should be different in ocean.js so this works for our case. 

Might be as simple as making `new ConfigHelper().getConfig()` return `ConfigHelperConfig` so we can remove our OR typing in market

mihaisc:
`getConfig` needs some cleaning. It actually returns `ConfigHelperConfig` because it returns one item form the `configs` array . So for now we can just return `new ConfigHelper().getConfig() as ConfigHelperConfig` in `getOceanConfig` (and change the return value from `Config` to `ConfigHelperConfig` ) in `utils/ocean.ts`. And this would actually be correct.
https://github.com/oceanprotocol/ocean.js/blob/b8fc15e4df6e700c809ff937b0ae712cbe00e222/src/utils/ConfigHelper.ts#L191

kremalicious (approved):
works perfectly!

PR #438, opened 2021-03-15
Removing showing updates on never updated assets

### LINKED ISSUE #436: Edit history shows updates on never updated assets
Edit history for any asset shows at least 2 entries, even if it has never been updated after initial creation:

Screen Shot 2021-03-11 at 17 39 08

This is caused by market implementation assuming that in the GraphQL query, `datatokens.updates` only contains updates after the creation, hence the additional query for `createTime` & `tx`. But the `datatokens.updates` has the metadata creation event as the first item, 

There are 2 ways to solve that:
1. in `market` assume the first item in `datatokens.updates` is the creation event, and output accordingly to switch text. Then `datatokens.createTime` & `datatokens.tx` can be removed.
2. in `ocean-subgraph`, remove the creation event from `datatokens.updates` as done in https://github.com/oceanprotocol/ocean-subgraph/pull/54

As [pointed out in `ocean-subgraph`](https://github.com/oceanprotocol/ocean-subgraph/pull/54#issuecomment-796705670) using the actual metadata creation (not the datatoken creation) transaction seems more in line how we display it in the UI so 1. would be preferred.

While we are at it, we could simply output both, datatoken creation, and metadata creation which in turn makes this more transparent too. Like `created datatoken 22 min. ago` & `created metadata 22 min. ago`

DimitarSD:
Fixes #436 
Closes https://github.com/oceanprotocol/ocean-subgraph/pull/54

Changes proposed in this PR:

- Removes `datatokens.createTime` & `datatokens.tx` from the `ReceiptData` query
- Removes showing updates on never updates assets (using only `datatokens.updates`)


kremalicious:
yeah, that package-lock: long story short, with npm v7 (which is default with Node.js v15) the lock file will automatically be updated to `lockfileVersion` v2 upon doing `npm install`. But npm v7 comes with some more challenges for us so we were holding off on too many dependency updates.

For now, what would work for this PR, is to get the old Node.js v14/npm v6 lock file, if you use `nvm`:

```bash
nvm use 14
npm i
git commit
git push
```

This should restore the v1 lock file version. Also we kinda forgot to document the need for npm below v7, since it only arises as problem when adding/changing dependencies.

Other than that, the actual fix works perfectly on first quick look 🥳

mihaisc:
I guess we can do it like this for now. In the future this will be a bug. You can have assets from outside the market that don't have this order. When we split the market and add multiple pools and other flows this will be inaccurate as well. You can create the ddo before the asset very easy.

kremalicious (approved):

PR #439, opened 2021-03-15
Start compute job

### LINKED ISSUE #343: Start compute job on data set

bogdanfazakas:
Fixes #343 .

Changes proposed in this PR:

- added in the use component from asset detaults a list of existing algorithms
- select an algorithm and start compute job

Screen Shot 2021-03-29 at 15 53 03


src/@types/Form.d.ts
```diff
@@ -2,7 +2,7 @@ export interface FormFieldProps {
   label: string
   name: string
   type?: string
-  options?: string[]
+  options?: any[]
```
kremalicious, line 5: in the `AssetSelection` component I created the `AssetSelectionAsset` typing. So we should be able to use `string[]` | `AssetSelectionAsset[]` here instead of `any`

src/utils/aquarius.ts
```diff
@@ -107,56 +108,15 @@ export async function getAssetsNames(
...
-  const query = {
-    page: 1,
-    query: {
-      query_string: {
-        query: `(service.attributes.main.type:algorithm) -isInPurgatory:true`
-      }
-    },
-    sort: { created: -1 }
-  }
+export async function getAlgorithms(
```
kremalicious, line 111: don't think we need this abstraction since method does not specifically get algorithms. So wherever `getAlgorithms()` is used, `queryMetadata()` can be used instead

src/utils/compute.ts
```diff
@@ -39,3 +43,44 @@ export async function transformComputeFormToServiceComputePrivacy(
 
   return privacy
 }
+
+export async function transformDDOToAssetSelection(
```
kremalicious, line 47: I like making this generic to all assets but when looking at this method, it probably should continue living in `utils/aquarius.ts`. Unless you have something in mind changing it to doing something specific to compute

src/utils/compute.ts
```diff
@@ -39,3 +43,44 @@ export async function transformComputeFormToServiceComputePrivacy(
...
+  ddoList: DDO[],
+  metadataCacheUri: string,
+  selectedAlgorithms?: PublisherTrustedAlgorithm[]
+): Promise<AssetSelectionAsset[]> {
+  const source = axios.CancelToken.source()
+  const didList: string[] = []
+  const priceList: any = {}
+  ddoList.forEach((ddo: DDO) => {
+    const did: string = web3.utils
```
kremalicious, line 56: missed that in the edit compute PR: `ddo.id` is the DID. No need to get it from the datatoken address.

src/components/organisms/AssetActions/Compute/index.tsx
```diff
@@ -0,0 +1,317 @@
...
+          options={['dataset-1', 'dataset-2', 'dataset-3'].map((x) => x)}
+          // onChange={handleSelectChange}
+        />
+      ) : (
+        <Formik
+          initialValues={getInitialValues}
+          validationSchema={validationSchema}
+          onSubmit={async (values, { resetForm }) => {
+            // move user's focus to top of screen
+            window.scrollTo({ top: 0, left: 0, behavior: 'smooth' })
```
kremalicious, line 280: this is unneeded here since we do not make the whole screen disappear like on publish. And here it would be a bad experience on mobile pushing user to top of screen after starting compute job

src/components/organisms/AssetActions/Compute/index.tsx
```diff
@@ -0,0 +1,317 @@
...
+        config.metadataCacheUri,
+        []
+      )
+    }
+    return algorithmSelectionList
+  }
+
+  useEffect(() => {
+    if (!ddo) return
+    async function setDatatokenSymbol(ddo: DDO) {
```
kremalicious, line 155: unneeded, it's available in the DDO under `ddo.dataTokenInfo.symbol` so we do not have to fetch anything

src/utils/aquarius.ts
bogdanfazakas: good point 👍

src/utils/compute.ts
bogdanfazakas: sure I'll move it, I did that because it has nothing to do with aquarius so i thought that it was more related to the compute, but you are right we can use for more generic purpose

src/utils/compute.ts
bogdanfazakas: hmm true, i'll remove the extra logic

src/components/atoms/Input/InputElement.module.css
```diff
@@ -100,6 +100,12 @@
   padding-left: 0.5rem;
 }
 
+.algorithmLabel {
```
kremalicious, line 103: can't quickly see how we use this but this seems too specific to be put into our generic input styles, and should probably live somewhere else

src/components/organisms/AssetActions/Compute/index.tsx
```diff
@@ -0,0 +1,300 @@
...
+          }}
+        >
+          <FormStartComputeDataset
+            algorithms={algorithmList}
+            ddoListAlgorithms={ddoAlgorithmList}
+            setSelectedAlgorithm={setSelectedAlgorithmAsset}
+            isLoading={isJobStarting}
+            isComputeButtonDisabled={isComputeButtonDisabled}
+            hasPreviousOrder={
+              hasPreviousDatasetOrder || hasPreviousAlgorithmOrder
```
kremalicious, line 278: this most likely will have to be tweaked cause now we have to consider 2 different previous orders. Which sounds like we should update the messaging accordingly in `ButtonBuy`, and then tweak what we pass here

src/components/organisms/AssetActions/Compute/index.tsx
```diff
@@ -0,0 +1,300 @@
...
+
+  return (
+    <>
+      <div className={styles.info}>
+        <File file={file} small />
+        <Price price={(ddo as DDO).price} conversion />
+      </div>
+
+      {type === 'algorithm' ? (
+        <Input
```
kremalicious, line 253: I'm in favor of removing this functionality from the current scope for timing reasons. We can replace this with some info message directing users to data sets to run compute jobs or something

src/components/organisms/AssetActions/Compute/index.tsx
```diff
@@ -0,0 +1,300 @@
...
+            dtSymbol={ddo.dataTokenInfo?.symbol}
+            dtBalance={dtBalance}
+            stepText="Starting Compute Job..."
+          />
+        </Formik>
+      )}
+
+      <footer className={styles.feedback}>
+        {isPublished && (
+          <Alert
```
kremalicious, line 290: let's use `SuccessConfetti` for this. After all those myriads of wallet transactions, users have earned some confetti

src/components/organisms/AssetActions/Compute/index.tsx
```diff
@@ -0,0 +1,300 @@
...
+          (selectedAlgorithmAsset as DDO).price,
+          selectedAlgorithmAsset
+        ))
+
+      const allowed = await ocean.compute.isOrderable(
+        ddo.id,
+        computeService.index
+      )
+
+      console.log('is dataset orderable ?', allowed)
```
kremalicious, line 193: would be super useful to add some debug logging, so after every step we could output like:

```js
Logger.log('[compute] isOrderable? ', isOrderable)
```

kremalicious:
we have to safeguard a bit more against older assets where `ServiceComputePrivacy` is not there. So for those cases we have to diverge from the typings and assume that some required values can be not present:

Screen Shot 2021-03-26 at 13 52 00

In this case there should be no console error, and the `AssetSelection` should show an empty screen

src/components/organisms/AssetActions/Compute/index.tsx
```diff
@@ -0,0 +1,304 @@
...
+  isBalanceSufficient,
+  dtBalance,
+  file
+}: {
+  ddo: DDO
+  isBalanceSufficient: boolean
+  dtBalance: string
+  file: FileMetadata
+}): ReactElement {
+  const { marketFeeAddress } = useSiteMetadata()
```
kremalicious, line 47: we have to pass this to the order methods. They both accept a `mpAddress` parameter, hich I think is meant for the `marketFeeAddress` (it's undocumented in ocean.js code)

kremalicious:
quick draft of what the goal should be for some sort of final price output:

Screen Shot 2021-03-29 at 11 38 02

src/components/organisms/AssetActions/Compute/index.tsx
```diff
@@ -288,6 +288,12 @@ export default function Compute({
         algorithmAssetOrderId,
         selectedAlgorithmAsset.dataToken
       )
+
+      if (!respone) {
```
kremalicious, line 292: an error will have a `response`. In my case a `500` was returned so check here would be along the lines of:

```js
if (!reponse || !response.ok || response.status !== '200') {
```

src/components/organisms/AssetActions/Compute/index.tsx
```diff
@@ -288,6 +288,12 @@ export default function Compute({
...
+        setError('Error starting compute job.')
```
kremalicious, line 293: would be great to get after this line also this pattern in here:

```js
Logger.error('[compute] Failed to start job!', error.message)
```

In my error case the body of the response has helpful message under `body.error` so would be great to output this somehow

bogdanfazakas: yeah exactly, I was testing now and didn't find a way to get to the error 😃

src/components/organisms/AssetActions/Compute/index.tsx
bogdanfazakas: I'm not sure we can have it exactly as above because response of type ComputeJob, and for eg the succes status is 10, we can also check the statusText to be "Job started"

bogdanfazakas:
> quick draft of what the goal should be for some sort of final price output:
> 
> Screen Shot 2021-03-29 at 11 38 02

Regarding this, the general idea is to add this sum section with price values for dataset and algorithm, but also keep the the Help text from ButtonBuy.tsx right? Also think we have to somehow extend the help text and also pass the info regarding the orders and datatoken of the algorithm and display that message for both asset maybe the user allready has datatoken and orders for both algo and data set an will not have to pay ocean again for neither of them

kremalicious:
All those different help text combinations are a good rabbit hole, I would be in favor of trying to reduce them for now, rather than making them more fine-grained. 

Or at least deal with this state first, in which every user will end up in after starting a compute job:

Screen Shot 2021-03-31 at 17 12 44

Basically selecting another algorithm than before makes the next try fail when selecting another algorithm:

Screen Shot 2021-03-31 at 17 15 19

kremalicious:
moved out remaining issues from here into #470 & #472 so we can merge this huge PR to get back into more manageable PRs.

kremalicious (approved):

PR #440, opened 2021-03-15
Add expected price impact for removing liquidity

### LINKED ISSUE #255: Add "Expected price impact" or similar when removing liquidity
A lot of removing liquidity transactions are failing because we have a minimumReceived.
During spikes, that minimumReceived is changing a lot, so it makes sense to have a slippage selector here as well

claudiaHash:
Fixes #255 .

Changes proposed in this PR:

- handle simple and advanced cases
- added number of ocean amount to remove
- compute number of shares needed
- calculate the slippage


src/components/organisms/AssetActions/Pool/Remove.module.css
```diff
@@ -60,3 +60,34 @@
 .output figure[class*='pool shares'] {
   display: none;
 }
+
+.slippage {
```
kremalicious, line 64: these are the exact same styles as https://github.com/oceanprotocol/market/blob/main/src/components/organisms/AssetActions/Trade/Slippage.module.css

If we do not want to use the existing `Slippage` component and make it usable for both use cases (which would be preferred), we should at least not repeat that many styles and rather make use of CSS modules and `compose` them from `Trade/Slippage.module.css` into here, e.g.:

```css
.slippage {
  composes: slippage from '../Trade/Slippage.module.css'
}
```

content/networks-metadata.json
```diff
@@ -0,0 +1,2324 @@
+[
```
kremalicious, line 1: this file should not be here, it is gitignored. So remove and commit the removal should remove it from PR too

mihaisc:
image
Balance should be the amount of ocean, not pool shares
You will receive ocean, not pool shares.

claudiaHash:
So considering what we discussed before, for the simple flow balance will be in ocean, but for the advance flow it will remain in pool shares. right?

mihaisc:
Apparently i was wrong, i updated the issue. We go back to the slider and implement what is described in the issue.

mihaisc:
Should be like this 
image

claudiaHash:
I corrected, so now we have:
Screenshot from 2021-03-18 11-02-26
Screenshot from 2021-03-18 11-14-20
I'll push the changes in a moment.
Also, kept the original helper content. For advanced, should we say "You will receive minimum" or keep it simple?

mihaisc:
`You will receive minimum`

src/components/organisms/AssetActions/Pool/Remove.tsx
```diff
@@ -149,13 +156,41 @@ export default function Remove({
...
+        (Number(amountOcean) * (100 - Number(slippage))) / 100
+      const minDatatokenAmount =
+        (Number(amountDatatoken) * (100 - Number(slippage))) / 100
+      setMinOceanAmount(`${minOceanAmount}`)
+      setMinDatatokenAmount(`${minDatatokenAmount}`)
+    }
+  }
+
+  useEffect(() => {
+    computeOutput()
```
kremalicious, line 183: the contents of this method should be within this `useEffect` which is the most simple way to prevent the need to wrap the method in a `useCallback` with it's own dependencies. Otherwise state could get out of sync, or appear not updated.

This will then also prevent this hooks dependency warning: 
Screen Shot 2021-03-23 at 15 14 55

src/components/organisms/AssetActions/Pool/Remove.tsx
```diff
@@ -149,13 +156,41 @@ export default function Remove({
     totalPoolTokens
   ])
 
+  async function computeOceansRemoved(amountPoolShares: string) {
```
kremalicious, line 159: small naming thing but `compute` is kinda reserved in our tech stack for something different. We could use `calculate` instead

kremalicious:
Functionality-wise seems to be all good now.

In both, simple & advanced we need to get the token numbers aligned again:

Screen Shot 2021-03-25 at 08 23 51

They need to be aligned like so where numbers are always on one row:

Screen Shot 2021-03-25 at 08 23 36

Most likely through making the font-size for those headings smaller. Which then implies doing the same on all the other screens, like Add/Trade so all those output headings stay consistent.

Other than that, #455 should then make the tests succeed again

content/networks-metadata.json
kremalicious: this file still needs to be removed

src/components/organisms/AssetActions/Pool/Remove.module.css
```diff
@@ -60,3 +60,28 @@
...
+
+.output p {
+  font-size: var(--font-size-small);
+}
+
+.slippage {
+  composes: slippage from '../Trade/Slippage.module.css';
+}
+
+.slippage select {
```
kremalicious, line 72: this is redundant, because it will be inherited from the above `compose`. So we can remove all declarations here for everything which is already defined in `../Trade/Slippage.module.css`:
-  `.slippage select`
- `.slippage strong`

And since we do not have any `input` in this component, `.slippage input` can be removed too

src/components/organisms/AssetActions/Pool/Remove.tsx
```diff
@@ -220,21 +251,40 @@ export default function Remove({
           </Button>
         </div>
       </form>
-
       <div className={styles.output}>
         <div>
           <p>{content.output.titleIn}</p>
-          <Token symbol="pool shares" balance={amountPoolShares} noIcon />
+          {isAdvanced ? (
```
kremalicious, line 257: was there a specific reason to replace this with a condition here? Cause both return the exact same component with the exact same props

kremalicious: ah, I remember, this is just some leftover from the gazillion refactors we went through in here, so should be good to just restore what we had originally

claudiaHash: I forgot to remove that condition when I replaced the maxShares with the amountPoolShares, in the simple case. I correct that immediately.

kremalicious (approved):
awesome, all working as expected. We can't fully test slippage functionality though because there's no simple way to simulate heavy price fluctuations during time of transaction

PR #444, opened 2021-03-18
Update search and publish

mihaisc:
Bump lib
Update publish 
Update search (removed `nativeQuery`)

Currently works on Ropsten

DO NOT MERGE - the merge needs to be in sync with aqua release on main to have as little downtime as possible

### DIFF
--- a/src/components/templates/Search/utils.ts
+++ b/src/components/templates/Search/utils.ts
@@ -78,7 +78,6 @@ export function getSearchQuery(
     page: Number(page) || 1,
     offset: Number(offset) || 21,
     query: {
-      nativeSearch: 1,
       query_string: {
         query: `${searchTerm} -isInPurgatory:true`
       }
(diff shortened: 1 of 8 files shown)


src/components/templates/Search/utils.ts
```diff
@@ -78,7 +78,6 @@ export function getSearchQuery(
     page: Number(page) || 1,
     offset: Number(offset) || 21,
     query: {
-      nativeSearch: 1,
```
bogdanfazakas, line 81: beside the native search delete, wouldn't it be better if we will use also the SearchQuery typings since it was updated in ocean js? 
here for eg https://github.com/oceanprotocol/market/blob/fb65c3f67761cb9b73c704c831715cb9b030f1da/src/components/templates/Search/utils.ts#L100 (and also use it in all other places like Home, Bookmarks etc)? Whit this we can also get rid of the any type hack

mihaisc: will change

mihaisc: I think i got all of them

bogdanfazakas (approved):
LGTM 👍

PR #450, opened 2021-03-23
Add custom network

### LINKED ISSUE #449: Add custom network
We should add an option for the user to automatically add other chains that we support, similar with custom token

https://docs.metamask.io/guide/rpc-api.html#other-rpc-methods

KatunaNorbert:
Fixes #449 .

Changes proposed in this PR:

- Allow adding custom network
- new announcement banner pattern


KatunaNorbert:
What if we use the `Alert` component to display?
Screenshot 2021-03-23 at 15 06 16

kremalicious:
whatever we do, not more banners in that place please. Any banner we add there diminishes the perceived importance of each warning.

What I think we are looking for is some announcement banner pattern which would solve some use cases along the way. My suggestion would be simply this, because one point of a minimal interface is to be able to use color alone to draw attention:

Screen Shot 2021-03-23 at 18 23 08

The condition for this would be:
- _available on Polygon_ linked to the blog post
- show on all pages (we might want to hide it on asset details though since adding network in metamask will also switch to it, making user in market end up with 404)
- if MetaMask is detected, show _Add Custom Network_ or _Add Polygon to MetaMask_ button

Then, when wallet or app is connected to Polygon we can ditch the double warning banner we have there right now and use the announcement banner for the same thing:

Screen Shot 2021-03-23 at 18 28 43

src/components/molecules/AnnouncementBanner.tsx
```diff
@@ -0,0 +1,30 @@
...
+}: {
+  text?: string
+  action?: AnnouncementAction
+}): ReactElement {
+  return (
+    <div className={styles.container}>
+      <div className={styles.banner}>
+        {text && <Markdown className={styles.text} text={text} />}
+        {action && (
+          <button className={styles.button} onClick={action.handleAction}>
```
kremalicious, line 23: we have `<Button style="text" size="small">...</Button>` so with that we should be able to do less style writing to overwrite default `button` style

content/site.json
```diff
@@ -17,7 +17,8 @@
       }
     ],
     "warning": "We are in beta. Please familiarize yourself with [the market](https://oceanprotocol.com/technology/marketplaces), [the risks](https://blog.oceanprotocol.com/on-staking-on-data-in-ocean-market-3d8e09eb0a13), and the [Terms of Use](/terms).",
-    "warningPolygon": "Polygon/Matic EVM support is in early stages. Use the Matic bridge to [get mOCEAN](https://docs.oceanprotocol.com/tutorials/polygon-bridge/).",
+    "warningPolygon": "Polygon/Matic EVM support is in early stages.",
```
kremalicious, line 20: Think this is now a good time to clean this content structure up a bit so content keys match component names, and we do not continue to add more and longer keys. 

Maybe like so:

```json
warnings: {
  "main": "We are in beta..."
  "polygonPublish": "Only republish data sets..."
},
"announcements": {
  "main": "Ocean Market is available on Polygon..."
  "polygon": "Polygon/Matic EVM support..."
}
```

This way we keep the main warning & announcement generic and are not forced to change keys in the future when we change messages

src/components/organisms/Header.tsx
```diff
@@ -1,10 +1,93 @@
...
+    chainId: 137,
+    name: 'Matic Network',
+    urlList: [
+      'https://rpc-mainnet.matic.network',
+      'https://rpc-mainnet.maticvigil.com/'
+    ]
+  }
+  const [text, setText] = useState<string>(warningPolygonNetwork)
+  const [action, setAction] = useState<AnnouncementAction>()
+  const addCustomNetworkAction = {
```
kremalicious, line 34: Feels like we should move out all that network-related logic into a dedicated component for only the network button, like `<ButtonNetwork>`. No need to tailor the `Header` for a specific announcement. Ideally, when we decide to remove an announcement we just remove the `AnnouncementBanner` import and are done, without having having to also remove all this network code.

Likewise, would be in favor of importing the `AnnouncementBanner` directly in `App.tsx` above the `Header`

kremalicious:
while testing, think we need another condition for users without a wallet to allow them to switch back to mainnet in the same place. This would leave us with logic for only the button:

1. no wallet, ETH mainnet: "Switch to Polygon"
2. no wallet, Polygon mainnet: "Switch to ETH"
3. with MetaMask, all networks except Polygon: "Add custom network"
4. any wallet, Polygon: "Get mOCEAN"
5. everything else: no button

Would say 1. - 3. should be handled in one new button component as described in comment above. 4. and 5. could be then handled in the announcement banner component.

src/components/organisms/Header.tsx
KatunaNorbert: Yeah, also was thinking that maybe I should move all the logic to `AnnouncementBanner` component. Moving it to a new button component would mean that we would have to make the network-related logic in 2 places, for the button in the new component and somewhere else for the text. Maybe it's bette to keep them together.

kremalicious: yup, let's start with that and then see if it makes sense abstracting further, so we have it out of `Header` at least. Also wasn't sure what exactly should be in the button when writing those use cases further down

src/components/molecules/AnnouncementBanner.tsx
```diff
@@ -0,0 +1,144 @@
...
+      'https://rpc-mainnet.maticvigil.com/'
+    ]
+  }
+  const [text, setText] = useState<string>(announcement.main)
+  const [action, setAction] = useState<AnnouncementAction>()
+  const addCustomNetworkAction = {
+    name: 'Add custom network',
+    handleAction: () => addCustomNetwork(web3Provider, network)
+  }
+  const addCustomTokenAction = {
```
kremalicious, line 42: sorry, misunderstanding. What I meant with the _Get mOCEAN_ action was to link that to the bridge tutorial, not add as custom token. 

Or, another more simple way might be to just include the sentence _Use the Polygon Bridge to get mOCEAN_ in the announcement message, and link that sentence to the bridge tutorial. Then remove the button for wallet + polygon condition

src/components/molecules/AnnouncementBanner.tsx
```diff
@@ -0,0 +1,144 @@
...
+          setText(announcement.polygon)
+          setAction(switchToEthAction)
+        }
+        break
+      case 'MetaMask':
+        if (networkId === 137) {
+          setBannerForMatic()
+        } else {
+          setText(announcement.main)
+          window.location.pathname.includes('/asset/did')
```
kremalicious, line 110: might save us from all this logic if we move that route detection out of here, and hide the whole announcement banner from `App.tsx`. And within `App.tsx` we should have `location.path` somewhere from the Gatsby `PageProps`. Cause with Gatsby and in often in React in general where we do SSR builds in addition to just running in browser, we should never use `window` directly, but rather props from the underlying router library which Gatsby exposes.

src/components/molecules/AnnouncementBanner.tsx
KatunaNorbert: ok, yeah I got this wrong

kremalicious (approved):
awesome, great job on making all those conditions work. That's shippable for our use case and we revisit the button isolation when the need arises for some other feature.

PR #453, opened 2021-03-24
feature/add toggle option for fixed & dynamic pricing

### LINKED ISSUE #451: Allow market owner to toggle option between Fixed and Dynamic pricing to chose the mechanism accordingly.
Have 2 env vars, that control if Fixed or Dynamic prices can be showed during the price creation page.

ALLOW_FRE_PRICING = boolean
ALLOW_DYNAMIC_PRICING = boolean

krisliew:
Fixes #451 

Changes proposed in this PR:

- Added 2 env variables: `GATSBY_ALLOW_FRE_PRICING`, `GATSBY_ALLOW_DYNAMIC_PRICING`
- The default value is set as `true` in `app.config.js`
- If the env variable value is set as `false`, that specific create pricing option will be hidden in the whole marketplace


.env.example
```diff
@@ -4,4 +4,6 @@ GATSBY_NETWORK="rinkeby"
 #GATSBY_INFURA_PROJECT_ID="xxx"
 #GATSBY_MARKET_FEE_ADDRESS="0xxx"
 #GATSBY_ANALYTICS_ID="xxx"
-#GATSBY_PORTIS_ID="xxx"
\ No newline at end of file
+#GATSBY_PORTIS_ID="xxx"
+#GATSBY_ALLOW_FRE_PRICING="true"
```
kremalicious, line 8: consistency, if we call it `GATSBY_ALLOW_DYNAMIC_PRICING` then the other one should be `GATSBY_ALLOW_FIXED_PRICING`. Only insiders know what `fre` stands for

app.config.js
```diff
@@ -38,5 +38,9 @@ module.exports = {
   },
 
   // Wallets
-  portisId: process.env.GATSBY_PORTIS_ID || 'xxx'
+  portisId: process.env.GATSBY_PORTIS_ID || 'xxx',
+
+  // Toggle pricing option
+  allowFREPricing: process.env.GATSBY_ALLOW_FRE_PRICING || 'true',
```
kremalicious, line 44: Same here, should be then `allowFixedPricing`

src/components/organisms/AssetContent/Pricing/FormPricing/index.tsx
```diff
@@ -9,6 +9,7 @@ import { PriceOptionsMarket } from '../../../../../@types/MetaData'
 import Button from '../../../../atoms/Button'
 import { DDO } from '@oceanprotocol/lib'
 import FormHelp from '../../../../atoms/Input/Help'
+import appConfig from '../../../../../../app.config'
```
kremalicious, line 12: also works, but we try to unify access to `appConfig` through our `useSiteMetadata()` hook which is populated on build time:

```tsx
```

This requires to add the new keys to the GraphQL query in that hook and then they become available

app.config.js
```diff
@@ -38,5 +38,9 @@ module.exports = {
   },
 
   // Wallets
-  portisId: process.env.GATSBY_PORTIS_ID || 'xxx'
+  portisId: process.env.GATSBY_PORTIS_ID || 'xxx',
+
+  // Toggle pricing option
```
kremalicious, line 43: If we do not start documenting all the app options in the readme, then we should be at least more helpful in describing what those options do. Maybe like:

```tsx
// Pricing options: used to show or hide the fixed and dynamic price options to publishers during the price creation.
```

krisliew: @kremalicious  should we open this as another issue to document the app options? because I think is important to ensure future market owner aware of the options available. Might be a low priority now but in case the future we might have more options available.

kremalicious (approved):
perfect. Can only think of a problem when both options are disabled (I don't know how tab component behaves with empty array) but this would defeat the purpose of the market anyway so we do not have to bother with that.

Congrats to your first contribution!

PR #458, opened 2021-03-25
Fix buy fre

mihaisc:
Properly fetch exchange id from the graph.

### DIFF
--- a/src/providers/Asset.tsx
+++ b/src/providers/Asset.tsx
@@ -44,6 +44,7 @@ const freQuery = gql`
   query FrePrice($datatoken: String) {
     fixedRateExchanges(orderBy: id, where: { datatoken: $datatoken }) {
       rate
+      id
     }
   }
 `
@@ -77,7 +78,7 @@ function AssetProvider({
     data: frePrice
   } = useQuery<FrePrice>(freQuery, {
     variables,
-    skip: true
+    skip: false
   })
   const {
     refetch: refetchPool,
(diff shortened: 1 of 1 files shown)


src/providers/Asset.tsx
```diff
@@ -77,38 +78,47 @@ function AssetProvider({
...
-    if (!frePrice || frePrice.fixedRateExchanges.length === 0) return
+    if (
+      !frePrice ||
+      frePrice.fixedRateExchanges.length === 0 ||
+      price.type !== 'exchange'
+    )
+      return
     price.value = frePrice.fixedRateExchanges[0].rate
+    price.address = frePrice.fixedRateExchanges[0].id
+    console.log(price)
```
kremalicious, line 114: missing console.log

kremalicious (approved):
confirmed this is working for issue we encountered

src/providers/Asset.tsx
```diff
@@ -77,38 +78,47 @@ function AssetProvider({
...
   useEffect(() => {
-    if (!frePrice || frePrice.fixedRateExchanges.length === 0) return
+    if (
+      !frePrice ||
+      frePrice.fixedRateExchanges.length === 0 ||
+      price.type !== 'exchange'
+    )
+      return
     price.value = frePrice.fixedRateExchanges[0].rate
+    price.address = frePrice.fixedRateExchanges[0].id
```
kremalicious, line 113: sidenote, I'm always not sure how React behaves when doing these object mutations with its state objects. Because of that, a better pattern is to modify anything in React state with the setState methods only. So in this case and all the similar cases this is the most React-safe way:

```js
setPrice((currentState) => ({
    ...currentState,
    value: frePrice.fixedRateExchanges[0].rate,
    address: frePrice.fixedRateExchanges[0].id
}))
```

We just don't run into issues here cause the `AssetProvider` is almost always mounted and unmounted before going to next asset

bogdanfazakas (approved):
working for me to 👍

claudiaHash:
> working for me to

+1

src/providers/Asset.tsx
mihaisc: cool finding.

src/providers/Asset.tsx
mihaisc: my old nemesis

Issue #366, opened 2021-02-07
UX issue: user can get stalled in Publish step due to surprising 'FILE' form behavior 


trentmc:
**Describe the bug**
In the Publish step, the 'submit' button is shaded-out / unclickable, and it isn't clear how to proceed to make it clickable.

**To Reproduce**
1. Go to market.oceanprotocol.com. 
2. Click button on top right to connect wallet (it can be rinkeby, mainnet, whatever). Make sure you have enough ETH.
3. Click on 'PUBLISH'
4. Fill out the whole form with test info. This includes: in the 'FILE' section, copy and paste, but do *not* hit Enter in the form or hit the gray 'ADD FILE' button.
5. Check the "I agree to these Terms and Conditions" box
6. Just below that box is the 'SUBMIT' button. But it's still shaded-out / unclickable. 
7. If I scroll up, there are no red warning labels next to any fields, good. Exception: there's a red 'REQUIRED' label next to the 'FILE' field, but that simply annoys me because I've already filled out the field. So I don't have any indication of what to do. So I feel stalled out, I don't know how to proceed. I fiddle with my wallet a bit, try a different browser, try a different os, nothing. (The way that it will go away is if I *do* click into the 'FILE' field then hit 'enter' or click the gray 'ADD FILE' button. But that is not obvious.)

**Expected behavior**

Best: For the 'FILE' text field act like all the other text fields in this form. There really shouldn't be a button 'ADD FILE', this should simply be auto-detected and verified.

OK: Or, (a) make it more obvious that the 'ADD FILE' text is actually a *button*. And b) if it's still not filled yet, the label should say more than just 'REQUIRED'; e.g. it should give an indicator of next step, such as 'REQUIRED, press ADD FILE button'.

**Screenshots**
Here's the perpetrator. Super frustrating. 
![file](https://user-images.githubusercontent.com/5305452/107144382-233b2e00-693b-11eb-81c9-9bdd9fa22df7.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.x, Win 10
 - Browser: Chrome, Firefox

kremalicious:
good points. Disabling the button by default was a workaround before proper live form validation was in place. Generally, disabling form submit buttons is a bad UX decision because users can't know the internal conditions under which button becomes enabled. 

[...]

PR #461, opened 2021-03-29
Issue 366/dataset publish ux

jamiehewitt15:
Closes: https://github.com/oceanprotocol/market/issues/366

This pull request implements the "Best" solution mentioned by @trentmc for issue #366 

Changes proposed in this PR:
- Info text has been changed to explain that the "ADD FILE" button should be clicked
- Warning message has been changed to make clear that the "ADD FILE" button needs to be clicked
- Button styling has been changed to make it look more like an actual button
- The "ADD FILE" button is automatically pressed as an onBlur event. So as soon as the user moves to a different field in the form the dataset file is automatically validated.


jamiehewitt15:
So you're suggestion the error message should stay there until the user has corrected the invalid field? The error message was setup on a timer before now (I assume that was a UX decision?) but I can change it.

jamiehewitt15:
Ok, I have made the suggested change:
1. The error message remains in place until the user inputs a URL that has a valid dataset.
2. Additionally: a success message will now appear when the user eventually inputs a URL with a valid dataset

jamiehewitt15:
@mihaisc were you able to take another look at this after I made the changes to the error and success messages last week?

mihaisc:
If you add a file and then remove it , you can't add another file again. The `ADD FILE` button doesn't do anything

jamiehewitt15:
The button is actually automatically pressed and the dataset is checked, there's no need to click it. As long as the input is valid, you'll get the success message.  

I think you may be talking about the fact that the button is inactive (greyed out) if you enter a URL without the proceeding `https://`. That's actually an existing part of the marketplace code (it's how it is on live at the moment) it's not introduced in this pull request. I can remove the URL validation that's making the button appear inactive though if you want?

mihaisc:
Nope,
fileAddBug
After you remove the first uploaded file there is no interaction on click, on focus lost, doesn't matter. Nothing in console as well.

jamiehewitt15:
@mihaisc ah ok, that's because you have entered the same URL twice. If you change any of the characters in the URL that you've entered you'll see that it works. Ocean Market doesn't allow you to enter the same URL twice in a row, it's how it is at the moment on live. See screen recording: 

https://www.loom.com/share/334e6a36080843dfbf25709f63608a40

I would suggest opening a separate issue for allowing people to enter the same URL twice, this issue was about validating the URL immediately after it is entered which has been done.

mihaisc (approved):

mihaisc:
Oh boy, had no idea we had that. It doesn't make any sense. So if i accidentally removed it, then i can't add it anymore.... not sure if this was intentional. Can you please create a issue for that.

jamiehewitt15:
Sure, I'll create a separate issue for it

jamiehewitt15:
I'll also create a separate issue for the `https://` issue that I mentioned above as well. If you enter a URL without the `https://` it won't work and it's not obvious to the user why. That could be quite frustrating... I think we could easily lose customers with that

jamiehewitt15:
These are the related issues: 
https://github.com/oceanprotocol/market/issues/490
https://github.com/oceanprotocol/market/issues/491

PR #462, opened 2021-03-29
Compute job history

### LINKED ISSUE #338: Fetch compute history from the graph
The current compute job history format is ok, the main problem is that it's very slow. We need to move this to the graph, `ocean.assets.getOrderHistory` needs to be fetched from [here](https://subgraph.rinkeby.oceanprotocol.com/subgraphs/name/oceanprotocol/ocean-subgraph/graphql?query=query%20PoolsQuery%20%7B%0A%20%20tokenOrders(where%3A%20%7Bpayer%3A%20%220x4d156a2ef69ffddc55838176c6712c90f60a2285%22%7D)%20%7B%0A%20%20%20%20id%0A%20%20%20%20amount%0A%20%20%20%20datatokenId%20%7B%0A%20%20%20%20%20%20id%0A%20%20%20%20%7D%0A%20%20%20%20serviceId%0A%20%20%20%20consumer%20%7B%0A%20%20%20%20%20%20id%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%0A&operationName=PoolsQuery)

claudiaHash:
Fixes #338  .

Changes proposed in this PR:

- fetching  compute job history from the Subgraph


kremalicious:
think some architectural change is required so we can ditch the fetch-each-asset-individually completely:

1. for all the DIDs we get from The Graph query, do a `queryMetadata()`, where the `query` includes a list of those DIDs, and with highest page count possible (so we do not have to deal with pagination)
2. it will return a `QueryResult` and `result.results` will have array of DDOs. Put that into a new component state.
3. in `getAssetDetails()` then iterate through that state, instead of querying Aquarius for each asset

Then it should load way faster.

src/components/pages/History/ComputeJobs.tsx
```diff
@@ -69,50 +94,91 @@ const columns = [
...
-              dateFinished: item.dateFinished,
-              assetName: assetName,
-              status: item.status,
-              statusText: item.statusText,
-              algorithmLogUrl: '',
-              resultsUrls: []
+        const assets = await getAssetDetails()
+        for (let i = 0; i < assets.length; i++) {
+          if (assets[i].type === 'compute') {
+            const computeJob = await ocean.compute.status(
```
kremalicious, line 161: somewhere within in this method the assets get lost, when doing `console.log(computeJob)` right after this, it's an empty array for me, where `assets` above actually has the assets:

Screen Shot 2021-03-31 at 12 34 26

That `[zeroXTransformer] Input transformation failed.` implies we do something wrong in the parameters we pass to `ocean.compute.status`

kremalicious:
I think we are stuck and the intended outcome of getting compute history from The Graph can only work out if we actually have the jobs in our subgraph. Because token orders !== compute jobs. Because a job can be started without a token order when a job on a dataset has been started before.

So correlating token orders with actual started jobs where we still have to fetch the jobs through ocean.js is defeating the purpose of this issue. The problem visualized:

My compute jobs through `ocean.compute.status`:

Screen Shot 2021-04-01 at 10 57 31

And then with this PR where we get all the token orders, make a list of DIDs out of it, then query Aquarius for it only including compute datasets:

Screen Shot 2021-04-01 at 10 59 36

So then `ocean.compute.status` would give us a completely different list.

mihaisc:
I'll take a look. At first sight the logic is wrong here but need to debug a bit to be sure.

mihaisc:
Currently blocked by https://github.com/oceanprotocol/ocean.js/pull/717

We are going to fetch all the jobs for the user from each provider. It's the least amount of calls and the fastest.

~1.5 seconds for me with 30+ jobs.

kremalicious:
I get multiple signature request for all jobs right away, even before list loads:

Screen Shot 2021-04-07 at 14 55 48

We should try to limit that signature to retrieving the detail logs, as we do right now on live.

As a comparison, it looks like we might show jobs multiple times:

Live:
Screen Shot 2021-04-07 at 14 57 29

This branch:
Screen Shot 2021-04-07 at 14 57 48

kremalicious:
Also seem to be getting different timestamps on reload, and they do not make sense for _Berlin Fish Fauna (2013)_ cause it didn't exist 6 months ago:

Screen Shot 2021-04-07 at 15 01 03

mihaisc:
Yes, because your testing doesn't make sense. My push is tested using a local version of ocean.js and not finished because i need the update in the provider so i didn't bother to wrap things up until i have that update.

kremalicious:
First quick test, might have started a bunch of jobs but not sure that many. And the `CREATED` & `FINISHED` times seem weird:

Screen Shot 2021-04-13 at 16 29 57

And not sure if we want to look at that in this PR, but none of the listed jobs have any logs for me

mihaisc:
is this on main or rinkeby?

mihaisc:
fixed, it was a bug related to network switching

kremalicious (approved):
another wild ride of what looked not that hard before starting it

PR #463, opened 2021-03-30
bump ocean.js v0.16.2, ocean.assets.isConsumable implementation to disable buy button and display relevant message 

### LINKED ISSUE #452: As a data seller, I would like to disable consume for certain assets during certain activities like maintenance or downtime
By default all published assets are shown in the catalog.

Let's check for an additional flag in the DDO

-isMaintenance

if exists and it's true, then consume is disabled and we need to show a banner.  All other functions are not affected (lp, etc)

Of course, during EditMetadata, publisher can update this field by checking/unchecking  a checkbox

krisliew:
Fixes #452 and #685

Changes proposed in this PR:
- bump ocean.js to [v0.16.2](https://github.com/oceanprotocol/ocean.js/releases)
  - set `datePublished` as `''` in `src/utils/metadata.ts` for `validate-remote`
- Use ocean.assets.isConsumable to check if dataset can be consume. Disable `buy` button if `false` is returned with relevant message below `buy` button.
- Updated `src/components/molecules/FormFields/Credential/index.tsx` to convert all address to lowercase for `allow` and `deny` list because `accountId` from `useWeb3()` have case sensitive issue. (Returning all lowercase on change metamask account)
 
To save your time on testing this PR, you may use the following DID to test 😄 (Rinkeby network)

1. Download test isOrderDisabled=true 
https://market-3628a6svt-oceanprotocol.vercel.app/asset/did:op:d1Dae50a32f613B282BFCcb50fE33Bf987514882 

2. Compute test isOrderDisabled=true
https://market-3628a6svt-oceanprotocol.vercel.app/asset/did:op:940F4d33C1886C050c5601ED6bd7F51d9084Cc16 

3. Compute test isOrderDisabled=false
https://market-3628a6svt-oceanprotocol.vercel.app/asset/did:op:6A41c8f617686B4Dff27c8A3ee5ECc4f175cb8a4 

4. Compute test allow list:
https://market-3628a6svt-oceanprotocol.vercel.app/asset/did:op:01530feD6aE9CFbfA602e05Ef51C447e8C145F1C 

5. Algo test allow list:
https://market-3628a6svt-oceanprotocol.vercel.app/asset/did:op:CFD8E0a3188A0ED6665929e29e3716e3C2B50df8

6. Algo test deny list:
https://market-3628a6svt-oceanprotocol.vercel.app/asset/did:op:5DA103145F09FEf6C60d65A9c7033A1b0458B26a 

- [x] Prerequisite PR at Ocean.js https://github.com/oceanprotocol/ocean.js/pull/786
- [x] Need bumps latest Ocean.js https://github.com/oceanprotocol/ocean.js/pull/869


mihaisc:
Please merge main into this so we can test.

krisliew:
@mihaisc @alexcos20  is `isConsumeDisable` variable name good enough? or should I update the flag name back to `isMaintenance`?

mihaisc:
i think `isMaintenance` is better for this purpose

alexcos20:
I would go towards:  isDisabled  & isRetired.  I will open a PR on oep

krisliew:
ok @alexcos20 lets finalise this, correct me if I am wrong below.
So we will be having 2 new variable `isDisabled` and `isRetired`  in the DDO.
- `isDisabled` is a temporary disable flag, and publisher can enable/disable it anytime.
- `isRetired` should be like a permanent disable flag for publisher to set?

kremalicious:
what happened here? The PR tries to revert `main` to some very old state. Needs a rebase or merge against `main`

mihaisc:
Considering we now have the correct metadata for this ( https://github.com/oceanprotocol/ocean.js/blob/727a55046c44c1841360f6aa2a4f94befe157860/src/ddo/interfaces/Status.ts#L25 ) we can move on implementing this.
Before this https://github.com/oceanprotocol/market/issues/638 issue we can create the new section "edit advanced settings" here and add the `isOrderDisabled` option here like described above
image

For the messaging on the asset please use the same mechanism as on https://github.com/oceanprotocol/market/pull/479, but change the status , and have a proper tooltip description.
image
As well please merge main before (and after if needed)

krisliew:
@mihai Should the "Buy" / "Download" button be disabled too?

mihaisc:
No do not disable it

src/components/molecules/AssetStatus.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { ReactElement, useEffect, useState } from 'react'
+import Status from '../atoms/Status'
+import Tooltip from '../atoms/Tooltip'
+import styles from './AssetStatus.module.css'
+
+export default function AssetStatus({
+  isDisable
```
mihaisc, line 7: Please use `isOrderDisabled`, it's more consistent

kremalicious: yup, and more so why a boolean and why is it optional? Why is the prop simply setting state when the whole component could be prevented from rendering in the parent component without passing a props?

What is this component supposed to do, what is it supposed to change inside when the prop changes? If nothing changes in component based on prop there seems to be no need for them.

src/components/molecules/AssetStatus.tsx
```diff
@@ -0,0 +1,27 @@
...
+    isDisable === true && setStatus('Disabled')
+  }, [isDisable])
+
+  return isDisable ? (
+    <div className={styles.wrapper}>
+      <Status className={styles.status} state="error" />
+      <span className={styles.text}>{status}</span>
+      <Tooltip
+        className={styles.info}
+        content="This dataset might be empty/offline, consume at your own risk."
```
kremalicious, line 23: this message does not fit the proposed state. "Disabling" is a dedicated user action whereas this message refers to "Unavailable" which might be untrue which this message reflects ("might be")

src/components/molecules/AssetStatus.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const [status, setStatus] = useState('')
+
+  useEffect(() => {
+    isDisable === true && setStatus('Disabled')
+  }, [isDisable])
+
+  return isDisable ? (
+    <div className={styles.wrapper}>
+      <Status className={styles.status} state="error" />
+      <span className={styles.text}>{status}</span>
```
kremalicious, line 20: so what's going to be the default `status` if `isDisable` is omitted as its's marked as optional?

krisliew:
@mihaisc @kremalicious  Do you have a good suggestion for showing `disabled` status at algorithm selection list for compute asset? I think is incorrect to pass whole DDO through [here](https://github.com/oceanprotocol/market/blob/e02babf2c2fd80ae67e686fb026406bc4dfcfac3/src/components/molecules/FormFields/AssetSelection.tsx#L13) as that would defeat the purpose of `transformDDOToAssetSelection()`... 🥵  (same scenario as connectivity check PR) 

Or perhaps modifying the [ocean.assets.isConsumable() ](https://github.com/oceanprotocol/ocean.js/blob/02693021b1d5195ad4c5e5b0d64210c96fd0be72/src/ocean/Assets.ts#L756)to accept DID and then retrieve DDO using [resolve()](https://github.com/oceanprotocol/ocean.js/blob/02693021b1d5195ad4c5e5b0d64210c96fd0be72/src/ocean/Assets.ts#L229) in ocean.js would solve this while keeping the market code clean.

image

krisliew:
Aside from that, I am planning to re-use this current component to handle https://github.com/oceanprotocol/market/issues/685 allow/deny list feature as well after this PR.
https://github.com/oceanprotocol/market/blob/321554d3315894fe0d6ed1e812a602e010639516/src/components/molecules/AssetStatus.tsx#L7-L18

mihaisc:
Good points. Resolving the did once more is not an option, we want to stop doing that. The best solution i think it would be to add `isOrderDisabled` in `AssetSelectionAsset`  and do the check in `transformDDOToAssetSelection`. And then just pass the boolean to the `AssetStatus` component

krisliew:
@mihaisc this mean that we are checking all the algorithm for `isOrderDisabled` before user is selecting one. This will be abit overkill if a compute asset's algorithm list have many algorithm right?

mihaisc:
Well it's just a boolean that we already have, it should be almost instant

krisliew:
ah, right. got it thanks 😄

mihaisc:
The asset is disabled but i can still start a job on it 
image

Shouldn't it be if asset is disable then you can't start a job. If asset is ok by algo is disabled you can't start job.

krisliew:
@mihai I don't think you can start the job, ocean.assets.order calls this function and will stop that. See here
https://github.com/oceanprotocol/ocean.js/blob/02693021b1d5195ad4c5e5b0d64210c96fd0be72/src/ocean/Assets.ts#L756-L774 
Which is why I am suggesting this approach too for `AssetStatus` to accept the returned `Consumable` as to make use of `status` and `message`: 

> Aside from that, I am planning to re-use this `AssetStatus` component to handle #685 allow/deny list feature as well after this PR.
> I will [update the ocean.js](https://github.com/oceanprotocol/ocean.js/blob/02693021b1d5195ad4c5e5b0d64210c96fd0be72/src/ocean/Assets.ts#L765) returned `consumable.message` if this approach is ok?
> 
> https://github.com/oceanprotocol/market/blob/321554d3315894fe0d6ed1e812a602e010639516/src/components/molecules/AssetStatus.tsx#L7-L18


From UX perspective, we should disable the `buy` button from beginning right? So they don't spend their Ocean only to find their  `order` failing later...

src/components/molecules/AssetStatus.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export default function AssetStatus({
+  isOrderDisabled
+}: {
+  isOrderDisabled: boolean
+}): ReactElement {
+  return isOrderDisabled === true ? (
+    <div className={styles.wrapper}>
+      <Status className={styles.status} state="error" />
+      <span className={styles.text}>Disabled</span>
```
kremalicious, line 14: this still does not account for changes in 2 other PRs which uses the same component. Just make it accept a `message: string` & `status: 'Disabled' | 'Unavailable'` prop and be done with it, make all changes in this component then based on those props if needed. The rest of the logic should happen outside in the parent component where component is used.

kremalicious:
And here CodeClimate is completely right and since it's the exact same code over multiple lines we should fix that:

Screen Shot 2021-06-23 at 09 40 52

mihaisc:
@kremalicious what design pattern should we follow (currently we use both) : disable button or hide button? I'm referring to the cases we know the action will fail.

src/components/molecules/AssetStatus.tsx
krisliew: yes thats what I am having in my mind too.
But ocean.js returned `status` is `number` here, do we modify ocean.js or handle that in market with a `if else`?
https://github.com/oceanprotocol/ocean.js/blob/02693021b1d5195ad4c5e5b0d64210c96fd0be72/src/ocean/Assets.ts#L756-L774

kremalicious:
To not change all flows I would go with whatever we have as established pattern right now to keep it consistent which would be disabled button.

Ideally if we have disabled buttons then the label of the disabled button should say why it's disabled (so in this case here we wouldn't need a dedicated "Unavailable" element, the button can just say that) but we don't need to solve this in here

src/components/molecules/AssetStatus.tsx
kremalicious: sorry, what do you mean with ocean.js? I mean to pass a `message` text which then is put as the tooltip. And the `status` is what's being output on screen by default, instead of the hardcoded copy in the line I added this comment to. Nothing involved with ocean.js

krisliew: Take a look at this too https://github.com/oceanprotocol/market/issues/685

jamiehewitt15 (approved):

jamiehewitt15:
It seems to work well to me. As far as I can tell the issues mentioned above seem to have been fixed

PR #466, opened 2021-03-31
The Graph sync status

### LINKED ISSUE #464: The Graph sync status
We need a way to show to users the current sync status of the displayed financial data, and if it is out of sync, display a warning to users, ideally by how much it is out of sync.

The principle is to compare the current block number (through web3) with the block number returned from a Graph query, and refetch and check continuously.

### UI

Exposing this in the UI with a subtle block number in the footer, connected to some visual status (we have a `Status` component for our visual system with red square, green circle), like UniSwap:

Screen Shot 2021-03-30 at 12 37 38

https://github.com/Uniswap/uniswap-interface/blob/main/src/components/Header/Polling.tsx

Then we need to find a global place to put a warning in case of out of sync detection and tell users what that means in its copy. Tending towards having another bar below our new `AnnouncementBanner`

KatunaNorbert:
Fixes #464 .

Changes proposed in this PR:

- show to users the current sync status of the displayed financial data, and if it is out of sync, display a warning to users

### New UI

Screen Shot 2021-04-12 at 11 25 07

Screen Shot 2021-04-10 at 13 33 11

114266679-cc63da80-99f7-11eb-972c-ebb73fe3bbff

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,7 +1,6 @@
 # Network, possible values: development, pacific, rinkeby, mainnet
 GATSBY_NETWORK="rinkeby"
-
-#GATSBY_INFURA_PROJECT_ID="xxx"
+GATSBY_INFURA_PROJECT_ID="a983b53583044593956054de049922fd"
 #GATSBY_MARKET_FEE_ADDRESS="0xxx"
 #GATSBY_ANALYTICS_ID="xxx"
 #GATSBY_PORTIS_ID="xxx"
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,8 @@
-# Network, possible values: development, pacific, rinkeby, mainnet
+# Default network, possible values: 
+# "development", "ropsten", "rinkeby", "mainnet", "polygon"
 GATSBY_NETWORK="rinkeby"
-GATSBY_INFURA_PROJECT_ID="a983b53583044593956054de049922fd"
+
+#GATSBY_INFURA_PROJECT_ID="xxx"
 #GATSBY_MARKET_FEE_ADDRESS="0xxx"
 #GATSBY_ANALYTICS_ID="xxx"
 #GATSBY_PORTIS_ID="xxx"
--- a/app.config.js
+++ b/app.config.js
@@ -5,7 +5,7 @@ module.exports = {
   // Ocean Protocol contracts are deployed for: 'mainnet', 'rinkeby', 'development'
   network: process.env.GATSBY_NETWORK || 'mainnet',
 
-  infuraProjectId: process.env.GATSBY_INFURA_PROJECT_ID || 'xxx',
+  infuraProjectId: process.env.GATSBY_INFURA_PROJECT_ID || 'a983b53583044593956054de049922fd',
 
   // The ETH address the marketplace fee will be sent to.
   marketFeeAddress:
--- a/app.config.js
+++ b/app.config.js
@@ -5,7 +5,7 @@ module.exports = {
   // Ocean Protocol contracts are deployed for: 'mainnet', 'rinkeby', 'development'
   network: process.env.GATSBY_NETWORK || 'mainnet',
 
-  infuraProjectId: process.env.GATSBY_INFURA_PROJECT_ID || 'a983b53583044593956054de049922fd',
+  infuraProjectId: process.env.GATSBY_INFURA_PROJECT_ID || 'xxx',
 
   // The ETH address the marketplace fee will be sent to.
   marketFeeAddress:
--- a/content/site.json
+++ b/content/site.json
@@ -18,7 +18,8 @@
     ],
     "warning": {
       "main": "We are in beta. Please familiarize yourself with [the market](https://oceanprotocol.com/technology/marketplaces), [the risks](https://blog.oceanprotocol.com/on-staking-on-data-in-ocean-market-3d8e09eb0a13), and the [Terms of Use](/terms).",
-      "polygonPublish": "Only republish data sets with a pool from ETH Mainnet into Polygon/Matic if the liquidity is **less than or equal to 1000 OCEAN in the original pool**. Doing otherwise will lead to [purgatory](https://github.com/oceanprotocol/list-purgatory) for the data set in Polygon/Matic."
+      "polygonPublish": "Only republish data sets with a pool from ETH Mainnet into Polygon/Matic if the liquidity is **less than or equal to 1000 OCEAN in the original pool**. Doing otherwise will lead to [purgatory](https://github.com/oceanprotocol/list-purgatory) for the data set in Polygon/Matic.",
+      "graphNotSynched": "Graph out of sync, possible data lack."
     },
     "announcement": {
       "main": "Ocean Market is [available on Polygon](https://blog.oceanprotocol.com/ocean-on-polygon-network-8abad19cbf47).",
--- a/content/site.json
+++ b/content/site.json
@@ -18,8 +18,7 @@
     ],
     "warning": {
       "main": "We are in beta. Please familiarize yourself with [the market](https://oceanprotocol.com/technology/marketplaces), [the risks](https://blog.oceanprotocol.com/on-staking-on-data-in-ocean-market-3d8e09eb0a13), and the [Terms of Use](/terms).",
-      "polygonPublish": "Only republish data sets with a pool from ETH Mainnet into Polygon/Matic if the liquidity is **less than or equal to 1000 OCEAN in the original pool**. Doing otherwise will lead to [purgatory](https://github.com/oceanprotocol/list-purgatory) for the data set in Polygon/Matic.",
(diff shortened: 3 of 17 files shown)


KatunaNorbert:
Good, so not my bad. Was strange for me too, tried it in many ways thinking that I missed something

mihaisc:
Pushed a fix, needs some refactoring and some cleaning but it works

KatunaNorbert:
How can I determine how much the Graph is out of sync, so I know when to show a warning? How much the difference between the 2 numbers should be?

mihaisc:
In theory it should be the same block number

KatunaNorbert:
Now the sync is checked on component mount, network change and wallet connect/disconnect. Should I make it to check on a certain interval? 
I also displayed the warning message inside the announcement banner but not sure how that should actually work. I guess we want to keep the current announcement so that adds up to 2 messages, if so, we have to differentiate them better.  And then should the announcement component support multiple messages(more than 2)? If It should than it need some changes.

mihaisc:
We need some messages on tooltip or something. Maybe show the block difference
I have no ideea what this means.
image

After i wrote the comment i remembered about the top message. If i didn't remember i wouldn't have seen it. We need something to warn people.

KatunaNorbert:
I think that's the `Announcement` component's job to display the warnings. But I see your point, the warning symbol down there can be confusing. Maybe we can drop the symbol change down and only show the warning in the announcement banner.

kremalicious:
refactored a bit, so we get a clear separation of concerns between banner, network stuff, graph sync stuff:

- new `atoms/AnnouncementBanner `: banner component reduced to presentation only, where its content is always passed as props
- revised `molecules/NetworkBanner`: the former `AnnouncementBanner ` now holds all the specific network detection logic, in the end also returns the `atoms/AnnouncementBanner`
- new hook `hooks/useGraphSyncStatus`: move all the graph fetching logic in there so we can use its status in multiple places in the app without all this props passing. This also decouples the `SyncStatus` component in footer from its logic
- in `App.tsx`, add the graph sync warning banner in another `atoms/AnnouncementBanner`, getting its values from the hook

Resulting in:

Screen Shot 2021-04-10 at 12 25 09

For the info in footer ended up with this:

Screen Shot 2021-04-10 at 12 41 01

.env.example
```diff
@@ -1,7 +1,6 @@
 # Network, possible values: development, pacific, rinkeby, mainnet
 GATSBY_NETWORK="rinkeby"
-
-#GATSBY_INFURA_PROJECT_ID="xxx"
+GATSBY_INFURA_PROJECT_ID="a983b53583044593956054de049922fd"
```
kremalicious, line 3: please never ever commit credentials. This infura Id should be now considered exposed and compromised, so will invalidate it in infura

app.config.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
   // Ocean Protocol contracts are deployed for: 'mainnet', 'rinkeby', 'development'
   network: process.env.GATSBY_NETWORK || 'mainnet',
 
-  infuraProjectId: process.env.GATSBY_INFURA_PROJECT_ID || 'xxx',
+  infuraProjectId: process.env.GATSBY_INFURA_PROJECT_ID || 'a983b53583044593956054de049922fd',
```
kremalicious, line 8: same here, committed credential

content/site.json
```diff
@@ -18,7 +18,8 @@
     ],
     "warning": {
       "main": "We are in beta. Please familiarize yourself with [the market](https://oceanprotocol.com/technology/marketplaces), [the risks](https://blog.oceanprotocol.com/on-staking-on-data-in-ocean-market-3d8e09eb0a13), and the [Terms of Use](/terms).",
-      "polygonPublish": "Only republish data sets with a pool from ETH Mainnet into Polygon/Matic if the liquidity is **less than or equal to 1000 OCEAN in the original pool**. Doing otherwise will lead to [purgatory](https://github.com/oceanprotocol/list-purgatory) for the data set in Polygon/Matic."
+      "polygonPublish": "Only republish data sets with a pool from ETH Mainnet into Polygon/Matic if the liquidity is **less than or equal to 1000 OCEAN in the original pool**. Doing otherwise will lead to [purgatory](https://github.com/oceanprotocol/list-purgatory) for the data set in Polygon/Matic.",
+      "graphNotSynched": "Graph out of sync, possible data lack."
```
kremalicious, line 22: right, good point getting copy in here. Quickly hacked it into `App.tsx` but feel free to find some way to inject the dynamic values into the text somehow

src/hooks/useGraphSyncStatus.ts
```diff
@@ -0,0 +1,117 @@
...
+      graphQuery
+    )
+    setGraphBlockNumber(response.data._meta.block.number)
+  }
+
+  useEffect(() => {
+    if (!config) return
+
+    getBlockFromGraph()
+    const timer = setInterval(() => getBlockFromGraph(), refreshInterval)
```
kremalicious, line 71: this was a `setTimeout` before but think the intention was to actually create an interval. Ideally we figure this out with `useSWR`, which gives us refetching on various user events for free so that in end we do not have to set manual intervals

kremalicious:
almost everything working nicely now, refactored some of the data flow and made everything more universal. When wallet is connected, head block is always coming from `web3.eth.getBlockNumber()`, set early in our `Web3Provider`. This block number is then used by `useGraphSyncStatus` to figure out if additional calls are needed to get head block, so only for users without a wallet:

- when on mainnet, get head block from `blocklytics/ethereum-blocks` graphql query
- for all other networks, create new web3 instance with `config.nodeUri`, and use `web3.eth.getBlockNumber()` again

But that only works for ETH networks, cause there is no infura for polygon, so slapping the infura id at the end of the `config.nodeUri` does not work here.

So we might end up having to create unique solutions to fetch the head block for non-ETH networks.

.env.example
KatunaNorbert: Right, sorry about that

src/hooks/useGraphSyncStatus.ts
```diff
@@ -12,91 +14,90 @@ const ethGraphQuery =
...
 
+  // Get and set head block
   useEffect(() => {
-    if (!config) return
+    if (!config || !config.nodeUri) return
 
-    getBlockFromGraph()
-    const timer = setInterval(() => getBlockFromGraph(), refreshInterval)
+    async function initBlockHead() {
+      const blockHead = block || (await getBlockHead(config))
```
KatunaNorbert, line 70: I think we should only use the `await getBlockHead(config)` to get the block number. I just tested and the block number is not changing inside the `Web3` provider when I change the network. So I get stocked with the initial block number

kremalicious:
phew, made sure the banner is never briefly appearing upon user network switching, trick was to work with a new `web3Loading` value from our `Web3Provider`.

Tested all combinations again, remaining issue stays Polygon without a wallet which gives us this:

Screen Shot 2021-04-12 at 11 41 28

It would not prevent app from working, nor would it falsely show the banner. But if the head block and subgraph are out of sync in Polygon there would be no warning banner.

src/hooks/useGraphSyncStatus.ts
kremalicious: sorry, just pushed some fixed, can you try again? The block retrieved via `web3` should always be the source of truth, hence I put that into our `Web3Provider`. So if wallets are connected this will always kick in. So the `await getBlockHead(config)` then is only the fallback. Cause right now, `getBlockHead()` only works for ETH mainnet, because `blocklytics/ethereum-blocks` is only ETH mainnet. So we would need to then start finding ways to get this for all ETH networks, which becomes quite cumbersome.

The ideal case is that we always get the block from web3, either with wallet, or infura-like service.

KatunaNorbert:
The bug still appeared for me so I added a loading state for the subgraph fetching too

src/hooks/useGraphSyncStatus.ts
KatunaNorbert: I get it. Tested it and works with `networkId` dependency added to the hook, now it's changing

src/hooks/useGraphSyncStatus.ts
kremalicious: nah, we decided to ditch it, just nice-to-have. I removed all interval/swr references in code. Thinking is that upon warning users will most likely reload anyway, and we should not put too much stress on the underlying subgraphs during times when there are no sync issues

content/site.json
kremalicious: so either we figure out how to get the copy currently in `App.tsx` into here, or we remove this copy line

KatunaNorbert: ok, I'll look into it

KatunaNorbert: Couldn't find a way for that, so we can remove the line and keep it as it is for now

kremalicious (approved):
perfect! So much to figure out along the way for something we do not want our users to ever see

PR #471, opened 2021-04-01
Add stats page

### LINKED ISSUE #355: public analytics
For the sake of transparency it would make sense exposing key traffic statistics to everyone.

Possible solutions:

1. use https://simpleanalytics.com for the `market` which would give us a public analytics page like this https://simpleanalytics.com/simpleanalytics.com
2. build a page within `market` fetching data from our Netlify project's analytics, and display them. Needs verification if that's actually possible 

In a later stage and most likely in a different issue, exposing key financial metrics for the whole market could find a place on this page too.

DimitarSD:
Fixes #355 .

Changes proposed in this PR:

- Adds a new stats page available at `/stats`

This PR is in some way related to the `market-stats` [PR #4](https://github.com/oceanprotocol/market-stats/pull/4) since the requests are directed to the API endpoints provided by `market-stats`


src/components/pages/Stats/index.tsx
```diff
@@ -0,0 +1,238 @@
...
+  ChartTooltipOptions
+} from 'chart.js'
+
+import { useUserPreferences } from './../../../providers/UserPreferences'
+import { darkModeConfig } from './../../../../app.config'
+import useDarkMode from 'use-dark-mode'
+
+import styles from './index.module.css'
+import moment from 'moment'
+import fetch from 'cross-fetch'
```
kremalicious, line 17: for all fetching either use `axios` or even way better use the `useSWR()` hook https://swr.vercel.app

DimitarSD: I thought we were going to use `cross-fetch`. I can easily switch to `useSWR` if needed

kremalicious: that is only over on ocean.js. All fetching code in market uses `axios`, and when using useSWR hook it should use axios too. See the `PricesProvider` https://github.com/oceanprotocol/market/blob/main/src/providers/Prices.tsx#L46-L50

Where you can then as seen there use the same `fetchData()` utility method https://github.com/oceanprotocol/market/blob/main/src/utils/index.ts#L43-L55

mihaisc: cross-fetch can be used with `useSWR`, check the graph sync pr. `axios` has CORS issues when accessing other endpoints. Not sure why stick with it

DimitarSD: In that case, it might be better if we update the [fetchData function](https://github.com/oceanprotocol/market/blob/main/src/utils/index.ts#L43-L55) to use cross-fetch instead of axios.

kremalicious:
Let's revisit after the v4 split as we ant to sprinkle various statistics throughout the whole interface, not just footer

PR #477, opened 2021-04-02
output datatoken symbol for each asset in asset selection

### LINKED ISSUE #472: output datatoken symbol in AssetSelection
In front of the DID, display the datatoken symbol in the AssetSelection component for each asset:

112846929-dbb56080-90a6-11eb-96a5-01c64b58b246

A bit like so maybe:

Screen Shot 2021-04-01 at 17 04 45

Super helpful when starting a compute job because the feedback text during the process refers to the datatoken symbol.

This is trickier than it looks like since the AssetSelection is a form field component and we use the `/assets/names` endpoint of Aquarius to the title, but the datatoken symbol does not come with it. So we might end up having to fetch all DDOs instead which might be an overhead so needs to be verified.

claudiaHash:
Fixes #472  .

Changes proposed in this PR:

- output the datatoken symbol for each asset in asset selection

### DIFF
--- a/src/components/molecules/FormFields/AssetSelection.tsx
+++ b/src/components/molecules/FormFields/AssetSelection.tsx
@@ -15,6 +15,7 @@ export interface AssetSelectionAsset {
   name: string
   price: string
   checked: boolean
+  symbol: string
 }
 
 function Empty() {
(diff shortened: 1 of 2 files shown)


bogdanfazakas (approved):
LGTM

Issue #474, opened 2021-04-01
deal with algorithms timeout


kremalicious:
In our spec and implementation there is no field to set a timeout during publish of an algorithm. But a timeout ends up in the access or compute service as it seems so it looks like we have to deal with it. So if timeout also applies to algos, we need to deal with it:

- add timeout field to algorithm publish form, just like for datasets
- when starting a compute job, consider this timeout in addition to the possible dataset timeout
  - we use _Buy for 1 day_ on the button for dataset timeout, not sure we can extend and combine this in some logical way with the timeout from selected algo
  - or we could show the timeout on the algorithm in the AssetSelection

PR #478, opened 2021-04-02
Help messages & price display for starting compute job

### LINKED ISSUE #470: help messages & price display for starting compute job
Our existing help messages which we adapt based on existence of a previous order need to be adapted for the compute case where we now have 2 orders (dataset + algo) we have to consider.

Screen Shot 2021-04-01 at 16 54 37

Likewise there should be some output of what user has to spend which most likely correlates with the help text:

112819187-9c761800-9084-11eb-92b9-46f4e18e96d5

bogdanfazakas:
Fixes #470 .

Changes proposed in this PR:

- added price info section with all amount ocean to be spend for a specific compute job
- updated help message on compute assets in BuyButton component

### DIFF
--- a/src/components/atoms/ButtonBuy.tsx
+++ b/src/components/atoms/ButtonBuy.tsx
@@ -68,14 +68,15 @@ function getComputeAssetHelpText(
     selectedComputeAssetTimeout === 'Forever'
       ? ''
       : ` for ${selectedComputeAssetTimeout}`
+
   const text =
     !dtSymbolSelectedComputeAsset && !dtBalanceSelectedComputeAsset
-      ? 'Please choose an algorithm to run you compute job'
+      ? ''
       : hasPreviousOrderSelectedComputeAsset
-      ? `You also bought the selected ${selectedComputeAssetType} already allowing you to use it without paying again${timeout}.`
+      ? `You already bought the selected ${selectedComputeAssetType}, allowing you to use it without paying again${timeout}.`
       : hasDatatokenSelectedComputeAsset
       ? `You also own ${dtBalanceSelectedComputeAsset} ${dtSymbolSelectedComputeAsset} allowing you to use the selected ${selectedComputeAssetType} by spending 1 ${dtSymbolSelectedComputeAsset}, but without paying OCEAN again.`
-      : `To be able to start compute using this ${selectedComputeAssetType}, you will buy 1 ${dtSymbolSelectedComputeAsset} and immediately spend it back to the publisher and pool.`
+      : `Additionally, you will buy 1 ${dtSymbolSelectedComputeAsset} for the ${selectedComputeAssetType} and immediately spend it back to the publisher and pool.`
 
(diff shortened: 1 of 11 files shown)


kremalicious:
first find, all the prices in output are different and do not reflect what we show on dataset/algo:

Screen Shot 2021-04-07 at 10 32 23

bogdanfazakas:
> first find, all the prices in output are different and do not reflect what we show on dataset/algo:
> 
> Screen Shot 2021-04-07 at 10 32 23

Hmm good finding, think that's because in compute component we get the price from the graph and send it to the BuyButton component and in AssetSelection we get it from the aquarius

mihaisc:
Fort [this asset](https://market-h5cvr2rjg-oceanprotocol.vercel.app/asset/did:op:8008Ef59A5B79353f15A06BdeB81cEd6c64ff7aC) the algo list fails to load (aquarius query fails). Probably because of this 
image

bogdanfazakas:
> Fort [this asset](https://market-h5cvr2rjg-oceanprotocol.vercel.app/asset/did:op:8008Ef59A5B79353f15A06BdeB81cEd6c64ff7aC) the algo list fails to load (aquarius query fails). Probably because of this
> image

Yeah it was fixed on the compute branch, will merge compute branch in here also

kremalicious:
made price output minimal, I think it is still clear that we add the 2 shown prices together:



for now left the full calculation code commented out, it's handy for debugging.

Except for the algo price differences, the remaining use case is combination of dataset or algo having previous order. In those cases they should be removed from `totalPrice`

kremalicious:
Unknown

Cause I think the timeout displayed in help message does not make sense:

Screen Shot 2021-04-07 at 13 32 00

I bought _Test Algo_ more than 1 day ago so just checking for previous order is not enough, we also need to check if previous order is within the timeout of the algo if I get this right. But we can leave all that to #474

mihaisc:
So the messaging regarding timeout will be handled in another pr or am i misunderstanding?

kremalicious:
with last night's commits this is the current state, could not find a good way to put all that info on screen at once, so moved to tooltip:

Screen Shot 2021-04-08 at 16 04 40

Screen Shot 2021-04-08 at 16 04 52

Also did all this timeout displaying stuff we talked about, and in case of previous orders made sure the price is not added:

Screen Shot 2021-04-08 at 16 05 52

Think the only thing left is that price difference thingy between graph & aquarius but in my testing on Rinkeby, this seemed to have solved itself. Prices seem always correct now so not sure we need to do something here

bogdanfazakas:
i would go for merging this, and for the changes I'm doing with the price for assets from the graph and the previous orders also I'll create a new PR with this subgraph utils.

kremalicious:
good idea with merging, also fine with this.

Suddenly prices are wrong again so we definitely need to get this from graph

Screen Shot 2021-04-08 at 17 09 06

kremalicious (approved):

PR #479, opened 2021-04-05
Feature/459 Check dataset connectivity on asset detail page

### LINKED ISSUE #459: Check dataset connectivity on asset detail page
When displaying the asset detail page, market should check if dataset is still available.

(ie: use [provider.fileInfo](https://github.com/oceanprotocol/ocean.js/blob/main/src/provider/Provider.ts#L136)

If files are not reachable, we should disable Download button

krisliew:
Fixes #459 

Changes proposed in this PR:
- On mount of consume and `Buy/Download` click, check dataset connectivity using `ocean.provider.fileinfo(did)`

### DIFF
--- a/src/components/organisms/AssetActions/Consume.tsx
+++ b/src/components/organisms/AssetActions/Consume.tsx
@@ -136,14 +136,14 @@ export default function Consume({
       }
 
       setIsDisabled(
-        (!ocean ||
-          !isBalanceSufficient ||
-          typeof consumeStepText !== 'undefined' ||
-          pricingIsLoading ||
-          !isConsumable) &&
-          !hasPreviousOrder &&
-          !hasDatatoken &&
-          !isFileValid
+        !isFileValid ||
+          ((!ocean ||
+            !isBalanceSufficient ||
+            typeof consumeStepText !== 'undefined' ||
+            pricingIsLoading ||
+            !isConsumable) &&
+            !hasPreviousOrder &&
+            !hasDatatoken)
       )
     }
     validateAsset()
@@ -159,16 +159,30 @@ export default function Consume({
   ])
 
   async function handleConsume() {
-    // TODO : check fileinfo
-    !hasPreviousOrder && !hasDatatoken && (await buyDT('1', price))
-    await consume(
-      ddo.id,
-      ddo.dataToken,
-      'access',
-      marketFeeAddress,
-      previousOrderId
-    )
-    setHasPreviousOrder(true)
+    let isFileValid: boolean
+    try {
+      const did = DID.parse(ddo.id)
+      const file = (await ocean.provider.fileinfo(did)) as any
+      isFileValid = file[0].valid
+    } catch (error) {
+      Logger.error(error)
+      isFileValid = false
+    }
+
+    if (isFileValid) {
+      !hasPreviousOrder && !hasDatatoken && (await buyDT('1', price))
+      await consume(
+        ddo.id,
+        ddo.dataToken,
+        'access',
+        marketFeeAddress,
+        previousOrderId
+      )
+      setHasPreviousOrder(true)
+    } else {
+      setIsDisabled(true)
+      toast.error('Dataset file endpoints is unavailable for consume.')
+    }
   }
 
   // Output errors in UI
--- a/src/components/organisms/AssetActions/Consume.tsx
+++ b/src/components/organisms/AssetActions/Consume.tsx
@@ -124,15 +124,20 @@ export default function Consume({
   }, [dtBalance])
 
   useEffect(() => {
-    setIsDisabled(
-      (!ocean ||
-        !isBalanceSufficient ||
-        typeof consumeStepText !== 'undefined' ||
-        pricingIsLoading ||
-        !isConsumable) &&
-        !hasPreviousOrder &&
-        !hasDatatoken
-    )
+    async function validateAsset() {
+      const isFileValid = await ocean.provider.fileinfo(ddo.id)
+      setIsDisabled(
+        (!ocean ||
+          !isBalanceSufficient ||
+          typeof consumeStepText !== 'undefined' ||
+          pricingIsLoading ||
+          !isConsumable) &&
+          !hasPreviousOrder &&
+          !hasDatatoken &&
+          !isFileValid
+      )
+    }
+    validateAsset()
   }, [
     ocean,
     hasPreviousOrder,
@@ -140,10 +145,12 @@ export default function Consume({
     consumeStepText,
     pricingIsLoading,
     isConsumable,
-    hasDatatoken
+    hasDatatoken,
+    ddo.id
   ])
 
   async function handleConsume() {
+    // TODO : check fileinfo
     !hasPreviousOrder && !hasDatatoken && (await buyDT('1', price))
     await consume(
       ddo.id,
--- a/src/components/organisms/AssetActions/Consume.tsx
+++ b/src/components/organisms/AssetActions/Consume.tsx
@@ -1,6 +1,6 @@
 import React, { ReactElement, useEffect, useState } from 'react'
 import { toast } from 'react-toastify'
-import { File as FileMetadata, DDO } from '@oceanprotocol/lib'
+import { File as FileMetadata, DDO, DID, Logger } from '@oceanprotocol/lib'
 import Button from '../../atoms/Button'
 import File from '../../atoms/File'
 import Price from '../../atoms/Price'
@@ -125,7 +125,16 @@ export default function Consume({
 
   useEffect(() => {
     async function validateAsset() {
-      const isFileValid = await ocean.provider.fileinfo(ddo.id)
+      let isFileValid: boolean
+      try {
+        const did = DID.parse(ddo.id)
+        const file = (await ocean.provider.fileinfo(did)) as any
+        isFileValid = file[0].valid
+      } catch (error) {
+        Logger.error(error)
+        isFileValid = false
+      }
+
       setIsDisabled(
         (!ocean ||
           !isBalanceSufficient ||


krisliew:
I am thinking about refactoring this by moving the check file endpoint code to `src\utils\provider.ts`. 
So it can be reuse for both `Consume.tsx` and `Compute.tsx ` in the future.

```
export async function checkFileEndpoint(
  id: string | DID,
  ocean: Ocean
): Promise<boolean> {
  try {
    const did = DID.parse(id)
    const file = (await ocean.provider.fileinfo(did)) as any
    return file[0].valid
  } catch (error) {
    Logger.error(error)
    return false
  }
}
```
But I got an error here from the `catch` block above, need some advice here please @kremalicious 
image

mihaisc:
Will look into this ( as well as the pr on ocean.js). I think there should be some clear messaging regarding this, not just disabling the button. People will be confused when seeing the disabled button without any reason

kremalicious:
as written over on https://github.com/oceanprotocol/ocean.js/pull/710, in case we plan on using the provider `fileinfo` endpoint in here, we should not rely on `ocean.provider`. Think it's fine for the `ocean.provider.isConsumable` part since we also disable buy button when no wallet is connected.

But why not show a status icon for everybody, no matter any wallet connection? For that we should hit the provider directly with our own `src/utils/provider.ts` methods

mihaisc:
Agree, in the market we should make use of the helper function to make a direct call to the provider and do the check.

TSS-LauKY:
The method on src/utils/provider.ts only support url as string type, are we going to replicate the logic from ocean.js to support the argument as` did: url.getDid()` ?

alexcos20:
can we keep this aside for a few days?
With the addition of allow/deny lists, we may find a better way of doing this in ocean.js

mihaisc:
Not sure this is related to allow/deny lists, this just checks the file

mihaisc:
So the solution here would be to remove the axios call from [here](https://github.com/oceanprotocol/market/blob/49caca234d6d7ad016778af9cf9d317c84e92dcd/src/utils/provider.ts#L23) and create another function (eg `getFileInfo`), replace the axios logic with the function, create another function `isFileValid` that uses the `getFileInfo` and checks the `valid` prop

src/components/organisms/AssetActions/Consume.tsx
```diff
@@ -124,35 +130,60 @@ export default function Consume({
...
-        !isConsumable) &&
-        !hasPreviousOrder &&
-        !hasDatatoken
-    )
+    async function validateAsset() {
+      const did = DID.parse(ddo.id)
+      const fileValid = await isFileValid(
+        did,
+        ddo.findServiceByType('access').serviceEndpoint,
+        axios.CancelToken.source().token
```
mihaisc, line 138: Not the proper way to handle cancel token, check here https://github.com/oceanprotocol/market/blob/1da7eae35a1bfe24bced84d99089638b772990b5/src/providers/Asset.tsx#L162

mihaisc:
Looks good. After the minor fix i'll do some testing. This will not be merged until we merge compute

kremalicious:
not sure what happened here, but 155 changed files and 56,556 lines of code changed can't be right. Needs a rebase or merge against current `main`

krisliew:
> not sure what happened here, but 155 changed files and 56,556 lines of code changed can't be right. Needs a rebase or merge against current `main`

Fixed it. Think something were messed up when I merge against current `main`  :(

jamiehewitt15:
I feel like the user experience of this could still be improved. It seems to allow you to click the "buy" button and then it checks the data connectivity after. Also, there seems to be some sort of glitch as I managed to buy a datatoken for a dataset that's unavailable after refreshing the page.

krisliew:
> I feel like the user experience of this could still be improved. It seems to allow you to click the "buy" button and then it checks the data connectivity after. Also, there seems to be some sort of glitch as I managed to buy a datatoken for a dataset that's unavailable after refreshing the page.

Appreciate the feedback @jamiehewitt15. Do you have the steps I can follow to reproduce the glitch you encountered?

jamiehewitt15:
1. I published a csv to web with Google sheets.
2. Published the dataset on Ocean market.
3. Created dynamic pricing. 
4. Removed the csv from the web in Google Sheets and therefore making it unavailable via the URL I published on Ocean Market. 
5. Refreshed the dataset page a couple of times in Ocean Market.
6. Clicked "Buy" and successfully bought a data token 
7. The purchase of the datatoken was successful but it didn't download a dataset and it won't allow me to click download. 

I was using the Moonbase Alpha network for the test but it should be the same on the other networks.

krisliew:
> 1. I published a csv to web with Google sheets.
> 2. Published the dataset on Ocean market.
> 3. Created dynamic pricing.
> 4. Removed the csv from the web in Google Sheets and therefore making it unavailable via the URL I published on Ocean Market.
> 5. Refreshed the dataset page a couple of times in Ocean Market.
> 6. Clicked "Buy" and successfully bought a data token
> 7. The purchase of the datatoken was successful but it didn't download a dataset and it won't allow me to click download.
> 
> I was using the Moonbase Alpha network for the test but it should be the same on the other networks.

@jamiehewitt15  Tried reloading and swapping tab multiple time, worked fine for mine on eth mainnet. My asset did `did:op:71f23C47d0Ad81aC6662C50398093750a1Cd9B61`. 
Plus, consume button is always disable (default) on mount too... https://github.com/oceanprotocol/market/blob/020d4fa05ad38da68762b85df6db0a78020f6fda/src/components/organisms/AssetActions/Consume.tsx#L58.

jamiehewitt15:
> 
> Did I miss anything? 

I'll run another couple of checks

jamiehewitt15:
I did another test run (this time on Rinkerby) and it let me buy the datatoken again, even though the dataset is unavailable. I can make a video if it helps

krisliew:
> I did another test run (this time on Rinkerby) and it let me buy the datatoken again, even though the dataset is unavailable. I can make a video if it helps

@jamiehewitt15 Yes please, a video will be helpful because I can't seem to reproduce that

jamiehewitt15:
@krisliew see here: https://www.loom.com/share/7591c2bfa9c5459eaf1bb773b1203599

krisliew:
> @krisliew see here: https://www.loom.com/share/7591c2bfa9c5459eaf1bb773b1203599

@jamiehewitt15 I found out why. When we stop publishing the google sheet, the file endpoint will temporary have an error page where fileinfo check returns `true`. 
image

But after google sheets "removed" the file endpoint, it will show this page where fileinfo check return `false`. 
image

Did you try reloading the page again, I think the buy button would be fully disabled without any issue.

jamiehewitt15:
I reloaded the page after the dataset was made unavailable and it still allowed me to click the "buy" button

krisliew:
@jamiehewitt15 can you share me or check your dataset file endpoint? what does it show?

kremalicious:
good direction, but if we render message & button at same time, let's remove the output of the loading state. Users are not interested in seeing this on every asset and in every interaction with algo list, but only when an error occurs.

On the other hand, if we deem the file check reliable enough and consider it super important than a better UI pattern would be waiting for the file check before rendering the button in which case outputting the loading state is justifiable. Because rendering button and message at same time could lead to situations where button is enabled because all other checks are done already, but the the file check might be pending

src/components/atoms/ButtonBuy.tsx
```diff
@@ -23,6 +26,40 @@ interface ButtonBuyProps {
   type?: 'submit'
 }
 
+function getConnectivityHelpText(
```
kremalicious, line 29: This is returning an actual React component, not just text string so naming and code style should be adapted accordingly where this method here should be a component like `ConnectivityHelp` with props instead of params.

src/components/organisms/AssetActions/Consume.tsx
```diff
@@ -59,6 +61,7 @@ export default function Consume({
   const [hasDatatoken, setHasDatatoken] = useState(false)
   const [isConsumable, setIsConsumable] = useState(true)
   const [assetTimeout, setAssetTimeout] = useState('')
+  const [fileConnectivity, setFileConnectivity] = useState<boolean>()
```
kremalicious, line 64: we are repeating this state setting and updating 3 times, suggesting it should be abstracted out for DRY reasons. Could be done inside a new `ConnectivityHelp` component under `atoms`

jamiehewitt15:
@krisliew I'm concerned that the check might be too unreliable. A lot of our most popular datasets are being marked as potentially unavailable...

krisliew:
I just looked through them, endpoint response `valid` returns `true` but some of the `contentLength` returned is either `""` or `"0"` which is why the warning is prompting (Altho i only checking for `"0"` now).

image

I think there should be a better way of handling this issue. Perhaps we should only apply this checking if the data set owner/publisher decide so as to cater this story #581 (like a toggle checkbox for data streaming on publish), or not add this at all since is possible that there will be many more endpoint (eg: google sheet csv) is not supporting the content-length properly (I assume)? 

@alexcos20 @kremalicious @jamiehewitt15 What do you think?

Reference to getting `contentLength` function: 
- fileInfo(): https://github.com/oceanprotocol/provider/blob/79de2a4aabd248ac2c72184d7067c58ac3ca7bbf/ocean_provider/routes/consume.py#L149
- contentLength: https://github.com/oceanprotocol/provider/blob/6abbf7d0e28f227ef8044e39ae4ce08911cfb099/ocean_provider/utils/url.py#L119

kremalicious:
let's dial back on the UI, info that an asset MIGHT not be reachable is not the most important info on the whole screen, but UI so far tries to be the most important thing on the page. At same time we can't add more walls of text under the button, of which more are coming in in other PRs. So adding more paragraphs under the main action button is simply not scalable.

I would propose this, where we can explain the check and its implications in the tooltip:

Screen Shot 2021-05-25 at 07 30 30

Nevertheless we should still try to make sure this can't be missed so simply doing the text in our alert red works too:

Screen Shot 2021-05-25 at 07 30 52

Similar on Compute. By placing the warnings in context, we get around having to repeat name/symbol/type in the visible warning text:

Screen Shot 2021-05-25 at 07 40 01

src/components/molecules/AssetStatus.tsx
```diff
@@ -0,0 +1,46 @@
...
+    }
+  }, [ddo])
+
+  if (fileConnectivity) return null
+
+  return (
+    <div className={styles.wrapper}>
+      <Tooltip content="This dataset might be empty/offline, consume at your own risk.">
+        <Status className={styles.status} state="error" />
+        <span className={styles.text}>Unavailable</span> ⓘ
```
kremalicious, line 42: trigger should not be a UTF-8 symbol but our `DefaultTrigger` svg icon from the `Tooltip` component. In this case here it would make sense only having the icon as a trigger, so not wrapping the whole text as a tooltip trigger. If we do that, there's no need to pass any `children`, in turn triggering the `DefaultTrigger`, like so:

```js
<span className={styles.text}>Unavailable</span> <Tooltip content="This dataset might be empty/offline, consume at your own risk." />
```

src/components/molecules/AssetStatus.tsx
```diff
@@ -0,0 +1,46 @@
...
+import React, { ReactElement, useEffect, useState } from 'react'
+import { useOcean } from '../../providers/Ocean'
+import { isFileValid } from '../../utils/provider'
+import Status from '../atoms/Status'
+import Tooltip from '../atoms/Tooltip'
+import styles from './AssetStatus.module.css'
+
+export default function AssetStatus({ ddo }: { ddo: DDO | DID }): ReactElement {
+  const { ocean } = useOcean()
+  const [fileConnectivity, setFileConnectivity] = useState<boolean>(true)
```
kremalicious, line 12: just small naming thing, would be more clear to name these `isFileValid, setFileValid` to stay consistent with the used method and to better signal the type of variable to other developers (`is` prefix always implies a boolean)

krisliew: We can't use `isFileValid` as it is colliding with the method name from `src/utils/provider.ts`.

src/components/molecules/FormFields/AssetSelection.tsx
```diff
@@ -82,6 +84,9 @@ export default function AssetSelection({
                   {...props}
                   disabled={disabled}
                   value={asset.did}
+                  onClick={(e: any) => {
```
kremalicious, line 87: this can lead to unintended consequences. Like, it just keeps `asset.checked` as `true` without deselecting it again. But instead of replicating formik functionality we should rather figure out to get selection state from existing functionality. Like:

- figure out if we can use `asset.checked` for keeping checked state instead of only using it for the default selected state
- the `value` prop in `AssetSelection` always holds the DID of the currently selected asset. So figure out how to use this to figure out what user has selected

src/utils/provider.ts
```diff
@@ -57,3 +74,19 @@ export async function fileinfo(
...
+  providerUri: string,
+  cancelToken: CancelToken
+): Promise<boolean> {
+  const response = await getFileInfo(url, providerUri, cancelToken)
+
+  if (!response || response.status !== 200 || !response.data) return false
+
+  if (!response.data[0] || !response.data[0].valid) return false
+
+  if (response.data[0].contentLength === '0') return false
```
kremalicious, line 89: this seems to still produce many false positives. Of the current 9 highest liquidity assets, 4 are being reported as unavailable assets which I highly doubt. If we ship this check in this state the top publishers will come to us and ask why we tell people their assets are not available when in fact they are.

krisliew: Yes, should we exclude the `contentLength` check in this PR?

src/components/molecules/AssetStatus.tsx
kremalicious: good point, didn't see that. State var definitely should be called `isFileValid` but contradicting my own first argument the method then could be like `checkFileValidity()` but sounds kinda weird, maybe just go with `checkFile()`. Naming is so hard 😄

src/utils/provider.ts
kremalicious: it seems so. Just tested and can confirm all highest liquidity assets are fine with this check removed. Still kinda weird casue upon publishing we seemed to have been able to correctly detect the size on all of them but they were published with our old `fileinfo` Node.js app endpoint. But don't see why `provider` would not be able to get the `contentLength`

krisliew: Yes...the checking gets crazier if you test it with empty string, `contentLength === ""`. Not sure why plenty of endpoint is returning `""`.

src/components/molecules/AssetStatus.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  useEffect(() => {
+    async function validateAsset(ddo: DDO | DID, source: CancelTokenSource) {
+      if (!ddo) return
+      if (ddo instanceof DID) ddo = await ocean.metadataCache.retrieveDDO(ddo)
+
+      const fileValid = await isFileValid(
```
kremalicious, line 19: we might have to move the provider call out of here. I also always have this habit of creating one component as combination of visual and functionality but that will lead to multiple calls to the same thing pretty quickly in our case.

Here we will have clashes with #631 which centralizes the provider file check into `AssetActions/index` and passes down results as props. This is a better pattern to prevent multiple provider calls by multiple components. So this component here then can receive that file check object and go from there.

Basically somebody has to refactor this, question is if we can easily do it in here already, or if @claudiaHash will be left having to refactor a lot

src/components/molecules/FormFields/AssetSelection.tsx
krisliew: Thanks, this is helpful 
I am thinking of a solution like this:
```
  const [selectedAlgorithm, setSelectedAlgorithm] = useState('')
  useEffect(() => {
    setSelectedAlgorithm(props.value)
  }, [props.value])
```
And then to display the status check:
```
  {selectedAlgorithm === asset.did && (
    <AssetStatus ddo={DID.parse(asset.did)} />
  )}
```

src/utils/provider.ts
krisliew: I had removed the contentLength check for now.

src/components/molecules/AssetStatus.tsx
krisliew: Agree with your point of centralising the provider file check at `AssetActions/index`.

I can refactor this here right now if the changes at #631 `AssetActions/index` is almost finalised, perhaps I can just copy that over?

krisliew: @kremalicious At `src/components/molecules/FormFields/AssetSelection.tsx` I think we still need to call provider file check each time a user select a different algorithm for compute asset. Unless we are doing the file check at `AssetActions/index` for all algorithm on mount which can be quite taxing on performance if the algorithm list is long. 
What do you think?

image

src/utils/provider.ts
alexcos20: can you get some dids ?
I will open an issue on provider to see why do we get such responses

krisliew: @alexcos20  
These are the DID of highest liquidity from ETH mainnet:

DID with `contentLength = ""` 
- did:op:a2B8b3aC4207CFCCbDe4Ac7fa40214fd00A2BA71 
- did:op:fcB47f5781F14Ed7E032BD395113b84C897aA23f

DID with `contentLength = "0"`
- did:op:7Bce67697eD2858d0683c631DdE7Af823b7eea38
- did:op:2a76F680279CE629a9F5E601BDa7246e06F226f0

src/components/molecules/AssetStatus.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { useOcean } from '../../providers/Ocean'
+import { checkFile } from '../../utils/provider'
+import Status from '../atoms/Status'
+import Tooltip from '../atoms/Tooltip'
+import styles from './AssetStatus.module.css'
+
+export default function AssetStatus({
+  did
+}: {
+  did: DID | boolean
```
kremalicious, line 13: what's the reasoning behind `DID | boolean`? Makes this prop a bit confusing and I would suggest rather passing the `did` or not based on whatever boolean prop is checked here.

src/components/molecules/AssetStatus.tsx
kremalicious: yeah, I see we have created a nice dead end where the proper way would be to make that status part of all AssetSelection asset data but in case of compute it can be overkill. So let's try at least that in download case we only have 1 provider call in tandem with #631

src/components/organisms/AssetActions/Consume.tsx
```diff
@@ -164,6 +167,7 @@ export default function Consume({
       <div className={styles.info}>
         <div className={styles.filewrapper}>
           <File file={file} />
+          <AssetStatus did={file.valid} />
```
kremalicious, line 170: this is the confusing usage, cause why would the did be a boolean?

As for the typing errors I see that `valid` was added over on ocean.js so this PR needs to have that new ocean.js version bumped for the build and tests to succeed

src/components/molecules/AssetStatus.tsx
```diff
@@ -0,0 +1,51 @@
...
+}): ReactElement {
+  const { ocean } = useOcean()
+  const [isFileValid, setFileValid] = useState<boolean>()
+
+  useEffect(() => {
+    async function validateAsset(did: DID, source: CancelTokenSource) {
+      const ddo = await ocean.metadataCache.retrieveDDO(did)
```
kremalicious, line 20: this won't work without a wallet connected. Please use the methods in `utils/aquarius.ts` which replicate these ocean.js methods, but directly hitting the Aquarius API

src/components/molecules/AssetStatus.tsx
```diff
@@ -0,0 +1,51 @@
...
+      if (!ddo) return
+
+      const fileValid = await checkFile(
+        DID.parse(ddo.id),
```
kremalicious, line 24: so the component accepts a `did` but then fetches the DDO again just to get the `did`?

src/components/molecules/AssetStatus.tsx
krisliew: I can change that but I think the provider file info call wouldn't work either without a wallet connected since we are using `config.providerUri` from `useOcean()` as well at `src/components/organisms/AssetActions/index.tsx`. This apply the same to https://github.com/oceanprotocol/market/pull/631

This mean to do the provider fileinfo check, we would need a wallet connected anyhow? 😞

src/components/molecules/AssetStatus.tsx
krisliew: The DID is for the compute asset's algorithm list, and the boolean prop is for accepting the `valid` prop from `AssetActions/index`.

src/components/molecules/AssetStatus.tsx
krisliew: What about the connectivity check for compute's algorithm list, can you provide an approach please?

src/components/molecules/AssetStatus.tsx
kremalicious: the `config` values are always present, without a wallet we load the default configured ones. But the `ocean` object is not and will only be initiated with a valid web3 provider, hence all methods under the `ocean` object are only available with a wallet. Which is why we replicate aquarius & provider methods for which a wallet is technically not required, like metadata query or the provider file check

krisliew: What about cases when a data asset use a different/custom provider? Can we get the custom providerUri from the DDO?

kremalicious: yeah, good point, we actually need to if we want to cover different providers but I see you are already passing the respective `serviceEndpoint` further down, so that is the correct way of getting the provider url for each asset. Guess we need to strip out the path from the `serviceEndpoint` so we only get the protocol + host

src/components/molecules/AssetStatus.tsx
kremalicious: so let's make that clear in code to get this component more flexible and effective.

1. instead of `did` prop, call it `asset` and have it accept `DDO | DID`. Then when you detect a DDO you skip the `retrieveDDO` call further down. This solves the aquarius double fetching for usages where parent component has the DDO already but still allows refetching in case we do not, like in algo list.
2. still not sure what the boolean is supposed to do, I see this line further down `else if (did === false) setFileValid(did)`. So if your boolean is false you set the file to be valid? If the intention is to signal the component to skip all checks and just set the file to valid, then would use an optional `isFileValid` prop

krisliew: For 2. I am naming the optional prop as `isValid` because we already having: https://github.com/oceanprotocol/market/blob/2ac7f7eb331bb2fa5d02d5199c6643a392d9e92a/src/components/molecules/AssetStatus.tsx#L16

src/components/molecules/AssetStatus.tsx
krisliew: In that case, do this need to be updated to pass `serviceEndpoint`

jamiehewitt15:
@kremalicious We discussed this morning the option that this could just be merged into the Daimler market rather than our own. That could be the best way forward to close this.

kremalicious:
We are testing all morning already. So far I could not get the warning to show up anywhere where I don't know if it's good or bad. Meaning we seem to not get any false positives anymore (which is good) but question arises if the check is realistically doing anything since we only left a check for a 200 response in the end. Do you have some test asset 

In terms of effective requests: for me when clicking on any algo in the AssetSelection list, I get 4 requests to Aquarius for getting the DDO where 2 of them fail for whatever reason:

Screen Shot 2021-06-10 at 11 31 45

On live there are 0 requests and in theory for this PR we should have only 1 additional one when clicking on an algo. Still trying to figure out what's going on here.

And PR now also includes the file metadata fetching which was also done in #631 but this PR has a slightly different behavior (rendering the empty message first instead of loading spinner). So after merge or rebase against `main` that file metadata loading mechanism from `main` should take over

kremalicious:
We are also hitting the provider multiple times. Just look at all the requests happening when clicking on an algo. I count 4 `GET` Aquarius requests and 4 `POST` provider requests where for each of them there should be only 1:

https://user-images.githubusercontent.com/90316/121504458-50c7df00-c9e2-11eb-8c57-21d2c7c05c02.mov

mihaisc:
i'll take a look

kremalicious:
Also getting the memory leak warnings suggesting the axios cancel token mechanism is not working properly. On every asset details page, where a compute assets gets 2 warnings:

Screen Shot 2021-06-10 at 12 31 34

mihaisc:
I get this one on other branches as well

kremalicious:
When tinkering around in code a more holistic solution which might work:

It seems way more effective to ditch the DDO fetching completely within `AssetStatus` as we only need the service endpoint in there. And this we should pass as prop. Then when fetching the data to construct the `AssetSelection` we add another prop `serviceEndpoint` for each algo. This then can be picked up by `AssetStatus` and we would not have any need to fetch the DDO again. Solving all aquarius double requests already, only pending figuring out why so many provider requests are happening

kremalicious:
oh, good find, should just check my own assets!

That file is actually available but it's pinned on IPFS via a gateway URL. Gateway URL works for me but not sure this all works with provider

jamiehewitt15:
@mihaisc @kremalicious What still needs to be done with this? Is the current plan still to merge this into the Ocean Market or is it better if it's just merged directly into the Daimler version of the market?

kremalicious:
the provider needs to be fixed. Basically we are not going to ship an unreliable check producing false positives on assets [we know they are available](https://github.com/oceanprotocol/market/pull/479#issuecomment-858635764).

So while we already removed the `contentLength` check in here because provider simply does not returns this for a lot of assets, we now find cases where even for the basic 200 check the provider produces wrong data. Which, again, needs to be investigated and fixed in provider, then checked again after new provider version with this PR.

mihaisc:
Provider should be ok now,  please rebase and check the issue that @kremalicious reported in the previous comment

kremalicious:
Revisit after v4 with changed provider

PR #482, opened 2021-04-07
AssetSelection: add search by datatoken symbol

kremalicious:
minor tweak so this works:

Screen Shot 2021-04-07 at 15 29 17


mihaisc:
Should we make this case insensitive? It's kinda weird to have caps on for this
image
image

mihaisc (approved):

PR #483, opened 2021-04-08
Feature/issue474 algorithms timeout

### LINKED ISSUE #474: deal with algorithms timeout

claudiaHash:
Fixes #474.

Changes proposed in this PR:

- timeout field added in algorithms publish form

### DIFF
--- a/content/pages/publish/form-algorithm.json
+++ b/content/pages/publish/form-algorithm.json
@@ -47,6 +47,16 @@
       "required": false
     },
     {
+      "name": "timeout",
+      "label": "Timeout",
+      "help": "Define how long buyers should be able to download the algorithm again after the initial purchase.",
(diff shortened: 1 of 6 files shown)


kremalicious (approved):
quickly tested publish & edit with timeout, works all like a charm

PR #487, opened 2021-04-09
fix ocean symbol on Polygon

### LINKED ISSUE #456: wrong OCEAN token symbol on Polygon
We have some places left where we do not switch to `mOCEAN` when market is connected to Polygon, and all other future networks with bridged OCEAN:

- `Price` component
- everything within pool & trade tab
- pool shares on History page

Currently, `OCEAN` is hardcoded within the pool & trade tabs because of https://github.com/oceanprotocol/ocean-subgraph/issues/13. With https://github.com/oceanprotocol/ocean-subgraph/pull/61 incoming, we should switch to getting both symbol names from GraphQL query.

For the `Price` component we should take the symbol from `config.oceanTokenSymbol` from `useOcean()`

KatunaNorbert:
Fixes #456 .

Changes proposed in this PR:

Switch ocean symbol depending on network in the following places:
- Price component
- everything within pool & trade tab
- pool shares on History page


KatunaNorbert:
Hardcoded the user address and subgraph2 url to get the new changes from the subgraph and pool shares on Polygon so I can test the symbol. These should be changed after testing and before merge.

kremalicious:
yeah, not sure about which subgraph exactly to use. Feels like we should solve this properly as was prepared with https://github.com/oceanprotocol/market/pull/457 but was waiting for some new graph deployment

KatunaNorbert:
I think we can merge this and #457 together before merging into main. Sorry but not sure what you mean by proper way, should we get every ocean symbol from the graph?

mihaisc:
Not sure if we can actually solve it "properly" here. Don't think this needs more attention, it's good enough for now. The subgraph was not released on main so we have to wait for that.
In the future with multi pools and other base tokens we will have a different mindset for handling pools so no need to handle it here.

kremalicious:
ok cool, then let's move forward with this. We should remove all the hardcoded URL changes, not use `subgraph2`, and no hardcoded accounts in this PR

kremalicious:
well, this needs a live subgraph update after all because in this PR we use changes from https://github.com/oceanprotocol/ocean-subgraph/pull/61 for the pool shares https://github.com/oceanprotocol/market/pull/487/files#diff-7b725b7c0d9d14eaa89b27b537b53a12ec596ce0984a1f3cde57a9c75d31ca1e

KatunaNorbert:
Than this will only fix the price component and trade tab or should fix the other ones too by using `config.oceanTokenSymbol`?

mihaisc:
this fix is needed as well https://github.com/oceanprotocol/ocean-subgraph/issues/200

mihaisc:
Before merging we need to remove the changes from `apollo.config.js` and `ocean.ts` . This merge need to be done in sync with the subgraph change since there are some minor changes in the subgraph that will break some pool functionalities.

I moved `BestPrice` from ocean.js to market since it has no use there at this point.

src/components/atoms/Price/PriceUnit.tsx
```diff
@@ -48,7 +49,8 @@ export default function PriceUnit({
         <>
           <div>
             {Number.isNaN(Number(price)) ? '-' : formatPrice(price, locale)}{' '}
-            <span className={styles.symbol}>{symbol || 'OCEAN'}</span>
+            {/* do we want a default symbol? */}
```
kremalicious, line 52: think it's fine without a default or fallback. We could define some base token in the app config but just querying the symbol seems more reliable anyway.

kremalicious (approved):

mihaisc (approved):

mihaisc:
do not merge, found some issues after merging main

PR #488, opened 2021-04-09
Subgraph helpers for assetSelection list and previous orders

bogdanfazakas:
Fixes # .

Changes proposed in this PR:

- set price for each asset in AssetSelection from the graph
- get previous orders for compute selceted assets from the graph


src/utils/subgraph.ts
```diff
@@ -0,0 +1,120 @@
...
+    const unixTime = new BigNumber(Math.floor(Date.now() / 1000))
+    if (unixTime.isLessThan(expiry)) {
+      return fatchedPreviousOrders?.data?.tokenOrders[0]?.tx
+    } else {
+      return null
+    }
+  }
+}
+
+export async function getAssetPrice(assets: DDO[]): Promise<any> {
```
kremalicious, line 88: this is nice, and I can see how this could replace relying on `ddo.price.value` in other places, like on asset teasers in all search query lists.

So we should be precise and also type the response, not just `any`. Would call method plural then, `getAssetPrices()`

src/utils/subgraph.ts
```diff
@@ -0,0 +1,120 @@
...
+    }
+  }
+  const freVariables = {
+    datatoken_in: frePriceAssets
+  }
+  const poolVariables = {
+    datatokenAddress_in: poolPriceAssets
+  }
+  const poolPriceResponse: any = await fetchData(poolQuery, poolVariables)
+  for (const poolPirce of poolPriceResponse.data?.pools) {
```
kremalicious, line 112: small typo, variable should be `poolPrice`, not `poolPirce`

bogdanfazakas: busted 😄 , i'll update that

src/utils/subgraph.ts
bogdanfazakas: Updating method name now, but not sure how can we be more strict with return type because it's an object so it helps us accessing the price using the dataset did in populating asset selection list without iterating the returned price list again
like this :  price: priceList[did]

kremalicious: yup, that's all fine, what I meant is that we know the structure of the object returned, so instead of typing it as `any` we should be precise. If I quickly see this right, we are also returning `DDO[]`, right? So should be typed as `Promise<DDO[]>` then

bogdanfazakas: no, actually we return a object not an array, eg object =  { 'didAsset1': priceValueAsset1, 'didAsset2': priceValueAsset2}

src/utils/subgraph.ts
```diff
@@ -0,0 +1,120 @@
...
+}
+
+export async function getAssetPrices(assets: DDO[]): Promise<any> {
+  const priceList: any = {}
+  const poolPriceAssets: string[] = []
+  const poolDTadressDID: any = {}
+  const frePriceAssets: string[] = []
+  const freDTadressDID: any = {}
+  for (const ddo of assets) {
+    if (ddo.price?.type === 'pool') {
```
mihaisc, line 95: we should stop relying on this from now on.

bogdanfazakas: true, started on it after the call pushing it right now, the query's will be a bit bigger because we will search all assets for both fre or poll but as far i see in local testing no big difference

mihaisc: you were doing that already, that's why there is no difference

src/utils/subgraph.ts
kremalicious: so then sounds like the typing should be `{ [key: string]: number }` or `{ [key: string]: string }` instead of `any`? We repeat that `any` then in many places so would make sense creating interface `PriceList` for it here and import that wherever we need it. But if we correctly type the return here there will be no need to do stuff like this in other places:

```ts
const priceList: any = await getAssetPrices(ddoList)
```

Cause then the typing is inferred from usage so this will be enough instead:

```ts
const priceList = await getAssetPrices(ddoList)
```

kremalicious:
awesome, confirmed all prices are in sync in this place now:

Screen Shot 2021-04-14 at 11 17 10

Getting this though, but I guess this is actually what we want now when checking for `isOrderable`?

Screen Shot 2021-04-14 at 11 18 44

Bit confusing to users as to what is wrong but at lest we don't send people into failed transactions so should not be a blocker. Ideally we have some visual pattern to signal to users right after selecting algo that the algo won't work, like dim it, or some tiny warning text on the row.

Case could be also made to hide algos where we know they are not orderable from the `AssetSelection` here altogether, or dim them, or group them in the list to the bottom with some group header _Not Usable_ or something.

src/utils/subgraph.ts
bogdanfazakas: yeah that could work, having an interface PriceList : { [key: string]: string } an using it

bogdanfazakas:
> awesome, confirmed all prices are in sync in this place now:
> 
> Screen Shot 2021-04-14 at 11 17 10
> 
> Getting this though, but I guess this is actually what we want now when checking for `isOrderable`?
> 
> Screen Shot 2021-04-14 at 11 18 44
> 
> Bit confusing to users as to what is wrong but at lest we don't send people into failed transactions so should not be a blocker. Ideally we have some visual pattern to signal to users right after selecting algo that the algo won't work, like dim it, or some tiny warning text on the row.
> 
> Case could be also made to hide algos where we know they are not orderable from the `AssetSelection` here altogether, or dim them, or group them in the list to the bottom with some group header _Not Usable_ or something.

yeah that error is displayed when asset is not orderable, in you case it's probably because your dataset is on other provider that the selected algorithm with service type compute
we add some checks in AssetSelection and pass the dataset provider and not display algorithms with compute services that are using diferent providers in the list

kremalicious:
haha, this is a product design nightmare. If the restriction of our product is that dataset and algo need to be on the same provider I would be in favor of only showing algos of the same provider in the list. Or we prevent them from showing up when publisher edits compute settings, then we will not run into the issue on the start compute side

mihaisc:
So you can use an algo from another provider if that algo has an `access` service not a `consume` service. Since the filtering is done client side you can remove the algos that have `compute` service with a different provider

kremalicious:
can't order anything.

On Rinkeby, trying to combine these, both on same provider:
- https://market-git-feature-sugraph-util-methods-oceanprotocol.vercel.app/asset/did:op:d5766cc73AE44977b4F7d72cDd3b33740101e830
- https://market-git-feature-sugraph-util-methods-oceanprotocol.vercel.app/asset/did:op:49A558D5DeE4b54B7fE4A578996e2Fc4E24847B3

Still giving me:

Screen Shot 2021-04-14 at 13 26 10

bogdanfazakas:
> can't order anything.
> 
> On Rinkeby, trying to combine these, both on same provider:
> 
> * https://market-git-feature-sugraph-util-methods-oceanprotocol.vercel.app/asset/did:op:d5766cc73AE44977b4F7d72cDd3b33740101e830
> * https://market-git-feature-sugraph-util-methods-oceanprotocol.vercel.app/asset/did:op:49A558D5DeE4b54B7fE4A578996e2Fc4E24847B3
> 
> Still giving me:
> 
> Screen Shot 2021-04-14 at 13 26 10

Strange, i also got another error for this asset combo, but when trying to buy DT, an oceanjs error 'Not enough Ocean Tokens'

mihaisc:
you can't buy this https://market-git-feature-sugraph-util-methods-oceanprotocol.vercel.app/asset/did:op:49A558D5DeE4b54B7fE4A578996e2Fc4E24847B3

mihaisc:
can you add this algo to trusted https://market-git-feature-sugraph-util-methods-oceanprotocol.vercel.app/asset/did:op:F1ffeB5eCA97da9144DC2964A5E2361F0A71DD99

mihaisc:
but this issue has nothing to do with this pr

kremalicious:
ah ok, so just tried with that _Row count_, but get the same. If nothing to do with this PR then I'm fine just moving ahead and merging, and fixing in another PR

Screen Shot 2021-04-14 at 14 17 04

bogdanfazakas:
> ah ok, so just tried with that _Row count_, but get the same. If nothing to do with this PR then I'm fine just moving ahead and merging, and fixing in another PR
> 
> Screen Shot 2021-04-14 at 14 17 04

can you try again now, for me started working suddenly maybe again some compute env issues

kremalicious:
nope, but this message seems interesting

```
ERROR: Algorithm [object Object] is not allowed by did:op:d5766cc73AE44977b4F7d72cDd3b33740101e830
```

Screen Shot 2021-04-14 at 14 27 01

kremalicious:
I think my bad, testing locally and despite fresh `npm install` and all still failing. But was able to run successfully on preview version 🥳 

Screen Shot 2021-04-14 at 14 34 04

kremalicious (approved):

PR #489, opened 2021-04-09
remove status and text status check on start job

bogdanfazakas:
Fixes # .

Changes proposed in this PR:

- removes check for status and status text on start compute job


kremalicious:
question is if an error would return a `response` too?

As it looks like the `statusText` is actual problem in our found issue so maybe we adapt that check for `Warming Up`, or not check for `statusText`

Screen Shot 2021-04-09 at 16 46 08

mihaisc:
I think we can ignore that, because you clearly started the job. Those errors can be seen in history. As well we can have a history in the asset similar with pool transactions

kremalicious (approved):

Issue #491, opened 2021-04-09
Cannot submit a URL without `https://` or `http://`  


jamiehewitt15:
**Describe the bug**
Cannot submit a URL / publish a dataset without the proceeding  `https://` or `http://`  

**To Reproduce**
1. Go to https://market.oceanprotocol.com/publish
2. Enter URL without `https://` or `http://` e.g.  www.example.com
3. It is not possible to submitt

**Expected behavior**
Any valid url should work without the proceeding `https://` or `http://`

jamiehewitt15:
Related pull request: https://github.com/oceanprotocol/market/pull/493

PR #493, opened 2021-04-09
Allowing urls without proceeding http or https

jamiehewitt15:
Closes: https://github.com/oceanprotocol/market/issues/491

Changes proposed in this PR:
- Changing the URL check so that users can enter URL's without needing to write `https://` or `http://`


jamiehewitt15:
I've removed the pre-submit URL validation. Previously, the "Add File" button was inactive until a valid URL was entered. Now the "Add File" button is active as soon as a string is entered. This should be more user friendly. We still check the file after the user has submitted the URL

KatunaNorbert:
The url is indeed accepted as valid without having to write `http` or `https`, but with removing pre-submit URL validation the new package is not used right? We only relay on `validateUrl()`.

jamiehewitt15:
@KatunaNorbert thanks for the feedback, I have removed the unused package. With this I was trying to follow what @kremalicious said about all buttons should be enabled by default so that the user can click on them and get an error message (otherwise the user doesn't know why the button is deactivated). 

If we leave the pre-submit URL-validation it could lead to users getting confused. The user could enter a URL without the proceeding `http` or `https` and they wouldn't know why the button is still deactivated.

KatunaNorbert:
Alright then, it's fine from my side

KatunaNorbert (approved):

mihaisc:
If all good you can merge it, don't need to wait. Usually the creator of the pr merges after the approve.

jamiehewitt15:
Thanks, I've merged it

PR #494, opened 2021-04-12
Removing tooltip animation onHover

### LINKED ISSUE #390: Liquidity graph isn't showing correct data when hovering with mouse
System: Mac OS Catalina
Browser: 86.0b4 (64-bit)

When I go to a data set with high liquidity and hover over the liquidity graph I get multiple problems:

- The indicator in the graph (red dot) is not appropriately following my mouse cursor
- Somehow data recorded in November is shown on a much broader scale than e.g. data for Jan 2021 or Feb 2021

See .gif for example

liquiditygraph

- I assume this problem also exists for the price graph. But I haven't tested that.

DimitarSD:
Fixes #390 

Changes proposed in this PR:
- removes the animation for the tooltip window onHover
- sets intersect to false for the tooltip window onHover

### DIFF
--- a/src/components/organisms/AssetActions/Pool/Graph.tsx
+++ b/src/components/organisms/AssetActions/Pool/Graph.tsx
@@ -77,6 +77,10 @@ function getOptions(locale: string, isDarkMode: boolean): ChartOptions {
     legend: {
       display: false
     },
+    hover: {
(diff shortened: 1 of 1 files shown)


kremalicious (approved):
amazing surgical change solving everything :-)

PR #501, opened 2021-04-12
fetch all data from graph

### LINKED ISSUE #500: Incorrect data in pool tab
Related to https://github.com/oceanprotocol/aquarius/issues/419 . In the market we are using the subgraph but we are getting some incorrect values from aquarius as well. Will fetch all needed data from the graph

mihaisc:
Fixes #500

image


kremalicious (approved):
oooh, good find!

PR #504, opened 2021-04-14
Add new custom form field for single value selection

### LINKED ISSUE #393: new custom form field for single value selection
General UX improvement affecting multiple places, so things can be selected with less clicks and available options are easier to grasp. 

All binary options and single value selections for which we use different form element types right now should follow this visual pattern, semantically and for accessibility reasons they should be `input` with type `radio`:

Screen Shot 2021-02-19 at 14 57 58

Current candidates this can replace:

- _Access Type_: Download or Compute
- _Algorithm Privacy_: Public or Private
- _Docker Image_: by thinking this new component more flexible we can then use it for outputting more than 2 options too, which then comes in handy for outputting the Docker images more nicely
- _Appearance_: probably that above custom UI then uses that new generalized component too to unify markup

Has the added benefit that we also get a space to add more help text for each option individually.

Maybe something like `<BoxSelection />`, to keep it generic so it could later also allow multi-selection with `checkbox`.

KatunaNorbert:
Fixes #393 .

Changes proposed in this PR:

- create a new custom field to be used for all binary options and single value selections


kremalicious:
Awesome so far!

Quick UI find, we should have that same space as shown here in chain selection between the 2 boxes:

Screen Shot 2021-04-20 at 10 29 15

Then, not sure how feasible that is, but would be great to have a title & an optional text for each option, both accepting a `ReactElement`, which should allow us to use same component for the chain selection in screenshot too. And for publish form it would allow us to explain download/compute a bit more individually.

src/components/molecules/UserPreferences/Appearance.tsx
```diff
@@ -1,42 +1,52 @@
...
   darkMode
 }: {
   darkMode: DarkMode
 }): ReactElement {
+  const options: BoxSelectionOption[] = [
+    {
+      name: 'Light',
+      checked: !darkMode.value,
+      title: (
+        <Dotdotdot clamp={1} tagName="span">
```
kremalicious, line 20: using `react-dotdotdot` and on all the other paces for every content in this new component seems unnecessary. A better way without any of the performance impact of all those truncate components on one screen would be to tweak the space in the boxes. I did that over on another PR for the Appearance boxes, which might be worth trying in here too https://github.com/oceanprotocol/market/pull/552/files#diff-0ec6d11806abee9f7ab3b6fbc4a8353ef37c8af4c1b5d0a01272b19cb2b699be

kremalicious (approved):
perfect, could not break it, working well in all of the places where we use it

PR #512, opened 2021-04-14
Highest liquidity list based on subgraph

### LINKED ISSUE #496: New `Highest Liquidity` component
We need a new component for Highest Liquidity were we show the top 9 assets, but there will be no All data sets with pool button. The visual component stays the same, we need to construct the data. We can fetch the pools from the graph and then for each pool fetch the ddo and create a list of ddos.

- fetch pools from the graph ordered by liquidity
- fetch ddo for each pool and create a list
- create a dummy `QueryResult` so that we can pass it to `SectionQueryResult`

claudiaHash:
Fixes #496 .

Changes proposed in this PR:

- fetching pools form the Sub-graph
- retrieving `ddo` for each pool
- creating a list of `ddos`
- display the list making use of `QueryResult` interface


jamiehewitt15:
Looks like some of the checks have failed, is it still a work in progress?

claudiaHash:
> Looks like some of the checks have failed, is it still a work in progress?

Sorry, I didn't see the message til now. Just made some changes which I hope will make all the tests pass.

jamiehewitt15:
It seems to work well. The pull request is still marked as a draft, is there anything left to do on this?

claudiaHash:
> It seems to work well. The pull request is still marked as a draft, is there anything left to do on this?

Nope, just made some checks on its functionality. I'll set it as ready for review right now.

src/components/pages/Home.tsx
```diff
@@ -47,36 +52,75 @@ function LoaderArea() {
...
-        query,
-        config.metadataCacheUri,
-        source.token
-      )
-      setResult(result)
-      setLoading(false)
+      if (queryType && queryType === 'graph') {
+        const ddoList: DDO[] = []
+        setLoading(true)
+        try {
```
kremalicious, line 76: this can be made into one `try/catch` where, if all the if/else can't be refactored, the if/else clauses happen within the `try` block

Then in all errors, we should log like so instead of plain `console.log`:

```js
Logger.error(error.message)
```

kremalicious:
For references, screenshots of the different assets returned:

this PR:

Screen Shot 2021-04-20 at 10 57 12

Live:
Screen Shot 2021-04-20 at 10 57 35

mihaisc:
The subgraph query should be like:
```
pools(orderBy: valueLocked, orderDirection: desc, first: 20) {
      id
      consumePrice
      spotPrice
      tx
      symbol
      name
      datatokenAddress
  valueLocked
    }
```
Sort by valueLocked not price

claudiaHash:
> The subgraph query should be like:
> 
> ```
> pools(orderBy: valueLocked, orderDirection: desc, first: 20) {
>       id
>       consumePrice
>       spotPrice
>       tx
>       symbol
>       name
>       datatokenAddress
>   valueLocked
>     }
> ```
> 
> Sort by valueLocked not price

Oh, ok, there is the problem. I sorted by consumePrice. Thanks!

jamiehewitt15:
I think there is a UI issue that may have caused the confusion here. On cards on the Market home page, where it says "62,861.072 OCEAN pool" - it looks like "62,861.072" is the number of tokens in the pool. With this interpretation it appears like the way that @claudiaHash did it initially is correct - and it would make sense that Data Union is not there because it looks like their pool is just 488.88.

I feel like it should be made clearer that 488.88 is the price and not the amount in the pool. If you agree, I'll create a separate issue for that.

claudiaHash:
I actually think what Mihai suggested is right, because now apart two assets ( I suspect there is a problem with the Subgraph in this case - the consumePrice: "-1" we discussed ), all the assets coincide with those on Live.

jamiehewitt15:
Yes, @mihaisc solution is correct. I was suggesting that there is also a separate issue with the UI that can lead to confusion here. I will open a separate issue to discuss it

jamiehewitt15:
Related issue that I was describing: https://github.com/oceanprotocol/market/issues/541

mihaisc:
You also need to filter for purgatory 
@jamiehewitt15 valid point, not related to this pr, will comment in the issue that you created.

image
https://market-aa4a0k1v4-oceanprotocol.vercel.app/
image

mihaisc:
Looks good

mihaisc:
Please rebase

mihaisc (approved):

kremalicious:
just double checking the new order, so it looks like the new order with `valueLocked` is the actual correct order, and I guess Aquarius was wrong all along?

live:
Screen Shot 2021-05-17 at 15 18 39

this PR:
Screen Shot 2021-05-17 at 15 19 22

kremalicious (approved):

mihaisc:
@kremalicious that's what it seems to be the case. I compared a few assets and aqua was wrong.

kremalicious:
@claudiaHash if you have another improvement as mentioned in standup, then feel free to just add here if it its not a bigger change

claudiaHash:
> @claudiaHash if you have another improvement as mentioned in standup, then feel free to just add here if it its not a bigger change

great, I'll push it in a moment :)

src/utils/subgraph.ts
```diff
@@ -298,3 +315,29 @@ export async function getAssetsBestPrices(
...
+  const ddoList: DDO[] = []
+  const fetchedPools = await fetchData(HighestLiquidityAssets, null)
+
+  if (fetchedPools.data?.pools?.length === 0) return null
+
+  for (let i = 0; i < fetchedPools.data.pools.length; i++) {
+    const did = web3.utils
+      .toChecksumAddress(fetchedPools.data.pools[i].datatokenAddress)
+      .replace('0x', 'did:op:')
+    const ddo = await retrieveDDO(did, metadataCacheUri, sourceToken)
```
kremalicious, line 332: only catching this now cause was wondering about the noticeable delay in loading. We should make this more effective by sending a list of DIDs to Aquarius in one query instead of 20 requests as it's doing right now. So the loop then is only there to construct the DID array for the Aquarius search query.

kremalicious (changes requested):

src/components/pages/Home.tsx
```diff
@@ -38,34 +29,57 @@ const queryLatest = {
...
-      const result = await queryMetadata(
-        query,
-        config.metadataCacheUri,
-        source.token
-      )
-      setResult(result)
+      setLoading(true)
+      try {
+        if (queryType && queryType === 'graph') {
+          const ddoList: DDO[] = await getHighestLiquidityAssets(
```
kremalicious, line 52: so in the end as requested we'll simply do a `queryMetadata`, only with a modified `query` passed to it. This would imply we do not need this if/else logic here at all, and rather have a helper method to construct the query and then pass it to `< SectionQueryResult  />` component further down. Then also no need for any new props like `queryType`

src/components/pages/Home.tsx
```diff
@@ -88,18 +101,9 @@ export default function HomePage(): ReactElement {
...
-        action={
-          <Button
-            style="text"
-            to="/search?priceType=pool&sort=liquidity&sortOrder=desc"
-          >
-            Data sets and algorithms with pool →
-          </Button>
-        }
-      />
+      {searchDids && (
```
kremalicious, line 104: the section needs to render at all times from the beginning to prevent bigger layout shifts and set user expectations by having the section title present at all times. If the component breaks because some data is missing it should be solved in the component, so as to have it use its loading state longer

src/utils/subgraph.ts
```diff
@@ -298,3 +316,26 @@ export async function getAssetsBestPrices(
 
   return assetsWithPrice
 }
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
```
kremalicious, line 320: if TypeScript complains about missing return type definition then it is right. So instead of disabling the rule, we should actually follow it. From what I'm seeing this method returns `Promise<string>` so the next line should be:

```ts
export async function getHighestLiquidityDIDs(): Promise<string>
```

src/utils/subgraph.ts
```diff
@@ -11,6 +11,10 @@ import {
 } from '../@types/apollo/AssetsFrePrice'
 import { AssetPreviousOrder } from '../@types/apollo/AssetPreviousOrder'
 import BigNumber from 'bignumber.js'
+import web3 from 'web3'
+import { CancelToken } from 'axios'
```
kremalicious, line 15: it looks like most of the added imports here are unused in this file

kremalicious:
And then not sure if it is the latest commits since my last comment, but when I check the assets now we again have a different order compared to what we had last time:

Screen Shot 2021-05-19 at 10 55 37

And in general: network switching does not work:

- without a wallet and switching networks, I get "No assets found" in this section for all networks other than ETH main
- with a wallet connected, section switches nothing when I switch networks

src/utils/subgraph.ts
claudiaHash: Resolved this and the one mentioned before. My concern related to this implementation is that the assets displayed aren't the correct ones. 

On live we got:
Screenshot from 2021-05-19 12-08-04

But on this PR we got:
Screenshot from 2021-05-19 12-07-12

I'm on this right now

kremalicious: mmmh, as posted in my last comment further down, I'm on the preview link and get what's in my screenshot, so that screenshot here shows different ones again. Are your latest changes pushed?

kremalicious: pretty sure we do not render the order we want because we send a search query to aquarius and its results will be ordered by its default sort order, unless we do something. E.g. we need to order it based on the order we have in the DID list somehow

claudiaHash: Saw this with some delay, sorry. The screenshots were made before changes were made. I pushed them seconds ago.

claudiaHash: > pretty sure we do not render the order we want because we send a search query to aquarius and its results will be ordered by its default sort order, unless we do something. E.g. we need to order it based on the order we have in the DID list somehow

yes, and what I missed is that I don't exclude the assets in purgatory in the aquarius query

src/components/pages/Home.tsx
```diff
@@ -77,6 +92,26 @@ function SectionQueryResult({
...
 
 export default function HomePage(): ReactElement {
+  const [searchDids, setSearchDIDs] = useState<string>()
+  const { config } = useOcean()
+
+  useEffect(() => {
+    getHighestLiquidityDIDs().then((results) => {
+      setSearchDIDs(results)
+    })
+  }, [config?.metadataCacheUri, searchDids])
```
kremalicious, line 102: since we query the subgraph and not aquarius in this effect we should switch dependency to `config?.subgraphUri`

src/components/pages/Home.tsx
```diff
@@ -46,26 +44,43 @@ function SectionQueryResult({
 }) {
   const { config } = useOcean()
   const [result, setResult] = useState<QueryResult>()
+  const [dids, setDIDs] = useState<string>()
 
   useEffect(() => {
     if (!config?.metadataCacheUri) return
-
     const source = axios.CancelToken.source()
+    getHighestLiquidityDIDs().then((results) => {
```
kremalicious, line 52: you probably tried to find solution for network switching and added refetch for pools here again? Cause they are already fetched and should also refetch down in the `HomePage` component

src/components/pages/Home.tsx
```diff
@@ -46,26 +44,43 @@ function SectionQueryResult({
...
+        Logger.log(error.message)
+      }
     }
     init()
 
     return () => {
       source.cancel()
     }
-  }, [config?.metadataCacheUri, query])
+  }, [config?.metadataCacheUri, query, dids])
```
kremalicious, line 83: dependencies here look ok so I would assume `query` does not get updated upon network switch. As a test could try moving the whole query into state instead of just the DIDs https://github.com/oceanprotocol/market/pull/512/files#diff-1df2f2876c20c90c294da5abb9fcf0812eec3448de2aea5eb20e5e5f4a7d6cc7R104-R113

This might solve some update issues upon network change

claudiaHash:
I made the changes mentioned, there is still a problem on loading. Having `[config.subgraphUri, searchDIDs, queryHighestAssets]` in the dependency array for HomePage component solves the UI problems, but totally overloads the network. Still try to figure out a solution for this.

mihaisc:
Not working for me 
image

claudiaHash:
> Not working for me
> image

I'll come in a minute with the fix for this

claudiaHash:
There's not a problem with passing data through components, but with network config. Talking to @KatunaNorbert we have reached this result where data is correctly loaded with or without a wallet connected in every network but `Polygon` and `Moonbase`. Still working on it.

KatunaNorbert:
Now it's working when connected to wallet. Still need to be fixed for changing the network when not connected to wallet

KatunaNorbert:
Now also working without wallet

claudiaHash:
The disparity between the correct order https://user-images.githubusercontent.com/90316/118495212-44be6980-b723-11eb-905f-198f3bc3abc1.png and what we have on this PR on `Mainnet`, is given by the fact that _Ocean Market 🌊 Pool trends and charts 📈_ - did:op:119F22fa21D4ee54c9911E1Bf876Fcf24A0ADFC1  is exceeded in value by _eBay DATASET - 10 Million Data Points (1,000,000 Product Listings)_  - did:op:862923EE25fFcEA0bF73273703808D1C00295cC2, as the subgraph data suggests:
imageedit_2_7209245529

bogdanfazakas (approved):
All seems to be good when testing now the PR, went also through the changes PR looks good👍

src/components/pages/Home.tsx
```diff
@@ -36,48 +27,99 @@ const queryLatest = {
...
+
+  useEffect(() => {
+    if (loading || (web3Loading && web3Provider)) return
+    getHighestLiquidityDIDs().then((results) => {
+      const queryHighest = {
+        page: 1,
+        offset: 15,
+        query: {
+          query_string: {
+            query: `(${results}) AND -isInPurgatory:true AND price.isConsumable:true`,
```
kremalicious, line 114: @claudiaHash I missed this here, I think we need to remove `price.isConsumable:true` in new PR cause this is not `price.isConsumable` from our app `price` but the one in Aquarius DDO. And that will be removed from there.

claudiaHash: Right, I'll remove it, test and then push the changes.

PR #514, opened 2021-04-14
Fix/order on polygon

mihaisc:
Ignore checking for previous orders in ocean lib when ordering an asset, we already do that in the market.


bogdanfazakas (approved):
LGTM

PR #515, opened 2021-04-14
filter ordered assets by service type

### LINKED ISSUE #509: Assets ordered for starting compute jobs show up under Downloads in History
Not sure what the best solution here is but it feels confusing because from user perspective they did not download anything, here the last 2 assets:

Screen Shot 2021-04-14 at 14 36 12

One solution could be to simply hide non-download orders in this list.

claudiaHash:
Fixes #509 .

Changes proposed in this PR:

- filter ordered assets by ddo service type

### DIFF
--- a/src/components/pages/History/Downloads.tsx
+++ b/src/components/pages/History/Downloads.tsx
@@ -2,14 +2,13 @@ import React, { ReactElement, useEffect, useState } from 'react'
 import Table from '../../atoms/Table'
 import { gql, useQuery } from '@apollo/client'
 import Time from '../../atoms/Time'
-import { OrdersData_tokenOrders as OrdersDataTokenOrders } from '../../../@types/apollo/OrdersData'
 import web3 from 'web3'
 import AssetTitle from '../../molecules/AssetListTitle'
 import { useWeb3 } from '../../../providers/Web3'
 import axios from 'axios'
 import { useOcean } from '../../../providers/Ocean'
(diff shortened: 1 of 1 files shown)


claudiaHash:
I think it's working as Mihai proposed, I'm just thinking now about a solution so we don't fetch the did twice.

mihaisc:
we can still see algos, but i think it's ok because you did sort of download them and you have access to download if you want again. So all good 
image

mihaisc:
> I think it's working as Mihai proposed, I'm just thinking now about a solution so we don't fetch the did twice.

Check how i did in compute history   (in the feature/compute branch) . I fetch all the did from the orders and then filter them. and in the row you pass directly the name and did

claudiaHash:
cool, thanks!

mihaisc:
Also please add a loading spinner , like other tabs

claudiaHash:
sure, on it

kremalicious:
the loader for me just comes up briefly, then the empty screen kicks in, and then the list comes in. We need to make sure the loader stays for as long as the list is compiled

claudiaHash:
checking it right now

kremalicious (approved):
perfect, all good against `main`. Not sure yet if we have to change something over in `feature/compute`, we'll see after merge

claudiaHash:
just checked it over in `feature/compute` and seems to work correctly

PR #517, opened 2021-04-15
Match dataset providers with algorithms providers

### LINKED ISSUE #510: matching providers for dataset + algo
As of right now we have the restriction that algorithms can only run on data sets when both of them were published with the same provider.

One way could be in our `AssetSelection` we only show algos from the same provider, which fits all use cases going through our market. Which would apply to the edit compute settings screen, and for starting compute job.

Another more time consuming way would be to come up with visual patterns and enough info on an asset row within `AssetSelection` to signal that this algo has another provider.

In both cases, sounds like we need to always check the providers before rendering `AssetSelection` list

bogdanfazakas:
Fixes #510  .

Changes proposed in this PR:

- check ddo's provider against algorithms with compute service providers in edit compute
- check ddo's provider with  algorithms with compute service providers in start compute tab asset selection list


mihaisc:
Don't merge this, i want to test later this day with multiple providers and see where the problem is exactly. This is not a blocker for c2d. You can just approve proper algos in the compute asset and all will work.

src/utils/aquarius.ts
```diff
@@ -115,14 +116,22 @@ export async function transformDDOToAssetSelection(
...
   }
   const ddoNames = await getAssetsNames(didList, metadataCacheUri, source.token)
   const algorithmList: AssetSelectionAsset[] = []
   didList?.forEach((did: string) => {
-    if (priceList[did]) {
+    if (
+      priceList[did] &&
+      (!didProviderEndpointMap[did] ||
+        didProviderEndpointMap[did] === datasetProviderEndpoint)
+    ) {
```
KatunaNorbert, line 134: I think we ca do this check before pushing the ddo id to `didList` so than we won't need the `didProviderEndpointMap`

kremalicious:
I always forget why exactly this is needed and we need to restrict it cause ocean.js has a test specifically for this

> should start a compute job with a dataset on provider1 and published algo on provider2
> https://github.com/oceanprotocol/ocean.js/blob/main/test/integration/ComputeFlow.test.ts#L1228

So based on that test this combination should actually work implying we should not restrict dataset & algos from same provider only

bogdanfazakas:
> I always forget why exactly this is needed and we need to restrict it cause ocean.js has a test specifically for this
> 
> > should start a compute job with a dataset on provider1 and published algo on provider2
> > https://github.com/oceanprotocol/ocean.js/blob/main/test/integration/ComputeFlow.test.ts#L1228
> 
> So based on that test this combination should actually work implying we should not restrict dataset & algos from same provider only

Yep so the restriction should apply only if both assets `dataset` and `algorithm` are of type `compute`, in this case the oceanjs method that we are using `isOrderable` is going to return a false value when trying to start the job, so at that time we decided to not display those kind of combinations to the users, neither on edit compute settings or on on start job tab
But you are right you can run jobs with assets on different providers if the algorithm is not compute type.

KatunaNorbert:
Is this still valide with the changes from #702 coming in?

mihaisc (approved):

PR #520, opened 2021-04-15
 fix package-lock

mihaisc:


kremalicious:
whoops, also ran into that locally :-) But think build fix in the end was missing package https://github.com/oceanprotocol/market/pull/376/commits/79497adbabf7ddec47701a244c685c1703bd1a78

but this change is good too

kremalicious (approved):

PR #522, opened 2021-04-15
fix sync status randomly fails

### LINKED ISSUE #508: Sync status randomly fails
This component fails randomly, can't replicate consistently. We need to investigate further
image

KatunaNorbert:
Fixes #508 .

Changes proposed in this PR:

- Fix Graph sync status randomly fails


KatunaNorbert:
I think I fixed it. Was a sync issue, that's why appeared so randomly. It was always fetching the block number from ETH main because the web3 and config was not mounted yet. And the response came back to late and overwrite the block head.
Added a default loading value for web3Loading so it is not fetching until that is initialized.

kremalicious (approved):
awesome investigation! Could not bring up the banner anymore in various combinations. Can't test the error state for real, but don't see why it wouldn't work after this change

PR #524, opened 2021-04-15
Update the file info UI after URL re-add

### LINKED ISSUE #490: Can't re-enter same URL after removing it (Publish)
Ocean Market doesn't allow you to enter the same URL twice in a row. See screen recording:

https://www.loom.com/share/334e6a36080843dfbf25709f63608a40

This means that if a user entered a URL and then removed it by mistake they wouldn't be able to add it again. That could be a frustrating user experience.

DimitarSD:
Fixes #490 

Changes proposed in this PR:

- check for re-adding the same URL


jamiehewitt15:
Looks like the tests are failing, is this still a work in progress?

DimitarSD:
Yeah, I just saw the test. I'm working on it.

jamiehewitt15:
Ok cool

.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
```
jamiehewitt15, line 1: What was the purpose of this? Was it meant to be included in this pull request?

jamiehewitt15 (changes requested):
It seems to work alright but you need to take a pull from origin main, as there are some changes to the publish page on main that aren't visible here. I'll check once more after you've done the pull

.vscode/launch.json
DimitarSD: Actually, it was not. It seems I didn't exclude it from the files. I used it for debugging. I removed it now.

jamiehewitt15: ok cool

jamiehewitt15:
Did you do a pull from origin main? There are some differences between the live version and you branch, for example the button on the main version is pink and the error messages and submission work differently: https://market.oceanprotocol.com/publish

jamiehewitt15:
It seems to work ok but I'm not sure why there are so many additional changes in other files now?

DimitarSD:
The changes are from `git rebase`, i.e. `fix/490` branch is up to date with the changes from the `main` branch.

jamiehewitt15 (approved):

jamiehewitt15:
Thanks Dimo, I've approved

DimitarSD:
I found some strange behavior with toast when we re-add the same URL. It needs some more testing before merging.

jamiehewitt15:
I've saw the issue with the toast messages appearing twice. Both the `onClick()` and `onBlur()` events were firing at the same time, so I removed the `onclick()` event (it's not needed if we have `onBlur`). I added `e.preventDefault()` as the `onClick()` event because otherwise the form will automatically attempt to submit

PR #530, opened 2021-04-16
fix compute history duplication

mihaisc:
Check the provider compute address to avoid duplication


mihaisc:
Not sure why this is happening. Deleted `node_modules` locally and did a `npm ci` and `npm run build`, all good.

mihaisc:
All good now

kremalicious:
alright! Way less items now for me in the list. Still get this console error but looks like we can ignore that?

Screen Shot 2021-04-16 at 14 20 53

kremalicious (approved):

mihaisc:
The error is ok. The provider doesn't exist anymore.

PR #531, opened 2021-04-16
Add datatoken to algo publish form

### LINKED ISSUE #525: Datatoken field is missing in algorithm publish
I know, we forgot to spec it but after testing a bit feels like a bug that this is missing since algos are tokenized as datatoken too. So add the same field we use for dataset publish also on algo publish

claudiaHash:
Fixes #525  .

Changes proposed in this PR:

- datatoken name and symbol added to algorithm publish form


bogdanfazakas:
I think the datatoken input should have an initial value and not require user to click refresh in order to set the values for it. We have similar flow for publish dataset also.
Screenshot 2021-04-16 at 14 29 13

claudiaHash:
sure, missed it. I'll fix it now

kremalicious:
fyi, empty datatoken info is actually our default behavior when no wallet is connected because ocean.js helper method can only be accessed once wallet is initiated, hence without a wallet that field stays empty. But as Bogdan said, when wallet is connected the field should be auto-populated with a first helper method call

bogdanfazakas:
Check also the use case @kremalicious presented, use the market without a wallet connected and go to publish algoritm and then you see the app will crash 
Screenshot 2021-04-20 at 09 58 21
For this use case the datatoken field must be empty as it is for the publish dataset form

kremalicious:
the underlying issue is that the datatoken name & symbol is created upon tab switching as intended, but the `initialValues` for the parent `Formik` form overwrite this during tab switch, emptying the field.

I removed the hacks in `FormAlgoPublish`, for me nothing crashes with it without a wallet and behaves like data set publish form.

We now somehow need to figure out how to prevent this `initialValues` overwrite for datatoken options when localStorage is also empty, or make the tab switch behave exactly like a first mount as this works on the dataset publish form

kremalicious:
more investigation: it's `enableReinitialize` behavior on `Formik` form in `Publish/index.tsx`. Without it the datatoken field behaves as expected, tab switch will make datatoken show up.

Screen Shot 2021-04-22 at 15 41 03

Problem is that this then conflicts with switching between localStorage values of dataset publish & algo publish form.

kremalicious (approved):
all good, could not break it anymore

PR #533, opened 2021-04-16
Fix/ Wrong Ocean amount to spend displayed for compute assets

### LINKED ISSUE #527: Yet another price calculation/help message combination
Lost track of all the combinations but I think this screenshot means we forgot the case where user has datatoken of the data set, as the help message says:

Screen Shot 2021-04-15 at 17 21 11

So in this case the dataset price should not be included in the displayed calculation, in these conditions shown in screenshot I will only pay 1 OCEAN for ordering the algo in the end

bogdanfazakas:
Fixes #527 .

Changes proposed in this PR:

- check also if the user already has datatokens for each asset when calculating total amount to spend and displaying ocean to spend for each asset in compute


kremalicious:
perfect, confirmed initial reported use case is fixed by this

Screen Shot 2021-04-16 at 14 24 05

kremalicious (approved):

PR #537, opened 2021-04-19
Fix buy asset button 

### LINKED ISSUE #529: Probable issue with previous order
From discord 
```
Below is current behaviour.
Let's say I have > 2.0 of the same datatoken in my wallet.
1) I buy the dataset and the UI button changes to Download. I download the dataset.

2) I go off and then navigate back to the same page. The UI button links to BUY again.

    But, perhaps I just want to re-download the dataset (for example, maybe I misplaced it).

    * If I click on BUY then reject the transaction, the UI button changes to Download again...
```

Did not investigate further but this seems to be an issue with fetching the previous order.

KatunaNorbert:
Fixes #529 .

Changes proposed in this PR:

- Fix buy asset button turns into download on unsuccessful buy


kremalicious:
great, confirmed the bug behavior is gone but there are now lots of new error messages. 

Like, when I reject first tx, I get all those errors (there should be only 1 in this case):
Screen Shot 2021-04-20 at 10 16 44

If I then click button again, I get even more errors shown:
Screen Shot 2021-04-20 at 10 17 22

Screen Shot 2021-04-20 at 10 17 39

Not sure what's going on here, but think we simply return too many errors. I see in the `consume()` method that we now return error message strings and at same time set a UI error. Maybe this leads to all those double errors?

KatunaNorbert:
Yeah, I saw that but I don't think is related to my changes. I simply return a string when calling the `consume` method. I wasn't sure if I should take care of that in this issue.
The problem seems to be from that `useEffect` where the errors are shown, having both pricingError and consumeError inside the dependency array, whenever one of then changes both are displayed

KatunaNorbert:
Now only one message is displayed. There is still an issue when declining the 3-rd confirm, button turns to `download` and no `error` but on refresh the button is `buy` because no previous order

mihaisc:
Please debug ocean.js as well. From what i can see the error is berried here https://github.com/oceanprotocol/ocean.js/blob/69a01c75dfc468b951b7af5ef17cbd4d04d1c32c/src/datatokens/Datatokens.ts#L451

KatunaNorbert:
The issue seems to be both on the market and ocean.js side. On the market I added some changes to verify the orderId, but the `order()` method doesn't return an error message when something fails, it returns null. It should return an error message in order to know what to display inside the market.

KatunaNorbert:
Blocked by https://github.com/oceanprotocol/ocean.js/issues/762

bogdanfazakas:
To properly test this I'll recomand using this version of market with a local build version of oceanJs from this PR https://github.com/oceanprotocol/ocean.js/pull/803 linked as a dependency.

KatunaNorbert:
Just tested it and looks good! Now we get the error message from `order` method

mihaisc:
Is this still blocked?

bogdanfazakas:
> Is this still blocked?

Yes, the corresponding PR (https://github.com/oceanprotocol/ocean.js/pull/803) in oceanJs was not merged yet. We need to wait to have a new oceanjs version, bump it in this Pr and should be ok

kremalicious:
So from what I can test it seems the original issue where tx rejections caused button to switch to _Download_ is gone so from that perspective this PR is good to go.

What I'm not sure is all that previous order stuff cause after buying ([I used this on Rinkeby](https://market-git-fix-consumebuttonfix-oceanprotocol.vercel.app/asset/did:op:D0d7b9eb435CE2350BFAFa63d3D99860409B699d)) I see correctly the _Download_ button after download:

Screen Shot 2021-06-14 at 13 56 11

but when I reload the page I'm forced to buy again from the start:

Screen Shot 2021-06-14 at 13 56 25

But that is also a behavior on current live so this PR seems to not fix this although I'm not sure it tried or should try.

as a side note, on live I get the file size, with this PR I do not, this is screenshot on live:
Screen Shot 2021-06-14 at 13 56 52

kremalicious (approved):
fixes original issue so would go ahead with this. Still no clue what the tests complain, maybe a fresh merge with `main` solves it

KatunaNorbert:
I fixed the issue, It was due to a wrong timeout validation.

kremalicious:
npm is fun. As this PR does not add/remove new packages maybe we should try to get the `package-lock.json` changes out of this PR. Like, copy/paste the current `main` version into your local version and commit https://raw.githubusercontent.com/oceanprotocol/market/main/package-lock.json

This should remove those changes from this PR

PR #540, opened 2021-04-20
Fetch asset price without relying on ddo's price

### LINKED ISSUE #497: Fetch price without relying on `ddo.price`
We need a helper function to fetch the price object for a did. It needs to check if it's a fixed price exchange or a pool (something may have been done here https://github.com/oceanprotocol/market/pull/488 )
After this we need to use it in the `AssetTeaser` to fetch the price there and update `useAsset`  to use that function

bogdanfazakas:
Fixes #497 .

Changes proposed in this PR:

- get price in asset provider from subgraph without relying on ddo.price
- get algorithm price using the same helper as above and remove code duplication


mihaisc:
This need to be updated with the price logic from [here](https://github.com/oceanprotocol/market/pull/544), the stuff regarding `spotPrice`/`consumePrice` and `isConsumable`

mihaisc (approved):

KatunaNorbert (approved):
Yep, looks good now

PR #543, opened 2021-04-21
Fix/ Terms and condition state

### LINKED ISSUE #538: publish: termsAndConditions stays `false`
1. Activate debug mode in user preferences
2. Go to `/publish`
3. tick the Terms & Conditions checkbox
4. see in debug at the bottom that option stays `false`, which will then also get populated in final DDO

Screen Shot 2021-04-19 at 17 30 57

bogdanfazakas:
Fixes #538 .

Changes proposed in this PR:

- fixed terms and condition state after checking the button
- fixed terms value set in the ddo and in the debug component
- added default check state for the terms checkbox


kremalicious:
quickly tested:

- confirmed that `true` is indeed set now when ticking the box
- it's broken on live so we should change the base of PR to `main`
- the `RESET FORM` button does not reset the terms checkbox. This is another bug on live too but for this PR leads to another bug: 
  1. tick the checkbox, value changes to `true`
  2. then hit `RESET FORM`, see that checkbox does not get de-selected
  3. de-select the checkbox, now the value logic is turned around, selecting checkbox returns `false` as value

bogdanfazakas:
> quickly tested:
> 
> * confirmed that `true` is indeed set now when ticking the box
> * it's broken on live so we should change the base of PR to `main`
> * the `RESET FORM` button does not reset the terms checkbox. This is another bug on live too but for this PR leads to another bug:
>   
>   1. tick the checkbox, value changes to `true`
>   2. then hit `RESET FORM`, see that checkbox does not get de-selected
>   3. de-select the checkbox, now the value logic is turned around, selecting checkbox returns `false` as value

Yeap missed that, I'll check this also and try to fix it

bogdanfazakas:
so changing the base to main gets all the commits from compute branch because the branch was created from compute, going to close this and open new one against main because we are not touching anything related to compute here

PR #544, opened 2021-04-21
fix price value

### LINKED ISSUE #542: Fail to buy data set when pool is has a low amount of dt
Getting the following error: 
image

Will investigate the exact cause.

mihaisc:
Fixes #542 
Example to test on rinkeby : https://market.oceanprotocol.com/asset/did:op:5445724541a15b257013588c1CDB878B5C2A33Bc
Example for isConsumable: https://market.oceanprotocol.com/asset/did:op:820e3Cf91a655cf276C25F0cB551F0E6AaF7B03C

Proper calculation of  `isConsumable`
Proper calculation of creator liquidity based on spot price not consume price.

### DIFF
--- a/src/components/organisms/AssetActions/Pool/index.tsx
+++ b/src/components/organisms/AssetActions/Pool/index.tsx
@@ -44,6 +44,7 @@ const poolLiquidityQuery = gql`
       id
       totalShares
       swapFee
+      spotPrice
       tokens {
         tokenAddress
         balance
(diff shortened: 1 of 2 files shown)


kremalicious (approved):
so many prices. Can confirm fix makes buying work on the mentioned asset. And it looks like switching to `consumePrice` removes the discrepancy between aquarius/graph prices

mihaisc:
Yes, we need to do this in compute branch as well @bogdanfazakas . Not sure yet what the exact difference is, will check out the graph.

bogdanfazakas:
> Yes, we need to do this in compute branch as well @bogdanfazakas . Not sure yet what the exact difference is, will check out the graph.

i'll take a look, so we should use everywhere the consume price for pool price assets

mihaisc:
Yes, only when you create the `price` object

bogdanfazakas (approved):
LGTM

PR #545, opened 2021-04-22
Fix/ Terms and condition state

### LINKED ISSUE #538: publish: termsAndConditions stays `false`

bogdanfazakas:
Fixes #538  .

Changes proposed in this PR:

- fixed terms and condition state after checking the button
- fixed terms value set in the ddo and in the debug component
- added checked property to checkbox input and fixed reset form issue

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16704,7 +16704,7 @@
       }
     },
     "ethereumjs-abi": {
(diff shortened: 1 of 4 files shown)


kremalicious (approved):
all working perfectly now

Issue #335, opened 2021-01-26
Missing search bar in 'All data sets'


mihaisc:
If navigate to 'All data sets' you get the search page without the search bar. We should have an empty search bar, no need to hide it. If you want to search for something you need to go back to the homepage and then search.
![image](https://user-images.githubusercontent.com/26000280/105875832-f582ea80-6006-11eb-8e7e-61e0f0ed5b10.png)
![image](https://user-images.githubusercontent.com/26000280/105875855-fb78cb80-6006-11eb-96ae-a7364b86f6d5.png)

kremalicious:
it was kinda intentional because of user intention. I know in the background we do a search but it's not a search to users, just a list of all assets. Likewise, those users on mobile should not see searchbar as the first thing, but the actual content, hence the searchbar was removed from the All Data Sets view

mihaisc:
Well, looking at amazon or ebay the search bar is on the top on mobile and it makes sense. There is not much value to view all the assets and not be able to search

jamiehewitt15:
I did a very small hack to make the search bar appear when the user clicks on "All Data Sets" https://github.com/oceanprotocol/market/pull/548

kremalicious:
The better solution is a proper place for a search bar which user can invoke from any screen as described in #665, whereas #672 provides the needed space for adding this into the menubar

PR #548, opened 2021-04-22
Making searchbar appear in all assets

jamiehewitt15:
Closes: https://github.com/oceanprotocol/market/issues/335

Changes proposed in this PR:
- Small change to the redirect URL to make sure that the search bar is still shown when the user clicks on "all data assets"

### DIFF
--- a/src/components/molecules/SearchBar.tsx
+++ b/src/components/molecules/SearchBar.tsx
@@ -17,7 +17,7 @@ export default function SearchBar({
   filters?: boolean
   size?: 'small' | 'large'
 }): ReactElement {
-  let [value, setValue] = useState(initialValue || '')
+  const [value, setValue] = useState(initialValue || '')
 
   function handleChange(e: ChangeEvent<HTMLInputElement>) {
     setValue(e.target.value)
@@ -25,7 +25,7 @@ export default function SearchBar({
 
   async function startSearch(e: FormEvent<HTMLButtonElement>) {
     e.preventDefault()
-    if (value === '') value = ' '
+    if (value === '') setValue(' ')
     const urlEncodedValue = encodeURIComponent(value)
     const url = await addExistingParamsToUrl(location, 'text')
     navigate(`${url}&text=${urlEncodedValue}`)
--- a/src/components/molecules/SearchBar.tsx
+++ b/src/components/molecules/SearchBar.tsx
@@ -17,7 +17,7 @@ export default function SearchBar({
   filters?: boolean
   size?: 'small' | 'large'
 }): ReactElement {
-  const [value, setValue] = useState(initialValue || '')
+  let [value, setValue] = useState(initialValue || '')
 
   function handleChange(e: ChangeEvent<HTMLInputElement>) {
     setValue(e.target.value)
@@ -25,7 +25,7 @@ export default function SearchBar({
 
   async function startSearch(e: FormEvent<HTMLButtonElement>) {
     e.preventDefault()
-    if (value === '') return
+    if (value === '') value = ' '
     const urlEncodedValue = encodeURIComponent(value)
     const url = await addExistingParamsToUrl(location, 'text')
     navigate(`${url}&text=${urlEncodedValue}`)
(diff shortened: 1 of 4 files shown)


KatunaNorbert:
Nice. I think you should do the same for the `ALL DATA SETS` button, right now when clicking that the search bar doesn't appear.

jamiehewitt15:
Sure, I'll do the same for the `ALL DATA SETS` button. 

I'm not exactly sure what you mean when you say you can't get all the data when searching, I wasn't able to recreate that. Is it something we should address in a separate issue?

KatunaNorbert:
So you go to the page where you see all data sets and search for something. After search you want to see again all the data sets, but you can't, you only see those returned by search. My intuition was to empty the search bar and hit search again to get all the data sets, but not working when the input is empty.
It only make sense now when we have the search bar on the page where you se all the data sets so I think should be handled here.

jamiehewitt15:
Thanks for the feedback @KatunaNorbert, I've changed the search function so that if you click on it with nothing in the box, it will show all of the datasets.

kremalicious:
personally, I did not consider the missing search bar in the search pages where users arrive via a link a bug. Main reason is that by adding a search bar at all times, we assume all users arriving there want to search. This is especially a problem on viewports where space is critical, e.g. on mobile here 50% of our users are. This is what they are thrown into, on every page of every result:

Screen Shot 2021-04-26 at 13 46 03

So the least we should do is to minimize the space this search field takes over, like using a small version, not wrapping input & 
button on mobile, whatever helps.

The proper solution would be to find a consistent, global place for a search bar, e.g. in the main menubar. Which could be triggered by a search icon, or be expanded on bigger viewports. Studying how the Rarible search works could be an inspiration https://rarible.com/search?query=ocean

jamiehewitt15:
@kremalicious Thanks for feedback, I see your point. One possible (quick) solution here is for the change I made to only be applied to desktop. So the search bar would be visible all of the time for desktop users but nothing would change for mobile users. I implemented that in the last commit to show how it could look. 

Otherwise, if that doesn't seem suitable then we can look at redesigning to top banner to have a mobile optimised search bar in a similar style to Rarible.

jamiehewitt15:
By the way, there was a change I did to the search function that Norbert requested as part of this (allowing users to submit an empty search and get all results). I have moved it into a separate pull request in case we want to merge it: https://github.com/oceanprotocol/market/pull/561

src/components/molecules/SearchBar.tsx
```diff
@@ -17,15 +17,15 @@ export default function SearchBar({
...
+  let [value, setValue] = useState(initialValue || '')
 
   function handleChange(e: ChangeEvent<HTMLInputElement>) {
     setValue(e.target.value)
   }
 
   async function startSearch(e: FormEvent<HTMLButtonElement>) {
     e.preventDefault()
-    if (value === '') return
+    if (value === '') value = ' '
```
kremalicious, line 28: never directly manipulate a React state object, but always use the setter methods to update the state

src/components/pages/Home.tsx
```diff
@@ -109,7 +109,11 @@ export default function HomePage(): ReactElement {
         action={
           <Button
             style="text"
-            to="/search?priceType=pool&sort=liquidity&sortOrder=desc"
+            to={
+              screen.width > 480
```
kremalicious, line 113: this checks for the device's screen width regardless of browser viewport size, but we should always check for the actual available viewport. Which rules out using the Screen API. If you absolutely have to check for viewport size in JS, use [`window.matchMedia`](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia). In general we should try to avoid viewport size detection in JS cause CSS is just better for that, e.g. gives viewport resize handlers for free.

But adapting the search string based on viewport might be not best solution. If the intention was to hide searchbar on smaller viewports then this should be done over in search page component

kremalicious:
let's close this as it's not the right solution for the problem of a somewhat hidden search in some view. The better solution is a proper place for a search bar which user can invoke from any screen as described in #665, whereas #672 provides the needed space for adding this into the menubar

PR #552, opened 2021-04-22
Support for Moonbase Alpha

### LINKED ISSUE #550: Integrating Market with Ocean.js 0.14.6 (Moonbeam)
This issues is for testing the market on the moonbeam testnet.

jamiehewitt15:
Closes #550

Changes proposed in this PR:

 -  Upgrading Ocean.js to v0.14.7
 - lock to @oceanprotocol/contracts v0.5.16, cause v0.5.20 renamed the `artifacts` folder so it's a mess
 - adapt explorer links for another blockscout-based network explorer
 - ocean.js `ConfigHelper().getConfig()` workaround to remove infura key from RPC URL just like with polygon
 - add Moonbase Alpha to chain user preferences (only visible when no wallet connected)


kremalicious:
Made all the UI production-ready for this particular network and also added it to the chain user preference for demo purposes, helps when demo participants just visit the URL:

Screen Shot 2021-04-22 at 21 56 51

Some findings:
- do not have any DEV or moonOCEAN or however they're called so did not test any chain flows, but I can see test assets were successfully published and all seems to be working from presentational side
- small naming thingy: ocean.js config names the `network` as `moonbeamalpha`, which does not exist. It's either _Moonbeam Testnet_ or [_Moonbase Alpha_](https://docs.moonbeam.network/networks/testnet/). So we should rename the config network name over there and then update this PR
- it's a testnet anyway so think it would be fine to roll this out without additional user hints, like help messages on how to get network tokens
- before live rollout, we should remove the chain user preference again to keep it focused on mainnets only

alexcos20:
my fault on that naming thingy.  but this would require a new release of contracts + new release of ocean.js.  should we just for the demo?

mihaisc:
I think it's good enough as it is for the demo.

jamiehewitt15:
Currently, I'm still getting an error message when trying to upload a dataset...

kremalicious (approved):

PR #558, opened 2021-04-26
Refetch compute status

### LINKED ISSUE #521: refetch compute status
When having the _Compute Jobs_ tab of the History open, the status of all jobs which are not finished yet should update without users having to reload the page.

KatunaNorbert:
Fixes #521 .

Changes proposed in this PR:

- refetch compute status


src/components/pages/History/ComputeJobs/index.tsx
```diff
@@ -229,9 +230,22 @@ export default function ComputeJobs(): ReactElement {
...
     }
+    let intervalId = intervalIdRef.current
+    if (data.tokenOrders.length === 0) {
+      intervalId && clearInterval(intervalId)
+      return
+    }
+    if (intervalId) return
     getJobs()
-  }, [ocean, account, data, config?.metadataCacheUri])
+    intervalId = window.setInterval(function () {
```
kremalicious, line 242: React has its own `setInterval` which can be used here so it's just `setInterval` instead of `window. setInterval`. In general almost never a reason to access browser-only `window` object, which can be problematic in a React/Gatsby app. This then might also allow ditching the `useRef` usage but not sure, having hard time figuring out what exactly is going on in that mega-method `getJobs()`.

Guess what would help as a code improvement is to get all logic for the refetching into it's own `useEffect()`. But that would imply moving `getJobs()` out of it, either have it wrapped in `useCallback` or moved into helper method returning values instead of setting state. So that in the end in the `useEffect` we only control when fetching occurs, including the interval fetching. Could be refactor rabbit hole but could be worth it for maintenance

kremalicious:
If I get this right we fetch all jobs all the time again in the interval, right? Just wondering if it would make more sense to query for a specific job (by passing the respective `jobId` to `ocean.compute.status`) because then we might be able to only refetch for those jobs where job is not finished/failed. which in turn would allow us to decrease the interval for it cause for users fast re-fetching is important only for those unfinished jobs

kremalicious:
tinkered around a bit, if we do not want to introduce further usability problems by always refetching and replacing the whole jobs list I think this is the data flow to use where we have to maintain the React state better. For this it might make sense to split-up all jobs fetching and updating, and job status fetching and updating for unfinished jobs.

1. on mount, fetch all jobs, put into `jobs` state
2. interval re-fetch all jobs, if returned jobs are different from `jobs` state, merge with existing state. This will prevent the user from disappearing jobs list and seeing a loading spinner upon every re-fetch
3. For jobs from `jobs` state where status is unfinished, interval re-fetch status for specific job, and merge with `jobs` state

This would then also allow us to handle 1. & 2. in one `useEffect` and the status fetching in another `useEffect`. And no fetching overhead in case all jobs are finished.

KatunaNorbert:
so we end up with the entire list fetching when the user clicks on a button and a fetching based on a 5-10 seconds interval for fetching only the jobs which are not finished(this will be se only if there are jobs which are not finished)?

mihaisc:
Honestly at this point i think just the button would be enough

bogdanfazakas:
Functionality wise all looks good, the refresh button style need some adjustments in my opinion, would be nice to follow other buttons guideline and use text instead of reload icon, and maybe to be positioned in the center of the component.

bogdanfazakas (approved):
Looks good to me 😃

PR #559, opened 2021-04-26
Fix/AssetSelection: Algo symbol & DID & Total price issues when no wallet connected

### LINKED ISSUE #554: AssetSelection: algo symbol & DID disappear for users without wallet
When clicking around in the compute asset action to figure out prices without a wallet there is this weird behavior where:

- DID is not visible
- upon clicking algo, also symbol disappears
- no total calculation happening so users can't check out final price

Screen Shot 2021-04-23 at 15 34 01

Screen Shot 2021-04-23 at 15 34 12

Screen Shot 2021-04-23 at 15 34 16

Tested on `feature/compute`, Polygon, this asset https://market-git-feature-compute-oceanprotocol.vercel.app/asset/did:op:196cc23a2e3D7c3469662d1802ca9f497e36c845

### Expectation

No matter the wallet status:

- DID is always visible
- symbol is always visible
- price calculation should always happen

bogdanfazakas:
Fixes #554  .

Changes proposed in this PR:

- fixed total price calculation for compute assets when using the market with no wallet connected
- fixed Algo symbol & DID disappear in Safari

### DIFF
--- a/src/components/molecules/FormFields/AssetSelection.module.css
+++ b/src/components/molecules/FormFields/AssetSelection.module.css
@@ -107,7 +107,9 @@
 
 .did {
   padding: 0;
-  font-size: var(--font-size-mini);
+  /* font-size: var(--font-size-mini); */
+  /* hack to make DotDotDot clamp work in Safari*/
+  font-size: 0.63rem;
   display: block;
   text-align: left;
   color: var(--color-secondary);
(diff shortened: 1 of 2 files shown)


kremalicious (approved):
all working perfectly now, good find with the Safari bug

Issue #560, opened 2021-04-26
Show all results when an empty search is entered


jamiehewitt15:
**Describe the solution you'd like**
If someone click "search" while the search bar is empty it would be nice if all results are shown.

PR #561, opened 2021-04-26
Issue 560 empty search

jamiehewitt15:
Closes: #560 

Changes proposed in this PR:
- Allowing users to submit an empty search and get all results

### DIFF
--- a/src/components/molecules/SearchBar.tsx
+++ b/src/components/molecules/SearchBar.tsx
@@ -25,7 +25,6 @@ export default function SearchBar({
 
   async function startSearch(e: FormEvent<HTMLButtonElement>) {
     e.preventDefault()
-    if (value === '') return
     const urlEncodedValue = encodeURIComponent(value)
     const url = await addExistingParamsToUrl(location, 'text')
     navigate(`${url}&text=${urlEncodedValue}`)
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "@coingecko/cryptoformat": "^0.4.2",
     "@loadable/component": "^5.14.1",
     "@oceanprotocol/art": "^3.0.0",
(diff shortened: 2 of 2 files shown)


kremalicious:
good idea, kinda feels like we could even trigger that on empty search field. Was thinking of use case where search box is filled with search text, then user empties field. This could already be the trigger to do the search instead of having user click the button?

jamiehewitt15:
Yeah, sounds like a good idea. I can change the trigger to an `onBlur()` event.

jamiehewitt15:
@kremalicious I have implemented it like you suggested, if you delete the content of the search box it will automatically search and show all of the results. This will only happen if the user has already searched for something, it won't happen on the home page for example.

I have included the search box on the "all results" page in this case as I thought it would lead to a bad user experience if the search box suddenly disappeared when the content is deleted. The search box is still absent when clicking "all datasets" from the home page, I suggest that is resolved in this pull request: https://github.com/oceanprotocol/market/pull/548

KatunaNorbert:
Nice, It's working good!

src/components/molecules/SearchBar.tsx
```diff
@@ -17,20 +17,32 @@ export default function SearchBar({
...
+      const url = await addExistingParamsToUrl(location, 'text')
+      navigate(`${url}&text=%20`)
+    }
+  }
+
+  function handleChange(e: ChangeEvent<HTMLInputElement>) {
+    setValue(e.target.value)
+    if (e.target.value === '') {
+      emptySearch()
+    }
```
KatunaNorbert, line 43: a shorter way to write this could be `e.target.value === '' && emptySearch()`

KatunaNorbert (approved):

PR #564, opened 2021-04-27
searching by full token symbol

### LINKED ISSUE #536: Search by Token Name
In order to find data assets as quick and comfortable as possible the full text search could include the token names.

Example:
LUMSTA-42 belongs to https://market.oceanprotocol.com/asset/did:op:a2B8b3aC4207CFCCbDe4Ac7fa40214fd00A2BA71 

image

claudiaHash:
Fixes #536 .

Changes proposed in this PR:

- search assets by full token name


KatunaNorbert:
Searching by token symbol works well. I think we should be more specific when replacing `-` with `AND` , although making sense for token symbol, in other scenarios can result different behaviour than the search we have right now. For example when searching for `1 - 1` worked like `1  or  -  or  1` now searching for `1 and 1`. Also searching for a text starting with `-` doesn't seems to find anything

claudiaHash:
> Searching by token symbol works well. I think we should be more specific when replacing `-` with `AND` , although making sense for token symbol, in other scenarios can result different behaviour than the search we have right now. For example when searching for `1 - 1` worked like `1 or - or 1` now searching for `1 and 1`. Also searching for a text starting with `-` doesn't seems to find anything

good points, didn't check for those cases. on it

kremalicious:
nice and tricky, not sure we can do more than hacks but I see 2 different ways of getting to a solution:

1. continue in our search to somehow detect symbol search, and react to it, either doing dedicated symbol search with it, or hack around it as done in this PR. Super tricky cause we deal with free text form and need to be careful not to destroy other searches
2. tweak the elastic search query so that symbol search returns a match or with multiple matches at least the actual match is in first place. There are [many options](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html) to try for `query_string` queries alone, like `fields`, `default_field`, `fuzziness`, etc. Obviously also requires careful testing to not destroy other seethes, but we might even end up with more precise searches for everything with this route

claudiaHash:
I think the solution lies in the second proposed way. With current modifications we can successfully search  for datatoken name, symbol, address and asset name, in all ways. The only problem with this solution is case sensitivity which I'm currently working on.

KatunaNorbert:
It seems that the case sensitivity only affects the datatoken symbol. So if it doesn't affect other fields making the text uppercase before using it in the quay I think will solve the problem.

mihaisc:
Go for the shortest working path. The search will need major improvements at some point because now is kinda broken, so no need to tackle more things in this pr.

src/components/templates/Search/utils.ts
```diff
@@ -100,12 +100,21 @@ export function getSearchQuery(
...
     query: {
       query_string: {
-        query: `${searchTerm} -isInPurgatory:true`
+        query: `*${searchTerm}* -isInPurgatory:true`,
+        fields: [
+          'dataTokenInfo.name',
+          'dataTokenInfo.symbol',
+          'service.attributes.main.name'
+        ],
+        default_operator: 'AND'
```
KatunaNorbert, line 117: wasn't this an `OR` operator before? And worth checking for other fields like owner or description which was working before and should be added to `fields` array

claudiaHash: Before we didn't have a specified operator, so by default `OR` was used indeed. In our search, I believe we need a match with all terms specified in search field, that's why I choose `AND`, as they explain in `default_operator` part : https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-simple-query-string-query.html. 

Sure, we can add the author and description as well.

claudiaHash: I see your point, but we need a solution which cooperates with strings introduced in the search field as token symbol, asset name, author,  description. There's no specific  separation between them. Thus, for a search token we won't search for token symbol specifically.

bogdanfazakas:
The current implementation seems to somehow break the current search, so for eg when searching for compute on this branch preview we got 199 result and on the live market there are 373 results, both on Rinkeby network. Also the filtering seems to be broken see the print screens for the same search for compute and filter only datasets on this branch even though there are lots of datasets containing compute keyword.
Screenshot 2021-05-12 at 12 23 43
Screenshot 2021-05-12 at 12 23 33

claudiaHash:
Didn't see this before, I'll check in a moment. Thanks!

KatunaNorbert:
Right, only the `text` should be uppercase not the entire `searchTerm`

claudiaHash:
This  implementation of search targets 5 fields (which are significant for user, in my opinion) : `data token name`, `data token symbol`, `asset name`, `asset author` and `asset description`. Thus, the results will slightly differ from what we got on live right now, since on live the search implementation considers every filed in the DDO. For example, for _compute_ keyword, this implementation won't give a match for ` "type": "compute" ` or provider description `"description": "Compute service with 16gb ram for each node." ` for an asset , as it does on live right now.

On the other hand, by simply eliminating the `fields` option in the query, we got the same results as the ones on live with the inconvenient mentioned above.

Of course, on both solutions `searching by token  symbol` works, with or without filters and sorting applied. :)

What do you guys think?

mihaisc:
I did some quick tests and it looks a lot better. We should do some more testing tough to be sure.

KatunaNorbert:
It's working fine. We can look deeper in how the search should ideally work in #619

claudiaHash:
Yes. I think we should consider it as a base and add on it what @jamiehewitt15 mentioned in #619 .

KatunaNorbert (approved):

PR #570, opened 2021-04-28
using BigNumber instead of Number

### LINKED ISSUE #555: Exit pool rounding bug
**Describe the bug**

Looks like when using MAX to remove liquidity, sometimes we have rounding errors:

looking at the contract (https://etherscan.io/address/0x5e9939f6d959ffe9b328243dfadbeed9c46ac197#readContract), you will see that user 0x67e84944b441cb0a3fb0be4f98974d7363308b9d has 13219061360452614720 shares (wei)

but on exit, he tries to spend  13219061360452615000 pool shares  (2nd argument from https://etherscan.io/tx/0x981ae2085790239f0f45b430843ffdf1d2c28b7be1c62cd03dce2c42249ea612 .  which is 0xb7738a450ddebb58 (hex)

Balance:     13219061360452614720 
Requested: 13219061360452615000 

Clearly, looks like a rounding issue

DimitarSD:
Fixes #555 .

Changes proposed in this PR:

- parse `poolTokens` using BigNumber instead of Number

For more info on why we use BigNumber instead of Number check [comment for issue #555](https://github.com/oceanprotocol/market/issues/555#issuecomment-828384938)


alexcos20:
please use Decimal.js, because BigNumber has no decimal support. it rounds everything to int

alexcos20 (changes requested):
please use Decimal.js, because BigNumber has no decimal support. it rounds everything to int

src/components/organisms/AssetActions/Pool/Remove.tsx
```diff
@@ -21,6 +21,7 @@ import UserLiquidity from '../../../atoms/UserLiquidity'
 import InputElement from '../../../atoms/Input/InputElement'
 import { useOcean } from '../../../../providers/Ocean'
 import { useWeb3 } from '../../../../providers/Web3'
+import BigNumber from 'bignumber.js'
```
kremalicious, line 24: not sure if we should prefer either package or if both can do the same but just when searching through code noticing that we use decimal.js in way more places than big number.js. At same time we specifically install decimal.js in package.json, but do not do that for big number.js, which gets installed from web3.js

DimitarSD: I'll test with decimal.js

> ideally also replace the bignumber.js usage in here and in Consume.tsx

What about `subgraph.ts` - we were using `bignumber.js` only there and in `Consume.tsx`

kremalicious: oh right, that's in `main`, I was searching in this branch. So also suggest rebasing or merging `main` into here.

Then looks like ocean.js also uses decimal.js for calculations within `OceanPool.ts` over there, so another reason to use same library over here

mihaisc: While i agree we should stick to one library, we should do that in another pr. We need this to fix some important bugs

mihaisc (approved):

mihaisc:
Do another test before merge please

src/components/organisms/AssetActions/Pool/Remove.tsx
kremalicious: yes, but here we are introducing a new usage so just want to prevent that. Agree this here should be isolated as bug fix and we don't need to change in other places, but should use same library ocean.js is using

src/components/organisms/AssetActions/Pool/Remove.tsx
```diff
@@ -180,7 +180,6 @@ export default function Remove({
 
     const amountPoolShares = (Number(e.target.value) / 100) * Number(poolTokens)
     setAmountPoolShares(`${amountPoolShares}`)
-    calculateAmountOfOceansRemoved(`${amountPoolShares}`)
```
DimitarSD, line 182: We have pretty much the same logic for `handleMaxButton` should we remove `calculateAmountOfOceansRemoved` from there too?

mihaisc: yes

DimitarSD (commented):
Updated `bignumber.js` to `decimal.js`

src/components/organisms/AssetActions/Pool/Remove.tsx
DimitarSD: I agree with Mihai, it'll be better to fix the current issue (including changing `bignumber.js` to `decimal.js` in `Remove.tsx`) and I'll create another PR for switching to `decimal.js` in the other files. Btw I updated the code to use `decimal` + removed part of the things we don't need anymore (`calculateAmountOfOceansRemoved` function).

mihaisc:
Looks ok

kremalicious (approved):

alexcos20 (approved):

PR #571, opened 2021-04-28
remove unnecessary objects

alexcos20:
Remove all provider attributes


bogdanfazakas (commented):
so publishing flow for algorithm succeed but aquarius is not indexing asset without the removed attributes

bogdanfazakas:
so it was not an issue with this PR, publishing compute algorithms works

bogdanfazakas (approved):

Issue #549, opened 2021-04-22
Add datatoken to metamask


mihaisc:
It would be cool to have a button to add the datatoken to metamask, like we have in the wallet
![image](https://user-images.githubusercontent.com/26000280/115729699-4e57ea00-a38e-11eb-924f-b33ce09a57db.png)
![image](https://user-images.githubusercontent.com/26000280/115729770-60d22380-a38e-11eb-8a02-30c22285b80a.png)

PR #574, opened 2021-04-29
Add datatoken to wallet for metamask users

jamiehewitt15:
Closes: https://github.com/oceanprotocol/market/issues/549

Changes proposed in this PR:
- Adding datatoken to wallet for metamask users

---

blocked by waiting for new MetaMask release https://github.com/oceanprotocol/market/pull/574#discussion_r625668640


jamiehewitt15:
I have implemented this for metamask users, however there is a bit of a compromise which can't be avoid. Metamask will only allow token symbols of 6 characters or less so I had to just use the first 6 characters as the token symbol

jamiehewitt15:
Also, something else to consider is - what the image should be for the datatokens? I have just used the Ocean logo at the moment

src/components/atoms/AddDatatoken.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const [providerInfo, setProviderInfo] = useState<IProviderInfo>()
+
+  useEffect(() => {
+    if (!web3Provider) return
+    const providerInfo = getProviderInfo(web3Provider)
+    setProviderInfo(providerInfo)
+  }, [web3Provider])
+
+  return (
+    <div>
```
kremalicious, line 19: is this wrapping `div` required? Cause when adding component another `div` is wrapped around it. So this component could return the button or `null` only. The goal should always be to keep the DOM free of markup which is not used, so in this case all users without MetaMask would still get that wrapping `div`, and then another one further down when component is used

src/utils/web3.ts
```diff
@@ -114,3 +101,34 @@ export function addOceanToWallet(
...
+      symbol: config.oceanTokenSymbol,
+      decimals: 18,
+      image:
+        'https://raw.githubusercontent.com/oceanprotocol/art/main/logo/token.png'
+    }
+  }
+  addTokenToWallet(web3Provider, tokenMetadata)
+}
+
+export function addDatatokenToWallet(ddo: DDO, web3Provider: any): void {
```
kremalicious, line 122: we don't need to pass around the whole DDO when only `ddo.dataTokenInfo` is needed. So could be simply something like this:

```tsx
import { DataTokenInfo } from '@oceanprotocol/lib/dist/node/ddo/interfaces/DataTokenInfo'

function addDatatokenToWallet(token: DataTokenInfo, web3Provider: any) {}
```

But then it looks like `addOceanToWallet()` & `addDatatokenToWallet()` are doing exactly the same thing, only with different values. So might be worth looking into a generic `addToken()` method accepting `address`, `symbol`, `decimals`, `image` as the parameters for it

jamiehewitt15: Yeah good point, I'll refactor that

src/components/atoms/AddDatatoken.tsx
jamiehewitt15: Yeah that makes sense, I'll change this. Thank you for the feedback

jamiehewitt15: This is done, it needed some element to wrap it so I used a fragment

src/utils/web3.ts
jamiehewitt15: this has also been done

kremalicious:
refactored a bit so that now we have `web3Providerinfo` on `useWeb3()` cause we were doing that in multiple places by now

src/components/atoms/AddDatatoken.tsx
```diff
@@ -0,0 +1,30 @@
...
+  address: string
+  symbol: string
+}): ReactElement {
+  const { web3Provider } = useWeb3()
+
+  async function handleAddToken() {
+    await addTokenToWallet(
+      web3Provider,
+      address,
+      // TODO: figure out how to add full symbol
```
kremalicious, line 19: this is actually a problem. There is no restriction on the length of a ERC-20 token symbol, but more than 6 is very uncommon and it looks like MetaMask opted to cap it there. But the `-13` numbers are part of the symbol so they need to be there. Not sure we can solve this apart from asking MetaMask to not be so picky

jamiehewitt15: One other thing we could do (but doesn't help the existing datatokens) is to change our function for creating the symbol on the publish page, so that it's limited to 6 characters

kremalicious: https://github.com/MetaMask/metamask-extension/issues/9243

https://github.com/MetaMask/metamask-extension/pull/5816

jamiehewitt15: Yeah, I got the same error. That's why I implemented it as a substring of 6 characters

jamiehewitt15: I just had a look at the code and it is still set to 12 characters: https://github.com/jamiehewitt15/metamask-extension/blob/bc24d9e696560780c13c94a39393cd8b6bf16147/ui/pages/add-token/add-token.component.js#L206

I'll open an issue on the Metamask GitHub and see if we can get any response

jamiehewitt15: I've created an issue, hopefully we can get some help on this: https://github.com/MetaMask/metamask-extension/issues/10973

kremalicious:
After some options, tending towards this, still trying to figure out if I can get away with having only an icon for the action:

Screen Shot 2021-05-03 at 16 03 34

jamiehewitt15:
> After some options, tending towards this, still trying to figure out if I can get away with having only an icon for the action cause space is precious up there:
> 
> Screen Shot 2021-05-03 at 16 03 34

Yeah, that looks good. I think that's probably the best place. 

If you feel like there is too little space in that top area, then it could maybe go somewhere on the side section, possibly next to the buy/download button:

kremalicious:
and of course while tinkering in browser ended up with another idea, to make the icon actually useful so it is more clear what is actually added and so it can be used standalone. Once added, that action is not so important anymore so that's a good visual compromise to keep the focus on the actual content:

Screen Shot 2021-05-03 at 20 30 32

Untitled

This also unifies the component usage, we are now left with one generic `AddToken` component used in both places shown in screenshot. Not fully happy about the logo URL passthrough (cause it's usually a big asset, and has to be remote for MetaMask, in our case from GitHub) but it keeps it most flexible also for forks.

src/components/atoms/AddDatatoken.tsx
kremalicious: hard to figure out their exact release process but from what I'm seeing the fix they already have in `develop` is not in latest 9.4.0 release https://github.com/MetaMask/metamask-extension/blob/v9.4.0/app/scripts/controllers/preferences.js#L823-L827

So guess we have to wait for a new MetaMask release

mihaisc:
Maybe say "Add datatoken to wallet" instead of "add to wallet" ?

kremalicious (approved):

PR #575, opened 2021-05-03
Issue 549 token length

jamiehewitt15:
Closes: #549

Changes proposed in this PR:
- Adding datatoken to wallet for metamask users


kremalicious:
now also removed the symbol shortening over in #574 so that as soon as there is a new MetaMask release it should start working. So we can close this PR

PR #578, opened 2021-05-04
Fix setting subgraph sync status

kremalicious:
Hopefully last fix for that random behavior of the sync banner when switching networks. Basically we stopped updating `isGraphSynced` whenever web3 or graph was loading and then did not fire the effect again once those two states were updated


kremalicious:
can be nicely tested right now by switching between networks in wallet, cause Rinkeby is out of sync and will show the banner

mihaisc (approved):

PR #579, opened 2021-05-04
Show the list of data sets an algorithm is allowed to run on

### LINKED ISSUE #566: List data sets an algorithm is allowed to run on
When browsing an algorithm on the AssetDetails screen, within the _Use_ tab there should be a list of data sets the algo is allowed to run on.

While we could go through great lengths figuring out how to also start compute from an algo page, it should be enough to simply list and link the datasets so users can start their job over there.

This enhancement can be one of the first improvements after launch, hence not put into any milestone yet

KatunaNorbert:
Fixes #566 .

Changes proposed in this PR:

- Show the list of data sets an algorithm is allowed to run on


src/components/molecules/FormFields/AssetSelection.tsx
```diff
@@ -77,7 +79,9 @@ export default function AssetSelection({
                 <input
                   id={slugify(asset.did)}
                   type={multiple ? 'checkbox' : 'radio'}
-                  className={styleClassesInput}
+                  className={`styleClassesInput ${
+                    hideRadio && styles.hideRadio
```
kremalicious, line 83: this logic can be in `styleClassesInput` like so:

```js
const styleClassesInput = cx({
  input: true,
  [styles.checkbox]: multiple,
  [styles.radio]: !multiple,
  [styles.hideRadio]: hideRadio
})
```

KatunaNorbert:
Can't we fix the aquarius search by id? Saw a comment that said ddo.id is not searchable, so I get all the assets with `publisherTrustedAlgorithms` and filter ones that contains the selected algorithm. It would be nice to let aquarius do that.

kremalicious:
this is great! Makes sense reusing the AssetSelection for this. Only weird UX thing now is the small link opening in new window/tab which was a solution to another problem we are not having in this context. So ideally in this context the whole title should be clickable, and be a route link so app doesn't reload after click. But not sure how feasible that is

src/components/organisms/AssetContent/AlgorithmDatasetsListForCompute.tsx
```diff
@@ -0,0 +1,40 @@
...
+      )
+      setDatasetsForCompute(datasets)
+    }
+    type === 'algorithm' && getDatasetsAllowedForCompute()
+  }, [type])
+
+  return (
+    <div className={styles.datasetsContainer}>
+      <span className={styles.text}>
+        Datasets algorithm is allowed to run on
```
bogdanfazakas, line 35: I think we should have the same html element and styling for this text on both compute and consume. Right now we span element here and a label in compute and have different styling for the text. See print screens 
Screenshot 2021-05-24 at 16 59 59
Screenshot 2021-05-24 at 16 59 48

kremalicious: yup, that title should be also a `h2` or `h3`. Visually, after we then make that top title consistent to compute, and remove that dataset title indent this should be good:

Screen Shot 2021-05-26 at 12 40 25

By visually removing the radios and external arrows I think we make it already sufficiently clear to users this list works differently.  We could enhance affordance by using some hover style, like switch row background to some slighty lighter color or switch dataset title color to `--brand-pink`

As for markup, ideally the whole `row` with that component is the `a` wrapping everything in the end so whole row is actionable, and no `label` or `input` should end up in this list.

kremalicious: not sure how much refactoring we want to do here but what's bugging me a bit is that obviously we also change the meaning of the whole component with it being named `AssetSelection` and create more asset lists. From UI development perspective this sounds like we should abstract out the actual visual elements into some other component and keep it as purely visual component without functionality only switching some markup. That would apply to the whole `div.row`.

And this list here is also no form field so another reason to rather have a new component which shares that visual row component with `AssetSelection`. Just dropping this here to have it in mind cause I'm sure we don't want to over engineer here

KatunaNorbert: I agree, with the changes specified above makes more sens to make a different component because we would basically have different elements, so doesn't make sense to just switch the entire returned element but rather have a different component. I went with using the `AssetSelection` because wasn't thinking that there would be so much differences.
I can create a different component for this then.

kremalicious (approved):
works like charm! Code duplication is justified as it's better to not further touch `AssetSelection` for now. Sorry for all the pushes, thought just quickly solving merge conflicts but I created a mess :-)

PR #586, opened 2021-05-06
Fix edge case calculation errors in remove liquidity screen

### LINKED ISSUE #584: Remove liquidity estimation fails on some pools
On some pools when trying to remove liquidity the estimation of OCEAN user would receive fails:

Screen Shot 2021-05-05 at 21 20 08

When hitting the 100% MAX in this case estimation works again.

I could reproduce this on ETH mainnet for this one so far:
https://market.oceanprotocol.com/asset/did:op:7Bce67697eD2858d0683c631DdE7Af823b7eea38

DimitarSD:
Fixes #584  .

Changes proposed in this PR:

- making sure any number has 18 decimals at all times, since `toPrecision()` alone led to errors in certain edge cases

### DIFF
--- a/src/components/organisms/AssetActions/Pool/Remove.tsx
+++ b/src/components/organisms/AssetActions/Pool/Remove.tsx
@@ -86,6 +86,12 @@ export default function Remove({
   async function handleRemoveLiquidity() {
     setIsLoading(true)
     try {
+      console.log(
+        'remove liquidity with',
+        amountPoolShares,
+        minDatatokenAmount,
+        minOceanAmount
+      )
       const result =
         isAdvanced === true
           ? await ocean.pool.removePoolLiquidity(
--- a/src/components/organisms/AssetActions/Pool/Remove.tsx
+++ b/src/components/organisms/AssetActions/Pool/Remove.tsx
@@ -81,7 +81,7 @@ export default function Remove({
   const [minOceanAmount, setMinOceanAmount] = useState<string>('0')
   const [minDatatokenAmount, setMinDatatokenAmount] = useState<string>('0')
 
-  Decimal.set({ toExpNeg: -18, precision: 16, rounding: 1 })
+  Decimal.set({ toExpNeg: -18, precision: 18, rounding: 1 })
 
   async function handleRemoveLiquidity() {
     setIsLoading(true)
@@ -176,10 +176,7 @@ export default function Remove({
       .dividedBy(100)
       .mul(new Decimal(poolTokens))
       .toString()
-
-    console.log('shares slider', Decimal.precision, amountPoolShares)
-
-    setAmountPoolShares(`${amountPoolShares}`)
+    setAmountPoolShares(`${amountPoolShares.slice(0, 18)}`)
   }
 
   function handleMaxButton(e: ChangeEvent<HTMLInputElement>) {
--- a/src/components/organisms/AssetActions/Pool/Remove.tsx
+++ b/src/components/organisms/AssetActions/Pool/Remove.tsx
@@ -175,9 +175,6 @@ export default function Remove({
       .mul(new Decimal(poolTokens))
       .toFixed(18, Decimal.ROUND_DOWN) // in some cases the returned value contain more than 18 digits which break conversion to wei
 
-      console.log('amountPoolShares = ', amountPoolShares)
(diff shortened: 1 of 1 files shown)


kremalicious:
In my tests it looked like using `toPrecision` in combination with rounding down is the solution https://github.com/oceanprotocol/market/issues/584#issuecomment-833579392

Or any reason why `toFixed` would be better here?

DimitarSD:
Using `toPrecision()` keeps breaking the code when having a low amount of shares. I'm using the exact same code from @kremalicious and I keep getting the same error. So far, I never got an error using `.toFixed` - it's because it counts the 0s

`toPrecision` digit range is.`sd: number: integer, 1 to 1e+9 inclusive` - it doesn't count 0s

Also, it seems that when adding a low amount of liquidity and when trying to remove it we get another conversion error with `toPrecision` - check issue #589

mihaisc:
Actually i think we need http://mikemcl.github.io/decimal.js/#toExpNeg

DimitarSD:
Neither of the two works and I keep getting the same error over and over again. I tested it with `toExpPos` and `toExpNeg`, in the cases with 

- `.toString()`
- `.toFixed(18, Decimal.ROUND_DOWN)` / `.toFixed(18)`
- `.toPrecision(18, Decimal.ROUND_DOWN)` / `.toPrecision(18)`

Best way to test: 
1) Switch the network to Polygon (or Rinkeby)
2) Choose a dataset for which you don't have an added liquidity (or if you have, remove it first). I've been using Maritime Word List on Polygon
3) Add the minimum amount of liquidity to a pool (0.0001 - or a close amount)
4) Try to remove the liquidity from the pool

Using `.toFixed(18, Decimal.ROUND_DOWN)` / `.toFixed(18)` is the only time when the code doesn't fail with one of the two errors (when working with min amount or converting to wei)

mihaisc:
ok, i'll take a look at it

mihaisc:
@DimitarSD please test this, it works for me

DimitarSD:
I'm getting only converting to wei error. Btw I'm testing by adding the min allowed amount for liquidity - `0.00001`

Screenshot 2021-05-13 at 14 31 44

mihaisc:
When are you substracting? Can you show me the code?

DimitarSD:
I didn't change anything in the code. I made a pull and ran the code with your latest changes. The only thing I did was to add 0.00001 liquidity to Maritime Word List on Polygon and then tried to remove it (hence the error)

mihaisc:
Can you replicate this on other networks or only polygon?

DimitarSD:
It's not ignored, `.toPrecision` doesn't count 0s in the beginning after the decimal point. 

If we have `0.004673258821970082` (total 18 digits after the decimal point) when using `.toPrecision(18)` it will count only 4673258821970082 (16 digits) - it will skip the first two 0s. And because we use `.toPrecision(18)` it will add two 0s at the end to make it 18. However, we then get 0.00467325882197008200 (20 digits), and when converted to a string and passed to wei conversion it breaks.

Every time we have 0s after the decimal point, and the total number of digits is close to 18 it will break (most of the time in edge cases and very small amounts)

mihaisc:
Not the best solution, but at the end of the day it's a form of rounding down.

DimitarSD:
When using the slider or MAX button now it works, however (again an edge case) when trying to remove 1% of 0.00001 or even 100% of 0.00001, it fails after the `APPROVE & REMOVE` is clicked. I'm not sure if this is actually coming from `ocean.js`

Screenshot 2021-05-14 at 14 33 12

mihaisc:
Ok, good catch. It was because ocean and dt values. Now it should be ok.

DimitarSD:
@mihaisc I get the same first error again if I try to remove 1% of 0.00001 (I don't have problems if I try to remove 100%). Also I noticed that certain percentages (like 9%) doesn't calculate properly and it basically add nothing:

Screenshot 2021-05-17 at 17 12 22

DimitarSD:
@mihaisc after the last updates I didn't manage to reproduce any of the bugs so far.

mihaisc (approved):

kremalicious (approved):
could not break it anymore in my limited manual test cases so looks like we made it 🥳

PR #587, opened 2021-05-06
hide Web3Feedback for private algorithms

### LINKED ISSUE #585: Hide feedback inside use tab for private algorithms
Inside the asset details page of a private algorithm the 'Insufficient Balance' warning message is displayed when there aren't enough Oceans to buy the algorithm.
Screenshot 2021-05-05 at 18 21 46

You can reproduce it by accessing the following link on Rinkeby network: 
https://market.oceanprotocol.com/asset/did:op:363EEdCC1787F0F792C9B1fa2370b62dA6E3162A

Expected behaviour:
There shouldn't be any other feedback beside the `Alert` component which contains al the needed information for this case

How to fix it:
This can be fixed by removing the `Web3Feedback` component when the algorithm is private.

claudiaHash:
Fixes #585 .

Changes proposed in this PR:

- hide Web3Feedback inside 'Use' tab for private algorithms

### DIFF
--- a/src/components/organisms/AssetActions/Compute/index.tsx
+++ b/src/components/organisms/AssetActions/Compute/index.tsx
@@ -102,7 +102,6 @@ export default function Compute({
   ] = useState<string>()
   const [datasetTimeout, setDatasetTimeout] = useState<string>()
   const [algorithmTimeout, setAlgorithmTimeout] = useState<string>()
-  const [isComputeService, setIsComputeService] = useState<boolean>(false)
 
   /* eslint-disable @typescript-eslint/no-unused-vars */
   const {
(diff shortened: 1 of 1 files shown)


src/components/organisms/AssetActions/Compute/index.tsx
```diff
@@ -491,7 +491,9 @@ export default function Compute({
             action={<SuccessAction />}
           />
         )}
-        <Web3Feedback isBalanceSufficient={isBalanceSufficient} />
+        {ddo.service[1].attributes.main.privacy.allowNetworkAccess && (
```
kremalicious, line 494: not sure this is the right check. Technically an algorithm is private if it has a `compute` service instead of an `access` service

KatunaNorbert: since we are in `compute` component I think just checking if it is an algorithm is enough

claudiaHash: I see, it's clear now, thanks! Now the checks should be correct.

claudiaHash: > since we are in `compute` component I think just checking if it is an algorithm is enough

Think you're right, I removed the redundant check.

KatunaNorbert:
Looks good!

KatunaNorbert (approved):

PR #590, opened 2021-05-10
Show tabs without relying on the ddo price type

### LINKED ISSUE #499: Show tabs without relying on `ddo.price.type`
We should decide what tabs we show in asset actions base on the `price` from `useAsset` . Maybe have some proper loading states in case it takes longer to fetch this info.

KatunaNorbert:
Fixes #499 .

Changes proposed in this PR:

- Use price type from useAsset() to show asset actions tabs


mihaisc (approved):
Looks good

Issue #592, opened 2021-05-11
WalletConnect integration: consecutive transactions never start


EncryptShawn:
When using wallet connect and trust wallet or metamask or pretty much any mobile wallet it seems, when publishing to the marketplace, the first transaction succeeds, the subsequent the transaction fail, the wallet is never prompted to authorize the additional transactions. On mainnet this caused me to lose $100 by only being able to pay the first two transactions and then the entire process is now in limbo despite the initial transactions having been executed.  

This should be easy to reproduce, I was able to recreate it at least 10 times on Rinkeby test net
Browsers used: Chrome, Firefox, Opera
Mobile Wallets used: Trust wallet, Walleth, Metamask

On Test-net I cannot ever get even a second transaction, only the first, then the marketplace sits there spinning doing nothing, indefinitely.

Speaking with the walleconnect devs, they say their app works like UDP and that its not uncommon for a transaction to get lost, it is  incumbent on the dapp to recognize it hasnt gotten a response and then resend the request.  It looks like Ocean app never recognizes the request has failed and never sends a new request, causing the transaction to be in limbo and initial funds to be lost.

*Request: Have Ocean dapp assess after some short timeframe without receiving a response from walletconnect, that the transaction request was lost and then resend it.

kremalicious:
can confirm, was able to reproduce in pretty much all multi-transactions where app is not able to pick up end of first tx, no matter if success or fail. 

Tested with mobile MetaMask on iOS connected to Rinkeby:
- trying to add liquidity to a pool
- approve tx is all successful in MetaMask
- market does not get anything back from `bridge.walletconnect.com`, so can't start next tx

In general, app is wallet-agnostic meaning we do standard web3 calls and don't do special things for particular wallets, relying on [`web3modal`](https://github.com/web3modal/web3modal) for handling wallet connections.

> dapp to recognize it hasnt gotten a response and then resend the request

Not sure I fully understand what "resending the request" means. We will not start another tx if we did not get a response for some time from the wallet and I'm sure most users would not appreciate this either. Cause app does not get a response from `bridge.walletconnect.com` even on successful tx as it looks like so we would just send tx after tx without getting anything back.

This is a regression maybe introduced with some dependency update, we are investigating further...

mihaisc:
Superseded by https://github.com/oceanprotocol/market/issues/1148

PR #594, opened 2021-05-12
Wallet fixes

kremalicious:
* improve logout to make sure walletconnect account is cleared correctly in `localStorage`
* restore network switching for WalletConnect by also listening for `chainId` changes
* fix switching between different connected wallets

As for #592, can't be fixed from market side as it might require splitting up all transactions in methods in ocean.js


mihaisc:
Network switching is not working for me with wallet connect. Using edge + metamask on android.

mihaisc:
When connecting with metamask i get this error. I get this error only the first time i try to connect metamask. It seems that it's not waiting for me to give permissions (click agree in the metamask popup).
image
I need to do a refresh for the wallet to connect.

mihaisc:
Ok, i'll do some local debugging to see what is going on.

mihaisc:
It seems that the issue is neither `handleChainChanged` or `handleNetworkChanged` are being called. `handleAccountsChanged` works as expected.

mihaisc:
Basically this is blocked until fixed in either web3Modal or wallet connect. Need to check both repos for an update from time to time.

PR #595, opened 2021-05-13
Making asset unavailable for purchase if the size is zero

jamiehewitt15:
Closes:

Changes proposed in this PR:
- Checking if the file size is 0
- Making the asset unavailable if the file size is zero

### DIFF
--- a/src/components/organisms/AssetActions/Consume.tsx
+++ b/src/components/organisms/AssetActions/Consume.tsx
@@ -167,7 +167,11 @@ export default function Consume({
         </div>
         <div className={styles.pricewrapper}>
           <Price price={price} conversion />
-          {!isInPurgatory && <PurchaseButton />}
+          {parseInt(file?.contentLength) === 0 ? (
+            <p>This asset is currently unavailable for purchase.</p>
+          ) : (
+            !isInPurgatory && <PurchaseButton />
+          )}
         </div>
       </div>
       <footer className={styles.feedback}>


src/components/organisms/AssetActions/Consume.tsx
```diff
@@ -167,7 +167,11 @@ export default function Consume({
         </div>
         <div className={styles.pricewrapper}>
           <Price price={price} conversion />
-          {!isInPurgatory && <PurchaseButton />}
+          {parseInt(file?.contentLength) === 0 ? (
```
kremalicious, line 170: `file` here refers to the file metadata object within the DDO, which is only set upon publish, retrieved when file is added. So this check here would only kick in if file has been published with `contentLength: 0` which is very unlikely

kremalicious:
Feels like this is generally a better direction for solving #452, which could make #463 obsolete. If we can figure out file availability automatically by hitting the respective `provider` endpoint (instead of basing logic on fixed `file` metadata) then we can remove button as is done in this PR. If we have this there would not be much reasons to have publisher pay gas fees to manually disable/enable the button.

src/components/organisms/AssetActions/Consume.tsx
bogdanfazakas: I think we can use this provide POST method (https://docs.oceanprotocol.com/references/provider/#apiv1servicesfileinfo) to get the file Content-Type and Content-Length using the assed did

kremalicious:
As outlined in #479 we can't rely on `content-length` being correct all the time as it looks like. Considering this, informing user instead of disabling action in #479 seems like a better direction so closing this

PR #596, opened 2021-05-13
Handle asset price in teaser

### LINKED ISSUE #563: Handle asset price in teaser
Stop using `ddo.price` in the asset teasers. Find a better way to fetch the price for multiple assets

KatunaNorbert:
Fixes #563 .

Changes proposed in this PR:

- use the subgraph to get the prices for the AssetTeaser components


mihaisc:
The loading state needs a bit more work. Because it only handles asset loading not price also.  So sometimes you get the "no results found" message for a short period of time.
priceLoad

KatunaNorbert:
Yep, I saw that, I'm thinking what would be the best solution for that. Since the loading component is outside the `AssetList` component we could move the price fetching there and show the loading component based on that. Or keep the price fetching inside the `AssetList` which enables us to reuse some code, but need to manage the loading state from here, maybe moving the loading component inside the `AssetList`

mihaisc:
Moving the loading component might be cleaner/easier, not sure.

KatunaNorbert:
Almost there. Seems to work as before now beside one thing, the loading for the number of results is not synchronised with the datasets display because the loading state of the prices is maintained  internally in `AssetList`. So thinking on a solution for that, maybe moving the loading state inside `AssetList ` might turn out not that feasible.

mihaisc:
I don't really mind that. This is a temporary solution anyway. Other opinions ?

claudiaHash:
Testing this, the assets are displayed as expected. If a quick and easy solution could be found for the assets number displaying would be great. However, I believe that's a minor issue, given that the goal for this task is achieved and it's a temporary solution anyway.

src/utils/subgraph.ts
```diff
@@ -151,53 +220,68 @@ export async function getPrice(asset: DDO): Promise<BestPrice> {
...
+
+  const bestPrice: BestPrice = transformPriceToBestPrice(
+    frePriceResponse.data.fixedRateExchanges,
+    poolPriceResponse.data.pools
+  )
+
+  return bestPrice
+}
+
+export async function getAssetsPrices(
```
bogdanfazakas, line 240: Some refactor on `getAssetPrices` would be nice and get rid of code duplications. Instead of having 2 methods that do the same thing but return different values you can have a base method that returns the graph `pool` and `fre prices`, and other 2 methods  responsible for our data processing, one returning `Promise<PriceList>` and the other one `Promise<AssetListPrices[]>.`
I would keep in the base method the following block https://github.com/oceanprotocol/market/blob/feature/use_asset_price_in_teaser/src/utils/subgraph.ts#L244-L265 return the values and move everything else out in those two new methods

bogdanfazakas:
Functionality wise looks good to me, just that minor issue with number of results that can be solved later in another PR

mihaisc (approved):

bogdanfazakas (approved):
Looks ok to me now 😃

Issue #582, opened 2021-05-05
Integrate Market with RBAC server


jamiehewitt15:
Part of this EPIC: https://github.com/oceanprotocol/multi-repo-issue/issues/76

- The market needs to be adapted so that it checks in the RBAC server if the users role has permission to carry out the specific action.
- The URL of the RBAC server will be contained within the config file. 
- If the URL is empty then all RBAC requests will be ignored by the market. 
- The RBAC server will not be used by Ocean Market but it will be used by many enterprise clients.

jamiehewitt15:
This is blocked until the Daimler team concludes on how they will made users to public keys.

mihaisc:
Is there any need for a UI to change something? Like rights?

jamiehewitt15:
The user roles won't be handled in the market, they will be defined in a separate service (such as Keycloak) which the RBAC server calls. This PR is about restricting access to elements within the market if the user doesn't have the right permission (browse, purchase, publish). 

This is for our enterprise clients and needs to be enabled via `app.config.js`. If the RBAC server isn't defined within the config file then there won't be any changes to the UI based on permissions. In Ocean Market we don't plan to implement any permission based restrictions

mihaisc:
Ok, thanks for clearing that up

PR #597, opened 2021-05-14
Issue 582 market rbac integration

jamiehewitt15:
Closes: #582

Changes proposed in this PR:
- Adding util function for calling the RBAC server 
- Creating new environmental variable for the RBAC server address
- Creating `Permissions` component to wrap elements that should be restricted due to access roles (browse, consume, publish)
- By default access is restricted if a URL for the RBAC server is defined in the environmental variables. 
- If a URL for the RBAC server is not defined then the permissions will default to allow access and this PR won;t change the user experience. For markets (such as Ocean Market) that don't have any role based restrictions, the URL env should be left blank. 
 
## Steps for testing:

1. Clone the RBAC server: 
```
git clone https://github.com/oceanprotocol/RBAC-Server.git
```

2. Create an `.env` file and add thefollowing URLs:
```
KEYCLOAK_URL="https://keycloak-int.data-marketplace.io/auth/realms/marketplace/protocol/openid-connect/userinfo"
KEYCLOAK_ADDRESS_URL="https://keycloak-int.data-marketplace.io/auth/realms/marketplace/custom-api/wallet"
DEFAULT_AUTH_SERVICE = "keycloak"
```

3. Run the RBAC server locally:
```
cd RBAC-Server
npm install
npm run build
npm run start
```

4. Run this branch of the market locally:
```
git checkout issue-582-market-rbac-integration
npm run start
```

5. In order to correct test the permissions you will need a role assigned to your Ethereum address within keycloak. Speak to the Daimler team about having your wallet address added to keycloak

 ###  Testing (without any permission restrictions):
In ocean market and other marketplace forks, there will be no need to have any permission based restrictions. This is made possible in this PR. If a `RBAC_URL` is not defined then the permission response will always be `true` and there will be no change in the user experience. 

To test:
1. Remove or comment out the  `RBAC_URL` from `.env.development`: https://github.com/oceanprotocol/market/blob/22c73020621baa81dd9b522b7a28be06b4fbd08d/.env.development#L4
2. Run the market with `npm run start` and use it as per normal


dungeonhao (approved):

jamiehewitt15:
@kremalicious @mihaisc Can you let me know if you have any feedback on this?

kremalicious:
Like the described focus on having this change not affect majority of users but not sure this is actually done. On first look my main concern would be that the rendering of most UI becomes dependent on wallet connection and successful RBAC server call. That is quite drastic for a feature which is not supposed to be used by majority of our users.

Cause having this as default view for everybody without a wallet is not ok:

Screen Shot 2021-06-04 at 12 18 15

mihaisc:
Or maybe condition this check by an env var. So if `true` do the checks like in the screenshot, if `false` leave the current behavior. Not sure how easy this is.

app.config.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
   // networks in their wallet.
   // Ocean Protocol contracts are deployed for: 'mainnet', 'rinkeby', 'development'
   network: process.env.GATSBY_NETWORK || 'mainnet',
+  rbacUrl: process.env.RBAC_URL || 'false',
```
kremalicious, line 7: 1. if you want to have access to this env var in the browser it needs to be prefixed with `GATSBY_`, [see here](https://www.gatsbyjs.com/docs/how-to/local-development/environment-variables/#client-side-javascript)
2. that `'false'` becomes redundant because if `process.env.RBAC_URL` is undefined it's enough to know that it's not there

jamiehewitt15:
> Like the described focus on having this change not affect majority of users but not sure this is actually done. On first look my main concern would be that the rendering of most UI becomes dependent on wallet connection and successful RBAC server call. That is quite drastic for a feature which is not supposed to be used by majority of our users.


It looks like this because the `RBAC_URL` is defined in `.env.development` (which is just for development & testing): https://github.com/oceanprotocol/market/blob/22c73020621baa81dd9b522b7a28be06b4fbd08d/.env.development#L4

If the `RBAC_URL` file is not defined the permission will always return `true` which means there is no change to the user experience from before, no wallet is required: https://github.com/oceanprotocol/market/blob/22c73020621baa81dd9b522b7a28be06b4fbd08d/src/utils/rbac.ts#L10

src/components/organisms/Permission.tsx
```diff
@@ -0,0 +1,42 @@
...
+  useEffect(() => {
+    const getData = async () => {
+      if (accountId !== undefined) {
+        const data = await rbacRequest(eventType, accountId)
+        updateData(data)
+      }
+    }
+    getData()
+  }, [eventType, accountId])
+  if (data === true) {
```
kremalicious, line 25: we could have another check here to solve the isolation issue. If we check here for undefined `appConfig.rbacUrl` and immediately return the `children` our main use cases should not be affected anymore. Then in the above `useEffect` immediately `return` too when `appConfig.rbacUrl` is not present

src/components/organisms/AssetActions/index.tsx
```diff
@@ -85,5 +86,9 @@ export default function AssetActions(): ReactElement {
       }
     )
 
-  return <Tabs items={tabs} className={styles.actions} />
+  return (
+    <Permission eventType="consume">
```
kremalicious, line 90: not sure this is intended but this will not just disable consume but also the pool and trade tab and price display and compute. If you only want to disable consume, I would suggest rather removing the actual consume button

src/components/organisms/Permission.tsx
```diff
@@ -0,0 +1,42 @@
...
+    return <>{children}</>
+  } else if (data === false) {
+    const message = `Sorry, you don't have permission to  ${eventType}. Please make sure you have connected your registered address.`
+    return <Alert title="Permission denied" text={message} state="error" />
+  } else {
+    return (
+      <>
+        <Alert
+          text="Please make sure your wallet is connected to proceed."
```
kremalicious, line 34: based on above logic this message will be also shown when request or response to/from RBAC server fails. Since we should stop just adding more `returns` would suggest rethinking the rendering logic a bit, e.g. always render the `Alert` as fallback but construct and switch its message based on conditions

src/components/organisms/AssetActions/index.tsx
jamiehewitt15: Yeah, that is intentional. This is following the specification shared by @dungeonhao 

Only users who have consume access should be able to:  
        ◦ Add Liquidity
        ◦ Trade
        ◦ Compute
        ◦ Download

kremalicious: ok, just making sure it's clear you are locking people out of all financial data for a dataset with this, not just the actual actions. Same goes for people who have added liquidity and then were added to some restriction list, preventing them from removing liquidity

mihaisc: This may be ok because this will be used in the enterprise world where there will be stricter conditions/contracts/rules.  But at the end of the day you are just locking the UI, because you can't prevent the user from interacting with the pools at an api level.

app.config.js
jamiehewitt15: 1. That's interesting because it seemed to be working ok in the browser. I have changed it to `GATSBY_RBAC_URL` now though:
https://github.com/oceanprotocol/market/blob/28e1b912c5c0c786d440b5f3b8e543b1a08431f8/app.config.js#L7

2. I've removed the `false` and am checking for undefined now: 
https://github.com/oceanprotocol/market/blob/28e1b912c5c0c786d440b5f3b8e543b1a08431f8/src/utils/rbac.ts#L9

kremalicious:
> If the RBAC_URL file is not defined the permission will always return true which means there is no change to the user experience from before, no wallet is required

I am looking at the Vercel preview as that's what would be deployed to live. And there I get the screen posted above so I guess Vercel picks up `.env` files. Which is why we usually don't commit them and gitignore them as we set env vars via Vercel/Netlify interface. Also no need to complicate things and just use the one `.env` so add to `.env.example`

src/components/organisms/Permission.tsx
jamiehewitt15: Good idea, I have implemented that here: https://github.com/oceanprotocol/market/pull/597/commits/22cacdb57747a700a84a5d3cc855b3fe7a50c1ca

Let me know if that was what you were thinking. As far as I can see that now makes the `if (url === undefined) ` check in rbac.ts redundant

jamiehewitt15:
> I am looking at the Vercel preview as that's what would be deployed to live. And there I get the screen posted above so I guess Vercel picks up `.env` files although it did not pick up the network variable. Which is why we usually don't commit them and gitignore them as we set env vars via Vercel/Netlify interface. Also no need to complicate things and just use the one `.env` so add to `.env.example`

Ok cool, I've removed the `.env.development` and updated `.env.example`

jamiehewitt15:
The latest vercel deployment doesn't require the wallet to be connected now

src/components/organisms/AssetActions/index.tsx
jamiehewitt15: @kremalicious It's a good point, I think the decision around that needs to come from the Daimler team though (@dungeonhao). If they are happy with this level of strictness or not. 

@mihaisc To my understanding we will also be implementing backend integrations with the RBAC server to stop the users interacting with the pools at API level.

mihaisc: that means it has to be done at a contract level and i doubt it, but maybe i'm wrong

src/components/organisms/Permission.tsx
jamiehewitt15: Good idea, I've updated the logic to reduce the return statements and switch the messages like you mention. 

It seems to work alright but the tests are currently failing due to a linting error which I'm not getting locally. Not exactly sure why the linting checks are different as I'm running the same version of ESLint locally: https://github.com/oceanprotocol/market/pull/597/checks?check_run_id=2746525029#step:7:215

jamiehewitt15: I solved the linting errors, it seems that ESLint had been upgraded on main since this branch was started

jamiehewitt15: @kremalicious let me know if you're happy with the refactoring of the logic to reduce the return statements and switch the alert messages. Also, let me know if you have any other feedback

mihaisc:
Did some quick tests with/without wallet, current functionality seems to be ok

jamiehewitt15:
Thanks @mihaisc

mihaisc (approved):

PR #607, opened 2021-05-18
fix assets from mainnet displayed on rinkeby inside the home page

### LINKED ISSUE #602: Assets from ETH mainnet displayed on Rinkeby
**Describe the bug**
I got the assets from ETH mainnet  on the home page when connected to Rinkeby network

**To Reproduce**
1. Go to home page
2. Keep refreshing until you get the assets from ETH mainnet

**Expected behavior**
We shouldn't get the assets from other networks

**Screenshots**
Screenshot 2021-05-17 at 13 11 43
Screenshot 2021-05-17 at 13 12 11

KatunaNorbert:
Fixes #602 .

Changes proposed in this PR:

-Fix the asset retrieving method from home page to get only the assets from the current network config

### DIFF
--- a/src/components/pages/Home.tsx
+++ b/src/components/pages/Home.tsx
@@ -12,6 +12,7 @@ import Button from '../atoms/Button'
 import Bookmarks from '../molecules/Bookmarks'
 import axios from 'axios'
 import { queryMetadata } from '../../utils/aquarius'
+import { useWeb3 } from '../../providers/Web3'
 
 const queryHighest = {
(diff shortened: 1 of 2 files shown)


KatunaNorbert:
Now we get the assets from the Home page only after the network is changed, if we are connected to a wallet

kremalicious (approved):
awesome, good find. Works very well and gives better experience and less calls when app is loaded with user wallet connected to other network than the app's default one

PR #611, opened 2021-05-19
Update to Gatsby v3

kremalicious:
_Highly WIP_

Update dependencies across the board to get them in a state where we can use automated dependency management with dependabot.

This includes lots of things including respective codebase updates for breaking changes:

### Gatsby v3

This one has most impact and requires touching almost every file in this PR because of [this](https://www.gatsbyjs.com/docs/reference/release-notes/migrating-from-v2-to-v3/#css-modules-are-imported-as-es-modules). Also major updates to all gatsby plugins.

All worth it for the way faster builds during local development alone.

- https://www.gatsbyjs.com/blog/gatsby-v3/
- https://www.gatsbyjs.com/docs/reference/release-notes/migrating-from-v2-to-v3

### Wepback v5

With Gatbsy v3 comes a new major webpack version in background, requiring us to specifically polyfill native Node.js module usage as done by ocean.js

### ESLint & Prettier config updates

Now making use of ESLint native parser, removing need for babel-eslint. And some prettier rule updates leading to code formatting updates


mihaisc:
This pr makes me nervous 😄

gatsby-node.js
```diff
@@ -42,9 +42,17 @@ exports.onCreatePage = async ({ page, actions }) => {
 
 exports.onCreateWebpackConfig = ({ actions }) => {
   actions.setWebpackConfig({
-    node: {
-      // 'fs' fix for squid.js
-      fs: 'empty'
+    resolve: {
+      fallback: {
+        // polyfill fixes for ocean.js
```
kremalicious, line 47: this one was interesting to find and affects pretty much all apps using latest webpack v5. I added fallbacks for modules where I assume we actually need them in the browser.

But ideally ocean.js users should not have to do this, and ocean.js should handle its usage of native Node.js modules (which are not available in browser) better, like using or injecting polyfills for release builds or use packages like we do with `cross-fetch` which provide polyfills by default

kremalicious:
closing as too much impact, we are stuck with Gatsby v2 for now

PR #612, opened 2021-05-19
 Too many layout shifts on asset details

### LINKED ISSUE #609: too many layout shifts on asset details
With the removal of getting data for initial rendering from the DDO we need to maintain the layout better and make sure to minimize layout shifts and repaints as much as possible, for better UX and performance.

One of those places where this becomes very noticeable is the asset details page whose initial layout shifts around a lot, isolated in a looping gif:

Untitled2

There are multiple patterns we could use to have layout elements in the right places before we have the data we want to put into this layout, like skeleton UI, longer loading spinner, etc. But in this specific case seems worth looking into:

- why do we get the title so late, leading to one main layout shift?
- why do we get the file info data so late, leading to layout shift because of document icon rendering too late?
- why do we get the description so late, leading to big layout shift within the main content box?
- price initially says "No price set", which is misleading

Everything except price is right there in the DDO so I don't see any reason why this should not be in the initial render. Only price and the tabs coming in late are justifiable

KatunaNorbert:
Fixes #609 .

Changes proposed in this PR:

- Fix too many layout shifts on asset details

### DIFF
--- a/src/components/templates/PageAssetDetails.tsx
+++ b/src/components/templates/PageAssetDetails.tsx
@@ -23,7 +23,7 @@ export default function PageTemplateAssetDetails({
     setPageTitle(isInPurgatory ? '' : title)
   }, [ddo, error, isInPurgatory, title])
 
-  return ddo ? (
+  return ddo && pageTitle ? (
     <>
       <Page title={pageTitle} uri={uri}>
         <Router basepath="/asset">


KatunaNorbert:
So it seems that is not related to the price and waiting for the asset title resolves all our problems

mihaisc:
Interesting, tested it with some network throttling and it seems to be working fine now.

mihaisc (approved):

PR #618, opened 2021-05-20
List of algorithms to select seems capped

### LINKED ISSUE #580: List of algorithms to select seems capped
When editing compute settings of a compute data set, the `AssetSelecttion` list of algorithms seems to be capped at 35 items.

Example on Rinkeby, when searching for all algos with a price like so:
https://market.oceanprotocol.com/search?serviceType=algorithm&sort=created&sortOrder=desc&priceType=all
Screen Shot 2021-05-04 at 21 22 11

So when editing compute settings, 58 algos should show up in AssetSelection yet in my case I count 35. 

If I remember correctly we do not do any `provider` match check yet as is worked on in #510, so there should be no other restrictions in place for algos to show up for selection. Might be that we specifically need to set some pagination parameters when querying Aquarius and we're just hitting some pagination default.

Worth checking if we have some unintended cap in place for `AssetSelection` within the start compute job component too.

KatunaNorbert:
Fixes #580 .

Changes proposed in this PR:

- show all the algorithms inside the select algorithms component

### DIFF
--- a/src/components/organisms/AssetActions/Compute/index.tsx
+++ b/src/components/organisms/AssetActions/Compute/index.tsx
@@ -126,7 +126,7 @@ export default function Compute({
     const algorithmQuery =
       trustedAlgorithmList.length > 0 ? `(${algoQuerry}) AND` : ``
     const query = {
-      offset: 200,
+      offset: 500,
       query: {
         query_string: {
           query: `${algorithmQuery} service.attributes.main.type:algorithm -isInPurgatory:true`
--- a/src/components/organisms/AssetActions/Compute/index.tsx
+++ b/src/components/organisms/AssetActions/Compute/index.tsx
@@ -126,7 +126,7 @@ export default function Compute({
     const algorithmQuery =
       trustedAlgorithmList.length > 0 ? `(${algoQuerry}) AND` : ``
     const query = {
-      page: 1,
+      offset: 200,
       query: {
         query_string: {
           query: `${algorithmQuery} service.attributes.main.type:algorithm -isInPurgatory:true`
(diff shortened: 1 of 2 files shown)


KatunaNorbert:
Indeed not all the algorithms are displayed in our `AssetSelection` component. There was a paginations set for the query so only first 100 assets was retrieved. I removed the pagination but there is still an offset which needs to be set and limits us. I set that to 200 just for testing purpose, but would be nice to have the option to get all the date if we don't want to implement a pagination for the `AssetSelection` component. Otherwise a pagination is needed

kremalicious:
Good finding. Realistically speaking, this is on Rinkeby right now:

Screen Shot 2021-05-26 at 13 02 11

So would bump the offset to maybe 500 and be done with it for now. Still nicer to just scroll instead of clicking through pages, and doing some smart infinite loading on scroll might be overkill for the usage we have right now for this component.

KatunaNorbert:
I agree, this seems the best solution for now. We also lose the use of search if we implement the pagination or smart loading

kremalicious:
testing on Rinkeby, when editing compute settings, on live I get `52` items in this list whereas with this PR I get a total of `68`:

Screen Shot 2021-05-26 at 15 40 27

What funky numbers :-) But if there are no other restrictions going on in the query for this list, then I should get `158` in this list based on the above search for algos.

KatunaNorbert:
The algorithms which doesn't have prices are filtered when we retrieve prices for them. So those would not appear in our `AssetSelection` component. So with the new changes there are probably 16 algorithms with prices in the extra 58 that we get now with the new offset

kremalicious:
got it, then those 2 lines of code changes are perfect

kremalicious (approved):

Issue #604, opened 2021-05-18
Add free pricing option


mihaisc:
![image](https://user-images.githubusercontent.com/26000280/118619033-e2ba3e80-b7cc-11eb-81b3-8d0f93afb719.png)

In the free tab we will have a checkbox ( `I agree that this asset will be able to be purchased at no cost` , but some better text) so that the user doesn't add free pricing by mistake.

kremalicious:
To keep the categorization logical from user perspective, free price should be an extension of fixed price cause that's what it is. E.g. with a checkbox/switch which makes the number input disappear or something

mihaisc:
Agree with the env, do we have this implemented for fre or amm?
Don't really agree bundling it with fre, technically it's not the same thing. Another thing is what happens if you have just amm and free price, how do you show free if it's bundled in fre? I'd rather have each pricing schema in a different component

krisliew:
Agree with @mihaisc point, I think is better to keep them decoupled so any update for fre or free pricing in future wouldn't have any impact on each other too.

PR #621, opened 2021-05-24
Free Pricing Option at create Pricing

krisliew:
Close: #604 

Changes proposed in this PR:

- [x] Add `free` pricing option at `create pricing` #604
- [x] Add new env var `GATSBY_ALLOW_FREE_PRICING` to toggle free pricing option (see #453) 

image


mihaisc:
Why is the free checkbox in fixed as well?
image

krisliew:
> Why is the free checkbox in fixed as well?
> image

Fixed it, thanks! 👍

src/utils/feedback.ts
```diff
@@ -52,6 +52,17 @@ export function getCreatePricingExchangeFeedback(
   }
 }
 
+export function getCreateFreePricingFeedback(
+  dtSymbol: string
+): { [key: number]: string } {
+  return {
+    99: 'Creating dispenser...',
+    0: `Proposing ${dtSymbol} minter...`,
+    1: 'Dispenser approving minter...',
```
kremalicious, line 61: I know this is what is technically happening in background but nobody knows what this means. Either we explain what is happening in the UI text in addition to those steps, or adapt those step messages

src/components/molecules/FormFields/WarnCheckbox.tsx
```diff
@@ -0,0 +1,33 @@
+import React, { ReactNode, FormEvent, ReactElement, ChangeEvent } from 'react'
```
kremalicious, line 1: why is this a new component? It's a simple checkbox so why create a dedicated component for one specific checkbox instead of using our established `InputElement`? Same goes for the above styles which are an exact copy of the checkbox styles from that `InputElement` needlessly duplicating code.

More importantly, why a user prompt for acknowledgement which is not present in any other pricing method, making this highly inconsistent?  To me there should be some representation of free (either big "Free" or a "0 OCEAN") and the datatoken still being represented, just like on fixed pricing.

src/utils/feedback.ts
krisliew: How about this below, is it better? 
`99: 'Creating ${dtSymbol} faucet...'`
`0: 'Setting faucet as minter ...'`
`1: 'Approving minter...'  `

src/components/molecules/FormFields/WarnCheckbox.tsx
krisliew: Got it, I will try to update this to a big "Free"

jamiehewitt15:
In general it seems to work well but I have noticed and edge case that could be improved: 
1. The user first tries to set the price to zero in the fixed pricing tab
2. User sees the error message (saying the price "Must be equal or more than 1") and then switches to the free pricing tab.
3. The error message remains in place on the free pricing tab and prevents the user from setting up free pricing. 
4. The user has to go back to the fixed pricing tab, set it to 1, and then return to the free pricing tab to set up free pricing. 

edge case 1

jamiehewitt15:
> In general it seems to work well but I have noticed an edge case that could be improved:
> 

This is fixed now

jamiehewitt15:
@krisliew It didn't work for me... 

It gave me the success message after creating the free pricing:
Free Pricing success - 2021-06-01_14-39

However, when I reloaded the page, it says no price set: 
Free Pricing - No price set - 2021-06-01_14-39

Also, I opened it in another tab using a different wallet and it also shows no price: 
No price - 2021-06-01_14-39

krisliew:
@jamiehewitt15 I am keeping this PR small just for the create free pricing function (easier review), u can find your asset with price set as free using deployment here https://github.com/oceanprotocol/market/pull/634

jamiehewitt15:
Ah ok thank you, I will also test out  #634 then. They should be merged together

jamiehewitt15 (approved):

jamiehewitt15:
It seems to work alright, I have approved, It shouldn't be merged until #634 is ready though

PR #625, opened 2021-05-25
use stableaqua URLs

KatunaNorbert:
Fixes #101 .

Changes proposed in this PR:

- Test new stableaqua URLs


kremalicious:
multinetwork Aqua is the new stable Aqua as it seems, so #628 takes over testing that too

PR #627, opened 2021-05-26
Check maximum datatoken to receive for pool priced assets in compute and consume

bogdanfazakas:
Fixes ##622 .

Changes proposed in this PR:

- check the maximum available datatoken to receive before buy compute or consume


src/components/organisms/AssetActions/Compute/index.tsx
```diff
@@ -190,6 +219,10 @@ export default function Compute({
   useEffect(() => {
     if (!ocean || !accountId) return
     checkPreviousOrders(ddo)
+    checkAssetDTMaxBuyQuantity(
+      price,
+      ddo.findServiceByType('metadata').attributes.main.type
+    )
   }, [ocean, ddo, accountId])
```
KatunaNorbert, line 226: shouldn't we include the price in the dependency array, so we recheck if the price changes?

bogdanfazakas: yep we can, at the moment we fetch the price only when the component mounts or ddo changes so `[ddo]` should be enough for the moment. It will be mandatory once we refetch the price from time to time

KatunaNorbert: ok, than it can be handled later

bogdanfazakas:
> Wouldn't be nicer for the user to show only the message that says there aren't enough DT in the pool when selecting the algorithm? The message is getting to long and we also have 2 messages in one which could be confusing:
> Screenshot 2021-05-27 at 16 50 37
> Screenshot 2021-05-27 at 16 53 27
> 
> Or we could disable the algorithm selection, this might make more sense

This check is made on both assets so we must display the according message for each of them. If we decide to stay with one message for this case we can reduce the message to show just the message with low DT amount if it is discovered for one asset because the other one doesn't make sense anymore.

bogdanfazakas:
@KatunaNorbert fixed the confusion and not displaying anymore the second message if one asset is low in liquidity 
you can reproduce it on this asset did:op:d5766cc73AE44977b4F7d72cDd3b33740101e830
Screenshot 2021-06-14 at 15 31 26

KatunaNorbert (approved):
Cool, looks good now

kremalicious:
this looks like it was successfully tested and all so would only need a merge or rebase for releasee

bogdanfazakas:
> this looks like it was successfully tested and all so would only need a merge or rebase for releasee

yeah it was also showcased in one of our demos. I will update the PR so it ca be merged then 😃

PR #628, opened 2021-05-27
Multinetwork Interface

### LINKED ISSUE #67: Drag & drop into decentralized storage
Goal should be to provide publishers with an easy solution to drag and drop some files during publish flow and in background they are added to a decentralized storage network.

This could start as providing an IPFS upload for sample files, since they can be public.

Storing actual data set files needs some more advanced solution to preserve data privacy.

### LINKED ISSUE #645: Allow marketowner to control what networks are supported
Since we are deploying contracts to multiple networks, it would be nice to have an ENV variable which controls what networks are supported for a market.

Maybe as a market owner I only want to support Mainnet & Rinkeby.
Or I have a enterprise POA and I only want to allow that.

Now, all networks from contracts/ocean.js are supported by default.

kremalicious:
- Closes https://github.com/oceanprotocol/multi-repo-issue/issues/67
- Closes #645
- Requires https://github.com/oceanprotocol/aquarius/issues/423

multinetwork

## Changes

- `Web3Provider`: now also handles ETH & OCEAN user balance without any ocean.js usage
- `OceanProvider`: this now only wraps the asset details view
- New networks selector in main menu which changes search queries across the app
- #675
- #672
- #667
- #690
- #693
- #676
- #736
- #708
- #745
- #740
- #746
- #686
- #729
- consolidate asset edit form actions ae988f083b7d7b50a34210107808d09cc3248bd9
- #761

## Main UI changes

Screen Shot 2021-07-15 at 15 53 07

Screen Shot 2021-07-16 at 13 51 06

Screen Shot 2021-07-16 at 14 06 07

Screen Shot 2021-07-15 at 14 57 43

Screen Shot 2021-07-15 at 14 31 45

Screen Shot 2021-07-22 at 12 27 01


mihaisc:
image
looks like a bug 😆 . This zone is getting kinda crowded and the icons now have too different meaning. Until now you had algo/data and access/compute that both represent the usage/type of the asset. The network is breaking the pattern.

Think about this:
> Network/Chain ? Do we need this on the teaser? Nice to have if we have room. If think the main goal is to get the user on the details page (and not discourage them if the asset is on a different chain, basically the asset teaser should actually tease the user not show all the data)

kremalicious:
> This zone is getting kinda crowded and the icons now have too different meaning
yes, agree, would be fine not showing network icon there as the mix of only icon or only text becomes inconsistent there. 

Only argument I have left for indicating network on asset teaser is that users immediately understand when arriving on front page that UI shows multiple networks. But icon could also be moved to bottom of asset teaser, we can decide after we have some user preference chainId switcher and know where it's going to be

kremalicious:
Alright, this is what we got in here where some basic mechanisms are in place:

New app configs:

- `metadataCacheUri`
- `chainIds`: List of chainIds which metadata cache queries will return by default.
- `chainIdsSupported`: List of all supported chainIds.

### Providers

#### `Web3Provider`

  - now also handles ETH & OCEAN user balance without any ocean.js usage
  - this completely decouples web3/user wallet & ocean in the app

#### `OceanProvider`
  - this now only wraps the asset details view and will initiate a connection based on `ddo.chainId` and the respective ocean.js ConfigHelper config when in AssetDetails view. As soon as Aquarius returns a `ddo.chainId` this should just work as is
  - one exception for config values: `metadataCacheUri` is an app config value now. This is because we need only the `metadataCacheUri` for most parts of the app and we don't have the `OceanProvider` around the whole app anymore
  - All still debatable, e.g. we could also wrap `OceanProvider` around whole app again and initially use it as a config store too. But so far it's nice it has such a reduced complexity

#### `UserPreferences`
  - this is the central place for holding the list of chainIds the interface should display
  - this is initially populated from the `chainIds` app config and a new UI list is populated from `chainIdsSupported`
  - based on current config this is then what all users get initially:
   Screen Shot 2021-06-09 at 11 56 01

  - so within any component this is how to know which assets to display:
 
    ```js
    const { chainIds } = useUserPreferences()
    ```
  - and finally all the places where we do a `queryMetadata` are now set up to fire again whenever that `chainIds` user preference changes which then makes #650 possible:

  Untitled

  - UI-wise it could make sense moving this user preference out into its own more prominent UI button

### Ocean config values

Basically `metadataCacheUri` does not change anymore, but all other values from ocean.js `ConfigHelper` are different in all networks. And most likely we need those values outside of `OceanProvider` and often a combination of them like when we want to display multiple TVL values from subgraph based on selected `chainIds`. But user wallet and ocean config are now completely detached so what we actually need is a mapping between the data in `ConfigHelper` and our list of user preferences `chainIds`. So multiple possible ways:

- wrap `OceanProvider` around whole app, store all `ConfigHelperHelper[]` in it
- since we only need the data we could simply export [all this](https://github.com/oceanprotocol/ocean.js/blob/main/src/utils/ConfigHelper.ts#L24-L145) in ocean.js so we can get it more easily in market
- again, only data is needed so we could also get this into Gatsby's data layer  on build time, make it part of `useSiteMetadata()` and be done with it

So once we have that data in the app somewhere we can have simple methods to return list of configs based on passed chainIds and such.

This then is pretty much needed for figuring out the History page where again. all queries need to change based on `chainIds` so we need e.g. multiple subgraph, node and maybe provider URIs on that page

kremalicious:
Temporarily created an interface in `OceanProvider` which types out `ddo.chainId` so all those typing errors in test runs are gone now. The last remaining one has a fix in #672 so once that's merged all tests on here should go green (except CodeClimate).

As for actually mocking it so that we can test functionality was not so straightforward but what I did is within our custom `retrieveDDO()` method I simply inject a `chainId` into the DDO response. But there is just no way to figure out the right chainId there so it's just hardcoded to `1`. But it works already and on asset details view makes the ocean provider correctly initiate itself.

mihaisc:
Removing tvl and build was intentional ?

image

image

kremalicious:
yup, whole `MarketStats` is commented out in app for now until we figure out how to deal with `ApolloProvider`. And also pending a decision on what exact TVL we want to show there, meaning which network and such. Ideally we don't have to redesign too much until we move those stats in v4 anyway

mihaisc:
> yup, whole MarketStats is commented out in app for now until we figure out how to deal with ApolloProvider. And also pending a decision on what exact TVL we want to show there, meaning which network and such. Ideally we don't have to redesign too much until we move those stats in v4 anyway

easy, total tvl of selected networks and a small tooltip where we breakdown for each of the selected network

I think we need to drop the Apollo context, it's useless in this situation. We need to create function that use the client and they are based on chainId (or network i mix them up).

mihaisc:
Aqua is ready to be tested https://multiaqua.oceanprotocol.com

mihaisc:
If i go to an asset and Metamask is set to another network i don't see all the tabs. I need to change network and refresh page.

image
image

KatunaNorbert:
For highest liquidity we need the changes to subgraph request to query multiple networks. @bogdanfazakas has done that in #693. That PR also makes the tabs appear because we there have the price for the asset

mihaisc:
Network not visible in teaser, probably broke in the global search pr
image

KatunaNorbert:
Yes, It is only visible on hover. I thought this is intentional to take from the UI load with the hover effect added

jamiehewitt15:
I noticed that the metadata history links don't work on the preview of this branch

alexcos20 (approved):

KatunaNorbert (approved):

PR #631, opened 2021-05-28
Update file metadata from provider

### LINKED ISSUE #623: Update file metadata from provider
- [ ] Get the type and the length from the Provider API call
- [ ] Update the UI to reflect the metadata

claudiaHash:
Fixes #623  .

Changes proposed in this PR:

-  file metadata fetched from provider
- UI updated with new data

### DIFF
--- a/src/components/atoms/File.tsx
+++ b/src/components/atoms/File.tsx
@@ -27,14 +27,7 @@ export default function File({
   small?: boolean
   isLoading?: boolean
 }): ReactElement {
-  const [loading, setLoading] = useState<boolean>(true)
-
-  useEffect(() => {
-    isLoading && setLoading(true)
-  }, [isLoading])
-
   if (!file) return null
-  console.log(isLoading, loading)
 
   const styleClasses = cx({
     file: true,
@@ -42,22 +35,26 @@ export default function File({
     [className]: className
   })
 
-  return !loading && (isLoading === undefined || isLoading === false) ? (
+  return (
     <ul className={styleClasses}>
-      {file.contentType || file.contentLength ? (
+      {isLoading === false || isLoading === undefined ? (
         <>
-          <li>{cleanupContentType(file.contentType)}</li>
-          <li>
-            {file.contentLength && file.contentLength !== '0'
-              ? filesize(Number(file.contentLength))
-              : ''}
-          </li>
+          {file.contentType || file.contentLength ? (
+            <>
+              <li>{cleanupContentType(file.contentType)}</li>
+              <li>
+                {file.contentLength && file.contentLength !== '0'
+                  ? filesize(Number(file.contentLength))
+                  : ''}
+              </li>
+            </>
+          ) : (
+            <li className={styles.empty}>No file info available</li>
+          )}
         </>
       ) : (
-        <li className={styles.empty}>No file info available</li>
+        <LoaderArea />
       )}
     </ul>
-  ) : (
-    <LoaderArea />
   )
 }
--- a/src/components/atoms/File.tsx
+++ b/src/components/atoms/File.tsx
@@ -1,22 +1,40 @@
-import React, { ReactElement } from 'react'
+import React, { useEffect, useState, ReactElement } from 'react'
 import { File as FileMetadata } from '@oceanprotocol/lib'
 import filesize from 'filesize'
 import classNames from 'classnames/bind'
 import cleanupContentType from '../../utils/cleanupContentType'
 import styles from './File.module.css'
+import Loader from '../atoms/Loader'
 
 const cx = classNames.bind(styles)
+
+function LoaderArea() {
+  return (
+    <div className={styles.loaderWrap}>
+      <Loader />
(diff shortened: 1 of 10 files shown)


src/components/atoms/File.tsx
```diff
@@ -1,30 +1,48 @@
...
+  console.log(isLoading, loading)
 
   const styleClasses = cx({
     file: true,
     small: small,
     [className]: className
   })
 
-  return (
+  return !loading && (isLoading === undefined || isLoading === false) ? (
```
kremalicious, line 45: the file icon should be rendered at all times so there will be no layout shift. Only its content should load and be replaced

kremalicious:
were you able to make this actually work, specifically the file size on some asset? Cause we run into issues over in #479 where `provider` seems to never return any `contentLength`.

claudiaHash:
For the great majority of the assets, I could, yes. But there are some assets like: `did:op:555a0DD8B24963e38FFF1a0c84F13f694Bd0f4F6` , `did:op:d5766cc73AE44977b4F7d72cDd3b33740101e830`, `did:op:1D9a0Ad91a5bD3d4091fD474c481A3079A846260` for which I've got a  `index: 0; valid: false` response although on live there is valid data for them. 
However, I ran  into some issues when I sent the did as a string, not as a DID.

krisliew:
> However, I ran into some issues when I sent the did as a string, not as a DID.

@claudiaHash When passing string argument, it can only accept file endpoint url for `getFileInfo()` 
See here too: https://github.com/oceanprotocol/provider/blob/79de2a4aabd248ac2c72184d7067c58ac3ca7bbf/ocean_provider/routes/consume.py#L149

src/components/organisms/AssetActions/index.tsx
```diff
@@ -2,28 +2,52 @@ import React, { ReactElement, useState, useEffect } from 'react'
...
 export default function AssetActions(): ReactElement {
   const { accountId } = useWeb3()
-  const { ocean, balance, account } = useOcean()
-  const { price, ddo, metadata } = useAsset()
+  const { config, ocean, balance, account } = useOcean()
+  const { price, ddo } = useAsset()
 
   const [isBalanceSufficient, setIsBalanceSufficient] = useState<boolean>()
   const [dtBalance, setDtBalance] = useState<string>()
+  const [fileMetadata, setFileMetadata] = useState<FileMetadata>()
```
KatunaNorbert, line 23: Looks like there should be a default value in here otherwise due to this `if (!file) return null`  check from the `File` component the component will not show the content. That's why you don't get the component. 
Might be set to an empty object or create an object with the data from the `ddo` until you get the data from provider

claudiaHash: Thought that the problem was elsewhere. Thanks, I'll fix it.

claudiaHash:
> > However, I ran into some issues when I sent the did as a string, not as a DID.
> 
> @claudiaHash When passing string argument, it can only accept file endpoint url for `getFileInfo()`
> See here too: https://github.com/oceanprotocol/provider/blob/79de2a4aabd248ac2c72184d7067c58ac3ca7bbf/ocean_provider/routes/consume.py#L149

Sorry, I don't think I expressed myself correctly. There are some assets without an url in their ddo ( like this one: did:op:6c5Ec27A1e6F14Ea45736fDF115D9b9279aF617a ) so, if we don't send the asset did as DID, we can't get a valid file info data.

krisliew:
> > > However, I ran into some issues when I sent the did as a string, not as a DID.
> > 
> > 
> > @claudiaHash When passing string argument, it can only accept file endpoint url for `getFileInfo()`
> > See here too: https://github.com/oceanprotocol/provider/blob/79de2a4aabd248ac2c72184d7067c58ac3ca7bbf/ocean_provider/routes/consume.py#L149
> 
> Sorry, I don't think I expressed myself correctly. There are some assets without an url in their ddo ( like this one: did:op:6c5Ec27A1e6F14Ea45736fDF115D9b9279aF617a ) so, if we don't send the asset did as DID, we can't get a valid file info data.

Yes, we definitely have to use the asset did (as `DID`) based on `getFileInfo()` implementation. 

For passing `string` param, it is used for `fileinfo()` method to check publish form file URL if i am not mistaken, see here at my PR
https://github.com/oceanprotocol/market/blob/8bc4665a864aff3deeab8adc7e0483699797426e/src/utils/provider.ts#L28-L34

mihaisc:
For the most assets it's ok. 
There are still some issue on some (not related to this pr, these are provider issues, but are reflected on the market) ; (pr vs main)

image
image

image
image

Not sure if we should merge until we make sure the provider information is 100%.

claudiaHash:
> Not sure if we should merge until we make sure the provider information is 100%.

If the structure of the data that comes from provider won't be changed, I think it could be merged. But maybe is safer to wait until the provider info is 100% reliable and merge it then.

mihaisc (approved):

PR #635, opened 2021-05-31
Fix trading tokens form validation

### LINKED ISSUE #620: Trading datatoken → OCEAN is broken because of form validation
1. go to Trade tab of an asset for which you have a datatoken in your wallet
2. switch the order to datatoken → OCEAN
3.  in the datatoken field enter any number within the shown maximum to spend range
4. Result:
    Screen Shot 2021-05-22 at 14 49 36

Encountered on Polygon asset. Looks like just some form UI validation error, all the calls to get the max/min numbers from pool are working as expected.

DimitarSD:
Fixes #620 .

Changes proposed in this PR:

- updates parameter for UI validation

### DIFF
--- a/src/components/organisms/AssetActions/Trade/FormTrade.tsx
+++ b/src/components/organisms/AssetActions/Trade/FormTrade.tsx
@@ -64,7 +64,7 @@ export default function FormTrade({
         .required('Required')
         .nullable(),
       datatoken: Yup.number()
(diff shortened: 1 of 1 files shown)


kremalicious (approved):
that moment you realize bug is not caused by some complex calculation error but we were just passing wrong values to the validation
giphy-1

Could not reproduce any validation error anymore in Trade tab with this change, so should be solved

PR #639, opened 2021-06-02
Remove price.isConsumable from elastic query

claudiaHash:
Changes proposed in this PR:

- removed price.isConsumable from elastic query
- updated subgraph query


src/utils/subgraph.ts
```diff
@@ -88,13 +88,13 @@ const PreviousOrderQuery = gql`
 `
 const HighestLiquidityAssets = gql`
   query HighestLiquidiyAssets {
-    pools(orderBy: valueLocked, orderDirection: desc, first: 15) {
+    pools(
+      where: { datatokenReserve_gte: 1 }
```
mihaisc, line 92: why do we need this condition for highest liquidity? Is this used in some other places than the home component for highest liquidity?

kremalicious: we found one asset with high liquidity where Aquarius had `isConsumable: ""` in the DDO so it was showing up in list after we removed `price.isConsumable` from the query. And a simple workaround to make sure no rug pull assets show up was to have a check for at least 1 datatoken left in pool

claudiaHash: After removing `price.isConsumable ` from elastic query, the list of highest liquidity assets was slightly different from what we knew to be correct. There was an asset: did:op:e73423A5E1334842f1eb0E0C8eBDb40E5270ede3 with `isConsumable: ""` and 0.123 datatokens in the pool, not enough to support any action. So should have had: `isConsumable: false`. Therefore, to replicate the checking that we previously had, I added it here, as @kremalicious suggested.
That's used just for home component, for highest liquidity.

mihaisc: ok

mihaisc (approved):

PR #640, opened 2021-06-03
Token approval split-up

### LINKED ISSUE #636: Token approval split-up
Goal is:
- split up token approval & action in all places into different user-initiated actions: Add Liquidity, Trade
- token approval itself presents a choice of specific amount (default) or infinite approval

UI-wise we need to prevent 3 buttons being on screen all the time for this:
- if token has not been approved there should be 2 buttons, _Approve TOKEN_ & _Action_. Upon clicking on _Approve_, button turns into 2 buttons: _xxx TOKEN_ / _Infinite_. User then has to decide and depending on the choice we do the approval flow.
- if token has infinite approval already, only the _Action_ button is visible

After UI and interaction implementation is there awe most likely have to refactor some ocean.js methods which call `approve` themselves.

Along the way this might fix #592

KatunaNorbert:
Fixes #636 .

Changes proposed in this PR:

- split up token approval, default to specific approvals
- new user preference to overwrite all approvals to be infinite
- fix various fetch and console errors:
  - d967f0b ExplorerLink console error fixes
  - e324fc0 dtItem fix, error fixes based on asset network match
  - 86654e4 enable add liquidity button at all times
  - 2efa1d2 fix getting to add liquidity screen without wallet connected
  - 2402ccf fix crash when switching coins after value input

### UI Changes

Screen Shot 2021-08-23 at 12 01 08

Screen Shot 2021-08-23 at 12 01 51


bogdanfazakas:
also saw there's an empty file added src/components/molecules/TokenApproval.module.css

KatunaNorbert:
Thanks @bogdanfazakas, I fixed the mentioned issues. Also some extra fixing of setting the datatoken as default token for trade and removed the loading when increasing the amount, looked like too much component shifting for me.

bogdanfazakas:
Got another error on add liquidity when switching between ocean token and dataset token then trying to set max amount. It looks like the problem is from the decimal conversion maybe this will be solved in this PR #653  not sure if we need to tackle it also here. 
Screenshot 2021-06-15 at 11 57 20

KatunaNorbert:
Right, that would be covered over there so I don't think we should fix it over here as well. The issue also appears on the live market and it brakes the app when switching from Ocean token to Datatoken for any value beside 0

DimitarSD:
@bogdanfazakas Yep, it should be fixed. I've added a couple of checks before the conversion to decimal including for `null`

bogdanfazakas (approved):
LGTM, the only thing i found was the one with switching from Ocean token to Datatoken for any value beside 0, but that will be fixed by the #653 PR  not by this one

kremalicious:
That was too much of a button maze so I slightly rethought the user flow to provide a more opinionated default flow:

- infinite approval is a user preference
- by default, specific approval kicks in, user is informed about infinite possibility 
- setting user preference then switches the approve buttons
- resulting in one layer less of deciding about buttons for user
- user does not have to decide every time anymore about specific or infinite

https://user-images.githubusercontent.com/90316/128189704-b8a412b2-8ce6-4da5-a3c9-8310500c9770.mov

But needs more testing now. I only refactored so that flow works on Add Liquidity but also tried to not touch too much. Already seeing on Trade that the passed `coin` is the wrong way around I think

src/components/molecules/TokenApproval.tsx
```diff
@@ -0,0 +1,128 @@
...
+  amount,
+  coin
+}: {
+  actionButton: JSX.Element
+  disabled: boolean
+  amount: string
+  coin: string
+}): ReactElement {
+  const { ddo } = useAsset()
+  // const [approveToken, setApproveToken] = useState(false)
```
kremalicious, line 66: was not sure if this is needed at all after I removed the initial switch from approve button to approve deciding buttons, hence I left it commented out throughout code in here. Can be removed once verified it's indeed not needed anymore

KatunaNorbert:
Looks good, I agree with the infinite option always displayed being too much.
What could be improved is the way we inform the user about infinite amount option, I think it's too hidden and should be more visible.
Also saw some smaller bugs that I'll start working on, like undefined value displayed by default as the token amount value.

kremalicious:
also running into some blank screen sometimes during adding liquidity which is caused by some number conversion errors which should all be fixed by #653 so can be ignored in here (a.k.a. just reload the page when it happens 😄) until we get those changes in here through `main`

src/components/organisms/AssetActions/Trade/Swap.tsx
```diff
@@ -52,25 +53,25 @@ export default function Swap({
...
   const [totalValue, setTotalValue] = useState<string>()
   const [tokenAmount, setTokenAmount] = useState<string>()
-  ///
+
   useEffect(() => {
-    if (!ddo || !balance || !values || !price) return
+    if (!ddo || !balance || !values?.type || !price) return
 
     async function calculateMaximum() {
       const dtAmount = values.type === 'buy' ? maxDt : balance.datatoken
```
kremalicious, line 65: after the number → decimal PR merge in here, this seems to be the only place left in this PR where we do calculations and we might want to use `Decimal` here but not sure if really needed. So `dtAmount` & `oceanAmount`

KatunaNorbert: cool, I'll take a look

mihaisc:
I don't really understand what is going on here.  I'm trying to add liquidity

1.  Allow infinite amount unchecked

I started testing with the pool from this did:op:5CD4756B5980A1877f0191c2f4A4ef6FfB686596
First i got a `Approve something` button (forgot the exact label) and i signed for `Approve OCEAN spend limit`, all good. After this i got a `Supply` button, tx signed and successful . After this i add another value and now i have only the `Supply` button , but i still have to sign twice, so what was the point? 
image

If i now go to another pool i never get that `Approve` button, just `Supply`

2. Allow infinite amount checked 

Is this supposed to to something? nothing change from the previous point.

KatunaNorbert:
Seems to be something wrong when approving `ocean`, I'll investigate

KatunaNorbert:
The problem is on Ocean.js where before comparing the token amount with allowed tokens, the token amount is converted to Wei resulting strange values and the comparison fails. Doe to this oceanjs thinks we don't have enough allowed tokens and the user needs to approve again.

Created an issue on Ocean.js: https://app.zenhub.com/workspaces/frontend-60acd67b133a660014278212/issues/oceanprotocol/ocean.js/1008

bogdanfazakas:
blocked by https://github.com/oceanprotocol/ocean.js/pull/1012.

KatunaNorbert:
Retesting this and seems to be fine, just one thing that's not clear for me. When swapping tokens do we have to approve both tokens or just the one we are swapping( as it is right now )?

kremalicious:
> When swapping tokens do we have to approve both tokens or just the one we are swapping( as it is right now )?

All good as it is right now, approval is always only for the token which leaves the user's wallet so user approves the contract to take those tokens out of wallet. Receiving the swapped token then from contract is normal transfer so nothing with approval required

bogdanfazakas:
I was also retesting this today and all worked well on my end 👍

KatunaNorbert:
From what I see nothing was broke by the rebase and everything it's working.
One thing that I realised now it's that when we select the `Allow infinite amount` from user preferences we still check if the token amount is approved and show the approve button only when that's false. Shouldn't we check in this case if infinite amount is approved and show the approve button even if there are a few token approved?

src/components/molecules/TokenApproval.tsx
```diff
@@ -0,0 +1,138 @@
...
+    if (!ocean || !tokenAddress || !spender) return
+
+    const allowance = await ocean.datatokens.allowance(
+      tokenAddress,
+      accountId,
+      spender
+    )
+
+    amount &&
+      Number(amount) > 0 &&
```
kremalicious, line 98: Just double checking, is it ok to keep `Number` here or we need to switch to decimal.js? In theory only numbers end up here which user has put into the form fields at one point in time

KatunaNorbert: I think keeping the `Number` doesn't affect the functionality because we only transform strings to Number and only use it for comparison, but we can change it to Decimal to be consistent

bogdanfazakas (approved):
works good on my end 👍 ,  i don't think we should change anything more than that number to decimal change mentioned above, the actual flow seems right for me

mihaisc:
I did a quick test and it seemed to work ok, didn't have time to do a proper check.

KatunaNorbert:
Shouldn't we also use the token approval split-up on buying an asset, I see that we have to approve over there too?

kremalicious:
Agree, feels like an oversight keeping a specific approval there. But it also would require us to first refatcor the `ocean.pool.buyDt()` method over in ocean.js, and then over here change the flows for consume & compute. We can do it but not in this PR I would say

kremalicious (approved):

PR #641, opened 2021-06-03
Search issues fix

### LINKED ISSUE #619: Search issues
We need a useful and relevant search. The current search is inaccurate and useless. Quick test, the asset is not on the first page:

image
image

This needs some research into elastic queries

claudiaHash:
Fixes #619  .

Changes proposed in this PR:

- search by datatoken address, owner address, asset id added to existent elasticsearch query


mihaisc (dismissed):
Looks good, not sure if you want to add more since it's still in draft.
Fixes just a small issue from #619.

claudiaHash:
> Looks good, not sure if you want to add more since it's still in draft.
> Fixes just a small issue from #619.

There are still some issues unresolved yet that I'm working on.

src/components/templates/Search/utils.ts
```diff
@@ -66,42 +66,72 @@ export function getSearchQuery(
...
-        ],
-        default_operator: 'AND'
+      bool: {
+        must: [
+          {
+            bool: {
+              should: [
+                {
+                  query_string: {
+                    query: `${searchTerm}`,
```
mihaisc, line 86: `searchTerm` differs in `query_string` you need to manually add `OR` between the words . in `match_prase` it's exactly the string the user inputs.

src/components/templates/Search/utils.ts
```diff
@@ -66,42 +66,72 @@ export function getSearchQuery(
...
+                      'publicKey.owner',
+                      'dataToken',
+                      'dataTokenInfo.name',
+                      'dataTokenInfo.symbol',
+                      'service.attributes.main.name^10',
+                      'service.attributes.main.author',
+                      'service.attributes.additionalInformation.description',
+                      'service.attributes.additionalInformation.tags'
+                    ],
+                    default_operator: 'AND',
```
mihaisc, line 98: remove this line, it shouldn't be `AND`

src/components/templates/Search/utils.ts
```diff
@@ -66,42 +66,72 @@ export function getSearchQuery(
       `(service.attributes.additionalInformation.categories:\"${categories}\")`
     : text || ''
 
+  if (searchTerm.startsWith('did:op:')) {
```
mihaisc, line 69: why do we need this?

src/components/templates/Search/utils.ts
```diff
@@ -66,42 +66,72 @@ export function getSearchQuery(
...
+    searchTerm = searchTerm.substring(7)
+  }
   // HACK: resolves the case sensitivity related to dataTokenInfo.symbol
   searchTerm = '*' + searchTerm.toUpperCase() + '*'
```
mihaisc, line 73: stop adding `*` at the start and end of the string

src/components/templates/Search/utils.ts
claudiaHash: I added 'AND' as default op because using OR would give a match for every word in search string. For "worldlist used by" I got matches for assets that only contain "used" in their description/ title/ etc

src/components/templates/Search/utils.ts
claudiaHash: Removed it because searching by id that included "did:op:" in the beginning gave no result. Since the address that follows differentiates the assets, I thought that could be a solution.

src/components/templates/Search/utils.ts
mihaisc: yes, please used the searchTerm formated as described above with 'OR' between each word. "worldlist used by" would become "worldlist OR used OR by", just in `query_string`, not in `match_phrase`

src/components/templates/Search/utils.ts
claudiaHash: Added them for the cases user doesn't type all the search word. When searching by token symbol I found the following problem:
Screenshot from 2021-06-07 17-22-43
Screenshot from 2021-06-07 17-22-55

src/components/templates/Search/utils.ts
claudiaHash: Didn't used 2 different search terms on `query_string` and `match_phrase ` there's where the problem came from. Please ignore the comments.

src/components/templates/Search/utils.ts
mihaisc: it should work without removing did:op. Comment it for now. It work ok in my tests without removing `did:op`

mihaisc: searching for `a2B8b3aC4207CFCCbDe4Ac7fa40214fd00A2BA71` is not working anyway

src/components/templates/Search/utils.ts
mihaisc: but it creates problems in other places. Searching for datatoken symbol will be fixed in https://github.com/oceanprotocol/aquarius/issues/492

mihaisc:
We'll tweak the query after these changes are implemented.

src/components/templates/Search/utils.ts
claudiaHash: Oh, ok then I'll remove it. That was the only issue that `*` resolved.

claudiaHash:
I added that `did:op:` hack because I found that if the query_string and match_phrase search terms don't match case wise (only in that particular case) we get a server error:
Screenshot from 2021-06-07 19-09-08

It worked after that: 
Screenshot from 2021-06-07 19-23-07

mihaisc:
Ok, i'll play with it a bit and see how relevant are the result and if i can improve the query even more.

claudiaHash:
Not sure if a `-` separated search input should give results for the whole string or if it's ok if terms are separated. Eg. `mayors-office` gives results that match:  "office", "mayors-office"

mihaisc:
it's ok, the secret is that we order them properly. We need to ad the sort criteria, after score. Currently they are sorted by created or if no sort filter is specified aquarius sorts them by id (it's not ok, an issue was created). The scope is to create multiple queries and order the matches. For example in the current queries `match_phrase` is the most strict and will be first in the results.

mihaisc:
I removed the `SearchQuery` return type because we can't use it and i'm not sure why even bother. The structure of the es query is too  complex and i see no value in trying to map in typescript

mihaisc:
Added `Relevance` sort that will be default. Do not test it because it's not working (or not working as it should)  for now. Investigating why.

claudiaHash:
Ok, currently working on previous version on a bug: 0 assets displayed when selecting _All datasets and algorithms_ on Home

mihaisc:
> Ok, currently working on previous version on a bug: 0 assets displayed when selecting All datasets and algorithms on Home

Can't replicate this issue neither local or remove deployment

claudiaHash:
Yes, after the last changes it doesn't happen anymore.

jamiehewitt15:
It would be good to include results where there is a partial match. For example, if you search "data union" you'll find that "dataunion.app" doesn't appear in the results

partial match 2021-06-09_14-05

jamiehewitt15:
Searching with a trailing space effects the search results, it would be good to trim the space before doing the search.

For example, with a trailing space: 
trailing space 2021-06-09_14-12

Without a trailing space: 
no trailing space 2021-06-09_14-12

mihaisc:
> It would be good to include results where there is a partial match. For example, if you search "data union" you'll find that "dataunion.app" doesn't appear in the results

true, i'll look into it

claudiaHash:
It seems to have a problem with special characters, at least for this two cases:
Screenshot from 2021-06-09 15-10-42  Screenshot from 2021-06-09 15-11-27

Screenshot from 2021-06-09 15-13-08
Screenshot from 2021-06-09 15-12-37

mihaisc:
> It would be good to include results where there is a partial match. For example, if you search "data union" you'll find that "dataunion.app" doesn't appear in the results

While this seems to be possible (check links below) it is far too complex to handle in this pr. It is nice to have indeed and I will create a new issue for it

https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-word-delimiter-tokenfilter.html
https://www.javacodegeeks.com/2018/03/elasticsearch-ignore-special-characters-query-pattern-replace-filter-custom-analyzer.html
https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-shingle-tokenfilter.html

mihaisc:
There already a lot of improvements here. I have commented `Relevance` sort until it's fixed. There are some other issues in aqua but i don't want to block this pr based on them. It's already good enough. Created 2 other issues based on things found here.

kremalicious:
using this new search strategy in the background is already super valuable and seems to work well after some quick tests. One important, because very prominent search which I found to be not working at all is the `owner` search, so the prominent publisher link on asset details page. When clicking that it looks to me as there is no search filtering happening at all, only lists recent items, e.g. https://market-git-fix-619-search-issues-oceanprotocol.vercel.app/search/?owner=0x655eFe6Eb2021b8CEfE22794d90293aeC37bb325&sort=created&sortOrder=desc

mihaisc:
Ok, thanks on that. I remember now, i wanted to remove those but didn't finish. This way there is a bug if you click on your link and then search for something else  (delete address and type something else )  nothing will happen.

mihaisc:
Ok, it should be working now ok.

claudiaHash:
> Ok, it should be working now ok.
The search after link clicking works perfectly. Since what I tested, the rest of the search works as it did before the last push. I think we could finally merge it.

jamiehewitt15:
It's a big improvement on before, all of the issues I mentioned above are fixed now.

jamiehewitt15 (approved):

PR #642, opened 2021-06-03
Put back in place removed docker hub image check

### LINKED ISSUE #572: Proxy for docker hub validation
We need to create a proxy for `https://hub.docker.com/v2/repositories` so we don't run into CORS issues when publishing an algorithm

bogdanfazakas:
Fixes #572  .

Changes proposed in this PR:

- re-added the validation for the custom images on docker hub
- used the newly created proxy from this [repo](https://github.com/oceanprotocol/dockerhub-proxy) instead of a direct call to docker hub repository api


bogdanfazakas:
> seems there are a lot of prettier errors.

yep had some problems with local eslint version and the those files were updated on merge trying to fix the problem now

kremalicious:
just checking, based on interface copy, using only `ubuntu` (as that's a public Docker image name) should work but it doesn't:

Screen Shot 2021-06-11 at 13 42 04

kremalicious:
The UX in case of error also feels a bit weird and surprising. Upon hitting publish at end of page I'm auto-scrolled to top and a toast error appears and in my case the docker image URL field is not even in viewport anymore. It would be way better to have the error show up where the user has to take action to actually publish, so in this case on the Docker image URL field.

Another alternative would be to use the same UI pattern as we use for adding files and samples, as then error would show just when user interacts with that field so we would be kinda fine keeping the toasts for the error in that case

bogdanfazakas:
> just checking, based on interface copy, using only `ubuntu` (as that's a public Docker image name) should work but it doesn't:
> 
> Screen Shot 2021-06-11 at 13 42 04
> 
> Like, what exactly do we expect there?
> 
> Screen Shot 2021-06-11 at 13 44 03

yep there seems to be a problem with the oficial images, debugging it now seems that we have to hardcode something for those images like ubuntu alpine etc. I previously tried with our public images like oceanprotocol/aquarius etc and for that works

We expect the name from docker hub so ubuntu in this case and oceanprotocol/aquarius for our aqua project

bogdanfazakas:
@kremalicious  it looks like we need to concat a 'library/{name}' for the official image names like ubuntu, alpine, redis etc otherwise we will have to let users know that they should add something like library/ubuntu in the name and that is not intuitive. I think we can implement this check and concatenation in the docker hub proxy repo rather than the market repo.

kremalicious (approved):
awesome, confirmed the latest changes over on `dockerhub-proxy` work with these changes. Used `ubuntu` & `xenial` and all was good. Using `ubuntu` & `whatever` correctly errored. I like the view scrolling workaround trick in case of errors, which is good enough

PR #646, opened 2021-06-07
Fix assets in purgatory not displayed

KatunaNorbert:
Fixes #644.

Changes proposed in this PR:

- Fix assets in purgatory are not displayed

PR #617, opened 2021-05-20
Bump all non-gatsby dependencies / npm v7

kremalicious:
Over in #611 the update to Gatsby v3 led to a myriad of problems suggesting we should hold off on that. In the meantime updating all other dependencies and all Gatsby packages to their latest working Gatsby pre-v3 versions, none of them being major changes.

All changes in app files is purely cosmetic because of ESLint & Prettier rule updates and should not affect functionality.

Also some other updates taken out of #611:

- fresh package-lock with npm v7 (created with `npm install --legacy-peer-deps`, verified that fresh `npm install` works without that flag with npm v7)
- ESLint config updates and simplifications
- Prettier formatting updates
- dependency cleanup for unused packages
- manual lockfile fix to work around npm/cli#2610
- add `.nvmrc` with Node.js v16, closes #616

Remaining major updates which should be handled individually in the future:

- chart.js / react-chartjs-2
- Gatsby v3 + all its plugins 
- graphql (had to lock to a specific version which is used by all packages, otherwise Gatsby build fails)

PR #647, opened 2021-06-07
Revert "Fix assets in purgatory not displayed"

mihaisc:
Reverts oceanprotocol/market#646

Thanks @bogdanfazakas  for noticing

> i'm not sure but i believe the package lock from this PR was locked using npm v6 and overwritten the changes from the previous merge of #617 Bump all non-gatsby dependencies / npm v7


bogdanfazakas (approved):
So we revert this and create another PR to solve the issue with that one line fix ?

mihaisc:
Actually you are right, this way we have to do another pr. Just fixing the lock file would be faster. I'll create another pr

PR #648, opened 2021-06-07
fix lock file

mihaisc:
Fix lock file after https://github.com/oceanprotocol/market/pull/646


mihaisc:
@bogdanfazakas can you check this out, it seems that my lock file is not ok for some reasom

bogdanfazakas (approved):
LGTM, i the test errors seem to be a GitHub issue i guess

PR #653, opened 2021-06-08
Update Number to Decimal

### LINKED ISSUE #624: Update from number to decimal
We need to update all places in the code where we use `Number` for conversion to use `Decimal`

- [x] Check all the files where we are using numbers 
- [x] Update them
- [x] Test and validate 
- [x] Edge cases testing

After the multi-network PR a couple of things need to done:
- [x] Rebase into the PR branch
- [x] Fix all rebase errors
- [x] Check for `Number` conversion that needs to be switched to `Decimal`
- [x] Test all changes during switching & update the PR regarding what has been changed and how to test it
- [x] Final PR review before merging into main

DimitarSD:
Fixes #624 

## Files

- [x] [Conversion.tsx](https://github.com/oceanprotocol/market/blob/main/src/components/atoms/Price/Conversion.tsx) - can’t be updated since the lib we’re using `coingecko` accepts only type number as a parameter
- [x] [PriceUnit.tsx](https://github.com/oceanprotocol/market/blob/main/src/components/atoms/Price/PriceUnit.tsx) - can’t be updated since the lib we’re using `coingecko` accepts only type number as a parameter
- [x] [Details.tsx](https://github.com/oceanprotocol/market/blob/main/src/components/molecules/Wallet/Details.tsx) - can’t be updated since the lib we’re using `coingecko` accepts only type number as a parameter
- [x] [Consume.tsx](https://github.com/oceanprotocol/market/blob/main/src/components/organisms/AssetActions/Consume.tsx) - I don’t think we need to change it here. We use `Number` only for a boolean check
- [x] [components/organisms/AssetActions/index.tsx](https://github.com/oceanprotocol/market/blob/main/src/components/organisms/AssetActions/index.tsx) - I don’t think we need to change it here. We use `Number` only for a boolean check
- [x] [FormComputeDataset.tsx](https://github.com/oceanprotocol/market/blob/main/src/components/organisms/AssetActions/Compute/FormComputeDataset.tsx) - I don’t think we need to change it here. We use it for UI display
- [x] [components/organisms/AssetActions/Compute/index.tsx](https://github.com/oceanprotocol/market/blob/main/src/components/organisms/AssetActions/Compute/index.tsx) - I don’t think we need to change it here since we’re using it for boolean check
- [x] [components/organisms/AssetActions/Pool/index.tsx](https://github.com/oceanprotocol/market/blob/main/src/components/organisms/AssetActions/Pool/index.tsx)
- [x] [components/organisms/AssetActions/Pool/utils.tsx](https://github.com/oceanprotocol/market/blob/main/src/components/organisms/AssetActions/Pool/utils.ts)
- [x] [FormAdd.tsx](https://github.com/oceanprotocol/market/blob/main/src/components/organisms/AssetActions/Pool/Add/FormAdd.tsx)
- [x] [components/organisms/AssetActions/Pool/Add/index.tsx](https://github.com/oceanprotocol/market/blob/main/src/components/organisms/AssetActions/Pool/Add/index.tsx)
- [x] [FormTrade.tsx](https://github.com/oceanprotocol/market/blob/main/src/components/organisms/AssetActions/Trade/FormTrade.tsx)
- [x] [components/organisms/AssetActions/Trade/index.tsx](https://github.com/oceanprotocol/market/blob/main/src/components/organisms/AssetActions/Trade/index.tsx)
- [x] [Output.tsx](https://github.com/oceanprotocol/market/blob/main/src/components/organisms/AssetActions/Trade/Output.tsx)
- [x] [Swap.tsx](https://github.com/oceanprotocol/market/blob/main/src/components/organisms/AssetActions/Trade/Swap.tsx)
- [x] [TradeInput.tsx](https://github.com/oceanprotocol/market/blob/main/src/components/organisms/AssetActions/Trade/TradeInput.tsx) - I don’t think we need to change it here
- [x] [Dynamic.tsx](https://github.com/oceanprotocol/market/blob/main/src/components/organisms/AssetContent/Pricing/FormPricing/Dynamic.tsx) - I don’t think we need to change it here; used for UI
- [x] [components/organisms/AssetContent/Pricing/FormPricing/index.tsx](https://github.com/oceanprotocol/market/blob/main/src/components/organisms/AssetContent/Pricing/FormPricing/index.tsx)
- [x] [Price.tsx](https://github.com/oceanprotocol/market/blob/main/src/components/organisms/AssetContent/Pricing/FormPricing/Price.tsx)
- [x] [PoolShares.tsx](https://github.com/oceanprotocol/market/blob/main/src/components/pages/History/PoolShares.tsx)

## Changes proposed in this PR:

- Switching from `Number` to `Decimal` where is possible and necessary.
- Update types of some variables to use `string` - having a unified approach to passing numbers across the different react elements

In some places like [File.tsx](https://github.com/oceanprotocol/market/blob/main/src/components/atoms/File.tsx#L34) and [Time.tsx](https://github.com/oceanprotocol/market/blob/main/src/components/atoms/Time.tsx#L21), we don't need to switch to `Decimal`.

## How to test

In some cases switching from `Number` to `Decimal` in the file is not enough since we're using different structures. Follow the commits and what changes have been made.

- Testing with edge cases (where is possible)

Things to test/look for:
- Before each conversion to `Decimal` we have `isValidNumber()` check since any value different from `number`, `string`, or `Decimal` will break the conversion.
- Some calculations are more complex (a lot of things are happening) and the syntax when using `Decimal` is a bit different. When reviewing check for valid flow of mathematical operations (best way: check the changes in the commit - old/new)

## Notes

- Since we're using `Decimal` passing `undefined`, `null` and etc to the function `new Decimal()` breaks the code. Unfortunately, when fetching data from other places it's impossible to completely avoid this. That's why before each `Decimal` conversion there is a check with `isValidNumber`, in the case it's not a valid number we set a default value which is updated as soon as the conversion to `Decimal` can pass
- In some file where we use `Number` in boolean checks I left it as it is. I don't think we need to change it there since it doesn't make much difference.


src/components/organisms/AssetActions/Trade/Output.tsx
```diff
@@ -29,8 +34,12 @@ export default function Output({
       setSwapFee(`${Number(swapFee) * 100}`)
```
mihaisc, line 34: is there a reason why this was not converted to decimal?

mihaisc:
I tested all the use cases that i could think of on rinkeby did:op:5CD4756B5980A1877f0191c2f4A4ef6FfB686596 and it worked ok. Can someone else do a quick test run on another asset?

KatunaNorbert:
I'll test it

KatunaNorbert:
Tested all the asset actions for did:op:B9ef87bcCb671c7f4B99b90b8c3cF6ACeC867199 on rinkeby and everything worked well, also the displayed values looking good comparing it to what we have on main

src/components/organisms/AssetActions/Trade/Output.tsx
DimitarSD: I missed this one, thanks for catching it! I'll go through all files again to check there aren't any missing (and needed) conversations like this one.

kremalicious:
fyi, merged `main` into here as the changeset of this PR was way too big (202 files changed, 30000+ lines of code added) so think somewhere during its last rebase something weird happened. It's now back at a manageable amount of reviewable code changes

src/components/organisms/AssetActions/Pool/index.tsx
```diff
@@ -22,8 +22,14 @@ import { useWeb3 } from '../../../../providers/Web3'
 import PoolTransactions from '../../../molecules/PoolTransactions'
 import { fetchData, getQueryContext } from '../../../../utils/subgraph'
 
+import { isValidNumber } from './../../../../utils/numberValidations'
+import Decimal from 'decimal.js'
+import { constants } from 'node:buffer'
```
kremalicious, line 27: I think this is unused throughout this file

src/components/organisms/AssetActions/Pool/TokenList.tsx
```diff
@@ -31,9 +32,9 @@ export default function TokenList({
         <div>
           <Token symbol="OCEAN" balance={ocean} />
           <Token symbol={dtSymbol} balance={dt} />
-          {conversion > 0 && (
+          {conversion.greaterThan(0) && (
             <Conversion
-              price={`${conversion}`}
+              price={`${conversion.toString()}`}
```
kremalicious, line 37: double string, the template literals already make it a string but to be sure we could remove them and only do:

```suggestion
              price={`${conversion}`}
```

But maybe that `toString()` method does something we actually need so in that case:

```suggestion
              price={conversion.toString()}
```

src/components/organisms/AssetActions/Pool/Add/FormAdd.tsx
```diff
@@ -74,15 +77,20 @@ export default function FormAdd({
         coin === 'OCEAN' ? ocean.pool.oceanAddress : ocean.pool.dtAddress,
         `${values.amount}`
       )
+      //
```
kremalicious, line 80: unused comment characters

kremalicious (changes requested):
just super small code review things I found, rest is all good. That `Decimal.set()` feels very repetitive, if we do not want to abstract out `Decimal` into a util file, we should at least make the values we pass there into a `const` somewhere so we only set this once

src/components/organisms/AssetActions/Pool/TokenList.tsx
mihaisc: I'd rather go with the recommended usage from Decimal, that is `.toString()`. Maybe it's the same, but why risk it and get some hidden, hard to find bugs.

DimitarSD: It's the same. We don't need the double string, `.toString()` is enough. In other places where we had the same thing, I removed the template literals.

src/components/organisms/AssetActions/Pool/index.tsx
DimitarSD: yep, there are also a lot of variables across the market's code that are never used as well.

kremalicious: so then remove then?

src/components/organisms/AssetActions/Pool/TokenList.tsx
kremalicious: I don't see the mentioned code change, please push

src/components/organisms/AssetActions/Pool/index.tsx
DimitarSD: Should I remove all of them in this PR? It might be better have a create a separate one for them

src/utils/numberValidations.ts
```diff
@@ -0,0 +1,8 @@
+export function isValidNumber(value: any) {
```
kremalicious, line 1:
```suggestion
export function isValidNumber(value: boolean) {
```

src/utils/numberValidations.ts
```diff
@@ -0,0 +1,8 @@
+export function isValidNumber(value: any): number {
```
kremalicious, line 1: Ignore my former suggestion, you are right that `any` is actually ok here as it’s a method checking the input so implies any input should be passable. But the return type based on actual code is really `boolean`

DimitarSD: Yeah, I'm not sure why I added `number`, thanks for noticing 😂 😂

kremalicious (approved):
Impossible to fully test everything so let’s just move ahead. Thanks for the detective work and getting this PR through many `main` merges

jamiehewitt15:
I've gone through all of the most obvious test cases with this: 
1. Publishing an asset
2. Creating a pool
3. Publishing an asset and creating fixed pricing
4. Swapping datatokens
5. adding liquidity to the pool
6. buying datatoken/ consuming the data asset

At the moment it's hard to confirm that everything is working because there seems to be an issue. After successfully publishing a dataset it says there is "no file info available" for the dataset and when I try to purchase it, I get the error message saying the URL is unavailable. I first noticed this while testing this branch but I have noticed it's the same on main, so the issue hasn't been caused here. 

No information available

I was using a simple [url to a csv](https://people.sc.fsu.edu/~jburkardt/data/csv/addresses.csv) in the test which seems to work fine, and the provider confirmed it was valid while publishing.

jamiehewitt15:
> At the moment it's hard to confirm that everything is working because there seems to be an issue

I created a separate issue for the problem I mentioned above: https://github.com/oceanprotocol/market/issues/804

jamiehewitt15:
I have tested this branch again and found a strange user experience that I haven't seen on main before. When creating a pool I got a toast error message saying "ERROR: Failed to create a new pool", however, when I refreshed the [page](https://market-gg4cu94r9-oceanprotocol.vercel.app/asset/did:op:0CBb54B0f33eE7fF3e15B9B91D2B9EdDeF484033) the pool creation had actually been successful. I'll do some more tests

jamiehewitt15:
I've done a few more tests and wasn't able to recreate this issue so I'll assume it was a one-off

> I have tested this branch again and found a strange user experience that I haven't seen on main before. When creating a pool I got a toast error message saying "ERROR: Failed to create a new pool", however, when I refreshed the [page](https://market-gg4cu94r9-oceanprotocol.vercel.app/asset/did:op:0CBb54B0f33eE7fF3e15B9B91D2B9EdDeF484033) the pool creation had actually been successful. I'll do some more tests

jamiehewitt15 (approved):

PR #654, opened 2021-06-09
Feature/free price support consume compute

### LINKED ISSUE #606: Update consume/compute to account for free price
Update consume and compute components to be compatible with the free pricing schema

krisliew:
Fixes #606 

Changes proposed in this PR:

- Fetch free pricing asset
- Free pricing asset dispense 1 DT if `isDispensable()` return `true` during `buyDT()`
- Button name as `Get` instead of `Buy` for consume asset for free pricing
- Button name as `Order Compute Job` instead of `Buy Compute Job` for free pricing


mihaisc:
image

Algo should be `Free` as well not 0

krisliew:
> image
> 
> Algo should be `Free` as well not 0

Correct me if I am wrong, I am showing the algorithm list price as `free` now using a simple workaround `asset.price === '0'` because:
- Publisher cannot add the algorithm asset in the list until the price of the algorithm asset is created (So we can conclude that all `asset.price` equivalent `0` is free priced asset in the list)
- Thought I could get the price.type and pass it down as prop to the algorithm list (`src/components/molecules/FormFields/AssetSelection.tsx`)  from https://github.com/oceanprotocol/market/blob/e26ed0e81abe9f7fb1cb39fee38ee142f3a7b0dc/src/utils/aquarius.ts#L109 but I couldn't (we might have to change this function since we need call subgraph to get that `price` object? don't seem right here 😅)

mihaisc:
hmm, let me check it out, hope it's not something more complicated

mihaisc:
I think it's ok since we already do all the checks before. @bogdanfazakas @kremalicious do you see any issues with this?

bogdanfazakas:
> I think it's ok since we already do all the checks before. @bogdanfazakas @kremalicious do you see any issues with this?

I think it's ok even though it is a bit inconsistent since we display `free` everywhere else and just the price calculation we have that `0` value, but trying to add free also there could be tricky and have a lot of use-cases since you can combine assets with price and free priced assets.

mihaisc:
the screenshot is old, https://market-hmkvspqcc-oceanprotocol.vercel.app/asset/did:op:5742F3ffa8A354b0483c97C8dF5f805e386D1B53

image

mihaisc (approved):

PR #655, opened 2021-06-09
Edit Advance Settings form for Edit Credentials & edit isOrderDisabled Flag

### LINKED ISSUE #638: Edit credentials
In asset details have a new button "Manage" or  "Edit Advanced Settings" ( other naming suggestions appreciated) . If this section will be displayed or not will be controlled by a env variable.

image

In this section we need to add/update/delete `allowList` and `denyList` defined here https://github.com/oceanprotocol/ocean.js/pull/787

In this section we will also edit  `isOrderDisabled` (not part of this issue, just for context).

krisliew:
Fixes #638

Changes proposed in this PR:

- New env variable `GATSBY_ALLOW_ADVANCE_SETTINGS` to show/hide the `Edit Advance Settings` button
- New env variable `GATSBY_CREDENTIAL_TYPE` to define credential type
- `Edit Advance Settings` consist of a form that allow user to 
  - update allow/deny list
  - update `isOrderDisabled` flag (Partial work done for #463)
  - sample screenshot 
09103E93-4F33-4506-8B53-D8993DCF3501


mihaisc (approved):

src/components/molecules/FormFields/Credential/Credential.module.css
```diff
@@ -0,0 +1,62 @@
+.chip {
+  border: 1px solid var(--border-color);
+  display: flex;
+  margin-top: 5px;
+  margin-bottom: 5px;
+}
+
+.input {
```
kremalicious, line 8: why do you duplicate all our input styles? Either use the `InputElement` component, or `compose`  the styles from the CSS module

kremalicious:
Without any testing so far, let's start with the basics, the word is called "advanced". So across all code changes in this PR where it only says "advance" it needs to be updated to say "advanced". Cause "advance" is a completely different word than "advanced".

If we say an ETH address is a credential, then a component collecting multiple credentials should also use plural. Otherwise it's just not clear if it only deals with one or multiple things.

Then just when looking at the UI it is rather terrible and needs an actual design and cleanup to stay consistent with the rest of the app. Like a top title and a description of what that screen does, proper spacing and so on. But I guess this could be neglected as we don't plan on making too many people ever see this.

All interface copy is a bit confusing too (what's a "credential"?). If we ask for wallet addresses it should just say that not just in some dimmed copy. I would suggest "Add ETH Address" or "Add Account". And then use "Remove... " for the deny part.

Same confusing and error copy on disable consume part. No user knows what consume as it's nowhere to be found in rest of app. So be specific that is disables download or compute.

And finally this PR introduces quite some new tech debt in form of duplication and complexity so would be preferable to clean this up as much as possible to keep our maintenance score stable on CodeClimate

jamiehewitt15:
I just did a test run through and it allowed me to buy the data token despite my address being on the deny list 
allow deny list 2021-06-11_13-02

jamiehewitt15:
Despite being on the "deny list" the download button is still active and there isn't any message saying I'm not allowed to download this asset... 

download data 2021-06-11_13-20

kianyee:
isConsumable issue is closed on ocean.js but yet to implement in market, this issue is more focus on to create an Advanced Settings to allow user update allow and deny list in DDO

PR #667, opened 2021-06-10
Adapt Aquarius queries to chainId

### LINKED ISSUE #650: Adapt all Aquarius queries based on passed chainIds
On frontpage & search template all search queries need to be moved into a method which accepts `chainIds` as param and based on those, the query needs to adapt and filter based on `ddo.chainid`. The goal is that whenever the list of `chainIds` changes, a new search query needs to happen.

The final source of truth in the app for the list of chain ids is `providers/UserPreferences` which exposes the selected chains.

KatunaNorbert:
Fixes #650 .

Changes proposed in this PR:

- Adapt all Aquarius queries based on passed chainIds


KatunaNorbert:
I guess this needs to wait until we get the `chainId` field on the `ddo`

KatunaNorbert:
There are some issue on the aquarius so there are some assets without chainId, but I think we can move on with reviewing and merging this PR if everything is fine from the market side because there are other PR's waiting for this changes. 
Also keep in mind while testing that in this PR the subgraph doesn't support multinetwork, so where we have subgraph and aquarius combinations might work just across main network due to subgraph.

claudiaHash:
Now that the bookmarks bug is fixed I think is good to go. For now, looks good to me, although using these changes in history might reveal others bugs to be fixed.

claudiaHash (approved):

src/components/molecules/AssetTeaser.tsx
```diff
@@ -11,7 +11,7 @@ import { useOcean } from '../../providers/Ocean'
 import styles from './AssetTeaser.module.css'
 
 declare type AssetTeaserProps = {
-  ddo: DDO
+  ddo: any
```
bogdanfazakas, line 14: you can stick to DDO since we added the chainId property in oceanjs

src/providers/Asset.tsx
```diff
@@ -14,14 +14,13 @@ import axios, { CancelToken } from 'axios'
...
 import { MetadataMarket } from '../@types/MetaData'
-import { DDO_TEMPORARY } from './Ocean'
 import { useWeb3 } from './Web3'
 import { useSiteMetadata } from '../hooks/useSiteMetadata'
 
 interface AssetProviderValue {
   isInPurgatory: boolean
   purgatoryData: PurgatoryData
-  ddo: DDO
+  ddo: any
```
bogdanfazakas, line 23: same here we can add back the DDO type, i think you can create a new PR with putting back the DDO this changes and merge it when we have a new version of oceanjs released

PR #672, opened 2021-06-10
Menu refactor

kremalicious:
Finding the perfect fluid layout flow with as less media query breakpoints as possible, and without switching components for "mobile" & "desktop" so that it stays flexible for adding more menu & action items in the future

Layout flow, not fully perfect as breaking is not fully based on content alone and we might have to tweak again when adding more menu items:

Untitled

Networks as its own action:

Screen Shot 2021-06-11 at 13 04 11

Highly tweaked wallet & menu actions components, the wallet component alone got lots of space-saving tweaks:

mobile view:
Screen Shot 2021-06-11 at 13 06 20

large viewport view:

Screen Shot 2021-06-11 at 13 06 55

when hovering over the network icon, name appears:

Screen Shot 2021-06-11 at 13 07 06


mihaisc (approved):
Awsome

PR #675, opened 2021-06-11
Change bookmarks and migration

### LINKED ISSUE #649: Bookmarks and migration
Bookmarks are part of `providers/UserPreferences` which also exposes `addBookmark` & `removeBookmark` methods.

Current data structure of bookmarks in `localStorage` as part of `ocean-user-preferences` item:

```json
"bookmarks": { 
  "mainnet" : ["did:op:a2B8b3aC4207CFCCbDe4Ac7fa40214fd00A2BA71"]
}
```

Where it's fine to simply remove the network reference and just flat out into an array of DIDs so we remove the complexity of getting the right network name in that place:

```json
"bookmarks": ["did:op:a2B8b3aC4207CFCCbDe4Ac7fa40214fd00A2BA71"]
```

Finally we should detect the old data structure in user's `localStorage` and if found, change it in `localStorage`. We did this for a former migration already which can now be replaced with a new migration.

### LINKED ISSUE #694: BSC integration

KatunaNorbert:
Fixes #649.

Changes proposed in this PR:

- Change bookmarks and migration to support multiple networks

### DIFF
--- a/src/components/molecules/Bookmarks.tsx
+++ b/src/components/molecules/Bookmarks.tsx
@@ -85,15 +85,13 @@ export default function Bookmarks(): ReactElement {
   const [pinned, setPinned] = useState<DDO[]>()
   const [isLoading, setIsLoading] = useState<boolean>()
 
-  const networkName = (config as ConfigHelperConfig)?.network
-
   useEffect(() => {
-    if (!appConfig.metadataCacheUri || !networkName || bookmarks === {}) return
+    if (!appConfig.metadataCacheUri || bookmarks === []) return
 
     const source = axios.CancelToken.source()
 
     async function init() {
-      if (!bookmarks[networkName]?.length) {
+      if (!bookmarks?.length) {
         setPinned([])
         return
       }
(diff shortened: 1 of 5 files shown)


kremalicious (approved):
amazing, works flawlessly. Tested migration by copying around existing `localStorage` structure from live market into preview version and all migration just worked. Meaning multiple network bookmarks were correctly combined into 1 array in `localStorage`. Right now only the mainnet assets will show up in bookmarks, but as soon as multi Aqua is there it should just work.

And with this PR we're down another typing error so we're almost at having missing `ddo.chainId` as last remaining type check error

KatunaNorbert:
So I guess this can be merged as it is and will fix the errors on the way

kremalicious:
yup, think we can only fix along the way and as long as Vercel deployment works all good cause that's doing a `npm run build` too. Currently trying to figure out how to mock `ddo.chainId` or something in `feature/multinetwork` so that we can get green tests again

KatunaNorbert:
Cool, that would be a good temporary solution and would enable us to keep going with the other issues where `ddo.chainId` is required on the frontend

PR #676, opened 2021-06-11
Create wallet network switcher

### LINKED ISSUE #657: User wallet network switcher
We need a component which solves multiple things:

- user needs to be informed when their wallet is connected to another network than the viewed asset so we can assume that component is only ever used as a child of `AssetProvider`
- we need a place to push our custom network configs like for polygon so users can easily switch

For that we can ~~steal~~ get inspired by Zapper:

Screen Shot 2021-06-09 at 12 57 40

We will then place this prominently on asset details page, most likely in the right column maybe even above everything else

KatunaNorbert:
Fixes #657 .

Changes proposed in this PR:

- Create a new component which enables user to switch the wallet network to the ddo's network

Screen Shot 2021-07-15 at 12 45 25

Screen Shot 2021-07-15 at 12 46 04


src/utils/web3.ts
```diff
@@ -19,6 +19,46 @@ export interface NetworkObject {
   urlList: string[]
 }
 
+const networks = [
```
kremalicious, line 22: during refactor I tried to unify all network metadata getting so ideally we'll get rid of this variable definition, or fill it with data from somewhere else. 

As for the name itself, `getNetworkDisplayName()` is your friend, making sure we control network name from one central place across whole app.

Then I created `useNetworkMetadata()` which exposes metadata about all networks.

And finally I'm still trying to figure out how to get all `ConfigHelperConfig` out of ocean.js into either this network metadata hook, or as part of `useOcean`. Which we then can use to iterate over as those are the networks the switcher should provide. Ultimately we also have to cross-reference it with `appConfig.chainIdsSupported`. So we could already create some loop to iterate over `appConfig.chainIdsSupported` and construct the data with `getOceanConfig()`

 So this then together should have all the metadata we need without hardcoding them here again

KatunaNorbert:
I'm trying to use out `addCustomNetwork` method to switch the user wallet network, but It looks like it is not working for default networks

kremalicious:
yes good point, there is (currently) no way to programmatically switch MetaMask's built-in networks, but they are all ETH test networks. So would be fine to only expose a switch button for mainnets in the network switcher and for everything else inform user to switch manually. Leaving us only with the switch from a non-ETH network to ETH mainnet. 

I mean, this is Zapper's workaround when I'm on Polygon and click their "Switch to Ethereum" button:

Screen Shot 2021-06-16 at 14 31 04

So if we want to show button at all times and assume it's going to work in future, we could go Zapper route and use our modal. But probably simpler to replace button with message for all ETH networks

KatunaNorbert:
I added the component to multiple places mention here #657 but not sure what would be the best place to pace it. In the `editComputeDate` or `editAdvancedSettings` might look better inside the form because the action button is over there, and also we don't have the preview component. Also the WalletNetworkSwitcher components needs some design advice.

I added the `Rinkeby` url for the `metadataCacheUri` so I can access edit asset pages, I don't have assets on mainnet

mihaisc:
So, first of all the message is wrong. `You are watching OCEAN on ETH Rinkeby but your wallet is connected to ETH` makes no sense in our context. That was just an example from zapper, where you were actually looking at the ocean token. A more accurate message would be `The current asset is on ETH Rinkby but your wallet is connected to ETH`.
image

I am a bit confused about the functionality, shouldn't be this in the web3 feedback component? Something like 
image

KatunaNorbert:
Right, makes more sense over there. I'll also fix the message

app.config.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   metadataCacheUri:
     process.env.METADATACACHE_URI ||
-    'https://aquarius.mainnet.oceanprotocol.com',
+    'https://aquarius.rinkeby.oceanprotocol.com',
```
kremalicious, line 4: should stay as it was, so fall back to mainnet when env var is not present

src/components/atoms/NetworkName.tsx
```diff
@@ -24,10 +24,12 @@ export function NetworkIcon({ name }: { name: string }): ReactElement {
 export default function NetworkName({
   networkId,
   minimal,
+  textOnly,
```
kremalicious, line 27: not sure about the need for this, it might be simpler to access `getNetworkDisplayName()` directly within `WalletNetworkSwither.tsx` if you just want to get the plain text

src/components/molecules/WalletNetworkSwither.tsx
```diff
@@ -0,0 +1,67 @@
+import React, { ReactElement, useState, useEffect } from 'react'
```
kremalicious, line 1: small typo in filename `WalletNetworkSwither.tsx`, missing a `c`: `WalletNetworkSwitcher.tsx`

src/components/molecules/WalletNetworkSwither.tsx
```diff
@@ -0,0 +1,67 @@
...
+import useNetworkMetadata from '../../hooks/useNetworkMetadata'
+import NetworkName from '../atoms/NetworkName'
+import { getOceanConfig } from '../../utils/ocean'
+import { useAsset } from '../../providers/Asset'
+
+export default function WalletNetworkSwitcher(): ReactElement {
+  const { networkId, web3Provider } = useWeb3()
+  const { networksList } = useNetworkMetadata()
+  const { ddo } = useAsset()
+  const DEFAULT_ETH_CHAIN_IDS = [1, 3, 4]
```
kremalicious, line 15: they should come from `appConfig.chainIds` which hold the default selected networks

```
```

KatunaNorbert: I think the name is confusing. There I wanted to keep track of the built in Metamask network to know when to show the switch button

kremalicious: ah right, now I get it, thanks! In other news, we might not have to do that anymore, as https://github.com/MetaMask/metamask-extension/pull/11184 seems finally be shipped in latest MetaMask version. I'll test a bit with other dapps and will report back

KatunaNorbert: Oh good news, looking forward for that

kremalicious: so the new Balancer v2 frontend does this where using their network switcher also switches wallet network and it works for ETH Mainnet:

Screen Shot 2021-07-12 at 16 55 08

And MetaMask docs now have this handy code snippet showing how to combine `wallet_addEthereumChain` with `wallet_switchEthereumChain` so we do not have to maintain a list of builtin networks:

https://docs.metamask.io/guide/rpc-api.html#usage-with-wallet-switchethereumchain

src/components/atoms/NetworkName.tsx
KatunaNorbert: right, wasn't sure if we need the icon too but if we stick witch the name only the function should do it

src/components/molecules/WalletNetworkSwither.tsx
KatunaNorbert: Cool, it's working :)

kremalicious (approved):
this works like a charm, great job! Think we found all the places where connection to asset network is important but we might find more during testing. Like, was not sure how adding datatoken as custom token behaves across networks with MetaMask so also based this on `isAssetNetwork`.

And for future we need to check other wallets and how network switching behaves with them as we currently show message and action to every wallet but a quick fix then is always to hide the switch button for certain wallets.

KatunaNorbert:
Also was thinking to add something to the Publish page, to make it clear for the user to what network is he publishing to, while having multiple networks selected. Also might be publishing to other network than the selected ones if the wallet is connected to different network. But this component doesn't really fit over there because you don't have an asset selected.

kremalicious:
yup! Had same idea and tackled this already as part of https://github.com/oceanprotocol/market/pull/729 where the publish page gets this network badge at the top

KatunaNorbert:
Oh, nice!

PR #677, opened 2021-06-14
updateMetadata() v3 workaround solution for free pricing

### LINKED ISSUE #674: Free pricing editMetaData workaround
editMetadata will not work if free price dispenser is created for the data asset

image
Etherscan error: https://rinkeby.etherscan.io/tx/0xd400f16164ba33cf9a6fc3de94b36c337deb2d2d08e36d272841d8e438cb85a5

Workaround solution is to set the minter back to the publisher --> editMetadata() --> set the dispenser back as the minter

1. cancelMinter() - (propose/accept) 
2. editMetadata() 
3. makeMinter() - (propose/accept)

krisliew:
Fixes #674 

Changes proposed in this PR:

`Edit Metadata` or `Edit Compute Settings` for `Free Pricing` Asset flow:
1. 2 transaction: propose publisher address as the new minter and to accept the proposed.
2. 1 transaction: `ocean.assets.updateMetadata` (no changes for here)
3. 2 transaction: propose dispenser address as new minter and to accept the proposed.

This will be a workaround solution until V4 is released where we will have publisher and minter role. Hence, publisher may renounce minter role for the dispenser free pricing feature while retaining publisher role for `Edit Metadata` or `Edit Compute Settings`.


jamiehewitt15 (approved):

jamiehewitt15:
I did a test run through and it works ok now

PR #686, opened 2021-06-16
Adapt history data fetching for multi network

### LINKED ISSUE #656: History page
All queries need to change based on `chainIds` user preference so we need e.g. multiple subgraph, node and maybe provider URIs on that page.

The exact place where to get the ocean.js `ConfigHelperConfig[]` values within the app is still pending, see [remarks here](https://github.com/oceanprotocol/market/pull/628#issuecomment-857575391).

claudiaHash:
Fixes #656 .

Changes proposed in this PR:

- adapt history tabs queries


src/components/molecules/PoolTransactions.tsx
```diff
@@ -61,28 +79,35 @@ const txHistoryQuery = gql`
...
   row: TransactionHistoryPoolTransactions,
   locale: string
 ) {
-  let title = ''
+  const source = axios.CancelToken.source()
+
+  const did = web3.utils
+    .toChecksumAddress(row.poolAddress.datatokenAddress)
+    .replace('0x', 'did:op:')
+  const ddo = await retrieveDDO(did, source.token)
```
KatunaNorbert, line 102: We don't need to retrieve the `ddo` anymore if we get the token symbol from the subgraph

src/components/molecules/PoolTransactions.tsx
```diff
@@ -131,19 +158,18 @@ async function getTitle(
...
 function Title({ row }: { row: TransactionHistoryPoolTransactions }) {
   const { networkId } = useWeb3()
-  const { ocean } = useOcean()
   const [title, setTitle] = useState<string>()
   const { locale } = useUserPreferences()
 
   useEffect(() => {
-    if (!ocean || !locale || !row) return
-
+    const config = getOceanConfig(4)
```
KatunaNorbert, line 165: do we still need the `config`?

src/components/pages/History/ComputeJobs/index.tsx
```diff
@@ -107,56 +117,76 @@ async function getAssetMetadata(
...
-  const { chainIds } = useUserPreferences()
-  const [result] = useQuery<ComputeOrders>({
-    query: getComputeOrders,
-    variables: {
-      user: accountId?.toLowerCase()
+
+  useEffect(() => {
+    async function initOcean() {
+      const oceanInitialConfig = getOceanConfig(networkId)
+      await connect(oceanInitialConfig)
```
KatunaNorbert, line 133: Doesn't this make the table to only show the compute jobs history for the wallet network?

src/components/molecules/PoolTransactions.tsx
claudiaHash: Busted :) Thanks, I removed it .

src/components/molecules/PoolTransactions.tsx
claudiaHash: True, I removed it from the "Setup" case so we really don't need it anymore.

src/components/pages/History/ComputeJobs/index.tsx
claudiaHash: That's a compromise @mihaisc suggested so we can have a functional and showable `ComputeJobs` tab.

KatunaNorbert: Ok, I get it

mihaisc: It actually doesn't, there is an issue with the architecture in ocean.js. I'll fix it later, for now it works like this.
history

src/utils/subgraph.ts
```diff
@@ -155,6 +155,30 @@ async function fetchData(
...
 }
 
+export async function fetchDataForMultipleChains(
+  query: TypedDocumentNode,
+  variables: any,
+  chainIds: number[]
+): Promise<any[]> {
+  let datas: any[] = []
+  for (const chainId of chainIds) {
+    console.log(chainId)
```
mihaisc, line 165: remove console.logs

KatunaNorbert:
When I don't have a wallet connected, the compute jobs history table remains stuck in this loading state:

Screenshot 2021-07-13 at 15 28 38

It should switch to a message like the other tabs

claudiaHash:
I'll check it right now

claudiaHash:
In case of no wallet connected, we show 3 different types of messages: 

Screenshot from 2021-07-14 12-55-15
Screenshot from 2021-07-14 12-55-21
Screenshot from 2021-07-14 12-55-27

I believe it would be more consistent if we show up a _"Connect your wallet to see your ..."_ in each tab, in case of no wallet connected and _"No results found"_  or _"Your pool transactions will show up here"_ in case of existing wallet connection but no data found. What do you guys think?

KatunaNorbert:
I agree, also noticed the different messages but not sure if we should do it in this PR

mihaisc:
Agree we should unify the messages. If wallet not connected : `Please connect your Web3 wallet.` , if wallet connected but no results : `No results found` . No need to customize message for each tab.

claudiaHash:
> Agree we should unify the messages. If wallet not connected : `Please connect your Web3 wallet.` , if wallet connected but no results : `No results found` . No need to customize messages for each tab.

Can I address it in this PR or should I open a new issue?

mihaisc:
Do it here, it's not that big of change

kremalicious:
works like a charm but when testing this now all together it felt rather unintuitive as the connection between network selection and table content is not so clear on this screen. What would help with that is to have a column with the `<NetworkName />` component for each row in all tables.

For the _Pool Transactions_ we need to also solve the `explorerUri` for each row, currently none of the explorer links are working for me.

For _Compute Jobs_ we need to figure out something when getting the results. @KatunaNorbert I think I found another place for #676 which is this screen which now would not be based on an asset, so extra fun:

Screen Shot 2021-07-15 at 13 45 19

Or does wallet network not matter here as it is only a wallet signature and no tx involved?

mihaisc:
> Or does wallet network not matter here as it is only a wallet signature and no tx involved?

Working ok from me, please check what errors do you get . Or am i misunderstanding the issue?
cpob

bogdanfazakas:
saw that the table order by time is not respected anymore (see gif), not sure but i think the order by time should still be respected
history

claudiaHash:
Yeah, good point, we need to reorder data by timestamp before displaying it in table. Data is taken network by network from each corresponding subgraph, hence the network order.

KatunaNorbert:
The `explorerUri` from `ExplorerLink` component for Pool Transactions still not working. There is still the wallet network id passed to the `ExplorerLink`

claudiaHash:
Good catch! Should be fine now.

KatunaNorbert (approved):
Everything seems to work fine now

src/components/pages/History/PoolShares.tsx
```diff
@@ -141,38 +155,70 @@ const columns = [
...
+        const result = await fetchDataForMultipleChains(
+          poolSharesQuery,
+          variables,
+          chainIds
+        )
+        for (let i = 0; i < result.length; i++) {
+          result[i].poolShares.forEach((poolShare: PoolShare) => {
+            data.push(poolShare)
+          })
+        }
```
bogdanfazakas, line 180: same as for the pool transactions with the refetch i think can be applied here

kremalicious:
> Working ok from me, please check what errors do you get . Or am i misunderstanding the issue?

nope, all good, was just unsure if network influences signature behavior in any way but wallet signatures are indeed independent of network

@claudiaHash sorry about the merge conflict, I merged @bogdanfazakas's PR already but see you both are already coordinating regarding refetch approach in here. That should be then really last thing we have to do in here before merge

claudiaHash:
No problem. Already talked about it with @bogdanfazakas and I'm working on it so this could be merged asap.

bogdanfazakas (approved):
look good to me now 🎉

Issue #665, opened 2021-06-10
Global search bar


mihaisc:
Move the search bar component in the header with a minimal design. The functionality stays the same.

Example https://codepen.io/912lab/pen/LsplC
![Animation](https://user-images.githubusercontent.com/26000280/121493716-c3d05600-c9e0-11eb-97b8-21f03bdb140e.gif)

kremalicious:
done in #690

PR #690, opened 2021-06-17
Global search bar

mihaisc:
Handles https://github.com/oceanprotocol/market/issues/665
Highly WIP. Functionality is sort of there.

What search pattern should we follow?

A. Easier way, like amazon 
image
image

B. More design work (on mobile)
image
image
image

Bonus feature : search should trigger on 'enter'


kremalicious:
B would be so nice but that's a lot to refactor I guess. So from behavior standpoint would go with A, but visually combine the two where the search field starts collapsed as search icon button only, and as viewport gets bigger search field can become more visible like Rarible does. On mobile the search field upon interaction can still take over the whole bar on top of other elements, but then searching still only happens when user hits enter.

But then we need to make sure when user arrives on search page that searchbar is open by default

mihaisc:
I think you managed to design a more complicated way 😆. Ok i'll try to get in in a place where just UI changes are needed

mihaisc:
draft 1 fullscreen
image

mihaisc:
Some more work needed on the button and behaviour
search1

mihaisc:
Needs more testing.

mihaisc:
There are bugs when searching by owner or tags (when you click on owner or tag in asset details) that will be fixed once we merge main in the multinetwork branch.

jamiehewitt15:
The styling is good. Only one small thing I would suggest is there should be more padding on the right of the History link. 

So it changes from this: 
history padding 0px 2021-06-30_16-50

To this:
history padding 25px 2021-06-30_16-50

jamiehewitt15:
It terms of the search functionality, I'm noticing a couple issues which I thought we had solved in: https://github.com/oceanprotocol/market/pull/641

- The results don't include a partial match. For example, if you search "data union" you'll find that "dataunion.app" doesn't appear in the results
- Searching with a trailing space effects the search results.

mihaisc:
@jamiehewitt15 search fixed were not merged, so basically this is the old search with multiple bugs. Those will be merged in the multinetwork branch.
I'll add some padding where you suggested

mihaisc:
Main is now merged.

> The results don't include a partial match. For example, if you search "data union" you'll find that "dataunion.app" doesn't appear in the results

This was not fixed, requires more work, it's a bit more complicated.

src/components/molecules/Menu.module.css
```diff
@@ -9,36 +9,89 @@
...
 
 .navigation {
-  order: 3;
-  width: 100%;
+  width: auto;
+  margin: 0;
+  text-align: left;
+  border: none;
+}
+/* .navigation {
```
jamiehewitt15, line 24: Is there any reason this is commented out? It would be cleaner to remove it if it's not needed

src/components/molecules/SearchBar.module.css
```diff
@@ -1,17 +1,70 @@
...
+.input {
+  height: 36px !important;
+  border-top-right-radius: 0px;
+  border-bottom-right-radius: 0px;
+}
+.searchInput {
+  flex-grow: 2;
+  margin-bottom: 0px;
+}
+/* @media screen and (min-width: 600px) {
```
jamiehewitt15, line 34: Not sure why styling has been added an commented out in this file? Might be better just to remove it

jamiehewitt15 (commented):
The functionality and styling look good to me. I just added a couple of minor comments on the code

mihaisc:
Removed comments, should be ok now

jamiehewitt15 (approved):

jamiehewitt15:
Looks good, I have approved

PR #693, opened 2021-06-22
Adapt price fetching for multi network using asset chainId and replace apollo with urql

### LINKED ISSUE #658: Price fetching needs to adapt to respective asset chainId
Right now no price display works when user is not connected to same network as the asset which is especially noticeable on frontpage. This needs to be refactored so that price fetching is always based on some passed `chainId` most likely from the DDO.

This pretty much requires some way to switch Apollo client endpoints based on `ddo.chainId`

bogdanfazakas:
Fixes #658 .

Changes proposed in this PR:

- replaces apollo client with urql graphQl client
- wip adapt price fetching to multi network


src/providers/UrqlProvider.tsx
```diff
@@ -0,0 +1,46 @@
...
+        'No subgraphUri defined, preventing UrqlProvider from initialization.'
+      )
+      return
+    }
+
+    const newClient = createUrqlClient(oceanConfig.subgraphUri)
+    urqlClient = newClient
+    setClient(newClient)
+    Logger.log(`[apollo] Client connected to ${oceanConfig.subgraphUri}`)
+  }, [networkId])
```
KatunaNorbert, line 42: Switching the `subgraphUri` based on `networkId` is no longer because we can have assets from different networks than the connected wallet.  When retrieving data for an asset It could be triggered by `ddo` change, but in other cases we might need to hit multiple subgraphs. So I don't know how reliable is to keep a client in provider, might be better to create a function which accepts a list of subgraphUri's and create the client when calling the function.

bogdanfazakas: yep this will be removed, it was left here just for testing reasons for the urql library with one of our subgraphs

kremalicious: the question how to get the `subgraphUri`s we need is also open question still, as getting them one-by-one with `getOceanConfig()` might be not really feasible anymore. For that reason ocean.js now exports the whole network metadata definition at once in case we need to iterate over that data: https://github.com/oceanprotocol/ocean.js/blob/main/src/utils/ConfigHelper.ts#L34

As for keeping the client in some provider here, we need to test to see. It depends on how quick the client can be initiated as to optimize for the most common user flow which is a user browsing assets on mainnets. So maybe makes sense keeping some connection in some sort of cache so browsing feels faster for user. But maybe it doesn't really matter as connection is quick anyway, then as Norbert suggests the utility method could be enough

kremalicious:
amazing, this already works super well for the original issue of solving e.g. asset teaser price fetching independent of user network. As this is also switching libraries we should now remove all Apollo stuff, like the provider, typings generation on build and such as it looks like we do not need to bother with urql and typings.

Then, if we want to already solve the pattern for fetching multiple subgraphs we could use `MarketStats` for this which I just added back. There we could either output a statistic line for each mainnet individually, or add them all together in one line. In both cases we would need to fetch multiple subgraphs from the same component. But think would be also fine to handle this in a next issue or in #686

KatunaNorbert:
The prices looks good, but the highest liquidity section doesn't seem right. It is not ordered correctly and some assets are different from what we have on main. But we could handled that in other PR if there are PRs waiting for these one to be merged

KatunaNorbert (approved):

bogdanfazakas:
> The prices looks good, but the highest liquidity section doesn't seem right. It is not ordered correctly and some assets are different from what we have on main. But we could handled that in other PR if there are PRs waiting for these one to be merged

Yep we need to discuss how the highest liquidity should properly work in the multi network context and fix that in another PR

PR #698, opened 2021-06-23
Create account teaser

### LINKED ISSUE #669: Account teaser
The account teaser will be a component similar to asset teaser.

It will have: 
- image : default will be the one from the wallet or fetch the one from 3box profile 
image
- address 
- name if there is one from 3box profile

The teaser will be used on the front page in the best sellers section and in search. Click event will redirect to `/account/0xxxxx`
It is dependent on https://github.com/oceanprotocol/market/issues/668 but can be done before if needed.

image (2)

We can have two versions of the teaser:
- small : just image, name, address (first image on the first row)
- large : add stats row (first image on second row, this will depend if the performance to fetch the stats is good enough)

claudiaHash:
Fixes #669 .

Changes proposed in this PR:

- created the `AccountTeaser` component


kremalicious:
closing this as it makes more sense to continue on #864 where the teaser is put into some actual context

PR #700, opened 2021-06-25
Fix refetch data token orders

### LINKED ISSUE #683: "Compute Jobs" tab in history page not fetching latest data
"Compute Jobs" tab in history page not fetching latest data, when switching between tab.

1. Open two browser tab, compute asset & history page.
2. Tab 1: At history page, Go to "Compute Jobs" tab first.
3. Tab 2: Run a compute job 
4. Tab 1: Switch between tab, when switch back to "Compute Jobs" tab, it wont display the latest compute job. 
4. Refresh the browser, then only it will fetch all the compute job (including the latest). 

This might happened because it will only trigger a subgraph query search once on "Compute Jobs" tab. 

Expected Result: 
When switching between tab to "compute jobs" at history page, expect to fetch latest data, instead of refreshing the browser to get the latest data. 

image

KatunaNorbert:
Fixes #683 .

Changes proposed in this PR:

- Refetch the token orders when getJobs function is called

### DIFF
--- a/src/components/pages/History/ComputeJobs/index.tsx
+++ b/src/components/pages/History/ComputeJobs/index.tsx
@@ -109,14 +109,12 @@ export default function ComputeJobs(): ReactElement {
     }
   })
 
-  console.log(data)
-
   async function getJobs() {
     if (!ocean || !account) return
 
     setIsLoading(true)
 
-    // await refetch()
+    await refetch()
     const dtList = []
     const computeJobs: ComputeJobMetaData[] = []
     for (let i = 0; i < data.tokenOrders.length; i++) {
--- a/src/components/pages/History/ComputeJobs/index.tsx
+++ b/src/components/pages/History/ComputeJobs/index.tsx
@@ -60,7 +60,11 @@ const columns = [
   {
     name: 'Finished',
     selector: function getTimeRow(row: ComputeJobMetaData) {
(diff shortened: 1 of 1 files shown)


KatunaNorbert:
To properly test this follow this steps:

- open 2 browser tabs one with compute jobs history tab open and one for starting the compute job
- start a compute job on a dataset that you haven't already run compute jobs on
- after starting the compute job go to the browser tab with compute jobs history and click the refresh button or switch to another tab from history page tabs(ex: Downloads) and back to the compute jobs tab
- you should see your compute job inside the compute jobs table

claudiaHash:
Tested it the way you suggested and in 2 cases it behaved strangely: first time when accessing the compute history browser tab or accessing the compute tab after switching tabs in history, after first `Refresh`, got this: 

Screenshot from 2021-06-25 11-25-36
Screenshot from 2021-06-25 11-32-34

It seems to copy the finished time from the last table asset. Just tested it on live and didn't encountered that there, so it might be related to this refresh approach. When accessing `Refresh` second time, table data was displayed correctly.

KatunaNorbert:
Interesting, I'll take a look

claudiaHash:
Seems to work fine now.

claudiaHash (approved):

PR #707, opened 2021-07-05
Add custom provider to publish form

### LINKED ISSUE #702: Add custom provider endpoint during publish
The user needs to have the option to set a custom provider url and not use the default OPF one. We can use the docker component (or similar) that we use in algorithm.
When adding a new url we need to validate that with ocean js, something like :
```
 const instance = new Provider()
 await instance.setBaseUrl(config.config.providerUri)  
// this is true or false if provider url is valid , eventualy we can check the version as well, instance.providerVersion
```

image

jamiehewitt15:
Fixes #702

Changes proposed in this PR:

- Adding UI component for inputting a custom provider in the publish form
- Validating the custom provider using `ocean.provider.isValidProvider(providerUrl)`


src/models/FormPublish.ts
```diff
@@ -43,6 +46,8 @@ export const initialValues: Partial<MetadataPublishFormDataset> = {
   description: '',
   timeout: 'Forever',
   access: '',
+  provider: '',
   termsAndConditions: false,
-  tags: ''
+  tags: '',
+  customProvider: ''
```
kremalicious, line 52: who do we collect a `provider` & a `customProvider` here? This is one field and we should capture different values for it, not every value in its own key?

src/components/molecules/FormFields/CustomProvider.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const [providerUrl, setProviderUrl] = useState<string>()
+  const { config } = useOcean()
+
+  function loadProvider() {
+    if (!providerUrl) return
+    async function validateProvider() {
+      let valid: boolean
+      try {
+        setIsLoading(true)
+        const ocean: Ocean = await Ocean.getInstance(config)
```
kremalicious, line 21: `ocean` is always there:

```js
const { ocean } = useOcean()
```

kremalicious:
Consistency and visual hierarchy:

Screen Shot 2021-07-16 at 09 41 14

- when custom is selected it looks like 2 fields on the same hierarchy but the custom URL input is a subset of the provider form field, so should also look like it. Making the input label smaller might already be enough.
- this file/URL field works different than all other file fields we use in the same form but looks the same as them. Elements looking the same should work the same 

And lots of copy issues. We would need to link people to a place where provider is explained, has multiple typos, and the example placeholder should not be some `localhost` URL.

And finally, why do all those changes only apply to data sets, but not algorithms?

src/components/molecules/FormFields/URLInput/Input.tsx
```diff
@@ -30,7 +32,7 @@ export default function FileInput({
         onClick={(e: React.SyntheticEvent) => e.preventDefault()}
         disabled={!field.value}
       >
-        {isLoading ? <Loader /> : 'Add File'}
+        {isLoading ? <Loader /> : `${submitText}`}
```
kremalicious, line 35: too much template literals wrapping, this should be enough if submitText is really always a string:

```suggestion
        {isLoading ? <Loader /> : submitText}
```

src/components/pages/Publish/index.tsx
```diff
@@ -133,14 +133,16 @@ export default function PublishPage({
...
+        values.provider === 'Custom' ? values.customProvider : null
       )
 
       const ddo = await publish(
         metadata as unknown as Metadata,
         serviceType,
         values.dataTokenOptions,
-        timeout
+        timeout,
+        values.provider === 'Custom' ? values.customProvider : null
```
kremalicious, line 145: right, now I get the capturing of 2 values to differentiate between default & custom selection. Maybe it would reduce some logic in some places if we only rely on an optional `customProvider` key which we could switch between a string and `null` depending on user selection which would reduce need to have 2 different keys. But in end also no big deal as we do not send this data around in end anyway

jamiehewitt15: I think you're right, it is better with just one field `customProvider`, I don;t think there is any need for the provider box selector when this is hidden behind advanced settings. I have removed the provider field

src/components/molecules/FormFields/URLInput/Input.tsx
jamiehewitt15: Thanks for the suggestion, this is fixed now

src/components/molecules/FormFields/CustomProvider.tsx
jamiehewitt15: Thanks for pointing that out. I've changed it to `const { ocean } = useOcean()`

src/models/FormPublish.ts
jamiehewitt15: I've now reduced it to just one field

jamiehewitt15:
@kremalicious Thanks a lot for your feedback. I've made the following changes: 

1. The custom provider input is now hidden behind an advanced settings section in the publish form.  
2. Advanced publishing settings can now be turned on/ off with an env var `GATSBY_ALLOW_ADVANCED_PUBLISH_SETTINGS`
3. Removing the separate `provider` field. Now there is just the custom provider URL input field and the publisher can leave it blank to use the default.   
4. The advanced publishing settings are available in both the dataset and algorithm publishing forms.
5. Adding a "learn more" link to the provider github in the helper text. 
6. Removing the template literals.
7. Changing the placeholder text for the custom provider field. 
8. Refactoring to reduce code duplication.

bogdanfazakas:
I think i found a small issue over here, when adding a custom provider on the publish dataset, the field value is not cleared when selecting publish algorithm tab like we do with all the other input fields.
publish-bug

jamiehewitt15:
@bogdanfazakas thank you for pointing that out. It should be working ok now

bogdanfazakas (approved):
Looks good to me! The PR needs a rebase and there are some merge conflicts that need some attention

src/@types/MetaData.d.ts
```diff
@@ -38,6 +38,7 @@ export interface MetadataPublishFormDataset {
   // ---- optional fields ----
   tags?: string
   links?: string | EditableMetadataLinks[]
+  customProvider?: string
```
kremalicious, line 41: let's be consistent across projects with variables and name this variable `providerUri` everywhere instead of `customProvider` (the component name is fine though)

src/components/molecules/FormFields/CustomProvider.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { toast } from 'react-toastify'
+import CustomInput from './URLInput/Input'
+import { useOcean } from '../../../providers/Ocean'
+import { InputProps } from '../../atoms/Input'
+
+export default function CustomProvider(props: InputProps): ReactElement {
+  const [field, meta, helpers] = useField(props.name)
+  const [isLoading, setIsLoading] = useState(false)
+  const [providerUrl, setProviderUrl] = useState<string>()
+  const { ocean, config } = useOcean()
```
kremalicious, line 12: this gets more tricky as I found after my Friday publishing investigation. Basically `ocean` is only initiated automatically in the asset details view so this `ocean` here will be `null`, just as `config`. While the `OceanProvider` now also wraps the publish form I did not add auto-connection for it here and initiate it only within the `usePublish` hook, so only after user hits _Submit_: https://github.com/oceanprotocol/market/blob/main/src/hooks/usePublish.ts#L47-L58

Now, if we require this in more components on publish form as this looks like here, it would make sense to initiate `ocean` in some higher levels like in `src/components/pages/Publish/index.tsx`.

(Note that this all compares to latest `main` so be sure to merge/rebase before continuing here)

src/components/molecules/FormFields/advancedSettings.tsx
```diff
@@ -0,0 +1,53 @@
+import React, { ReactElement, useState, FormEvent, ChangeEvent } from 'react'
+import Input from '../../atoms/Input'
+import Button from '../../atoms/Button'
+import { FormContent, FormFieldProps } from '../../../@types/Form'
+import { Field } from 'formik'
+import appConfig from '../../../../app.config'
+import styles from './advancedSettings.module.css'
```
kremalicious, line 7: consistency, our React component names and their file names are in `PascalCase` so this component should be `AdvancedSettings.tsx` & `AdvancedSettings.module.css`

src/components/molecules/FormFields/advancedSettings.tsx
```diff
@@ -0,0 +1,53 @@
+import React, { ReactElement, useState, FormEvent, ChangeEvent } from 'react'
+import Input from '../../atoms/Input'
+import Button from '../../atoms/Button'
+import { FormContent, FormFieldProps } from '../../../@types/Form'
+import { Field } from 'formik'
+import appConfig from '../../../../app.config'
```
kremalicious, line 6: always get this within components like so:

```js
```

That's a Gatsby-only thing as we import this into Gatsby's GraphQL layer already on build, so making sure we do not have duplicate json in our final js bundle

kremalicious (changes requested):
merge `main` and address above code comments

src/components/molecules/FormFields/CustomProvider.tsx
jamiehewitt15: I'm not sure about this, I did a console.log of ocean here: https://github.com/oceanprotocol/market/blob/cbdb60edb8b87b38dc635182e0f782a0ecd0690d/src/components/molecules/FormFields/CustomProvider.tsx#L13

And it is defined: 
ocean defined custom provider

src/@types/MetaData.d.ts
jamiehewitt15: This has been done

src/components/molecules/FormFields/advancedSettings.tsx
jamiehewitt15: This has been done

src/components/molecules/FormFields/advancedSettings.tsx
jamiehewitt15: Thanks for pointing this out, I've updated it

src/components/molecules/FormFields/CustomProvider.tsx
jamiehewitt15: It seems to work as it is, let me know if I'm missing anything?

jamiehewitt15:
@kremalicious I have done another pull from main, let me know if you have further comments

src/components/molecules/FormFields/CustomProvider.tsx
kremalicious: can confirm, all good. This all works on publish page since #768 was merged

kremalicious (approved):

PR #708, opened 2021-07-05
Restrict AssetSelection list to only show same network assets

### LINKED ISSUE #651: AssetSelection needs to be restricted to only show same network assets
We can't run compute jobs across networks so simplest way to prevent users from running into issues is to limit the assets in `AssetSelection` always to the same network based on `ddo.chainid`

bogdanfazakas:
Fixes #651  .

Changes proposed in this PR:

- WIP added check for dataset chainId for each asset in asset selection list


src/utils/aquarius.ts
```diff
@@ -126,16 +126,19 @@ export async function getAssetsNames(
 export async function transformDDOToAssetSelection(
   datasetProviderEndpoint: string,
   ddoList: DDO[],
-  selectedAlgorithms?: PublisherTrustedAlgorithm[]
+  selectedAlgorithms?: PublisherTrustedAlgorithm[],
+  ddoChainId?: number
```
kremalicious, line 130: just checked, all works in the end but feels super ineffective as I assume we are first getting all algos of all networks and then filter them? So a big impact optimization would be to rather modify the Elasticsearch query to begin with which should be way faster in the end and we can ditch the js filtering. But just first thoughts, maybe I'm missing something why we do it like this

package-lock.json
```diff
@@ -1781,6 +1781,25 @@
       "resolved": "https://registry.npmjs.org/@emotion/unitless/-/unitless-0.7.5.tgz",
       "integrity": "sha512-OWORNpfjMsSSUBVrRBVGECkhWcULOAJz9ZW8uK9qgxD+87M7jHRcvh/A96XXNhXTLmKcoYSQtBEX7lHMO7YRwg=="
     },
+    "node_modules/@endemolshinegroup/cosmiconfig-typescript-loader": {
```
kremalicious, line 1784: I'm always so afraid of `package-lock.json` changes as for me it already feels like I'm constantly dealing with its merge conflicts which are a pain to resolve, and when a PR does not actually upgrade packages we should try to keep those changes out. So can we get this change somehow out of here, like copying over current `package-lock.json` from `feature/multinetwork`?

src/utils/aquarius.ts
bogdanfazakas: Yep you are right, I have thought about this also, I'll try to figure out a way of getting just the algos instead of getting them from all networks and filter them

package-lock.json
bogdanfazakas: Yup sure I'll copy it from feature `feature/multinetwork,` the lock file was updated when I've bumped the `oceanlib` version to have acces to the `DDO.chainId` but then it was bumped later also on the feature/multinetwork branch so not needed anymore

kremalicious (approved):
perfect, working nicely and list loads way faster for me after recent query changes.

PR #716, opened 2021-07-07
ocean js update

jamiehewitt15:
Closes: #718 

Changes proposed in this PR:
- Updating ocean.js to the latest version

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "@coingecko/cryptoformat": "^0.4.2",
         "@loadable/component": "^5.15.0",
         "@oceanprotocol/art": "^3.0.0",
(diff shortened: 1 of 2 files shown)


jamiehewitt15:
This needs more testing. I had to use `npm install --force` to get it to work locally and then it seemed to break again during the demo...

kremalicious:
yeah, running into that too sometime. Most of the time can be solved by using `--legacy-peer-deps` when installing something specific although `--force` might do something similar I guess:

```bash
npm install @oceanprotocol/lib --legacy-peer-deps
```

kremalicious:
feel free to close this as we require a new ocean.js version for #725 so will be done in there

PR #729, opened 2021-07-13
Multinetwork UI additions

### LINKED ISSUE #721: visual fixes
- searchbar fixes, maybe pick up patterns from #665 description
- network name on asset teaser, asset details
- ...

kremalicious:
Fixes #721

- modified menubar layout flow with less breakpoints
- add BSC to default selected chains
- minimal visual indicator for amount of selected chains on top of networks button
- searchbar visual tweaks
- network name fixes and contextual output: asset teaser, asset details, publish page
- beta → v3 badge, use the title badge for versioning rather than as stability indicator as that is handled with warnings already
- pragmatic mobile search, don't really have time for something more sophisticated

Screen Shot 2021-07-15 at 15 53 07

Screen Shot 2021-07-15 at 14 57 43

Screen Shot 2021-07-14 at 12 44 48

Screen Shot 2021-07-15 at 14 31 45


mihaisc:
Don't think that `v3` is a good idea. What is the point of it?

mihaisc:
There is an interval where the search box is a bit off
image

kremalicious:
> Don't think that v3 is a good idea. What is the point of it?

Thinking was: we say it's beta in the warning at the very top already. And realistically we only get out of beta with what's currently labeled as `v4`. It also opens UI possibities where this badge could be used for showing non-migrated assets between v3 & v4. In that sense we could of course leave as is and with the release of v4 just remove the badge

mihaisc:
On mobile the search button icon seems to not be full width 
image

mihaisc:
image
there is no more X to clear the text

mihaisc:
Should we add some space between `into` and the network 
image

PR #735, opened 2021-07-14
Show compute jobs history in asset detail page

### LINKED ISSUE #673: Show most recent compute job status in asset detail page
The message below will be gone when user try to refresh or re-enter the page again after compute job started.
User may tend to forget they have just started the job and they might trigger another job again
Screenshot 2021-06-11 at 12 03 32 PM

 
Show most recent compute job status in asset detail page

KatunaNorbert:
Fixes #673 .

Changes proposed in this PR:

- display most recent compute job status in asset detail page


claudiaHash:
When I access for the first time `Show` button on asset details page, I  get _'No assets found'_  few seconds after, and then I see the correct results. Besides that, it looks good too me.

mihaisc:
I have a minor UI quarrel but not sure how to address it. I have 8 pages that are identical , adding another column would be nice, like date, but i'm not sure if there is any room

image

KatunaNorbert:
I was trying that but than the 3rd column will not be visible, it could be done with some UI changes, maybe making the table columns thinner

mihaisc:
Leave it like this for now. What Claudia said happens to me as well

KatunaNorbert:
That's a bug inside the `ComputeJobs` component, I'm just using that without making any changes. You can see it inside the history page

KatunaNorbert:
But I could fix it here, shouldn't be a difficult one

KatunaNorbert:
Also appears on `Downloads` tab

mihaisc:
No, wait, this things were mostly fixed in multinetwork , wait for that to merge, then rebase and check after

KatunaNorbert:
Oh, right, waiting for multinetwork than

KatunaNorbert:
> When I access for the first time `Show` button on compute job details on asset details page, I get _No assets found_ and few seconds after, I see the correct results. Besides that, it looks good too me.

This is fixed now

KatunaNorbert:
That's what @mihaisc also suggested to add more info to the table, and I was trying that but when I add more column there isn't any scrolling or shrinking mechanism working. But I see there is on other table, so there might be something blocking that, I'll try to fix it.
I see there was another column added to compute jobs table that changed the selected column values.

KatunaNorbert:
I've fixed the scroll but the display is not to suggestive, you can't see the 3-rd column and scrollbar so the user might no know there is another column in the table and he can scroll:
Screenshot 2021-08-04 at 12 22 49

To fix it I was thinking of either making the columns width smaller so the user know there is another column and he can scroll to see the column:
Screenshot 2021-08-04 at 12 22 15
or 
Make all 3 column fit, which than affects the visibility of cells contents but the user doesn't have to scroll and he sees all the column:
  
Screenshot 2021-08-04 at 12 22 33

kremalicious:
[!](https://user-images.githubusercontent.com/37236152/128157243-b9afb5bd-2ddb-472d-94ae-a79a00cb3a60.png)

Yes to this, good signal to users that there is an overflow. And if we want to get more space we could also limit the action link to just _DETAILS_

kremalicious:
I also run into this small visual thing where all the jobs look the same as they have been started around same time. But might not have that big of an impact in real-life scenarios. Guess the only way would be to show a shortened `jobId` column or the used algo symbol but then we get even less space so think we can launch with current state

kremalicious (approved):

mihaisc (approved):

PR #736, opened 2021-07-14
Highest liquidity correct assets order

### LINKED ISSUE #727: Highest liquidity pool in the multi network context
Issue: The assets are not in the correct order 
Solution: Order the assets based on the liquidity

bogdanfazakas:
Fixes #727 .

Changes proposed in this PR:

- order assets in highest liquidity component correctly in multi network context

### DIFF
--- a/src/components/pages/Home.tsx
+++ b/src/components/pages/Home.tsx
@@ -27,7 +27,7 @@ async function getQueryHighest(
   // TODO: this query needs to adapt to chainIds
   const queryHighest = {
     page: 1,
-    offset: 9,
+    offset: dids.length,
     query: {
       query_string: {
         query: `(${dids}) AND (${transformChainIdsListToQuery(
--- a/src/components/pages/Home.tsx
+++ b/src/components/pages/Home.tsx
@@ -24,7 +24,6 @@ async function getQueryHighest(
   chainIds: number[]
 ): Promise<[SearchQuery, string]> {
   const dids = await getHighestLiquidityDIDs(chainIds)
-
   // TODO: this query needs to adapt to chainIds
   const queryHighest = {
     page: 1,
@@ -89,7 +88,7 @@ function SectionQueryResult({
       try {
         setLoading(true)
         const result = await queryMetadata(query, source.token)
-        if (queryData && result.totalResults > 0 && result.totalResults <= 15) {
+        if (queryData && result.totalResults > 0) {
           const searchDIDs = queryData.split(' ')
           const sortedAssets = sortElements(result.results, searchDIDs)
           // We take more assets than we need from the subgraph (to make sure
--- a/src/components/pages/Home.tsx
+++ b/src/components/pages/Home.tsx
@@ -91,9 +91,6 @@ function SectionQueryResult({
         if (queryData && result.totalResults > 0) {
           const searchDIDs = queryData.split(' ')
           const sortedAssets = sortElements(result.results, searchDIDs)
-          // We take more assets than we need from the subgraph (to make sure
-          // all the 9 assets with highest liquidity we need are in OceanDB)
-          // so we need to get rid of the surplus
           const overflow = sortedAssets.length - 9
           sortedAssets.splice(sortedAssets.length - overflow, overflow)
           result.results = sortedAssets
--- a/src/components/pages/Home.tsx
+++ b/src/components/pages/Home.tsx
@@ -24,7 +24,6 @@ async function getQueryHighest(
   chainIds: number[]
 ): Promise<[SearchQuery, string]> {
   const dids = await getHighestLiquidityDIDs(chainIds)
-  // TODO: this query needs to adapt to chainIds
   const queryHighest = {
     page: 1,
     offset: dids.length,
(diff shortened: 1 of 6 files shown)


KatunaNorbert (approved):
LGTM

src/utils/subgraph.ts
```diff
@@ -426,26 +430,27 @@ export async function getHighestLiquidityDIDs(
...
-    if (fetchedPools.data?.pools?.length === 0) return null
-    for (let i = 0; i < fetchedPools.data.pools.length; i++) {
-      if (!fetchedPools.data.pools[i].datatokenAddress) continue
-      const did = web3.utils
-        .toChecksumAddress(fetchedPools.data.pools[i].datatokenAddress)
-        .replace('0x', 'did:op:')
-      didList.push(did)
-    }
+  }
+  highestLiquidiyAssets.sort((a, b) => a.valueLocked - b.valueLocked).reverse()
```
claudiaHash, line 447: Seems like this ordering leads to the correct result.

claudiaHash (approved):
Looks good to me!

src/components/pages/Home.tsx
```diff
@@ -24,11 +24,10 @@ async function getQueryHighest(
   chainIds: number[]
 ): Promise<[SearchQuery, string]> {
   const dids = await getHighestLiquidityDIDs(chainIds)
-
   // TODO: this query needs to adapt to chainIds
```
kremalicious, line 27: I think we did that so can remove the TODO

src/utils/subgraph.ts
```diff
@@ -16,6 +16,10 @@ import {
   AssetsFreePrice_dispensers as AssetFreePriceDispenser
 } from '../@types/apollo/AssetsFreePrice'
 import { AssetPreviousOrder } from '../@types/apollo/AssetPreviousOrder'
+import {
+  HighestLiquidiyAssets_pools as HighestLiquidiyAssetsPools,
```
kremalicious, line 20: not sure where exactly this is coming from but throughout code small typo with a missing `t` like `HighestLiquidiyAssetsPools` which should be `HighestLiquidityAssetsPools `

src/utils/subgraph.ts
kremalicious: tending towards sneaking in small change towards ordering by `oceanReserve` in here as we're refactoring and touching it. I'm undecided by now but current implementation indeed feels as only being technically correct, whereas the arguments for `oceanReserve` are all good points from user standpoint like @w1kke & @DMats brought up and we generally tend towards user perspective. And realistically it should only lead to noticeable differences between new and very old pools.

src/utils/subgraph.ts
bogdanfazakas: right missed that 😃 , it comes from the typing generated by the highest liquidity query, I'll update that also

src/components/pages/Home.tsx
bogdanfazakas: sure gonna remove that

src/utils/subgraph.ts
bogdanfazakas: sure we can try ordering by oceanReserve, i tried to reproduce the actual order in our live market where we sort asset by valueLocked see https://github.com/oceanprotocol/market/blob/main/src/utils/subgraph.ts#L127-L128.
I'm gonna push the changes so we can have a look and decide if we stick to order by valueLocked or go with oceanReserve

mihaisc:
`oceanReserve` would make sense. Our ui is kinda of a mess right now

Screenshot 2021-07-14 172018
Screenshot 2021-07-14 172327

So we should add a label to TVL and change Pool Creator Liquidity to Pool Creator Statistics to be consistent

bogdanfazakas:
> `oceanReserve` would make sense. Our ui is kinda of a mess right now
> 
> Screenshot 2021-07-14 172018
> Screenshot 2021-07-14 172327
> 
> So we should add a label to TVL and change Pool Creator Liquidity to Pool Creator Statistics to be consistent

order by oceanReserve should be fine, going to add also the label for TVL and change the Pool Creator Liquidity to Pool Creator Statistics

kremalicious:
Other than points mentioned implementation is perfect but as an enhancement it seems there could still be value in some split-up of this section showing liquidity categorized based on network. Main reason being discoverability, as there currently is much more liquidity in ETH mainnet, pushing down assets of other networks in that section. But can all be handled in a new issue with some initial design thinking

mihaisc:
You can simply select your desired network, not sure if this is a problem. The discoverability issue we can solve after the split, we shouldn't focus on it now

bogdanfazakas:
> Other than points mentioned implementation is perfect but as an enhancement it seems there could still be value in some split-up of this section showing liquidity categorized based on network. Main reason being discoverability, as there currently is much more liquidity in ETH mainnet, pushing down assets of other networks in that section. But can all be handled in a new issue with some initial design thinking

The discoverability issue can really be a problem, but as Mihai mention with this implementation the users can select the desired networks and see the highest liquidity assets from those networks only and since we have the same implementation for the recently publish I would go with this mixed assets and have a new issue in where we decide how we are going to split-up this section or sections from home page

PR #739, opened 2021-07-15
Download ordered asset with forever timeout multiple times

### LINKED ISSUE #726: Unable to re-download ordered asset with forever timeout
Step to reproduce: 
1) Create a dataset with access type `Download` , `timeout` Forever
2) Create fixed pricing. 
3) Go through full process of buy  --> download
3) Refresh Page
4) User need to buy the asset again. 

Possible issue: https://github.com/oceanprotocol/market/blob/b9bd6a3a25f027ea73661cc17f48c7d16f358a28/src/components/organisms/AssetActions/Consume.tsx#L78

Possible Solution:   `if (assetTimeout === 'Forever' || assetTimeout === '0')`

claudiaHash:
Fixes #726 .

Changes proposed in this PR:

- download ordered asset with forever timeout multiple times fix

### DIFF
--- a/src/components/organisms/AssetActions/Consume.tsx
+++ b/src/components/organisms/AssetActions/Consume.tsx
@@ -74,7 +74,7 @@ export default function Consume({
     if (!data || !assetTimeout || data.tokenOrders.length === 0) return
 
     const lastOrder = data.tokenOrders[0]
-    if (assetTimeout === 'Forever' || assetTimeout === '0') {
+    if (assetTimeout === '0') {
       setPreviousOrderId(lastOrder.tx)
       setHasPreviousOrder(true)
     } else {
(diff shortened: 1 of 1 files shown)


src/components/organisms/AssetActions/Consume.tsx
```diff
@@ -74,8 +74,7 @@ export default function Consume({
     if (!data || !assetTimeout || data.tokenOrders.length === 0) return
 
     const lastOrder = data.tokenOrders[0]
-
-    if (assetTimeout === 'Forever') {
+    if (assetTimeout === 'Forever' || assetTimeout === '0') {
```
KatunaNorbert, line 77: This should work only with this check: `assetTimeout === '0'` because `assetTimeout` is in seconds

claudiaHash: Yes, it does. Not sure why the checking was done using `Forever` from the beginning and hope removing it doesn't affect anything else we did not observed yet.

KatunaNorbert (approved):
Good to go

src/components/organisms/AssetActions/Consume.tsx
KatunaNorbert: Before we were converting to string on init but than it was changed to fix the timeout error and this part somehow was missed

PR #740, opened 2021-07-16
Adapt subgraph url queries in pool tab for multinetwork

### LINKED ISSUE #737: Wrong url for subgraph queries in pool tab
Queries in the pool tab always use mainnet subgraph

bogdanfazakas:
Fixes #737  .

Changes proposed in this PR:

- Updates Pool Liquidity query for multi network use case
- Updates Pool History query for multi network use case

### DIFF
--- a/src/components/organisms/AssetActions/Pool/Graph.tsx
+++ b/src/components/organisms/AssetActions/Pool/Graph.tsx
@@ -17,8 +17,10 @@ import { darkModeConfig } from '../../../../../app.config'
 import Button from '../../../atoms/Button'
 import { Logger } from '@oceanprotocol/lib'
 import { useAsset } from '../../../../providers/Asset'
-import { gql, useQuery } from 'urql'
+import { gql, OperationContext, OperationResult } from 'urql'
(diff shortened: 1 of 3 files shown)


kremalicious:
testing right now, I still get an interval refresh of mainnet subgraph even when going to a polygon asset, and without switching to Pool tab. Looks like it's fetching the pool transactions from both subgraphs. So we should figure out where those auto-refetches are coming from and stop it somehow

Screen Shot 2021-07-20 at 10 43 33

bogdanfazakas:
> testing right now, I still get an interval refresh of mainnet subgraph even when going to a polygon asset, and without switching to Pool tab. Looks like it's fetching the pool transactions from both subgraphs. So we should figure out where those auto-refetches are coming from and stop it somehow so only the ones for the network of the current asset are left.
> 
> Screen Shot 2021-07-20 at 10 43 33

yep right missed that, i'll try to clear intervals when leaving assets detail page ora pool tab

kremalicious (approved):
this looks like it is all fetching the right things now, nice!

Issue #251, opened 2020-11-15
deal with datatoken cap in publish flow


kremalicious:
In ocean.js the default cap when creating a datatoken is `1000` https://github.com/oceanprotocol/ocean.js/blob/main/src/datatokens/Datatokens.ts#L84

Which is not transparent to users. When users then want to mint more datatokens during create pricing flow, it will fail with `DataTokenTemplate: cap exceeded`.

Before solving this in a more holistic way on the contract level, we need an interim solution. Suggestions:

- in publish form, show the cap in the datatoken field
- optionally make it editable as text field
- needs a help text describing connection to minting in dynamic price creation

trentmc:
[Better IDOs](https://github.com/oceanprotocol/multi-repo-issue/issues/36), a key part of V4, will implicitly handle this.

Because that's coming, I hesitate to do this issue beforehand, especially since it will raise many user Q's. That is, I recommend closing this ticket. @kremalicious is this ok with you? And cc @alexcos20

trentmc:
As there was no pushback on closing this, I'm... closing it:)

Reopen if you feel otherwise.

PR #742, opened 2021-07-17
set minimum price to 0.01

kremalicious:
Testing implications of lower minimum OCEAN price

### DIFF
--- a/src/models/FormPricing.ts
+++ b/src/models/FormPricing.ts
@@ -4,7 +4,7 @@ import * as Yup from 'yup'
 export const validationSchema: Yup.SchemaOf<PriceOptionsMarket> =
   Yup.object().shape({
     price: Yup.number()
-      .min(1, (param) => `Must be more or equal to ${param.min}`)
+      .min(0.01, (param) => `Must be more or equal to ${param.min}`)
       .required('Required'),
     dtAmount: Yup.number()
       .min(9, (param) => `Must be more or equal to ${param.min}`)
@@ -19,7 +19,7 @@ export const validationSchema: Yup.SchemaOf<PriceOptionsMarket> =
     weightOnOcean: Yup.string().required('Required'),
     swapFee: Yup.number()
       .min(0.1, (param) => `Must be more or equal to ${param.min}`)
-      .max(10, 'Maximum is 10%')
+      .max(10, (param) => `Maximum is ${param.max}%`)
       .required('Required')
       .nullable()
   })


realdatawhale:
Hey @kremalicious , though I'll just shoot you a comment here. Thanks for the Rinkeby tokens. 

I've tried to create the pricing on the pool, however it looks that I'm unable to mint the amount of Data Tokens that I'm looking for. I'll provide upwards of $OCEAN 5000+ initial liquidity at an IDO price of 0.02 - 0.05. So the amount of Data Tokens available will be between 50K - 120K. Is it possible to change this, so I'm able to mint that much?
I think the total supply is currently capped at 1000. 

Let me know if you can help us for this.

kremalicious:
Change in this PR would be totally fine so far.

But the datatoken cap indeed is hardcoded right now, or more precisely the `market` does not capture or pass it, making `ocean.js` fall back to `1000`. The tricky part here is that the cap would need to be set right in the publish form and not during price creation as creating the metadata also mints the initial data tokens. And it's hard to make majority of users understand this when they just create metadata and don't think about pricing yet. Which is why we were holding off on just adding it as a new field in the UI in publish form. The ideal solution here would be to split up metadata publishing & datatoken creation, which we already have in mind but no ETA. On the other hand in v4 this all seems to be [not an issue anymore](https://github.com/oceanprotocol/market/issues/251#issuecomment-732760746).

Some options I can think of to make your use case possible in a timely way where we can pick one:

1. Hardcoded custom cap which we pass to `ocean.js` during publish. So add to this PR, and you would use its preview link for publishing your assets. Quickest solution for everybody but you would have to live with the same cap on every asset.
2. Expose cap as a new optional input field in publish form for everybody. This might get feasible already to be put behind a new collapse-by-default _Advanced Settings_ area in publish form introduced in #707. So this is not as quick as 1 as we need to finalize #707 first and then see more UI & copy work will be required.
3. Wait for v4, but most likely not an option for you.

realdatawhale:
Hi @kremalicious ,

thanks a lot for presenting all the options. I’m inclined towards Option 1, as it seems to be the most timely solution. Also I feel Option 2 will be obsolete when v4 launches and since the micro-cap IDO is a trial & error, I don’t want to create too much work for your team.

Before going ahead, could you please help to clarify the below for my understanding?
- will the hardcoded custom cap be linked to my publishing address? Say, I settle on a supply of 250K for this IDO, then for any future IDOs, this would be my personal maximum supply cap correct? 

thanks again for helping and supporting 🙏

realdatawhale:
Thanks, just tried with the vercel link, mint still exceeds max cap on Rinkeby
Screen Shot 2021-08-02 at 5 21 59 PM

kremalicious:
Well you have to publish a new asset, only that will have the cap hardcoded in here. The cap is set during the datatoken creation which is happening before the price creation. So you can't set it during price creation because the datatoken is already minted. Price creation is creating the pool (or fixed rate exchange), not the datatoken itself.

So nothing in the DDO regarding the cap there will be affected by anything you do on that existing asset https://market-git-feature-micro-cap-oceanprotocol.vercel.app/asset/did:op:81DC6629b05F43D522049382a013EB7E2772cC99

Screen Shot 2021-08-02 at 15 29 27

realdatawhale:
Working! Thank you :-)

realdatawhale:
Just created the price & supply of 100'000 Data Tokens! Exciting! 

@kremalicious , please advise how do we go about the custom ticker ?

kremalicious:
> please advise how do we go about the custom ticker

to allow this in the UI would be a bit more complex to setup, hardcoding could work but also needs quick investigation as we auto-generate a new one on initial render of form. But we are not going to update it for you for every planned IDO anyway. As I said, ideally you would create the initial asset programmatically which you seem to have done in the past already and would give you all the flexibility you need. Once it's published you can then just pick it up and continue in the UI normally to create the price.

realdatawhale:
Hi @kremalicious , appreciate the offer to hardcode it for this one. Once it's done, will I have to create it via the link or will it show up immediately? If this is the case, let's wait until I'm fully ready with the dataset in about 15 days or so.

Thanks

kremalicious:
Revisit after v4

PR #743, opened 2021-07-19
Remove 'Forever' timeout for compute data sets from publish form

### LINKED ISSUE #733: Publishing Compute Asset with "Forever" `timeout` value will be = 3600 instead of "Forever" or "0"
Since timeout = 0 is passed in at here: 

https://github.com/oceanprotocol/market/blob/bea4f46c9ae65b0049d1d077c0343b04a12d4579/src/hooks/usePublish.ts#L86

https://github.com/oceanprotocol/market/blob/bea4f46c9ae65b0049d1d077c0343b04a12d4579/src/components/pages/Publish/index.tsx#L127

https://github.com/oceanprotocol/market/blob/bea4f46c9ae65b0049d1d077c0343b04a12d4579/src/utils/metadata.ts#L20-L23

Forever timeout will always be `3600` after publishing DDO.

Steps to reproduce:
1) Publish a compute dataset with timeout `forever`
2) Check published ddo `timeout` value

claudiaHash:
Fixes #733 .

Changes proposed in this PR:

- timeout value fix


claudiaHash:
After checking around and resolving what @krisliew  suggested in #733 : https://github.com/oceanprotocol/market/blob/bea4f46c9ae65b0049d1d077c0343b04a12d4579/src/hooks/usePublish.ts#L86
I also checked the createComputeService in ocean.js and there seems to be the problem: https://github.com/oceanprotocol/ocean.js/blob/77dbfb836aa122687fa4a1a7bb4100ec3e5333a7/src/ocean/Compute.ts#L349

When publishing data in market, we use `createComputeService` for creating `computeService` metadata for a `ddo`, which then comes as services in the the new ddo published. https://github.com/oceanprotocol/market/blob/bea4f46c9ae65b0049d1d077c0343b04a12d4579/src/hooks/usePublish.ts#L94

krisliew:
the issue for ocean.js `createComputeService()` https://github.com/oceanprotocol/ocean.js/issues/918

claudiaHash:
> the issue for ocean.js `createComputeService()` [oceanprotocol/ocean.js#918](https://github.com/oceanprotocol/ocean.js/issues/918)

We discussed this issue yesterday and came to the conclusion that is better to eliminate the `Forever` timeout option when creating a dataset with `Compute` asset type.

soonhuat:
@claudiaHash 
Does that mean `Forever` timeout will be remove for algorithm compute too? since it goes through the same createComputeService method?

claudiaHash:
No, it won't. Only  when creating a `dataset` with `Compute` type, the `Forever` option won't exist for timeout. The change will affect the options user can select for timeout in this situation, in the Publish form.

src/hooks/usePublish.ts
```diff
@@ -83,7 +83,7 @@ function usePublish(): UsePublish {
           break
         }
         case 'compute': {
-          if (!timeout) timeout = 3600
+          if (!timeout) timeout = 0
```
jamiehewitt15, line 86: I'm not sure why you are changing the timeout to 0 in the case where there isn't a timeout value? To my understanding, 0 will default to forever?

claudiaHash: Yes, exactly, as it's also suggested in: https://github.com/oceanprotocol/market/blob/b398106238b2e805a7209ffa37218762a4e7f36d/src/utils/metadata.ts#L20 
But you're right, timeout should be equaled to 0  as 'Forever' only for datasets with download asset type, as Alex suggested in the grooming session. I'll revert the change in a moment.

jamiehewitt15:
I tried to publish a dataset on the preview of this branch but got an error:

Failed publish 2021-08-03_16-54
Failed publish 2 2021-08-03_16-54

claudiaHash:
Using the latest deployment of the branch I just published 2 datasets, one with download  and the other one with compute access type and seemed to work for both cases. 

Screenshot from 2021-08-11 12-44-20

Hope it's not something related to my setup. Did someone else also tried?

kremalicious (approved):

jamiehewitt15 (approved):
Tested it and it's working great

PR #745, opened 2021-07-19
Check graph sync status

### LINKED ISSUE #728: Check graph sync status on asset details

mihaisc:
Fixes #728

Added errors in Web3Feedbak and removed them from asset page 

image

image

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -40,17 +40,17 @@ export default function App({
   const { warning } = useSiteMetadata()
   const { accountId } = useWeb3()
   const { isInPurgatory, purgatoryData } = useAccountPurgatory(accountId)
-  const { isGraphSynced, blockHead, blockGraph } = useGraphSyncStatus()
+  // const { isGraphSynced, blockHead, blockGraph } = useGraphSyncStatus()
 
   return (
     <Styles>
       <div className={styles.app}>
-        {(props as PageProps).uri.includes('/asset') && !isGraphSynced && (
+        {/* {(props as PageProps).uri.includes('/asset') && !isGraphSynced && (
           <AnnouncementBanner
             text={`The data for this network has only synced to Ethereum block ${blockGraph} (out of ${blockHead}). Please check back soon.`}
             state="error"
           />
-        )}
+        )} */}
 
         {(props as PageProps).uri === '/' && (
           <AnnouncementBanner text={warning.main} />
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -45,13 +45,6 @@ export default function App({
   return (
     <Styles>
       <div className={styles.app}>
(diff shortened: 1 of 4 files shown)


src/hooks/useGraphSyncStatus.ts
```diff
@@ -59,7 +59,7 @@ export function useGraphSyncStatus(): UseGraphSyncStatus {
   const { block, web3Loading } = useWeb3()
   const [blockGraph, setBlockGraph] = useState<number>()
   const [blockHead, setBlockHead] = useState<number>()
-  const [isGraphSynced, setIsGraphSynced] = useState(true)
+  const [isGraphSynced, setIsGraphSynced] = useState(false)
```
kremalicious, line 62: won't this result in the faulty graph sync status showing up all the time first, and then disappearing? Should be the other way around, should only come up when there is an error detected, not by default

mihaisc: no, this was just for testing and forgot to change it, good catch

src/components/molecules/Web3Feedback.tsx
```diff
@@ -18,24 +19,29 @@ export default function Web3Feedback({
...
+      : 'warning'
 
   const title = !accountId
     ? 'No account connected'
     : // : !ocean
     // ? 'Error connecting to Ocean'
     accountId && isAssetNetwork === false
     ? 'Not connected to asset network'
+    : isGraphSynced === false
+    ? `The data for this network has only synced to Ethereum block ${blockGraph} (out of ${blockHead}). Please check back soon.`
```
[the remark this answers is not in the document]
mihaisc: fair enough, will move this into the description. If i take another look i kinda swapped the title and the description

mihaisc:
Messages fixed
image

src/components/App.tsx
```diff
@@ -40,17 +40,17 @@ export default function App({
...
   const { accountId } = useWeb3()
   const { isInPurgatory, purgatoryData } = useAccountPurgatory(accountId)
-  const { isGraphSynced, blockHead, blockGraph } = useGraphSyncStatus()
+  // const { isGraphSynced, blockHead, blockGraph } = useGraphSyncStatus()
 
   return (
     <Styles>
       <div className={styles.app}>
-        {(props as PageProps).uri.includes('/asset') && !isGraphSynced && (
+        {/* {(props as PageProps).uri.includes('/asset') && !isGraphSynced && (
```
kremalicious, line 48: so far the place we put the warning seems prominent enough so we should not need all the commented out stuff in here, so safe to just remove

kremalicious (approved):
perfect, all good. Only that minor removal of commented out code in `src/components/App.tsx` would be cool before merging

PR #746, opened 2021-07-19
Statistics footer

### LINKED ISSUE #741: statistics footer
Add simple changes to `MarketStats` component getting it into a launchable state so we do not have to start crazy refactoring or new screens, like:

- add multiple lines, one for each selected network
- add TVL/pools/OCEAN of all selected networks together and display as one
- base it only on user network (current behavior but feels unintuitive right away)
- more ideas?

Not super pretty but simple and provides good user value until this gets all moved to a new `/pools` page:

Screen Shot 2021-07-16 at 15 08 53

KatunaNorbert:
Fixes #741 .

Changes proposed in this PR:

- show market statistics for multiple selected chainIds inside the footer

### DIFF
--- a/src/components/atoms/Tooltip.tsx
+++ b/src/components/atoms/Tooltip.tsx
@@ -28,9 +28,7 @@ export default function Tooltip({
   trigger,
   disabled,
   className,
-  placement,
-  link,
-  reference
+  placement
 }: {
   content: ReactNode
   children?: ReactNode
@@ -38,8 +36,6 @@ export default function Tooltip({
   disabled?: boolean
   className?: string
   placement?: Placement
-  link?: string
-  reference?: string
 }): ReactElement {
   const [props, setSpring] = useSpring(() => animation.from)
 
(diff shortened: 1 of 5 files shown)


kremalicious:
nice, functionality-wise this seems perfect on first look. I'll tinker around locally a bit to figure out how to deal with this mega tooltip visually

kremalicious:
alright, would say this looks good enough until we move the stats to other screens in future anyway:

Screen Shot 2021-07-22 at 12 27 01

Did not touch core functionality but refactored everything to prevent repaints and layout reflows once values are fetched, which should be a general principle we follow. A loading spinner especially for bigger layout sections really is just an easy dev solution so we should try to not add more if we can handle it without expensive repaints, which in end always provides better user experience too.

As in this case most of the sentence will not be replaced I render the whole thing by default with `0` as fallback for all values, then only those numbers change once values are fetched, which on most viewports will prevent layout reflows.

kremalicious (approved):

PR #751, opened 2021-07-22
Update CODEOWNERS

kremalicious:
We all own the code by now 😀

We require approval by code owners on each PR so with this change everybody in this list is requested for review automatically on each PR ready for review. 

And with that, we can also think about bumping the requirement from 1 code owner to 2 code owner approvals for each PR as more eyes on each change are always preferable

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @mihaisc @kremalicious
+* @mihaisc @kremalicious @claudiaHash @bogdanfazakas @KatunaNorbert @jamiehewitt15 @DimitarSD


mihaisc (approved):

jamiehewitt15:
> And with that, we can also think about bumping the requirement from 1 code owner to 2 code owner approvals for each PR as more eyes on each change are always preferable

Requiring 2 code owner approvals seems like a good idea to me. It will ensure that we all get involved in the review process.

jamiehewitt15 (approved):

bogdanfazakas (approved):
The 2 code owner approvals sounds good since there are going to be 7 people notified 😃

Issue #678, opened 2021-06-14
Improve search  (even more)


mihaisc:
Improve search to handle various word groups, synonyms and special characters 

For example, if you search "data union"  (as well as data-union) you'll find that "dataunion.app" doesn't appear in the results.

This work is mostly done in elasticsearch.

Few suggested ideas : 
https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-word-delimiter-tokenfilter.html
https://www.javacodegeeks.com/2018/03/elasticsearch-ignore-special-characters-query-pattern-replace-filter-custom-analyzer.html
https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-shingle-tokenfilter.html

PR #755, opened 2021-07-23
Improve search queries

jamiehewitt15:
Closes: #678

Changes proposed in this PR:

- Improving the search functionality. For example, previously if you searched for "data union" it would not return "DataUnion.App", now this has been corrected.


claudiaHash:
The new search works great! Although, I think there might be an issue when combining searching with filtering. For example, selecting `ALGORTIHMS` I got the following results:

Screenshot from 2021-07-26 12-51-04

mihaisc:
The issue we still add in query `service.attributes.main.type:algorithm`, i saw this in the multinetwork branch. In this case is incorrect and in multi is not optimal ( at this point the difference is minimal ). The correct way is to add a term 
```
 {
       "term":{
               "service.attributes.main.type": "algorithm"
          }
}
```

jamiehewitt15:
@claudiaHash Thank you for pointing that out, I have fixed the filters now. 

@mihaisc I used `match` instead of `term` as that's the recommendation of the documentation when using text fields:

```
Avoid using the term query for text fields.

By default, Elasticsearch changes the values of text fields as part of analysis. This can make finding exact matches for text field values difficult.

To search text field values, use the match query instead.
```
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-term-query.html

kremalicious (approved):
perfect functionality-wise. Crazy how bad our search was before :-)

KatunaNorbert (approved):
looks good!

kremalicious:
right, the merge conflict. @jamiehewitt15 could you take another look at this please? As we now need to combine your changes with the query for multiple chainIds we now have in `main` and you're the elastic search expert now 😄 

Screen Shot 2021-07-30 at 14 38 18

mihaisc:
Ah yes, @jamiehewitt15  while merging could also change the chain id query to use match/tern in the boolean query (like we do with type and isPurgatory), and not leave it as a query string. 😁

kremalicious (changes requested):
as mentioned, merge `main` and adapt query. Then we should all test it again

kremalicious (approved):
works flawlessly for me, now also with combinations of different networks selected and such, amazing improvement overall!

Issue #714, opened 2021-07-07
Handle publish errors better


mihaisc:
In the publish form you can get errors like 

![Error - 2021-07-07_13-50](https://user-images.githubusercontent.com/26000280/124750833-9cf93700-df2e-11eb-846f-672b084a6b1b.png)

The suggestion is to have a more user friendly message `Something went wrong . Woops!` and then a button that show the full error `Show full error`

jamiehewitt15:
I got this error while testing the moonbase alpha netwoprk ([slack link](https://oceanprotocol.slack.com/archives/CR8MHU7S8/p1625655119024100)) I'll see if I can reproduce it

PR #757, opened 2021-07-27
Hiding error message

jamiehewitt15:
Closes: #714 

Changes proposed in this PR:
- Publishing error messages are hidden by default
- User must click "More Info" to see the error message 

## To Test
The easiest way to test this is to publish an asset and then reject the metamask transactions.

The default view is with the error message hidden:
Hidden error

If you click "More Info" the error message shows:
Show error - 2021-07-27_16-40


kremalicious (approved):
nice and simple, could use some smaller design thinking to make it more clear it's an error by default but works as expected and no need to bother to further design errors

KatunaNorbert (approved):
working as expected, nice solution

PR #761, opened 2021-07-28
Fix broken explorer links for multinetwork

### LINKED ISSUE #759: Explorer links are broken
All uses of `ExplorerLink` seem broken on asset details:

- metadata history
- user profile link
- pool link
- datatoken link

KatunaNorbert:
Fixes #759 .

Changes proposed in this PR:
Fix explorerLinks for:
- metadata history
- user profile link
- pool link
- datatoken link

### DIFF
--- a/src/components/organisms/AssetActions/Pool/index.tsx
+++ b/src/components/organisms/AssetActions/Pool/index.tsx
@@ -65,7 +65,7 @@ export default function Pool(): ReactElement {
   const data = useStaticQuery(contentQuery)
   const content = data.content.edges[0].node.childContentJson.pool
 
-  const { accountId, networkId } = useWeb3()
+  const { accountId } = useWeb3()
   const { ocean } = useOcean()
   const { isInPurgatory, ddo, owner, price, refreshInterval, isAssetNetwork } =
     useAsset()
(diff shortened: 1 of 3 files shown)


src/components/organisms/AssetActions/Pool/index.tsx
```diff
@@ -278,15 +278,15 @@ export default function Pool(): ReactElement {
...
                 path={`address/${price?.address}`}
               >
                 Pool
               </ExplorerLink>
               <ExplorerLink
-                networkId={networkId}
+                networkId={ddo.chainId}
                 path={
-                  networkId === 137 || networkId === 1287
+                  ddo.chainId === 137 || ddo.chainId === 1287
```
kremalicious, line 289: forgot about this check. We recently switched Polygon to use `polygonscan.com` as explorer in ocean.js ConfigHelper, which behaves just like etherscan. So we can remove the check for `137` as it's now `token` there too, e.g. https://polygonscan.com/token/0x282d8efce846a88b159800bd4130ad77443fa1a1

In fact all explorer links we now have in ConfigHelper behave like etherscan, but still 2 left we need as check: `1287` & `2021000`

jamiehewitt15:
Just tested this out and the links seem to be working now

bogdanfazakas:
Functionality wise looks good, handle the requested change from Matthias and it should be good to go

src/components/organisms/AssetActions/Pool/index.tsx
KatunaNorbert: Right, I checked Polygon and Mainnet and no need for check. Not sure what `2021000` is

kremalicious: I have this file bookmarked just for that cause I never can remember chainIds :-) https://github.com/oceanprotocol/ocean.js/blob/main/src/utils/ConfigHelper.ts

`2021000` is Gaia-X testnet but so far we are not dealing with it but we can still add the check for it here as its expected this will always stay blockscout-based explorer so will need `/tokens/` here

KatunaNorbert: Cool, thanks

kremalicious (approved):
perfect, I think we found all possible explorer combinations, thanks for picking this up so quickly

Issue #724, opened 2021-07-12
Make author field editable


kremalicious:
Based on deltaDAO research a quick win to move in direction of being GDPR-compatible is to make the author field editable, and possibly optional so it can also be cleared. This might need changes in Aquarius' metadata validation

jamiehewitt15:
The plan is to add the `author` field to editable metadata within Ocean.js

kremalicious:
goals from user perspective in the end:

- publisher can edit existing author as part of _Edit Metadata_ flow
- publisher can add author as part of _Edit Metadata_ flow
- publisher can remove the author as part of _Edit Metadata_ flow
- this all implies author in main publish flows becoming optional

PR #763, opened 2021-07-29
Author metadata update

jamiehewitt15:
Closes: #724

Changes proposed in this PR:
- Adding author to editable metadata fields
- makes use of https://github.com/oceanprotocol/ocean.js/pull/936


content/pages/edit.json
```diff
@@ -44,6 +44,13 @@
         "options": ["Forever", "1 day", "1 week", "1 month", "1 year"],
         "sortOptions": false,
         "required": true
+      },
+      {
+        "name": "author",
+        "label": "New Author",
+        "placeholder": "e.g. Mrs McJellyfish",
+        "help": "Give proper attribution for your data set.",
+        "required": true
```
kremalicious, line 53: here I'm not sure how Formik behaves with this to cover this use case:

> publisher can remove the author as part of _Edit Metadata_ flow

So would clearing the field still make it submittable or would form error be shown?

kremalicious (changes requested):
already merged `main` into this PR. Now we just need to figure out where the `author` should be required or optional

content/pages/edit.json
jamiehewitt15: Formik won't let you submit an empty form field for the author as it is a required field. The submit button is disabled if you clear the author field

jamiehewitt15:
> already merged `main` into this PR. Now we just need to figure out where the `author` should be required or optional

I would say it should be required. If it is optional users can delete it and leave it blank

mihaisc:
Optional, users need the option to remove it. That what the whole point. This is a gdpr requirement

jamiehewitt15:
@mihaisc I've updated it so that the author is now optional

mihaisc (approved):

kremalicious (approved):
awesome, all use cases worked for me

For reference, this is what it looks like after author field has been cleared. Was first thinking to hide it altogether in that case but it's actually nice for transparency as this implies it has been cleared later
Screen Shot 2021-08-03 at 17 04 44

jamiehewitt15:
Thanks, @kremalicious I agree that it's good to keep the author field there, otherwise, users will look for it and be unsure why it's not there. In terms of increasing transparency, we could write _Removed_ instead of leaving a blank space. 

Also, going one step further in terms of transparency. It would also be good if the metadata history could what has been changed. Although this might be a lot of work: 
Metadata history - 2021-08-03_17-37

PR #765, opened 2021-07-30
Add price impact indicator to the trade component

### LINKED ISSUE #430: Add slippage indicator as proxy for data set pool quality of liquidity
**Challenge:** I want to swap OCEAN for datatokens (or vice versa), and a pool's liquidity is low, and I'm super concerned about losing $$ because of slippage. I don't have visibility into the tradeoff of amount traded versus price slippage. 

I *do* know the pool's liquidity (and some other stats). Liquidity is a useful signal towards dataset pool quality, as are others. However not seeing the exact slippage for my trade feels like flying partly blind when swapping, and is concerning.

**Proposed solution:** add slippage indicator to the UI, like Uniswap and other DEXes do.

**Related:** this was borne out of an issue with a different scope: #391

## Scope

Add a "price impact" metric that is based on a pool's slippage to the UI. An example from Uniswap:

image

Note how they use color coding to predict the trade's safety (red: danger, green: ok).

KatunaNorbert:
Fixes #430 .

Changes proposed in this PR:

- change the current price impart imput to slippage
- add price impact indicator


KatunaNorbert:
Trying to figure out how to get the fiat price for the datatoken, any suggestions?

kremalicious:
price of datatoken is the actual consume price of the asset which is always shown in OCEAN. Like, the price you pay when buying an asset is the datatoken price. So from there should be simple to convert that to fiat. Unless I'm missing something our `price.value` should also be the datatoken price but remember we had some different prices somewhere in subgraph along the lines of sport price and consume price or something and I always forget which is which

mihaisc:
Yes, spotPrice  != consumePrice

KatunaNorbert:
This is still in work progress, but the impact value can be tested.

src/components/organisms/AssetActions/Trade/PriceImpact.tsx
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    init()
+  }, [totalValue, tokenAmmount, spotPrice])
+
+  return (
+    <div className={styles.priceImpact}>
+      <strong>Expected price impact</strong>
+      <strong
+        className={parseInt(priceImpact) > 5 && styles.alert}
```
jamiehewitt15, line 58: It could be nice if the text was in green when the slippage estimate is very low. This could help give the user confidence to go ahead with the trade

src/components/organisms/AssetActions/Trade/FormTrade.tsx
```diff
@@ -79,6 +79,11 @@ export default function FormTrade({
     try {
       const impact = new Decimal(100 - Number(values.slippage)).div(100)
       const precision = 15
+      console.log(impact)
```
jamiehewitt15, line 82: Should probably remove the console logs from the production version

jamiehewitt15 (commented):
Seems to work very well. I left a couple of small comments

src/components/organisms/AssetActions/Trade/PriceImpact.tsx
mihaisc: i'm also not sure of the 5% limit, it might be too small for our pool sizes.

bogdanfazakas (approved):
LGTM also all seems to work fine for me and all discussions were handled 👍

kremalicious:
Still so many UX things to do while testing around but think it's better than having nothing right now. Like, consider this case:

Screen Shot 2021-08-18 at 12 30 17

We know tx will fail even before hitting the button, so we should tell that to user in some "are you really sure you want to do this" kind of thing. Our slippage setting does not allow this high impact either so we would need to deal with this too. In general, established user pereception is that price impact above 1-2% is crazy high and all DEX tools mark it visually in red, yet for us we would need some dynamic definition of what's considered too high based on pool size or something. But at least in above example it should rightfully prevent people from loosing money, as going into a trade with a 100% slippage is kinda madness

kremalicious (approved):

mihaisc:
> Still so many UX things to do while testing around but think it's better than having nothing right now. Like, consider this case:
> 
> Screen Shot 2021-08-18 at 12 30 17
> 
> We know tx will fail even before hitting the button, so we should tell that to user in some "are you really sure you want to do this" kind of thing. Our slippage setting does not allow this high impact either so we would need to deal with this too. In general, established user pereception is that price impact above 1-2% is crazy high and all DEX tools mark it visually in red, yet for us we would need some dynamic definition of what's considered too high based on pool size or something. But at least in above example it should rightfully prevent people from loosing money, as going into a trade with a 100% slippage is kinda madness

It will not fail , slippage and price impact differ

Issue #713, opened 2021-07-06
Account page - Jobs


mihaisc:
Move current component to new layout. This will be visible only when account = connected wallet

Issue #712, opened 2021-07-06
Account page - Shares


mihaisc:
Move current component to new layout.

Issue #711, opened 2021-07-06
Account page - Transactions


mihaisc:
Move the current component in the new layout.

Issue #710, opened 2021-07-06
Account page - Published


mihaisc:
![image](https://user-images.githubusercontent.com/26000280/124588199-335f2700-de61-11eb-992f-893f7af907e1.png)

Same filters/sort as in search component (maybe we can reuse that) .Show assets from selected chains.

PR #774, opened 2021-08-04
Account profiles

### LINKED ISSUE #668: Account page
Url : `/account/0xxxxx`
This page will sort of replace the history page.

This will be a place where a user can view all information and stats regarding his interaction with the market.
There will be public sections and private ones. For example published assets is a public information, compute jobs and results is private.
The general design should be dashboard like.
We need to add here all the information from the history page but structure it better.
We can add other statistics as well based on orders, financial data etc.

Probably this will need some designing (not only UI) before implementing it

Design: https://www.figma.com/file/qJ9LShmjORMMFqIQ1rM2ob/Account

### LINKED ISSUE #844: explorer link on asset page is leading to a 404
Screen Shot 2021-09-07 at 22 00 39

https://market.oceanprotocol.com/asset/did:op:5eF1e1D65e058A343e0106065cc3926cf416A1F4

links to: https://market.oceanprotocol.com/asset/undefined/address/0x304fFD0750Fa2D642fd87480D0A1e632DA8bE593 while it probably should lead to etherscan

bogdanfazakas:
- closes #668
- fixes #844

Changes proposed in this PR:

- replaces history page with profile page
- replaces history routes from /history to /profile
- fetched  the published assets (required in #710), pool transactions (required in #711), pool shares (required in #712), downloads using on accountId from the url or connected walled id if accessing my account section 
- display history compute jobs tab only when account id equals connected walled id and fetch jobs ( required in #713)
- link up publishers on asset details to their new profiles within app
- #776
- refactor ComputeJobs/index.tsx, splitting up presentation & logic into multiple helper methods
- #841
- #849
- #860, addressing https://github.com/oceanprotocol/multi-repo-issue/issues/117

Screen Shot 2021-09-13 at 16 42 40


kremalicious:
I mean, with 1644b79 already better than no title at all 😄 

Screen Shot 2021-09-02 at 16 34 09

But something more changed in subgraph resulting in all swap tx in `PoolTransactions` component to miss the `type: "in"` tx somehow, which really becomes obvious with these account profiles. So needs to be fixed here, or outside in another PR against `main` as on live the swap titles are just empty

mihaisc:
Tx are fixed but we have a nice scroll bar 
image

kremalicious (approved):
Amazing all-hands teamwork on this one, great job! 🥳

mihaisc (approved):

PR #776, opened 2021-08-06
Account metadata header

### LINKED ISSUE #709: Account page - Account
Main page (click to enlarge) :

[acc1](https://user-images.githubusercontent.com/26000280/124445706-73061000-dd88-11eb-95d2-f26f86647db4.png)
Mobile: 
[image](https://user-images.githubusercontent.com/26000280/124586613-512b8c80-de5f-11eb-8fc2-66dd8001a868.png)
All these are on the selected chains (from the header component)
Nr. of published  : query in aquarius based on owner 
Nr. of sold: sum of orders 
TVL in own assets is basically a sum of all the publishers liquidity in pools of assets that he has published

bogdanfazakas:
Fixes #709  .

Changes proposed in this PR:

- create account metadata header above the history section
- add profile information like name/profile image/description/ links/ nr of published assets/ total nr of orders and total liquidity in own assets


src/components/pages/Account/AccountHeader.tsx
```diff
@@ -0,0 +1,237 @@
...
+                    rel="noreferrer"
+                  >
+                    Read more
+                  </a>
+                </span>
+              ) : (
+                ''
+              )}
+            </div>
+            <div className={styles.publisherLinks}>
```
claudiaHash, line 224: Displaying this section might be conditioned if there are any links to show. In my case there aren't any and the bullet item is empty: 

Screenshot from 2021-08-23 14-30-41

kremalicious:
All working amazingly well already, but then had one of those lemme-just-quickly-design-something here turning into moving around everything, both visually and within codebase, sorry about that. 

Highly reduced the CSS needed to display what we have, and split up everything into subcomponents as this was just way easier to design with and in general will be easier to maintain. Also moved over that `NumberUnit` component we already use on `site` as we have the exact elements on screen that component already deals with.

All together now visually resulting in:

Screen Shot 2021-08-24 at 14 45 54

Another breaking change is the route change from `/account` → `/profile`. As for you vs. my interface, this is [always fun to look at](https://medium.com/@jsaito/is-this-my-interface-or-yours-b09a7a795256) so in our case we had "My Account" while we were using "Your ..." in other places so we always need to make sure to stick to either "My" or "Your". 

But after another round of thinking the naming "Your Account" didn't fully make sense and found that simply "Profile" better describes what we are building here, as the ETH account is only the base for what we display there. So in end we are building a profile for each account.

Did not check all the states with it yet but the loading state seems already good enough. We should further figure out something with adding/changing accountId in the route address but can be figured out in main PR or additional one

kremalicious (approved):

KatunaNorbert:
Screenshot 2021-08-24 at 16 51 41

This 'Please connect your web3 wallet' it's starting to get repetitive, and could get worse with more components, we might want to find a general message or view for in case no wallet is connected.

kremalicious:
Good point. Ideal case would be to resemble the final layout with received data as much as possible when no data is present at all but also uninspired what those fallback values should be. 

Another way would be to hide everything on this page without wallet, and instead show some kind of account explore thing listing multiple account teasers users can explore.

And finally,  habe we just want to hide the menu link altogether without a wallet, something already in discussion over in #810

KatunaNorbert:
If there are multiple chains selected the other lines beside the first one looks a bit off, because the elements will have a left margin. Will mostly be visible on mobile because of the smaller space. A quick fix would be to add margin to all the element from the div and remove the gap between account image and other info.

Screenshot 2021-08-24 at 17 07 14
Screenshot 2021-08-24 at 17 18 37
Screenshot 2021-08-24 at 17 08 08
Screenshot 2021-08-24 at 17 17 02

kremalicious:
made the links smaller, and gave it another layout flow on mobile so it stays extendible if we add more networks

Screen Shot 2021-08-24 at 16 37 06

Screen Shot 2021-08-24 at 16 38 01

KatunaNorbert:
> made the links smaller, and gave it another layout flow on mobile so it stays extendible if we add more networks
> 
> Screen Shot 2021-08-24 at 16 37 06
> 
> Screen Shot 2021-08-24 at 16 38 01

Cool, that's significantly better.
Just thinking that on mobile if we center the content of the Account component it would look better if we center the other components too, something like this:

Screenshot 2021-08-25 at 11 02 40

kremalicious:
Agree, that looks more balanced when centered on mobile.

By now I’m almost in favor of removing the icons, as they do not provide any help apart from decoration and we won’t find any metaphor which would make them clear in themselves so users need to read the label anyway to get their meaning so they are distracting user from getting the meaning of each item. And without any usefulness they take away precious space like for the TVL. If we wanna keep them then they need to be more dimmed so that the number value becomes the most prominent element.

If we want to have something nicely visual but also useful we could look into figuring out some small tiny graphs for each metric showing their values over time below or as some background in a next iteration

kremalicious:
And then to me it seems the sold number is not correct, e.g. I doubt directory dataset was only downloaded 3 times:

https://market-git-issue709-account-metadata-header-oceanprotocol.vercel.app/profile/0xbbd33AFa85539fa65cc08A2e61a001876D2f13FE

Can also see that this account downloaded assets from 0x31A7f70b8812daEf0A6C7F19575DB09f726F0DD1, but on that profile it says 0 sold:

https://market-git-issue709-account-metadata-header-oceanprotocol.vercel.app/profile/0x31A7f70b8812daEf0A6C7F19575DB09f726F0DD1

So think we need to look into how we get this number again

bogdanfazakas:
> And then to me it seems the sold number is not correct, e.g. I doubt directory dataset was only downloaded 3 times:
> 
> https://market-git-issue709-account-metadata-header-oceanprotocol.vercel.app/profile/0xbbd33AFa85539fa65cc08A2e61a001876D2f13FE
> 
> Can also see that this account downloaded assets from 0x31A7f70b8812daEf0A6C7F19575DB09f726F0DD1, but on that profile it says 0 sold:
> 
> https://market-git-issue709-account-metadata-header-oceanprotocol.vercel.app/profile/0x31A7f70b8812daEf0A6C7F19575DB09f726F0DD1
> 
> So think we need to look into how we get this number again

Yep that value is not yet correct, we were waiting for a release of the `subgraph` to easily get that but decided that we can have a value before the release time by having some calculations done on the market, i was thinking and trying to get today all the `token_orders` where the `account_id` value is either the `payer` or the `consumer`, right now we have a value that represents all the token orders where the `account_id` is the `payer`.

kremalicious:
Alright, most remaining UI changes done and gonna merge this finally so we can continue finding bugs on the main epic PR. Already moving the remaining sold value thingy into new issue 

Screen Shot 2021-09-01 at 13 31 26

Issue #782, opened 2021-08-09
Fixing "More Info" button styling


jamiehewitt15:
The buttons are currently too close together:

![Buttons too close](https://user-images.githubusercontent.com/20739535/128685096-e91d35af-3fc3-45ba-87d6-8a7e49271b0a.png)

PR #783, opened 2021-08-09
Fixing more info button padding

jamiehewitt15:
Closes: #782 

Changes proposed in this PR:
- Adding additional padding around the "More Info" text button.

Before it looked like this: 
Buttons too close

Now it looks like this:
fixed button spacing

### DIFF
--- a/src/components/molecules/MetadataFeedback.module.css
+++ b/src/components/molecules/MetadataFeedback.module.css
@@ -38,3 +38,6 @@
 .action {
   margin-top: calc(var(--spacer) / 1.5);
 }
+.moreInfo {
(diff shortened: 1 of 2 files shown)


mihaisc (approved):
Pretty sure @kremalicious will not like that `!important` but this is too minor to matter.

kremalicious (approved):
yeah, using specificity is always better than taking out styles from the cascade flow with `!important` but no need to deal with that here. I would even consider the need to use `!important` here a bug of our CSS modules implementation as I would expect to be able to overwrite original styles of a component but for whatever reason this does not work with our implementation

Issue #773, opened 2021-08-04
History links not working for some assets 


mihaisc:
Some metadata history links are not working , tx is undefined

examples: https://market.oceanprotocol.com/asset/did:op:b98B96B2DDe7e1237928400cFCD54327eF58fff4

https://market.oceanprotocol.com/asset/did:op:7FBEDfF92F39ae900393B17370E192D508458FDB

jamiehewitt15:
This seems to be a multinetwork issue. If you are not on the same network as the asset it will have undefined in the URL. If you are on the same network it works. 

Interestingly, it seems to work fine with no wallet connected.

jamiehewitt15:
The issue is that the graphQL query isn't working when you're on the wrong network.
```
const [result] = useQuery({
  query: getReceipts,
  variables: { address: ddo?.dataToken.toLowerCase() }
})
const { data } = result
console.log('Data:', data) /// Data: undefined
```

Still not sure why it works when a wallet isn't connected

jamiehewitt15:
I noticed that the DDO already contains the id of the creation transaction, so the easiest way to solve this is by using: 
```
ddo.event.txid
```

kremalicious:
Totally missed that `ddo.event` was added which is not documented anywhere and also not clear why this is just the creation transaction and not all events. Regardless, if it's there and really the creation transaction then we can use it but as this is just the creation tx can't see how this would solve the link for update transactions

kremalicious:
I think we have to ditch `useQuery()` from `urql` as this will always use the client initiated in `UrqlProvider` which is based on the `networkId`. Might be worth rethinking that provider to also deal with a DDO but this could get messy in terms of logic. For me, `did:op:7FBEDfF92F39ae900393B17370E192D508458FDB` also does not work without wallet.

What we have done for getting the asset pool data is to pass another `OperationContext` and initiate the subgraph connection from components. E.g. here: https://github.com/oceanprotocol/market/blob/main/src/components/organisms/AssetActions/Pool/index.tsx#L103-L116

So we could do the same for edit history, and looks like it would make sense replacing the auto-connection of `UrqlProvider` with some `connect` method in there which is then shared among all the components which need to query subgraph independent of wallet network.

kremalicious:
Same is true for `Consume.tsx` btw which also uses `userQuery` and I get a bunch of console errors from this on every asset details. So apart from edit history and consume component we do not use `useQuery` and its provider anywhere

[...]

PR #784, opened 2021-08-10
Fix metadata history

jamiehewitt15:
Closes: https://github.com/oceanprotocol/market/issues/773

Changes proposed in this PR:
- Getting the transaction Id from the DDO rather than using the graphQL query
- Removing unused `networkId`


jamiehewitt15:
This works as it is, however it gives a type error if `event` isn't defined in the `DDO.d.ts`

I have edited Ocean.js here: https://github.com/oceanprotocol/ocean.js/pull/965

kremalicious:
made obsolete by #794

PR #785, opened 2021-08-11
Fix isBalanceSufficient 

### LINKED ISSUE #779: OR || operator won't work in the case statement
### Steps to reproduce bug:
1) Have 1 Ocean balance in wallet
2) Visit an asset page that is price exactly at 1 Ocean
3) You will see insufficient balance msg

### Root cause: OR || operator won't work in the case statement 

https://github.com/oceanprotocol/market/blob/cf44ab7f51e7dbfdb4c437ae589bcd649975e042/src/utils/compareAsBN.ts#L8-L14

### Solution: 
``` 
 switch (compare) { 
   case 1: // balance is greater than price
   case 0: // balance is equal to price
     return true 
   default: 
     return false 
```

krisliew:
Fixes #779 

- Remove OR  `||` operator in case statement

### DIFF
--- a/src/utils/compareAsBN.ts
+++ b/src/utils/compareAsBN.ts
@@ -8,7 +8,8 @@ export default function compareAsBN(balance: string, price: string): boolean {
   const compare = aBN.comparedTo(bBN)
 
   switch (compare) {
-    case 1 || 0: // balance is greater or equal to price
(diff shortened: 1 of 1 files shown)


kremalicious (approved):
nice minimal change, thanks!

PR #786, opened 2021-08-11
Display user's pool transactions history values

### LINKED ISSUE #772: Your Pool Transactions history values not found on asset details page
The Your Pool Transactions table shows `No results found` inside the asset details page pool tab, although it appears in the pool transactions table from the History page. 
Also multiple subgraphs are queried based on selected chainIds for the transactions, which doesn't make sense since we know on what network the asset is and we can only query the specific subgraph. 

**Screenshots**
Screenshot 2021-08-04 at 13 25 30

Screenshot 2021-08-04 at 13 26 03

claudiaHash:
Fixes #772  .

Changes proposed in this PR:

- displaying user's pool transaction history values fixed
- query only the subgraph corresponding to the asset's network

### DIFF
--- a/src/components/molecules/PoolTransactions/index.tsx
+++ b/src/components/molecules/PoolTransactions/index.tsx
@@ -129,9 +129,11 @@ const columnsMinimal = [columns[0], columns[3]]
 
 export default function PoolTransactions({
   poolAddress,
+  poolChainId,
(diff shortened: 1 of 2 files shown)


kremalicious (approved):
perfect, works like a charm

PR #788, opened 2021-08-11
fix checkIsConsumable dependencies

### LINKED ISSUE #787: checkIsConsumable throws lots of errors
Looks like `checkIsConsumable` is using `ocean` object but ocean is not initiated resulting in long list of errors on every asset:

Screen Shot 2021-08-11 at 13 11 32

kremalicious:
- fixes #787
- do nothing when `ocean` is not present, or when asset network is not the same as user network as this would give us a wrong `ocean` to work with, making the check fail


mihaisc (approved):
good enough for now, ideally you wouldn't need a wallet for this check  (from a UX standpoint). But i guess the impact is not that big for now

bogdanfazakas (approved):

PR #794, opened 2021-08-11
Fixing metadata history

### LINKED ISSUE #773: History links not working for some assets

jamiehewitt15:
Closes #773 
Supersedes #784

Changes proposed in this PR:

- Getting chainId from DDO

### DIFF
--- a/src/helpers/wrapRootElement.tsx
+++ b/src/helpers/wrapRootElement.tsx
@@ -2,6 +2,7 @@ import React, { ReactElement } from 'react'
 import Web3Provider from '../providers/Web3'
 import { UserPreferencesProvider } from '../providers/UserPreferences'
 import PricesProvider from '../providers/Prices'
+import UrqlProvider from '../providers/UrqlProvider'
 
 export default function wrapRootElement({
   element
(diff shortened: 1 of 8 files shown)


src/helpers/wrapRootElement.tsx
```diff
@@ -11,7 +11,7 @@ export default function wrapRootElement({
 }): ReactElement {
   return (
     <Web3Provider>
-      <UrqlProvider>
```
jamiehewitt15, line 14: I tried removing `UrqlProvider` from here but that causes some errors

kremalicious:
What we need to prevent is to have multiple providers nested on one screen. Did lots of tinkering and I think the best way is to always pass a specific query context, and have the provider only there to have an initial client as that's required.

So something like this in the `EditHistory` component:

```js
const queryContext = getQueryContext(ddo?.chainId)
const [result] = useQuery({
  query: getReceipts,
  variables: { address: ddo?.dataToken.toLowerCase() },
  context: queryContext,
  pause: !ddo || !queryContext
})
```

 Currently going this direction and trying to get the asset details screen to render again

kremalicious:
Should work now™. My above code sample actually worked but not many ways to instruct `useQuery` to stop rendering as I could not get that `pause` functionality to stop re-rendering the component, but this would be pattern to use:

```js
const [queryContext, setQueryContext] = useState<OperationContext>()

useEffect(() => {
  if (!ddo) return
  const queryContext = getQueryContext(ddo.chainId)
  setQueryContext(queryContext)
}, [ddo])

const [result] = useQuery({
  query: getReceipts,
  variables: { address: ddo?.dataToken.toLowerCase() },
  context: queryContext,
  pause: !ddo || !queryContext
})
const { data } = result
```

kremalicious (approved):

mihaisc (approved):

PR #796, opened 2021-08-13
GDPR Compliance

moritzkirstein:
This PR delivers the following changes and features to Ocean Market to address GDPR compliance:

- **Input disclaimers**: Input fields support a disclaimer component now. They display a given text when certain values are entered. This feature is enabled on the dataset and algorithm publish forms.

- **Input help**: Help texts on the author and description input fields (publish forms) have been adjusted in relevance to GDPR requirements.

- **Multi-Language Privacy Policies**: Ocean Market now has its own privacy policies. Market users can select a preferred language for the policies. Languages included are English, German, French & Spanish.

- **Privacy Preference Center**: A customizable privacy preference center has been added. This allows for easy and fast GDPR compliant forking of the repository. The preference center is disabled per default. For more details a new configuration section has been added to the README

- **Imprint**: An imprint has been added to Ocean Market. This needs further attention as some additional details must be provided (refer to the prepared imprint provided in the pull request).

- **History Printing**: A print button has been added to the history page to address the right to data portability (Art. 20 GDPR).


kremalicious:
solid work so far, thanks for the PR. Some quick notes after first look:

- let's try to minimize duplication as much as possible. I would advocate for having only 1 `PageMarkdown` template and add some specific components to it based on the `/privacy` route or so.
- I would opt for the route `/privacy` instead of `/privacypolicy`
- as for language switcher on those pages, would be much more clear and discoverable and less clicks required if we simply output the `select` options as simple route links in a row. There is no actual form here required so let's not use form elements for simple links.
- switches in cookie settings look nice but have multiple accessibility problems which is why we simply use checkbox inputs for binary options. Switches would also introduce some inconsistencies as not clear why we then use switches and checkboxes, both essentially doing the same. But checkboxes do not pose any problems to e.g. colorblind users and they are just damn clear about their state.
- I do not agree with the imprint. This is following a rule which exclusively applies to sites hosted in Germany only, and is almost always useless information duplication for (german) lawyers as the contact data should be part of privacy policy anyway.
- the scroll up button to me is just codebase and performance overhead and does not provide any better functionality, and it's covering up content on mobile. Users know how to scroll so would remove it from codebase 
- don't put plain links as text into the content. Something like `https://market.oceanprotocol.com` in plain text without being linked is useless to users. Should be short linked text if within copy, like `[market.oceanprotocol.com](https://market.oceanprotocol.com)` or simply link `[Ocean Market](https://market.oceanprotocol.com)`

src/providers/CookieConsent.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+    setConsentStatus(initialValues)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
+
+  useEffect(() => {
+    if (!privacyPreferenceCenter || !window || !navigator) return
+
+    if (
+      (window.doNotTrack && window.doNotTrack === '1') || // Safari >7.1.3, Edge
```
kremalicious, line 110: Do Not Track never worked as expected and the standard has been abandoned, see https://en.wikipedia.org/wiki/Do_Not_Track

So we don't need to bother with detecting it

mihaisc: It actually works in edge, really good i might say.

kremalicious: I think this was the beginning of the end of that standard when Microsoft had decided in Edge to turn it on by default so immediately it was not reflecting user intention anymore, but browser vendor decisions, making it kinda useless. Most browsers by now have switched to some built-in tracking blocker anyway

moritzkirstein: I just committed some changes including the removal of Do Not Track detection

README.md
```diff
@@ -386,6 +389,51 @@ Ocean Market also includes a number of advanced features that are suitable for a
...
+
+### Privacy Preference Center
+
+Additionally, Ocean Market provides a privacy preference center for you to use. This feature is disabled per default since we do not use any cookies on our deployment of the market. However, if you need to add some functionality depending on cookies, you can simply enable this feature by changing the value of the `privacyPreferenceCenter` variable in the `app.config.js` config file. This will enable a simple cookie banner stating the use of necessary cookies. The content of this banner can be adjusted within the `content/gdpr.json` file. The `gdpr.json` file contains the minimum required values to configure the preference center. However, if you need more customizability for example to configure optional cookies for your market, you can also add these here. To find out how to configure this, we provide an example within the `content` directory. To quickly preview the example you can enable it with the following command:
+
+```
+cd content && mv gdpr.json gdpr_essential.json && cp gdpr.example.json gdpr.json
```
jamiehewitt15, line 405: It would be nice to add this script to package.json and replace it with something shorter here. For example:

```suggestion
npm run gdpr:preview
```

kremalicious: right, I did not get the intent of all the different json files as the app config seems to control showing it or not so that would be enough of a mechanism so we can reduce some complexity here. Would be fine to just have one `content/gdpr.json` in version control and that's it. No renaming and such. Readme makes it sufficiently clear that the app config will then pick up values from there.

app.config.js
```diff
@@ -60,5 +60,17 @@ module.exports = {
...
+  // this should be the slug of your default policy markdown file
+  defaultPrivacyPolicySlug: '/privacy/en',
+
+  // This enables / disables the use of a GDPR compliant
+  // privacy preference center to manage cookies on the market
+  // If set to true a gdpr.json file inside the content directory
+  // is used to create and show a privacy preference center / cookie banner
+  // To learn more about how to configure and use this, please refer to the readme
+  // https://github.com/deltaDAO/Ocean-Market/tree/main#-gdpr-compliance
+  privacyPreferenceCenter: false
```
jamiehewitt15, line 75: For this to be consistent for the other feature toggles (e.g. `allowAdvancedSettings`), this should be:

```suggestion
  privacyPreferenceCenter:  process.env.GATSBY_PRIVACY_PREFERENCE_CENTER || 'false'
```

and then `.example.env` should be updated to include:
```
# GATSBY_PRIVACY_PREFERENCE_CENTER="true"
```

README.md
```diff
@@ -386,6 +389,51 @@ Ocean Market also includes a number of advanced features that are suitable for a
...
+
+### Privacy Preference Center
+
+Additionally, Ocean Market provides a privacy preference center for you to use. This feature is disabled per default since we do not use any cookies on our deployment of the market. However, if you need to add some functionality depending on cookies, you can simply enable this feature by changing the value of the `privacyPreferenceCenter` variable in the `app.config.js` config file. This will enable a simple cookie banner stating the use of necessary cookies. The content of this banner can be adjusted within the `content/gdpr.json` file. The `gdpr.json` file contains the minimum required values to configure the preference center. However, if you need more customizability for example to configure optional cookies for your market, you can also add these here. To find out how to configure this, we provide an example within the `content` directory. To quickly preview the example you can enable it with the following command:
```
jamiehewitt15, line 402: I would suggest it would be better to handle the feature toggle in the `.env` as this would be consistent with how we have implemented all of the other optional configurations

```suggestion
Additionally, Ocean Market provides a privacy preference center for you to use. This feature is disabled per default since we do not use any cookies on our deployment of the market. However, if you need to add some functionality depending on cookies, you can simply enable this feature by changing the value of the `GATSBY_PRIVACY_PREFERENCE_CENTER` environmental variable to `"true"` in your `.env` file. This will enable a simple cookie banner stating the use of necessary cookies. The content of this banner can be adjusted within the `content/gdpr.json` file. The `gdpr.json` file contains the minimum required values to configure the preference center. However, if you need more customizability for example to configure optional cookies for your market, you can also add these here. To find out how to configure this, we provide an example within the `content` directory. To quickly preview the example you can enable it with the following command:
```

jamiehewitt15:
It all looks really good and it is great to see the effort that has gone into this. One potential room for improvement would be to make the Privacy Preference Center smaller and less intrusive.  This could be achieved by hiding all of the individual cookie preference options behind a "Show all options" button. The default view could therefore be smaller and simpler. I imagine most users will either click "Accept All" or "Only Essential" and they will probably want to get past this banner as quickly as possible

README.md
jamiehewitt15: Yeah, a single gdpr.json file would simplify it

moritzkirstein: I agree it's unnecessary overhead. I removed the example file with one of the latest commits.

moritzkirstein:
Thanks for your suggestion @jamiehewitt15 
I just merged a few commits addressing the styling of the PPC. An option to switch between two different "initial" stylings of the cookie banner is now provided: Setting the `style` prop of the `PrivacyPreferenceCenter` component to `'small'` will enable a smaller cookie banner and only show the default styled preference center when the user requests further customization (screenshot below).

@kremalicious - thanks for your reviews and comments! I addressed the provided feedback in previous commits.

I believe all feedback is merged into the PR now. Please feel free to re-review the changes and provide feedback where necessary.

grafik

jamiehewitt15:
@moritzkirstein thank you very much for all of your on this. The `small` styling for the banner is really nice and much less intrusive, so that's a great addition. As far as I can see all of the comments above have been addressed. I have left just one additional suggested change on the readme.

README.md
```diff
@@ -386,6 +389,47 @@ Ocean Market also includes a number of advanced features that are suitable for a
...
+    case CookieConsentStatus.REJECTED:
+    case CookieConsentStatus.NOT_AVAILABLE:
+    default:
+      // example logic
+      deleteCookie(`YOUR_COOKIE_NAME`)
+      break
+  }
+})
+```
+
```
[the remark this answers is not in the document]
moritzkirstein: Thanks for the suggestion. I think this looks a lot cleaner!

mihaisc:
@moritzkirstein can you please merge the conflicts

moritzkirstein:
@mihaisc I just committed the necessary changes to merge the conflicts. Please feel free to review and/or merge the PR

PR #799, opened 2021-08-17
Adding Cypress for integration testing

### LINKED ISSUE #128: add pragmatic testing

DimitarSD:
Fixes #128 .

- use Cypress for easy scripting of end-to-end testing
- setup and isolate under tests/integration
- GitHub Action setup: run barge, run integration test
- new npm scripts
- use the latest changes from the main


DimitarSD:
@kremalicious @mihaisc  In `web3modal` module, file `injected.ts` there is a line of code `await provider.request({ method: 'eth_requestAccounts' })` where when using HDWalletProvider it throws an error. 

```javascript
const ConnectToInjected = async () => {
  let provider = null;
  if (typeof window.ethereum !== 'undefined') {
    provider = window.ethereum;
    console.log('provider  =  ', provider);
    try {
      await provider.request({ method: 'eth_requestAccounts' })
    } catch (error) {
      throw new Error("User Rejected");
    }
  } else if (window.web3) {
    console.log('window.web3  /  =  ', window.web3)
    provider = window.web3.currentProvider;
    console.log('window.web3.currentProvider === ', provider)
  } else {
    throw new Error("No Web3 Provider found");
  }
  return provider;
};

export default ConnectToInjected;
```

There isn’t a `request` method that we can use from HDWalletProvider. Using `web3modal?.connect()` in the market breaks the code as well. Also the events are not available, the ones we're looking for in the market:

```ts
useEffect(() => {
    if (!web3Provider || !web3) return

    web3Provider.on('chainChanged', handleChainChanged)
    web3Provider.on('networkChanged', handleNetworkChanged)
    web3Provider.on('accountsChanged', handleAccountsChanged)

    return () => {
      web3Provider.removeListener('chainChanged', handleChainChanged)
      web3Provider.removeListener('networkChanged', handleNetworkChanged)
      web3Provider.removeListener('accountsChanged', handleAccountsChanged)
    }
  }, [web3Provider, web3])
```

mihaisc:
One of our problems might be that people don't really care about EIP 1193 https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md#implementations. @kremalicious if you remember when we tried the other providers supported by web3Modal each one had it's own implementation/api

DimitarSD:
I did some research, and there are two possible solutions for this one to move forward:

1) cloning HDWalletProvider repo and adding what we need (the proper structure/standard for web3 that Metamask and some of the other wallets use) or
2) automatically adding the Metamask extension to the browser cypress use (the 1st solution I started working on at the beginning using a 3rd party lib)

The main issue is that the object we get from HDWalletProvider doesn't contain what we're trying to access in terms of properties, and from there everything breaks. I'm open to other solutions, although I'm not sure if there is another approach to this problem.

PR #802, opened 2021-08-19
Display highest liquidity assets fix

### LINKED ISSUE #800: Highest liquidity list not working for all non-ETH chains
When deselecting all networks except Polygon I end up with this:

Screen Shot 2021-08-18 at 13 21 59

For me, only selecting ETH Mainnet works to show anything in Highest Liquidity list.

claudiaHash:
Fixes #800.

Changes proposed in this PR:

- display highest liquidity assets according to selected network(s)
-  no network selected case fix

### DIFF
--- a/src/components/pages/Home.tsx
+++ b/src/components/pages/Home.tsx
@@ -94,8 +94,6 @@ function SectionQueryResult({
         try {
           setLoading(true)
           const result = await queryMetadata(query, source.token)
-          console.log('QUERY: ', query)
-          console.log('RESULT: ', result.results)
           if (queryData && result.totalResults > 0) {
             const searchDIDs = queryData.split(' ')
             const sortedAssets = sortElements(result.results, searchDIDs)
--- a/src/components/pages/Home.tsx
+++ b/src/components/pages/Home.tsx
@@ -104,7 +104,7 @@ function SectionQueryResult({
           setResult(result)
           setLoading(false)
         } catch (error) {
-          Logger.log(error.message)
+          Logger.error(error.message)
         }
       }
     }
--- a/src/components/pages/Home.tsx
+++ b/src/components/pages/Home.tsx
@@ -21,10 +21,10 @@ import styles from './Home.module.css'
 async function getQueryHighest(
   chainIds: number[]
 ): Promise<[SearchQuery, string]> {
-  const dids = await getHighestLiquidityDIDs(chainIds)
+  const [dids, didsLength] = await getHighestLiquidityDIDs(chainIds)
   const queryHighest = {
     page: 1,
-    offset: dids.length,
+    offset: didsLength,
     query: {
       query_string: {
         query: `(${dids}) AND (${transformChainIdsListToQuery(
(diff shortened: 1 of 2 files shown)


src/components/pages/Home.tsx
```diff
@@ -80,20 +81,33 @@ function SectionQueryResult({
...
         setResult(result)
         setLoading(false)
-      } catch (error) {
-        Logger.log(error.message)
+      } else {
+        try {
+          setLoading(true)
+          const result = await queryMetadata(query, source.token)
+          console.log('QUERY: ', query)
+          console.log('RESULT: ', result.results)
```
bogdanfazakas, line 98: delete console logs 😃

claudiaHash: sure :) also, not totally working yet, it's just a draft.

claudiaHash:
Seems that using the 'network-only' policy on requests solves the problem (thanks  for hint, @bogdanfazakas ) , although it will ignore  cached results. So I think 'cache-and-network' would work best for us, because it  _returns cached results but also always sends an API request, which is perfect for displaying data quickly while keeping it up-to-date_ as they specify in their documentation (https://formidable.com/open-source/urql/docs/basics/document-caching/)

kremalicious:
Good job on the investigation, I was the one who put in that caching change but obviously without fully realizing all implications. I tried to limit some requests, as on the Trade tab they were going crazy but in end was caused by other things

claudiaHash:
>  I was the one who put in that caching change but obviously without fully realizing all implications.

No problem, I think it's impossible to anticipate all the implications a change like this could have. At lest, I think we found the policy that fits best for this request, as I briefly tested the Trade tab also.

src/components/pages/Home.tsx
```diff
@@ -80,20 +81,31 @@ function SectionQueryResult({
...
+            const searchDIDs = queryData.split(' ')
+            const sortedAssets = sortElements(result.results, searchDIDs)
+            const overflow = sortedAssets.length - 9
+            sortedAssets.splice(sortedAssets.length - overflow, overflow)
+            result.results = sortedAssets
+          }
+          setResult(result)
+          setLoading(false)
+        } catch (error) {
+          Logger.log(error.message)
```
kremalicious, line 107:
```suggestion
          Logger.error(error.message)
```

src/components/pages/Home.tsx
```diff
@@ -80,20 +81,31 @@ function SectionQueryResult({
...
-      try {
-        setLoading(true)
-        const result = await queryMetadata(query, source.token)
-        if (queryData && result.totalResults > 0) {
-          const searchDIDs = queryData.split(' ')
-          const sortedAssets = sortElements(result.results, searchDIDs)
-          const overflow = sortedAssets.length - 9
-          sortedAssets.splice(sortedAssets.length - overflow, overflow)
-          result.results = sortedAssets
+      if (chainIds.length === 0) {
```
kremalicious, line 84: Nice, this solves something I was about to write an issue for. This use case that all chains are unselected was never thought about and simply showing empty results is way better than before

kremalicious (approved):

mihaisc:
If you quickly change networks you break it , maybe we should add a debounce (or something similar) like in pool liquidity removal. 
image

bogdanfazakas (approved):
G2G

claudiaHash:
Great! I think we should address the problem @mihaisc discovered in a separate issue.

PR #803, opened 2021-08-20
fix fileinfo fetching

kremalicious:
- fix fileinfo fetching without wallet connected
- we can’t rely on config from `useOcean` cause that’s only there when there is an “Ocean connection” meaning a wallet connected

### DIFF
--- a/src/components/organisms/AssetActions/index.tsx
+++ b/src/components/organisms/AssetActions/index.tsx
@@ -14,10 +14,11 @@ import { useWeb3 } from '../../../providers/Web3'
 import Web3Feedback from '../../molecules/Web3Feedback'
 import { getFileInfo } from '../../../utils/provider'
 import axios from 'axios'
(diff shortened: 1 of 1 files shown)


claudiaHash (approved):
Looks good to me!

jamiehewitt15:
I published [this asset](https://market-prukzwkgh-oceanprotocol.vercel.app/asset/did:op:4C88502Cf9E8b8FDa6Ab32ceB0cD5aC63b6dd007) but I'm still finding that the file info isn't working on Rinkeby

File info not working

kremalicious:
this fix is unrelated to the issue with the provider described in #804. This here only fixes the problem that fileinfo isn't displayed on all assets when user has no wallet connected, sorry for confusion as I did not create dedicated issue for this

KatunaNorbert (approved):

PR #805, opened 2021-08-23
Account history published page

### LINKED ISSUE #710: Account page - Published

bogdanfazakas:
Fixes #710  .

Changes proposed in this PR:

- adds filter by asset type (dataset and algorithms) on published section from account's history page

### DIFF
--- a/src/components/pages/Account/History/PublishedList.tsx
+++ b/src/components/pages/Account/History/PublishedList.tsx
@@ -44,7 +44,7 @@ export default function PublishedList({
       try {
         const source = axios.CancelToken.source()
 
-        queryResult || setIsLoading(true)
+        setIsLoading(true)
         const result = await queryMetadata(queryPublishedAssets, source.token)
         setQueryResult(result)
       } catch (error) {
--- a/src/components/pages/Account/History/PublishedList.tsx
+++ b/src/components/pages/Account/History/PublishedList.tsx
@@ -58,11 +58,16 @@ export default function PublishedList({
 
   return accountId ? (
     <div>
-      <ServiceFilter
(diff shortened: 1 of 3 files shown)


claudiaHash (approved):

KatunaNorbert:
It's working well. I think a loading icon while fetching data when switching between asset type would make the user experience better, like we have on search screen.

jamiehewitt15:
> It's working well. I think a loading icon while fetching data when switching between asset type would make the user experience better, like we have on search screen.

Yeah, I agree, this would be a good addition. It loaded quite slowly when I tested it

bogdanfazakas:
> It's working well. I think a loading icon while fetching data when switching between asset type would make the user experience better, like we have on search screen.

yep good find, loading was displayed only the first time accessing the page, going to change that to every time the data is fetched from aqua

jamiehewitt15:
The loading spinner is good now.

I noticed another small part of the user experience issue that could be improved. When I clicked on the "algorithms" filter (from an account that hasn't published any algorithms) both filters disappeared. When a user has no results to display, neither filter is visible. This creates an odd user experience because my first instinct after clicking the "algorithms" filter was to remove the filter so I could see the datasets I had published from that account.

publish history filters

I would suggest it would be better to keep both filters visible even if there aren't any results to display.

jamiehewitt15 (approved):
Works great now. Thanks for making the changes

bogdanfazakas:
> Works great now. Thanks for making the changes

Thanks thanks for pointing out the strange behaviour i've missed.

PR #812, opened 2021-08-25
Web3Feedback UI messages fix

### LINKED ISSUE #777: Web3Feedback UI message
when `isGraphSynced` and `isBalanceSufficient` is equal to `false` in the code fragment below,

https://github.com/oceanprotocol/market/blob/af3bb86073a8522886341413da93532ae93a649b/src/components/molecules/Web3Feedback.tsx#L36-L55

This will appear which can be confusing to user, 

image

### Suggesting solution:
Sync up both const variable `title` and `message` to check for `isGraphSynced` first before `isBalanceSufficient`.
So it show something like this instead, 

image
^ Think above message have room to improve as well to be dynamic based on connected network instead of "Ethereum"?

KatunaNorbert:
Fixes #777 .

Changes proposed in this PR:

- look into splitting out some messages like the `isBalanceSufficient` case, as that is only needed on `Use` tab but we display it in context of all tabs
- when `isGraphSynced` and `isBalanceSufficient` is equal to `false` insufficient balance message with graph not synced title is showed
- when both `isAssetNetwork` and `isGraphSynced` are `false` prioritize the assetNetwork warning sign

### DIFF
--- a/src/components/atoms/ButtonBuy.tsx
+++ b/src/components/atoms/ButtonBuy.tsx
@@ -21,6 +21,7 @@ interface ButtonBuyProps {
   dtBalanceSelectedComputeAsset?: string
   selectedComputeAssetLowPoolLiquidity?: boolean
   selectedComputeAssetType?: string
+  isBalanceSufficient: boolean
   isLoading: boolean
   onClick?: (e: FormEvent<HTMLButtonElement>) => void
   stepText?: string
@@ -38,6 +39,7 @@ function getConsumeHelpText(
   lowPoolLiquidity: boolean,
   assetType: string,
   isConsumable: boolean,
+  isBalanceSufficient: boolean,
   consumableFeedback: string
 ) {
   const text =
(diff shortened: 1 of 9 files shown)


KatunaNorbert:
We need to prioritize some messages in case we have to display 2 messages, for example the graph is out of sync but the user doesn't have enough balance, which message should we display?

I guess the graph sync message has the lower priority since it doesn't stops the user to buy like the wrong network, insufficient balance or other does

kremalicious:
Yup, those decisions make the `Web3Feedback` component so complex in code. So would be in favor of removing complexity from it by moving some messages out as we now know some of them are really only needed in a very specific context, like insufficient balance on Use tab. Based on that, we can move the message and the logic for it out of `Web3Feedback`, and visually and in codebase put it within the Use tab content box. In there we could show it either just below that help text we have there or switch the help text for insufficient  balance message.

Might require more refactoring then current solution in here but has advantages

KatunaNorbert:
That might help, I'll try to refactor as it is and then see after that how can we reduce the complexity.
Also I think the icons are not right, we show warning for wrong network and insufficient balance and error for graph not synced, shouldn't be the other way around? Warning shouldn't stop the action

KatunaNorbert:
Moved `isBallanceSufficient` message handle inside the ButtonBuy component looks like the right spot for it and went with only show this message if it occurs.

Screenshot 2021-08-25 at 16 24 03

kremalicious (approved):

PR #813, opened 2021-08-25
Add Dockerfile

mariacarmina:
Changes proposed in this PR:

- closes [27](https://github.com/oceanprotocol/privateissues/issues/27);
- dockerized market.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,41 @@
+##
+## Copyright 2021 Ocean Protocol Foundation
+## SPDX-License-Identifier: Apache-2.0
+##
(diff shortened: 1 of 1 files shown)


mariacarmina:
After `docker run` command seems that the container is running the market app 
image

kremalicious:
at exactly is the use case for this? We do not use Docker images to run the live app. The changes in this PR will start the app in development mode not suitable for production usage so I assume this is only used for development? If so, it should be documented in README why this file exists

kremalicious:
Does not fit any valid use case

PR #817, opened 2021-08-27
Add service filters download and compute to the existing filters

### LINKED ISSUE #814: Add service filters : `compute` , `access`
Filter by service type. In the image i named `access` as `download` because it's easier to understand for users.
You can mix between the filters, meaning you can select `data set` and `access` and you will get only assets that can be downloaded and have a data set.
image

The following query will work , add it in the `must` section of the query, after `isInPurgatory`
```
 {
      "term": {
            "service.type": "compute"
          }
}
```
OR
```
 {
      "term": {
            "service.type": "access"
          }
}
```

claudiaHash:
Fixes #814 .

Changes proposed in this PR:

- add service access type filters `download` and `compute` to the existing filters

### DIFF
--- a/src/components/templates/Search/filterService.tsx
+++ b/src/components/templates/Search/filterService.tsx
@@ -56,7 +56,29 @@ export default function FilterPrice({
     navigate(urlLocation)
   }
 
+  async function applyFilter(filter: string, filterType: string) {
+    console.log('APPLY FILTER: ', filter, filterType)
+    let urlLocation = ''
+    if (filterType.localeCompare('accessType') === 0) {
+      console.log('IS ACCESS TYPE')
+      urlLocation = await addExistingParamsToUrl(location, ['accessType'])
+    } else {
+      console.log('IS SERVICE TYPE')
+      urlLocation = await addExistingParamsToUrl(location, ['serviceType'])
+    }
+
+    if (filter && location.search.indexOf(filterType) === -1) {
+      filterType === 'accessType'
+        ? (urlLocation = `${urlLocation}&accessType=${filter}`)
+        : (urlLocation = `${urlLocation}&serviceType=${filter}`)
+    }
+
+    filterType === 'accessType' ? setAccessType(filter) : setServiceType(filter)
+    navigate(urlLocation)
+  }
+
   async function handleSelectedFilter(isSelected: boolean, value: string) {
+    console.log('IS SELECTED, VALUE: ', isSelected, value)
     if (
       value === FilterByAccessOptions.Download ||
       value === FilterByAccessOptions.Compute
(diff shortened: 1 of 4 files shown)


kremalicious:
All works nicely. But here we should look into the duplication Codeclimate complains about, as `accessType` and `serviceType` are essentially doing the same. So maybe we can come up with only one handler for `serviceType` instead of separating this into 2 things with which we then do exactly the same with minimal differences

claudiaHash:
Sure, I'll manage it right now.

mihaisc:
Maybe some space between the two categories ? Might look a bit better.

claudiaHash:
Not sure how to refactor more since we use 'serviceType'  and 'accessType' filters both separate and together, if for example we want all datasets with compute access type. So I think we need to differentiate them in order to obtain a proper search query. What do you guys think?

KatunaNorbert:
I think we can change a little bit how those filters buttons work. Right now for example if we take in consideration the access type either you have no access type selected or have all selected you get the same result. We could change that, either to be able to select only one and if no one selected to get both, or have them both selected by default and if deselect them you won't get any asset.

src/components/templates/Search/filterService.tsx
```diff
@@ -91,8 +148,31 @@ export default function FilterPrice({
           </Button>
         )
       })}
+      &nbsp;&nbsp;&nbsp;
```
bogdanfazakas, line 151: i think it would be better to handle this spacing using css rather than the HTML's non-breaking space

bogdanfazakas:
Functionality wise works great, maybe we should try and tackle the issues mentioned above. Another thing that we need to decide on is if we will have the new added filters on the profile's published assets page where we rely on the existing filters updated components in this PR, otherwise we will have to handle that in another manner.

kremalicious:
> new added filters on the profile's published assets page...

yes! They were one of the problems I encountered when trying to improve assets loading over in #841 and would be a good candidate to replace them with js filtering. That is already kinda prepared as the `ProfileProvider` already holds all published assets https://github.com/oceanprotocol/market/pull/841/files#diff-2fd7fca45bccc84bae93cf604acfb3ee47fa57584a37e7c19b3a543869fcdc13R167-R171.

This way the filter mechanism in this PR would be kept for only search, which probably prevents the component from getting more complex with #774

kremalicious (approved):
All working well on my side. I'm sure we could find a more elegant css-only solution for that empty separator div but no need to bother here I think as it might get small refactor once merged into account profiles epic

bogdanfazakas (approved):
As mentioned above works good for me 👍 , and as @kremalicious suggested we can handle the refactors in a later stage for eg on the profile epic

PR #825, opened 2021-08-30
Fix Your Liquidity linked to wallet network 

### LINKED ISSUE #819: Your Liquidity only works when connected to asset network
Consider this case where I am connected to Polygon but asset is on BSC and I have liquidity in there. Then I get this with empty _Your Liquidity_:

Screen Shot 2021-08-27 at 12 27 05

The expectation would be that I always see my liquidity there, all based on `ddo.chainId` and independent of my wallet network. Also have to reload page after switching networks so _Your Liquidity_ shows up but that should not be an issue if it works from the beginning

KatunaNorbert:
Fixes #819 .

Changes proposed in this PR:

- Make Your Liquidity work independent of user wallet network

### DIFF
--- a/src/components/organisms/AssetActions/Pool/index.tsx
+++ b/src/components/organisms/AssetActions/Pool/index.tsx
@@ -68,7 +68,7 @@ const poolLiquidityQuery = gql`
 `
 
 const userPoolShareQuery = gql`
-  query PoolLiquidity($id: ID!, $shareId: ID) {
+  query PoolShare($id: ID!, $shareId: ID) {
     pool(id: $id) {
       id
       shares(where: { id: $shareId }) {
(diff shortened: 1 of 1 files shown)


kremalicious:
Works like a charm. Think we can ignore this weird super small percentage difference:

PNG image

KatunaNorbert:
Interesting, not sure why is that but I'll took a look

KatunaNorbert:
Ok so the problem is that se use the `spotPrice` to calculate `totalCreatorLiquidityInOcean` and in other places like `totalUserLiquidityInOcean` we use `price.value`
Is there any reasons for that or should I change it to `price.value`?

mihaisc:
`spotPrice` should be used everywhere. `price.value` should be used only when consuming, should not be used in any calculation

KatunaNorbert:
Ok, got it. I'll make the changes

kremalicious:
Perfect, now the final TVL is matching. I guess the `18.918 % of pool` vs `18.91 % of pool` is really only a display thing if final numbers match:

Screen Shot 2021-08-31 at 13 41 26

kremalicious (approved):

mihaisc (approved):

PR #831, opened 2021-09-01
Fix pagination breaks on to many results

### LINKED ISSUE #719: Pagination breaks if there are too many total results
It's a bug with the es query, it has limitations 10k results. Need to research scroll or search after (maybe wait for es upgrade )

KatunaNorbert:
Fixes #719 .

Changes proposed in this PR:

- Fix pagination breaks on to many results

### DIFF
--- a/src/components/molecules/PageHeader.tsx
+++ b/src/components/molecules/PageHeader.tsx
@@ -23,9 +23,7 @@ export default function PageHeader({
     <header className={styleClasses}>
       <h1 className={styles.title}>{title}</h1>
       {description && (
-        <p className={styles.description}>
-          <Markdown text={description} />
-        </p>
+        <Markdown text={description} className={styles.description} />
       )}
     </header>
   )
--- a/src/components/molecules/PageHeader.tsx
+++ b/src/components/molecules/PageHeader.tsx
@@ -1,6 +1,7 @@
 import React, { ReactElement } from 'react'
 import classNames from 'classnames/bind'
 import styles from './PageHeader.module.css'
+import Markdown from '../atoms/Markdown'
 
 const cx = classNames.bind(styles)
 
@@ -21,7 +22,11 @@ export default function PageHeader({
   return (
     <header className={styleClasses}>
       <h1 className={styles.title}>{title}</h1>
-      {description && <p className={styles.description}>{description}</p>}
+      {description && (
+        <p className={styles.description}>
+          <Markdown text={description} />
+        </p>
+      )}
     </header>
   )
 }
--- a/src/components/molecules/Pagination.tsx
+++ b/src/components/molecules/Pagination.tsx
@@ -56,7 +56,7 @@ export default function Pagination({
 
   return totalPageNumbers && totalPageNumbers > 1 ? (
     <ReactPaginate
-      pageCount={totalPageNumbers}
+      pageCount={totalPageNumbers > 476 ? 476 : totalPageNumbers}
       // react-pagination starts counting at 0, we start at 1
       initialPage={currentPage ? currentPage - 1 : 0}
       // adapt based on media query match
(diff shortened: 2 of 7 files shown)


src/pages/search.tsx
```diff
@@ -32,10 +33,19 @@ export default function PageGatsbySearch(props: PageProps): ReactElement {
       }`
 
   return (
-    <Page title={title} uri={props.uri}>
+    <Page
+      title={title}
+      description={
+        totalPagesNumber &&
+        totalPagesNumber > 476 &&
+        'More than 10k results were found, displaying the first 10k. Please refine your search'
```
jamiehewitt15, line 41: In general, it works very nicely and is a good solution. I think we could refine the message slightly as it says "60876 results for data" and then "More than 10k results were found" which seems like repetition. 

Pagination message

I would suggest a shorter message "Results are limited to the first 10k, please refine your search."

```suggestion
        'Results are limited to the first 10k, please refine your search.'
```

KatunaNorbert: Thanks for the feedback. It might sound a bit repetitive and also not sure if that number '60876'  of results makes any sense in this context.
So the title could be `More than 10k results for data` and in description 'First 10k results are displayed. Please refine your search' or simply 'Please refine your search' for description

kremalicious: agree, if we only display 10000 results then the title should reflect that, maybe `>10000 results for data`. As for the message I'm afraid people will miss it as their eyes most likely will jump to asset list results right away. Not sure how feasible it is to put an `<Alert state="info">` into that `description`, if we markdown-process that already we could at least make it bold with `**Results displayed are limited to the first 10k, please refine your search.**`

KatunaNorbert: showing an `alert` components doesn't seem too feasible without modifying some components because right now it waits for the description as a string. I don't think it's worth modifying it for this scope, because in most of the cases the user will refine the search when he sees this much results.

src/components/molecules/PageHeader.tsx
```diff
@@ -21,7 +22,11 @@ export default function PageHeader({
   return (
     <header className={styleClasses}>
       <h1 className={styles.title}>{title}</h1>
-      {description && <p className={styles.description}>{description}</p>}
+      {description && (
+        <p className={styles.description}>
+          <Markdown text={description} />
```
kremalicious, line 27: This results in invalid markup (can't have `p` add child of `p`):

Screen Shot 2021-09-13 at 15 25 10

So goal here would be to reduce nested markup as much as possible, ideally the `Markdown` component outputs as `p` and gets the `className={styles.description}`

KatunaNorbert: oh right, haven't seen that

kremalicious (approved):

src/components/molecules/Pagination.tsx
```diff
@@ -56,7 +56,7 @@ export default function Pagination({
 
   return totalPageNumbers && totalPageNumbers > 1 ? (
     <ReactPaginate
-      pageCount={totalPageNumbers}
+      pageCount={totalPageNumbers > 476 ? 476 : totalPageNumbers}
```
bogdanfazakas, line 59: it would be nice if the 476 value would be wrapped in a constant that we can use across all affected components, so if the maximum result value changes we don't need to find all the 476 values and replace them with the new total page number value and would be easier to understand in a eventual later update

bogdanfazakas (approved):
Functionality wise all good with the PR, take a look at the previous comment and see if it would be feasible otherwise good to go

src/components/molecules/Pagination.tsx
KatunaNorbert: yep this can be done, I think we can add a constant inside aquarius file and import it in those places

PR #834, opened 2021-09-02
Fix issues consuming free assets

### LINKED ISSUE #827: Issues consuming free assets
When first consuming the asset it works fine you get the datatoken, you consume the asset, you download. After that you have the download button that when you click you will get a "not enought dt" (or something similar) . The whole flow needs some testing.

KatunaNorbert:
Fixes #827 .

Changes proposed in this PR:

- Fix download free assets returns 'Not enough datatokens' error

### DIFF
--- a/src/components/atoms/Alert.tsx
+++ b/src/components/atoms/Alert.tsx
@@ -23,6 +23,7 @@ export default function Alert({
   action?: {
     name: string
     style?: 'text' | 'primary' | 'ghost'
+    disabled?: boolean
     handleAction: (e: FormEvent<HTMLButtonElement>) => void
   }
   onDismiss?: () => void
@@ -48,6 +49,7 @@ export default function Alert({
           size="small"
           style={action.style || 'primary'}
           onClick={action.handleAction}
+          disabled={action.disabled}
         >
           {action.name}
         </Button>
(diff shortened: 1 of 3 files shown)


KatunaNorbert:
Free price asset: `did:op:58D141514A59E97bCAba46547Ce5976d7b91Aef2`

jamiehewitt15:
I test this a few times and haven't seen the "not enough datatokens" error message. However, I've noticed that flow makes you go through the process of requesting a datatoken each time, I'm not sure if this is right? Ideally, the user would only have to request the datatoken once and then when they revisit the page it would just show the download button.

KatunaNorbert:
@jamiehewitt15 I couldn't reproduce that. Do you have previous orders of the asset?

KatunaNorbert:
I also disabled the `CREATE PRICING` button when wallet is not connected to asset network because it let me to go through the process and got a success message but the price was not created

jamiehewitt15:
> @jamiehewitt15 I couldn't reproduce that. Do you have previous orders of the asset?

I checked again with a different account and it wasn't happening so must be due to the previous orders, somehow

jamiehewitt15 (approved):

kremalicious (approved):

PR #836, opened 2021-09-03
[Free Pricing] Check `minterApproved` before cancelMinter

krisliew:


src/utils/freePrice.ts
```diff
@@ -7,15 +7,19 @@ export async function setMinterToPublisher(
...
+      accountId
+    )
+    if (!response) {
+      setError('Updating DDO failed.')
+      Logger.error('Failed at cancelMinter')
+    }
+    return response
   }
-  return response
+  return true
```
kremalicious, line 22: what is the reasoning behind this double return returning multiple schemas (a boolean and a response, which I guess should be `TransactionReceipt`), and returning a boolean instead of the actual response? Likewise, the return type of the method should be correctly typed as something other than `any`.

So would suggest moving around logic a bit so you don't end up with all those nested statements returning different schemas:

```js
if (!status?.minterApproved) return

const response = await ocean.OceanDispenser.cancelMinter(
 dataTokenAddress,
 accountId
)

if (!response) {
 setError('Updating DDO failed.')
 Logger.error('Failed at cancelMinter')
 return
}

return response
```
 return true

kremalicious:
I kinda forgot, but do we really have 5 transactions for editing metadata? Even if I would just edit the title? If so, this really should be investigated as I doubt we need these many transactions unless users update everything, metadata & price I guess?

jamiehewitt15:
> update: yup, just checked and this is crazy imposing all those tx onto users no matter what they edit but looks like it's a required workaround for free priced assets? Another reason they should not exist in our marketplace

Yeah, updating metadata is very awkward when free pricing is used. It's because the minter role has to be transferred back from the dispenser to the publisher and then back to the dispenser after the metadata is published.

src/utils/freePrice.ts
krisliew: I kept it as `any` because `cancelMinter` is returning `SubscribablePromise<DispenserCancelMinterProgressStep, TransactionReceipt>;` 

While the boolean return `true` is for https://github.com/oceanprotocol/market/blob/4c653b6ae86cd4c957d1b4f3d5a3b43016cec3f6/src/components/organisms/AssetActions/Edit/EditAdvancedSettings.tsx#L97

krisliew: I have updated function return type to `boolean`

kremalicious: > I kept it as any because cancelMinter is returning SubscribablePromise<DispenserCancelMinterProgressStep, TransactionReceipt>;

yes, but you are `await`ing it and do not subscribe to the `SubscribablePromise` so in your method you are producing a simple `Promise`. So the return type with my above example would be `Promise<TransactionReceipt>` which you usually get imported with `import { TransactionReceipt } from 'web3-core'`.

When then using this method you know if it succeeded or not by simply checking for a response, as all the checks in my example will return early so they will be `null`. This way you keep the methods for a wider range of use cases as maybe in another context you actually need the tx hash in the `TransactionReceipt` to display some explorer link. It's then also consistent with the other helper method `setMinterToDispenser` as both should return the same thing

krisliew: Thanks for the detail explanation, I have updated return schema to `Promise<TransactionReceipt>` for `setMinterToDispenser`. As for `setMinterToPublisher` It require an addition `boolean` return schema for the code fragment mentioned above. 

I think market is ok to keep it returning 2 different schema temporary for now, since this function won't be required in coming v4.

Issue #769, opened 2021-08-02
Search gives unexpected result for data token symbol


TimDaub:
Try the following:

1. Take any known data set's symbol, e.g. RUMNAR-48, and type it into the search of market.oceanprotocol.com.
2. The result will surprise you; mainly because in this specific case RUMNAR-48 is not presented as a confident first result

See screenshot:

<img width="1153" alt="Screen Shot 2021-08-02 at 15 28 27" src="https://user-images.githubusercontent.com/2758453/127869457-d9d3fe20-257f-415a-a93c-8efcc5a7a11f.png">

It'd in the search results - but what'd be less confusing or more pleasing to the users is showing it as result number 1 and potentially also removing the other false-positive results.

kremalicious:
we have a greatly enhanced search query in #755 but can confirm that your exact example behaves the same there. Fun thing is that for any other picked symbol I really get only 1 result over there but not for `RUMNAR-48`

This specific issue might be also solved in combination with #699 where then I would expect based on relevancy to at least get `RUMNAR-48` as first result

jamiehewitt15:
This happens because elastic search removes the hyphen when it analyses the text. I think getting `RUMNAR-48` as the first result should be a satisfactory result, I don't think we necessarily need to remove all the other results. 

If this doesn't get solved in #699, I'm happy to work on another search PR to address this

alexcos20:
@TimDaub - can you check again ?

TimDaub:
![Screenshot_20211005-122140](https://user-images.githubusercontent.com/2758453/136005534-59bc905e-f182-45e5-bb97-4e85fa017cbf.jpg)

The screenshot shows that RUMNAR-48 comes first. Anyways IMO when a symbol is searched, I think RUMNAR-48 should be specificallly highlighted to signal the user that indeed the token symbol that hey're looking for exists.

FYI: There are now 8 false positives instead of 3. So maybe this fix was a regression?

kremalicious:
searching for any datatoken symbol always presents this as the first result so not sure what exactly the issue here is anymore. So closing

PR #837, opened 2021-09-03
Bump aquarius version

### LINKED ISSUE #679: Add relevance sort
Add back `Relevance` sort when it is fixed in aquarius.
Basically just remove the comments.

https://github.com/oceanprotocol/aquarius/issues/493

mihaisc:
The new version of aqua (not sure what version it will be) changes a few things in the endpoint structure and request/response.
I updated the models and enabled sorting by score.

ToDO: 

- [x] test and approve pr
- [x] remove `aquarius2` from `app.config.js`
- [ ] swap aquarius with aquarius 2
- [ ] merge pr

Testing with https://www.npmjs.com/package/@elastic/elasticsearch seems to be promising as importing only the types does not import the whole library thus we keep it compact.  --- No go, the types are made for the library to use, too cumbersome for us.

Another alternative would be https://www.npmjs.com/package/@types/elasticsearch , but will not be handled in this issue.

For now i just left it with `any` so we can migrate to the new version of aqua.

Closes #679 as well


src/components/templates/Search/index.tsx
```diff
@@ -33,14 +34,13 @@ export default function SearchPage({
 
   useEffect(() => {
     if (!appConfig.metadataCacheUri) return
+
+    const source = axios.CancelToken.source()
```
kremalicious, line 38: We have to use the cancel token within this `useEffect`, in its`return` callback so that unmounting it really cancels the request

src/components/templates/Search/index.tsx
```diff
@@ -33,14 +34,13 @@ export default function SearchPage({
 
   useEffect(() => {
     if (!appConfig.metadataCacheUri) return
```
kremalicious, line 35: If we don't pass `appConfig.metadataCacheUri` anymore we also don't need to check for it up here

kremalicious (changes requested):

kremalicious:
Functionality-wise all seems to work well, I could not break it and results indeed seem more relevant. 

As side note regarding #769, result is now on 2nd place so improvement from before
https://market-git-feature-bump-aquarius-oceanprotocol.vercel.app/search?sort=_score&sortOrder=desc&text=RUMNAR-48

And indeed lots of breaking changes for end users of our remote Aquarius. Gonna have a hard time explaining why a simple patch release breaks all existing search implementations

mihaisc:
Regarding #769 i did some quick tests and it's more complicated that i thought, we need to rethink a bit the query and maybe the types in es.

kremalicious (approved):

bogdanfazakas (approved):
works good on my side, there is a conflict after the latest merges into main that needs some attention

PR #840, opened 2021-09-06
Fix Profile sold value

### LINKED ISSUE #829: Sold value is wrong
New account header developed in #776 has a wrong _Sold_ value.

From [over there](https://github.com/oceanprotocol/market/pull/776#issuecomment-906241156):

> To fetch the correct number of sold you need to fetch the list of published did/datatokenId and for each get the list of orders. Sum that and you have total sold.

```graphql
{
  tokenOrders(where: { datatokenId: "0x009d43f9ec69b9ba40aef305c7216e793239627c" }) {
    payer {
      id
    }
    datatokenId {
      id
    }
  }
}
```

bogdanfazakas:
Fixes #829  .

Changes proposed in this PR:

- updated sold value calculation logic

### DIFF
--- a/src/components/pages/Profile/Header/Stats.tsx
+++ b/src/components/pages/Profile/Header/Stats.tsx
@@ -52,17 +52,21 @@ export default function Stats({
         const result = await queryMetadata(queryPublishedAssets, source.token)
         setPublishedAssets(result.results)
         setNumberOfAssets(result.totalResults)
+        const nrOrders = await getAccountNumberOfOrders(
+          result.results,
+          chainIds
+        )
+        setSold(nrOrders)
       } catch (error) {
         Logger.error(error.message)
       }
     }
     getPublished()
 
-    async function getAccountSoldValue() {
-      const nrOrders = await getAccountNumberOfOrders(accountId, chainIds)
-      setSold(nrOrders)
-    }
-    getAccountSoldValue()
+    // async function getAccountSoldValue() {
+    //
+    // }
+    // getAccountSoldValue()
   }, [accountId, chainIds])
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


kremalicious:
Definitely more complete numbers than before, just double checking so let's use one example and look at chain.

This profile with 1 published asset for me says 5 sold:
https://market-git-fix-issue829-wrong-sold-value-oceanprotocol.vercel.app/profile/0x655eFe6Eb2021b8CEfE22794d90293aeC37bb325

Then all _Start Order_ with 0.001 and 0.999 transfer are combined 1 sale:
https://etherscan.io/token/0x7Bce67697eD2858d0683c631DdE7Af823b7eea38

Screen Shot 2021-09-07 at 10 29 52

And it all checks out when counting through the pages, great! Although there is one more lone _Start Order_ where only 0.001 was transferred so not sure what exactly happened there and if this was a sale or not as then would be 6 sold:
https://etherscan.io/tx/0x9d0e9a851cd7ad6c486bdea15970bcfd815ca43443d868e123752939efb8af66

src/components/pages/Profile/Header/Stats.tsx
```diff
@@ -52,17 +52,21 @@ export default function Stats({
...
       }
     }
     getPublished()
 
-    async function getAccountSoldValue() {
-      const nrOrders = await getAccountNumberOfOrders(accountId, chainIds)
-      setSold(nrOrders)
-    }
-    getAccountSoldValue()
+    // async function getAccountSoldValue() {
```
kremalicious, line 66: we can remove those comments

bogdanfazakas: busted 😄 , forgot to delete it

mihaisc (approved):

PR #841, opened 2021-09-07
add total user liquidity, new Profile provider

### LINKED ISSUE #839: TVL value is incomplete
Checking out [this profile](https://market-git-feature-accountpage-oceanprotocol.vercel.app/profile/0xbbd33AFa85539fa65cc08A2e61a001876D2f13FE) and if you switch to Pool Shares with multiple networks selected in network selector, you will see that numbers don't add up and the TVL display in account header is only accounting for ETH Mainnet in this case:

Screen Shot 2021-09-06 at 12 38 40

Expectation is that selecting/deselecting networks also recalculates TVL number. As far as I can see the effects are setup to refetch on `chainIds` changes but seems `getAccountLiquidityInOwnAssets()` helper has a bug

kremalicious:
- closes #839
- [new `ProfileProvider`](https://github.com/oceanprotocol/market/pull/841#issuecomment-915067025) wrapping profile pages, reducing lots of requests
- new stat: _Total Liquidity_
- renamed stat: _Liquidity in Own Assets_
- #849

Example:

Screen Shot 2021-09-07 at 13 22 21

https://market-git-feature-accountpageliquidity-oceanprotocol.vercel.app/profile/0xbbd33AFa85539fa65cc08A2e61a001876D2f13FE

### DIFF
--- a/src/components/atoms/Tooltip.module.css
+++ b/src/components/atoms/Tooltip.module.css
@@ -9,6 +9,10 @@
   font-size: var(--font-size-small);
 }
 
+.content p {
+  margin: 0;
+}
+
 .icon {
   width: 1em;
   height: 1em;
--- a/src/components/atoms/Tooltip.tsx
+++ b/src/components/atoms/Tooltip.tsx
@@ -5,6 +5,7 @@ import { useSpring, animated } from 'react-spring'
 import styles from './Tooltip.module.css'
 import { ReactComponent as Info } from '../../images/info.svg'
 import { Placement } from 'tippy.js'
+import Markdown from './Markdown'
 
 const cx = classNames.bind(styles)
 
--- a/src/components/molecules/AssetListTitle.tsx
+++ b/src/components/molecules/AssetListTitle.tsx
@@ -1,5 +1,4 @@
 import { DDO } from '@oceanprotocol/lib'
-import { useOcean } from '../../providers/Ocean'
 import { Link } from 'gatsby'
 import React, { ReactElement, useEffect, useState } from 'react'
 import { getAssetsNames } from '../../utils/aquarius'
--- a/src/components/molecules/AssetListTitle.tsx
+++ b/src/components/molecules/AssetListTitle.tsx
@@ -42,7 +42,7 @@ export default function AssetListTitle({
 
   return (
     <h3 className={styles.title}>
-      <Link to={`/asset/${did || ddo.id}`}>{assetTitle}</Link>
+      <Link to={`/asset/${did || ddo?.id}`}>{assetTitle}</Link>
     </h3>
   )
 }
--- a/src/components/molecules/AssetTeaser.tsx
+++ b/src/components/molecules/AssetTeaser.tsx
@@ -13,11 +13,13 @@ import { BestPrice } from '../../models/BestPrice'
 declare type AssetTeaserProps = {
   ddo: DDO
   price: BestPrice
+  noPublisher?: boolean
 }
 
 const AssetTeaser: React.FC<AssetTeaserProps> = ({
   ddo,
-  price
+  price,
+  noPublisher
 }: AssetTeaserProps) => {
   const { attributes } = ddo.findServiceByType('metadata')
   const { name, type } = attributes.main
@@ -34,7 +36,9 @@ const AssetTeaser: React.FC<AssetTeaserProps> = ({
           <Dotdotdot clamp={3}>
             <h1 className={styles.title}>{name}</h1>
           </Dotdotdot>
-          <Publisher account={owner} minimal className={styles.publisher} />
+          {!noPublisher && (
+            <Publisher account={owner} minimal className={styles.publisher} />
+          )}
         </header>
 
         <AssetType
--- a/src/components/molecules/Bookmarks.tsx
+++ b/src/components/molecules/Bookmarks.tsx
@@ -5,10 +5,7 @@ import { Logger } from '@oceanprotocol/lib'
 import Price from '../atoms/Price'
 import Tooltip from '../atoms/Tooltip'
 import AssetTitle from './AssetListTitle'
-import {
-  queryMetadata,
-  transformChainIdsListToQuery
-} from '../../utils/aquarius'
+import { getAssetsFromDidList } from '../../utils/aquarius'
 import { getAssetsBestPrices, AssetListPrices } from '../../utils/subgraph'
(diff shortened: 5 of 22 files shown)


kremalicious:
As seen in example would work with 2 rows of stats but should be meaningful ones and right now would be fine to just leave out _Published_ as that's obvious from the first history tab anyway

kremalicious:
Also noticed that in every pool share we get a `poolid.valueLocked` which looks like what we construct in [our helper method](https://github.com/oceanprotocol/market/blob/feature/account_page/src/utils/subgraph.ts#L592-L601) anyway: 

Screen Shot 2021-09-07 at 13 29 42

This looks like we can simply use this value and correlate with user's pool share number to get user liquidity. But then again, isn't that TVL anyway?

kremalicious:
That ineffective triple fetching of pool shares was bugging me so started to move some calls into a new `ProfileProvider` which wraps the profile page exposing this interface:

```ts
interface ProfileProviderValue {
  profile: Profile
  poolShares: PoolShare[]
  isPoolSharesLoading: boolean
  assets: DDO[]
  assetsTotal: number
  isEthAddress: boolean
}
```

Works well for 3box & pool shares subgraph query which then also speeds up loading of pool shares tab. It centralizing some fetching and refetching so this can stay out of components.

My goal was actually to have it hold the published assets too so that when users switch between tabs back to _Published_ app would not refetch from Aquarius but had to drop that plan as we create too many different asset structures (`assets`, `assetsWithPrices` and so on) and the current `AssetList` is setup to always fetch no matter what. Started refactoring all this but quickly went down rabbit hole so just dropped this. It would only make sense if the provider would hold all the required assets and the subcomponents filter against it instead of sending new queries to Aquarius. Only then we would have a UI where there is only an initial loading but then tabbing through tabs would be instant. But not feasible without huge refactor.

mihaisc:
Value locked is indeed tvl, not liquidity

mihaisc:
when i switch to the pool shares tab i seem to get a lot of requests. From a quick glance it seems it calls the `names` enpoint twice for the same did (not preflight, actual posts)
image

kremalicious:
> but in multiple places i see the cancelToken is used wrong

yes, but to me it seems the linked usage is actually correct, what's important is to create `cancelToken` outside of helper method in the respective `useEffect`, which I've also done with the published assets lists in this PR in c8e0b69775856b80ab3204bdb341afa5ff4a71f3. But also having problems getting this done for `AssetListTitle` which still produces console errors upon tab changing

> Also maybe we can do something with the 3box profile. Basically we load it for the header, and then load it for every asset.

took the easy way and simply removing publisher on asset teasers on account pages, removing those calls

> but in pool transactions we fetch both name and ddo for each row, generating a lot of calls. We need to aggregate the did's and make one call to fetch the names.

yup, also logged this in #842 and would be quick win without much refactor.

But we have lots of other places coming together here on account page which have rather ineffective data handling, where data is fetched multiple times and such. Still tinkering but it really is tricky as we have lots of subcomponents doing their own fetching with their own data model and while we could make them not fetch when some existing data is passed via props, having hard time unifying data models fetched outside and what component fetching requires.

I think only holistic way to solve this is to come up with a bunch of data structures for e.g. all assets which combine Aquarius & subgraph data like asset prices, and handle fetching, refetching from Aquarius in some global provider. And then subcomponents work and filter against that instead of fetching data from Aquarius. Bonus would be way faster interface flows as not every click here in history would initiate some refetching from Aquarius directly

kremalicious:
phew, think I was able top remove all the asset name fetching as we always have the full DDOs fetched before that so now that is the source for all asset titles in all tables. Remaining ineffectiveness is then the DDO fetching per row which we should handle independently in #842

mihaisc:
It is not correct, the token is just added in the body, check the request body and you will find it there. The correct usage is this: 
```
await axios.post(
      `${metadataCacheUri}/api/v1/aquarius/assets/names`,
      { didList } ,
      { cancelToken } 
    )
```

kremalicious:
oh! Now I get it, thanks. So we actually had 2 false usages of cancel token:

1. wrong init of token source in the method instead of component.
2. wrong passing into `axios` methods

mihaisc:
Yep, you are right, we need to refactor everything. Like you said we need to implement some sort of combination of mvc with mvvm that suits us. Separate the calls to different data providers, create a unified view model and have components that only display data.

bogdanfazakas:
Nice improvements with the profile provider 😃 . While testing discovered that the profile statistics (liquidity in own assets and sold) are having a strange comportment when select different network combinations from the network switcher, also for the profile provided in the description when selecting also Rinkeby test network the pool shares return no results and an error is thrown in the console even tho there should have been shares. I will test this also on the main epic's branch and if the issues appears also there I will create separate issues so don't add more to this PR

kremalicious:
right, that is an open question what to do with test networks. I would argue to not include them in any money-based statistics, just like we do not list them in the stats footer as they are test networks. So for the stats here would remove their value from _Liquidity in Own Assets_ and _Total Liqudity_, but keep them in _Sales_ and _Published_ when they are selected. Or maybe we should ditch them from the profile stats completely although that would create some mismatch between header & tables

kremalicious:
And another thing I'm trying to solve is this where some periodic refetching outputs `undefined` in console, 19 times, and I just don't know what is causing this as console logs are quite unhelpful. 

Screen Shot 2021-09-09 at 12 49 10

mihaisc:
Shouldn't we handle the network like we do everywhere else, so we show what the user selects. Not sure why we should handle the networks differently. So default flow would be if you are  1st time on market then only main networks are selected (this is in the network component) and you see the profile for the main networks. But you should be free to customize this. This is what means to "support" the networks.

kremalicious:
> when select different network combinations from the network switcher, also for the profile provided in the description when selecting also Rinkeby test network the pool shares return no results

yup, can reproduce and not sure yet for which value we have to wait more. Maybe we have to check for length differences of `chainIds` in some places, or we have to add some debounce somewhere as it seems there's a delay between ticking the box in network selection and update of `chainIds` in `UserPreferencesProvider`. Also quickly selecting/unselecting networks leads to incomplete data somehow.

> Shouldn't we handle the network like we do everywhere else, so we show what the user selects.

It's also way easier as we don't have to do anything right now. Despite above mentioned network switch problems, all components and the provider are setup to change on based on network selection.

PR #846, opened 2021-09-08
Error when switching to a network that is not added in Metamask

### LINKED ISSUE #832: Error when switching network to a network that is not added in Metamask as custom RPC
### Steps to reproduce:
1) Make sure metamask is connected to Ethereum Mainnet
2) Delete BSC Mainnet from your metamask RPC if you have that custom RPC.
3) Go to an asset page that is published on BSC Mainnet (https://market.oceanprotocol.com/asset/did:op:2a82D20D65565Be093576Ed9000c4b853D005be3)
4) Click `Switch Network` 
5) Error:
image

### Expected behaviour:
Not sure why but code should have added the custom RPC as `wallet_addEthereumChain` below
https://github.com/oceanprotocol/market/blob/4123dae431f67548741b7062ab0bae6fb31d5c2c/src/utils/web3.ts#L99-L141

### Possible cause: 
Correct me if I am wrong but I think is likely the RPC endpoint is not responding to RPC calls of `eth_chainId`?  
See https://docs.metamask.io/guide/rpc-api.html#description-6 for more information.

KatunaNorbert:
Fixes #832.

### DIFF
--- a/src/utils/web3.ts
+++ b/src/utils/web3.ts
@@ -43,7 +43,7 @@ export function getNetworkConfigObject(node: any): NetworkObject {
     name: node.chain,
     symbol: node.nativeCurrency.symbol,
     chainId: node.chainId,
-    urlList: [node.nodeUri]
+    urlList: node.rpc
   }
   return networkConfig
 }
--- a/src/utils/web3.ts
+++ b/src/utils/web3.ts
@@ -43,7 +43,7 @@ export function getNetworkConfigObject(node: any): NetworkObject {
     name: node.chain,
     symbol: node.nativeCurrency.symbol,
     chainId: node.chainId,
-    urlList: [node.providerUri]
+    urlList: [node.nodeUri]
   }
   return networkConfig
 }
--- a/src/utils/web3.ts
+++ b/src/utils/web3.ts
@@ -10,19 +10,22 @@ export interface EthereumListsChain {
   networkId: number
   nativeCurrency: { name: string; symbol: string; decimals: number }
   rpc: string[]
-  faucets: string[]
   infoURL: string
+  faucets: string[]
+  explorers: [{ url: string }]
 }
 
 export interface NetworkObject {
   chainId: number
   name: string
+  nativeCurrency: string
+  explorers: [{ url: string }]
   urlList: string[]
 }
 
 const configGaiaX = getOceanConfig(2021000)
 
-export const networkDataGaiaX = {
+export const networkDataGaiaX: EthereumListsChain = {
   name: 'GAIA-X Testnet',
   chainId: 2021000,
   shortName: 'GAIA-X',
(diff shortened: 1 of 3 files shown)


KatunaNorbert:
There is the same problem with Polygon network, but there it seems like the `nodeUri` from oceanJs is wrong, so I'll create an issue for that

KatunaNorbert:
Also worth testing for: `Polygon Mumbai` and `Moonbase Alpha` but no assets published on those networks yet

kremalicious:
That `providerUri` → `nodeUri` change we definitely need.

As mentioned over in https://github.com/oceanprotocol/ocean.js/issues/1010 using the infura URL with the infura key used on live `market` should work™. But also raises question if we want that, or if it would be better to inject something different from `oceanConfig.nodeUri` like `https://rpc-mainnet.maticvigil.com` just for some cases

kremalicious:
WhatWouldBalancerDo: 
- https://github.com/balancer-labs/frontend-v2/blob/develop/src/lib/config/polygon.json
- https://github.com/balancer-labs/frontend-v2/blob/a19a6a1e6b388c361708d85a636d76467bf85111/src/services/web3/utils/helpers.ts#L4-L26
- https://github.com/balancer-labs/frontend-v2/blob/a19a6a1e6b388c361708d85a636d76467bf85111/src/services/web3/utils/helpers.ts#L28-L67

Instead of the infura `rpc` they use for their app, they define another `publicRpc` to be used for adding network to wallet. Which then would be a good candidate to handle https://github.com/oceanprotocol/ocean.js/issues/1010 by adding optional `nodeUriPublic` or something to some network configs in ocean.js

mihaisc:
Do we need to complicate ourselves with ocean.js, this is a purely market feature. Not sure how to handle this.

KatunaNorbert:
Right, getting it from network metadata seems to be the best solution

src/utils/web3.ts
```diff
@@ -43,7 +43,7 @@ export function getNetworkConfigObject(node: any): NetworkObject {
     name: node.chain,
     symbol: node.nativeCurrency.symbol,
     chainId: node.chainId,
-    urlList: [node.providerUri]
+    urlList: node.rpc
```
kremalicious, line 46: typings should have caught this but `const networkConfig` should be `const networkConfig: NetworkObject`, then TypeScript would correctly complain here as `urlList` is defined as an array at top of file in `NetworkObject`.

KatunaNorbert: but `urlList` is an array, just now we get the array from `EthereumListsChain` and no need to be created over here as it was before

kremalicious: Ah you're right, my bad. Did not see that node.rpc is an array too

kremalicious:
It seems to work now, could add Polygon after clicking button on Polygon asset with fresh MetaMask. But I ended up with this incomplete network where it's the wrong symbol (should be MATIC), block explorer is missing, and we should be consistent to our presented network names (think there's a helper method for that) so in this case it should be _Polygon_ and NOT _Matic(Polygon)_:

Screen Shot 2021-09-13 at 16 33 14

KatunaNorbert:
The error occurs because it tries to switch networks first and there you got this error and after that it tries to add the new network.
Good catch about the incomplete network configuration, I'll look for a fix.

bogdanfazakas (approved):
working well on my end, I also get the error mentioned above not sure how we can get rid of it

claudiaHash (approved):
same for me!

PR #848, opened 2021-09-08
Tokens for swapping are wrong in Trade page

### LINKED ISSUE #845: Trade: tokens for swapping are wrong
On some pool assets, seen in ETH Mainnet & BSC, only the datatoken shows up in the swap interface, where in this case at the top it should be OCEAN:

Screen Shot 2021-09-08 at 12 08 09

Examples:
- https://market.oceanprotocol.com/asset/did:op:5a376380D8114D871A7d59D05C888e090a929862
- https://market.oceanprotocol.com/asset/did:op:2a82D20D65565Be093576Ed9000c4b853D005be3

Not clear yet if only a UI display thing or if indeed we would try to swap datatoken against datatoken in those examples.

KatunaNorbert:
Fixes #845 .

### DIFF
--- a/src/utils/subgraph.ts
+++ b/src/utils/subgraph.ts
@@ -121,7 +121,7 @@ const AssetPoolPriceQuerry = gql`
       datatokenAddress
       datatokenReserve
       oceanReserve
-      tokens {
+      tokens(where: { isDatatoken: false }) {
         symbol
       }
     }


claudiaHash (approved):
Surprisingly short solution, but seems to work great.

kremalicious (approved):
it works! 1 line changes are the best

Issue #842, opened 2021-09-07
pools shares DDO fetching is ineffective


kremalicious:
We do a `retrieveDDO` for each and every pool share in this method:
https://github.com/oceanprotocol/market/blob/feature/account_page/src/components/pages/Profile/History/PoolShares.tsx#L180-L199

Should be refactored to rather use an Elasticsearch query to get the needed assets in one request. This will dramatically speed up the pool shares table.

PR #849, opened 2021-09-09
Add total downloads, speed up downloads fetching

kremalicious:
Add new stats number for total downloads account has done. All fetching for it centralized in `ProfileProvider` so as a bonus clicking on _Downloads_ tab is instant as data was alreeady fetched in background.

Screen Shot 2021-09-09 at 17 28 16

So those individual `retieveDDO()` became really irresponsible so also addressed what's described in #842, replacing multiple retrieveDDO for each row with one single request. The helper methods for this then can be used to solve #842 too

### DIFF
--- a/src/utils/aquarius.ts
+++ b/src/utils/aquarius.ts
@@ -256,12 +256,16 @@ export async function getAssetsFromDidList(
   cancelToken: CancelToken
 ): Promise<QueryResult> {
   try {
+    // TODO: figure out cleaner way to transform string[] into csv
     const searchDids = JSON.stringify(didList)
       .replace(/,/g, ' ')
       .replace(/"/g, '')
       .replace(/(\[|\])/g, '')
       // for whatever reason ddo.id is not searchable, so use ddo.dataToken instead
       .replace(/(did:op:)/g, '0x')
+
+    // safeguard against passed empty didList, preventing 500 from Aquarius
+    if (!searchDids) return
 
     const query = {
       page: 1,
--- a/src/utils/aquarius.ts
+++ b/src/utils/aquarius.ts
@@ -250,14 +250,11 @@ export async function getPublishedAssets(
   }
 }
 
-export async function getDownloadAssets(
+export async function getAssetsFromDidList(
   didList: string[],
-  tokenOrders: OrdersData[],
   chainIds: number[],
   cancelToken: CancelToken
-): Promise<DownloadedAsset[]> {
-  const downloadedAssets: DownloadedAsset[] = []
-
+): Promise<QueryResult> {
   try {
     const searchDids = JSON.stringify(didList)
       .replace(/,/g, ' ')
@@ -266,7 +263,7 @@ export async function getDownloadAssets(
       // for whatever reason ddo.id is not searchable, so use ddo.dataToken instead
       .replace(/(did:op:)/g, '0x')
 
-    const queryDownloads = {
+    const query = {
       page: 1,
       offset: 100,
       query: {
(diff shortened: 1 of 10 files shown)


src/utils/aquarius.ts
```diff
@@ -240,3 +249,57 @@ export async function getPublishedAssets(
...
+  cancelToken: CancelToken
+): Promise<DownloadedAsset[]> {
+  const downloadedAssets: DownloadedAsset[] = []
+
+  try {
+    const searchDids = JSON.stringify(didList)
+      .replace(/,/g, ' ')
+      .replace(/"/g, '')
+      .replace(/(\[|\])/g, '')
+      // for whatever reason ddo.id is not searchable, so use ddo.dataToken instead
```
kremalicious, line 266: this I'm not sure as I discovered this pattern here in our `Bookmarks` component. So worth checking if we can simply search for `ddo.id` by now with all the latest Aquarius changes

claudiaHash:
On the profile stats everything works nicely. On the Downloads section table seems to be a problem: when switching from one page to another I firstly can see the results but seconds later I got No results found. It doesn't seem to depend on which or how many networks are selected.

Screenshot from 2021-09-09 19-04-00

kremalicious:
could you try again with latest preview please? Just refactored everything again and also had this problem of empty results but for me seems gone now locally

claudiaHash:
> could you try again with latest preview please? Just refactored everything again and also had this problem of empty results but for me seems gone now locally

yes, everything it's fine now!

kremalicious:
so far so good, all much snappier but one weird find which also Claudia has: there's a difference in downloads after this PRs change and what is currently on live. For me all checks out when only main networks are selected but when Rinkeby is put in the mix on live my downloads list is longer than with this PR.

I suspect the new pattern replacing individual `retrieveDDO()` does not search everything

kremalicious:
So I guess Aquarius has some mechanism to prevent duplicates in its query responses. Cause here we would actually need duplicates as we send a list of DIDs based on token orders, and there will be duplicates in that DID list which is intended. But Aquarius response then kills them all, returning each asset only once. Resulting then in wrong downloads count.

That might be a reason we actually cannot replace `retrieveDDO()` with one single query, unless Elasticsearch/Aquarius has something in place to return duplicates. This most likely would also affect #842 and all other places we use that ineffective fetching pattern

kremalicious:
So 040eba4 should do it, I now get the same list in this PR and on live

claudiaHash (approved):

mihaisc:
Now that i see the total i realize these are not downloads, they are orders. The difference? You can order and then click download 20 times and actually download the file 20 times , but you will have only one entry in the table/total.

mihaisc (approved):

kremalicious:
good point. I mean we could be precise and label it as _Download Orders_ which has risks of raising questions in users but would also inspire learning about our download process. On other hand we labeled it as _Downloads_ the whole time.

Regardless, @bogdanfazakas are you doing something with #841 branch? Did this here as additional PR against it cause you needed some branch without further updates, so let me know if merging would be fine. Or was it the epic PR #774 which should stay as is for bug fixing?

mihaisc:
We can call it _Orders_ or _Ordered_. Definitely not  _Download Orders_, this is most confusing of both variants.

kremalicious:
naming is so hard. Cause orders are happening for either `access` or `compute` service so using only _Orders_ in a way would be misleading. Tending towards keeping it as _Downloads_ and sounds like we might wanna add some tooltips to some or all stats labels if we want to, as most likely we won't find one word matching tech and user expectation.

Also as a note, pretty sure there's a potential to speed up compute jobs too as they are just different token orders and it's using almost same subgraph query and just filters results differently. `ProfileProvider` in here already has user token orders at some point so we could use it in compute jobs too, but didn't touch all this to keep PR focused

mihaisc:
Yes on both statements. The first part of fetching the compute jobs is the same as downloads but with a different filter. We should delay refactoring the compute jobs because we will have to do that once we merge the prs in ocean.js, so it would be a waste of time to tackle it twice.

PR #854, opened 2021-09-14
Fix liquidity in own assets and sales broken statistics on network switching 

### LINKED ISSUE #852: Liquidity in Own Assets andSales statistics brakes randomly on network switching while on a profil page
**Describe the bug**
When switching networks while on a profile page the liquidity in own statistics and sales statistics randomly became 0 even thou for that specific combination the value should be bigger than 0

**To Reproduce**
1. Select only ETH mainnet network from network switcher
2. Go to a profile page for eg (https://market-nezhlvcae-oceanprotocol.vercel.app/profile/0x655eFe6Eb2021b8CEfE22794d90293aeC37bb325)
3. Select another mainnet network
4. Check if the mentioned statistics became 0 
5. Otherwise keep select and deselecting mainnet network until the issue is reproduced

https://user-images.githubusercontent.com/13447142/133213962-c36f97e3-27f3-467c-8a2a-1791b6ee0195.mov

bogdanfazakas:
Fixes #852 .

Changes proposed in this PR:

- removed `chainId` as a dependency on `useEffect()` used for number of **sales** and the **liquidity in own assets** statistics since we already have there the user assets dependency and the values should change only when the published assets is changed.

### DIFF
--- a/src/components/pages/Profile/Header/Stats.tsx
+++ b/src/components/pages/Profile/Header/Stats.tsx
@@ -59,7 +59,7 @@ export default function Stats({
       }
     }
     getSales()
-  }, [accountId, chainIds, assets])
+  }, [accountId, assets])
 
   useEffect(() => {
     if (!assets || !accountId || !chainIds) return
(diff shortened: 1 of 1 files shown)


KatunaNorbert (approved):
Looks good!

claudiaHash (approved):
Good to go!

kremalicious:
nice, one of those times when we have to ignore the react hooks rules cause we know better 😄 

Screen Shot 2021-09-15 at 12 11 47

Issue #722, opened 2021-07-12
Switch typings generation from Apollo to urql


kremalicious:
Addendum to #608, we need to switch the typings generation from Apollo to urql

PR #860, opened 2021-09-16
ENS names

### LINKED ISSUE #858: Deal with ENS names
This should work for kicking off profile page fetching but think we might prevent fetching in this case:
https://market-git-feature-accountpage-oceanprotocol.vercel.app/profile/officialoceandao.eth

As a generic solution, if an ETH address is attached to an ENS name, it should take over in all display which is an industry standard as per [ENS design guidelines](https://docs.ens.domains/dapp-developer-guide/front-end-design-guidelines). So instead of `0xxx` we show e.g. `domain.eth` in profile page, user wallet, publisher line. So all logic for showing wallet addresses in UI would be:

```
ENS name || ETH address
```

3box names are still valid for the main profile title with this logic:

```
 3box profile name || ENS name || ETH address
```

A good start: https://docs.ens.domains/dapp-developer-guide/resolving-names

kremalicious:
All conception and process thinking in #858

- fetch ENS names against [ENS subgraph](https://thegraph.com/legacy-explorer/subgraph/ensdomains/ens?selected=playground) in multiple places:
  - user wallet display (reasoning against using ENS js libraries for this in #858)
  - publisher line on asset teasers, asset details
  - use for decentralized vanity profile URLs, enhance profile data with it
  - logic on `/profile` route supports both, ETH address or ENS name as end of URL path and will resolve accordingly to kick off all child components
- new `ens.ts` util methods for getting name or resolved addresses
- closes #858
- might close https://github.com/oceanprotocol/multi-repo-issue/issues/117

some preview example profiles:

- https://market-git-feature-accountpageens-oceanprotocol.vercel.app/profile/dataunion.eth
- https://market-git-feature-accountpageens-oceanprotocol.vercel.app/profile/cloutcontracts.eth

as for original reported URL in #858, this now works and everything on the profile is mapped to resolved address as you can see on the explorer links:

- https://market-git-feature-accountpageens-oceanprotocol.vercel.app/profile/officialoceandao.eth

Screen Shot 2021-09-16 at 14 27 20

Example for ENS name + 3box info:

133775275-451b9bd8-c178-47d3-8529-f6809ce58829

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -8,10 +8,10 @@ execSync(`node ./scripts/write-repo-metadata > repo-metadata.json`, {
 })
 
 // Generate GraphQl typings for urql
-execSync(`npm run graphql:graphTypes`, { stdio: 'inherit' })
+// execSync(`npm run graphql:graphTypes`, { stdio: 'inherit' })
 
 // Generate Apollo typings
-// execSync(`npm run apollo:codegen`, { stdio: 'inherit' })
+execSync(`npm run apollo:codegen`, { stdio: 'inherit' })
 
 // Fetch EVM networks metadata
 execSync(
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -8,10 +8,10 @@ execSync(`node ./scripts/write-repo-metadata > repo-metadata.json`, {
 })
 
 // Generate GraphQl typings for urql
-// execSync(`npm run graphql:graphTypes`, { stdio: 'inherit' })
+execSync(`npm run graphql:graphTypes`, { stdio: 'inherit' })
 
 // Generate Apollo typings
-execSync(`npm run apollo:codegen`, { stdio: 'inherit' })
+// execSync(`npm run apollo:codegen`, { stdio: 'inherit' })
 
 // Fetch EVM networks metadata
 execSync(
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@coingecko/cryptoformat": "^0.4.4",
+        "@ensdomains/ensjs": "^2.0.1",
         "@loadable/component": "^5.15.0",
         "@oceanprotocol/art": "^3.2.0",
         "@oceanprotocol/lib": "^0.17.3",
@@ -30,6 +31,7 @@
         "dotenv": "^10.0.0",
         "ethereum-address": "0.0.4",
         "ethereum-blockies": "github:MyEtherWallet/blockies",
+        "ethereum-ens": "^0.8.0",
         "filesize": "^6.3.0",
         "formik": "^2.2.9",
         "gatsby": "^2.32.13",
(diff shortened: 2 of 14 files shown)


kremalicious:
ah great, moving forward with the subgraph variant in here would require us to figure out a way to generate subgraph typings from our own subgraphs, and external ones. Before that, deployments and tests will fail.

mihaisc:
In this case we can just write them. Test fail because we still use that apollo command, we should remove that at some point.

kremalicious:
those typings... So yes to kicking out Apollo and switching typings generation to urql as defined in #722. It's already prepared in code, the commands are there and I quickly tried to switch it in here but then needs lots of changes to source the newly generated typings but the generation itself seems to work flawlessly already.

As for now, just need to figure out a way to manually type out that ENS subgraph query so Apollo then does not try to grab that query from our defined Rinkeby subgraph to generate typings. Any ideas on where exactly to put some `Ens.d.ts` file and how it should be typed? Already tried something like this in `src/@types/Ens.d.ts`:

```ts
export interface UserEnsDomains_domains_name {
  __typename: 'UserEnsDomains'
  accountId: string
}
```

But Apollo codegen does not seem to pick that up during its typing generation.

@bogdanfazakas do you remember anything regarding this when looking into all this graphql typing generations which could help here?

bogdanfazakas:
> those typings... So yes to kicking out Apollo and switching typings generation to urql as defined in #722. It's already prepared in code, the commands are there and I quickly tried to switch it in here but then needs lots of changes to source the newly generated typings but the generation itself seems to work flawlessly already.
> 
> As for now, just need to figure out a way to manually type out that ENS subgraph query so Apollo then does not try to grab that query from our defined Rinkeby subgraph to generate typings. Any ideas on where exactly to put some `Ens.d.ts` file and how it should be typed? Already tried something like this in `src/@types/Ens.d.ts`:
> 
> ```ts
> export interface UserEnsDomains_domains_name {
>   __typename: 'UserEnsDomains'
>   accountId: string
> }
> ```
> 
> But Apollo codegen does not seem to pick that up during its typing generation.
> 
> @bogdanfazakas do you remember anything regarding this when looking into all this graphql typing generations which could help here?

yep the generation was working but it was not replaced yet due to missing types and some types that were generated differently. Regarding the domain name types I don't remember specifically but i can take a look now over the decimation to see if I find something useful

mihaisc:
I think apollo tries to check all the queries based on the config from `apollo.config.js`

kremalicious:
"solved" the typings for now but following #722 definitely would be better choice

Slice 1

kremalicious:
> Should we update the links in the market to use ens instead of address?

yup, good idea, URL already gets replaced in address bar automatically though

mihaisc:
Indeed , did not notice that.

mihaisc:
ensBug1
if i navigate to the url directly it works ok 
image

kremalicious:
> if i navigate to the url directly it works ok

that should have been fixed by changing to linking to ens name URLs as it saves one roundtrip in the logic

mihaisc:
So if someone has the url with address instead of ens it will fail to load proper data?

bogdanfazakas:
> So if someone has the url with address instead of ens it will fail to load proper data? This is not ok behavior.

for me it switched automatically to ens when tried the following url  https://market-git-feature-accountpageens-oceanprotocol.vercel.app/profile/0x655efe6eb2021b8cefe22794d90293aec37bb325

mihaisc:
It seems that the bug in my gif was happening only when navigating to a url with address from the market. If i navigate directly to the url it's ok.  Guess it's sort of fixed now, since we don't have those url in the market.

mihaisc (approved):
Looks ok except for the minor issue described [here](https://github.com/oceanprotocol/market/pull/860#issuecomment-922715831)

bogdanfazakas (approved):
Works great for me also. In regards to the typing issue I'll try to tackle it on the switch to urql PR

Issue #660, opened 2021-06-09
New front page 


mihaisc:
The front page (landing page) will have the role to present the market to the user.

Maybe we can have a section for news/announcements
Have call to action for publish, maybe other
Have one section for each category (
- assets  with action to /assets
- pools with action to /pools
- best sellers (accounts)

mihaisc:
![Inkedtest_LI](https://user-images.githubusercontent.com/26000280/130212671-af095447-30ad-4376-939d-d42c744ba7cd.jpg)

The order is not relevant, it's just the order i thought them up

1. We can pick some nice text from oceanprotocol.com and also embed a video of one of the demos @kremalicious made (couldn't find then on youtube, hope we have them saved somewhere)
2. Fetch total tv for each main network, 1 subgraph call for each
3. Fancy call to action, maybe we can add something extra there
4. Best selling assets . Fetch top 3 for each main net and the just show the top 3.  It will be possible in the upcoming subgraph update
5. Top tvl pools. Fetch top 3 pool for each main net and the order them. I think it would look nice to have some teasers for the pool as well. We can show the tokens, tvl , ?
6. Quick tutorials . Nothing from here exists, these titles are just quick ideas. Ideally we need a page for each with nice explanations for the concept. flow, metadata, etc. 
7. Best sellers. Fetch top 3 from all networks.

kremalicious:
- initial launch demo: https://www.youtube.com/watch?v=KiDg0ry6oV4
- CtD launch demo: https://www.youtube.com/watch?v=RxrkUd9pds0

CatalinUXS:
hello, here is the first version of the design:

https://www.figma.com/file/e5KaRp3fayJYPQR9JUBYPK/Homepage-draft?node-id=2%3A35

PR #864, opened 2021-09-21
Publishers with most sales, using AccountTeaser component

### LINKED ISSUE #828: Create "Publishers with most sales" component
Linked issue: https://github.com/oceanprotocol/market/issues/660

Top publishers will be sorted by TVL

claudiaHash:
Fixes #828 and #669  .

Changes proposed in this PR:

- create Account Teaser component and use it 
- create Publishers with most sales component


mihaisc:
The fetching and calculation of sales is very inefficient. It will be solved with https://github.com/oceanprotocol/ocean-subgraph/issues/230 in just one query (per chain)

claudiaHash:
Right, and we won't have to do extra operations to get the maximum values.

kremalicious:
For reference, this is what I get after loooooong loading so lots to do on the design side:

Screen Shot 2021-09-23 at 11 33 18

I would only leave _Sales_ & _Published_ as metrics on there, or work on some minimal variant of the Stats component as current implementation breaks pretty quickly with larger numbers. But getting this section loaded as fast as possible has highest priority

src/components/atoms/ExplorerLink.tsx
```diff
@@ -16,7 +16,7 @@ export default function ExplorerLink({
 }: {
   networkId: number
   path: string
-  children: ReactNode
+  children?: ReactNode
```
kremalicious, line 19: Was it intended to make `children` optional? Cause we should not have any empty link which this would produce when no `children` is passed. Also there should be no further actions on the teasers as they themselves are fully clickable so no actions inside of that link can be actually used so most likely we don't want an ExplorerLink inside the account teasers

src/components/molecules/AccountTeaser.tsx
```diff
@@ -0,0 +1,93 @@
...
+import Stats from '../pages/Profile/Header/Stats'
+import styles from './AccountTeaser.module.css'
+import ProfileProvider from '../../providers/Profile'
+
+declare type AccountTeaserProps = {
+  account: string
+  showStatistics: boolean
+}
+
+const Blockies = ({ account }: { account: string | undefined }) => {
```
kremalicious, line 20: we have a component for this now under `src/atoms/Blockies`

src/components/organisms/AccountList.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const { chainIds } = useUserPreferences()
+  const [loading, setLoading] = useState<boolean>(true)
+
+  useEffect(() => {
+    isLoading && setLoading(true)
+    if (!accounts) return
+    setLoading(false)
+  }, [accounts])
+
+  function handlePageChange(selected: number) {
```
kremalicious, line 47: we can ditch all pagination stuff in this section

src/components/atoms/ExplorerLink.tsx
claudiaHash: yess, that was my bad, forgot it there

kremalicious:
Honestly, I would ditch the stats completely, or just use 1 number reflecting the section order principle which is _Sales_. This then also prevents us from again having to mix up different use cases in one component and introducing all this `showInAccountTeaser` logic. Using `NumberUnit` directly instead of making the collection component in profile header fit this new use case might be better direction anyway. But in end all stats can be seen after clicking account and the intention is to actually bring people to those new profile pages so reducing stats even more, or completely leave it out would create a better incentive to also check out the profile.

In this context, and I guess also on profile page an interesting metric would be sale revenue, meaning the amount generated by all the sales converted to fiat which then would be nice meaningful metric to only show on account teaser. But not even sure we can calculate that right now so no need to solve in here

kremalicious:
So for design we essentially only have this [from here](https://github.com/oceanprotocol/market/issues/660#issuecomment-908153697):

Screen Shot 2021-09-23 at 14 23 41

I think that's all just too much in one place, if we only use a name (which should be either truncated address, ENS name, or 3Box name, in that order) & 1 key metric we have way more space to show more accounts, like 10 or so, and liven up the frontpage with a differently layed out section.

In the end for general layout, always a fan of simply ~~stealing~~ getting inspired by rarible.com:

Screen Shot 2021-09-23 at 14 22 02

mihaisc:
We can use the small version of the account teaser.  And stop working on this, i already marked it as blocked. The current implementation is wrong.

mihaisc:
After the subgraph is deployed the query would look something like this:
```
{
  users(first: 1000, orderBy: nrSales, orderDirection: desc, where: {nrSales_not: 0}) {
    id
    nrSales
  }
}
```

Basically you have all the information in this query for the small teasers

KatunaNorbert:
Looks like these two are not in the right order having all main networks selected:
Screenshot 2021-11-16 at 10 13 02

src/components/molecules/AccountTeaser.module.css
```diff
@@ -0,0 +1,41 @@
...
+  font-size: var(--font-size-mini);
+  height: 90%;
+  color: var(--color-secondary);
+  position: relative;
+}
+
+.link span {
+  font-size: var(--font-size-large);
+  margin-right: calc(var(--spacer) / 3);
+  margin-top: calc(var(--spacer) / 4);
```
KatunaNorbert, line 27: I'll suggest removing this margin and rather use flex on parent component (`align-items:center`) for centring the child components

src/components/molecules/AccountTeaser.module.css
```diff
@@ -0,0 +1,41 @@
...
+}
+
+.header {
+  display: flex;
+  flex-direction: row;
+}
+
+.sales {
+  font-size: small;
+}
```
KatunaNorbert, line 41: I think it might look better to remove the margin we get o sales from `p` element, and use flex to center the name and number of sales, something like this
Screenshot 2021-11-16 at 10 44 44
:

mihaisc:
For me it takes like 30 seconds to load and does hundreds of calls to the subgraph. Something is not ok

claudiaHash:
> For me it takes like 30 seconds to load and does hundreds of calls to the subgraph. Something is not ok

The delay comes from the supplementary sort we have to do: selecting more networks, we get the top publishers from each network. the publisher placed on #2 on Rinkeby, for instance, could be on #7 on Polygon. therefore, I had to sum the number of sales for the publishers found and sort the publishers by the sale sums obtained, to get the correct ranking

mihaisc:
But that just means 1 call for each chain, why do you need so many? So get all the top users from all chains and then sort them by `nrSales`

mihaisc (approved):

KatunaNorbert (approved):

PR #866, opened 2021-09-22
Fix scrollbar showing on table on large screens

### LINKED ISSUE #856: Data tables have too many scrollbars
As [described here](https://github.com/oceanprotocol/market/pull/774#issuecomment-918867006) we get scrollbars way too often in our table components. They are supposed to scroll like that when viewport or container does not have enough space for its content, the overflow pattern. This is happening in multiple places throughout app so not just applies to account profiles changes.

1. In linked case we have enough space so there should be no scrollbar. Looks like we need to explore some of the underlying component's options, or our own css styles for it.
2. When there is a scrollbar it should not be so obvious although this highly depends on browser/OS/input used. So we should explore styling the scrollbars within data tables slightly, e.g. making the track & thumb into colors based on our color scheme and minimizing its height

KatunaNorbert:
Fixes #856 .

### DIFF
--- a/src/components/atoms/Table.module.css
+++ b/src/components/atoms/Table.module.css
@@ -23,6 +23,10 @@
   color: var(--color-secondary);
 }
 
+div [class*='rdt_TableCell'] {
+  min-width: 80px;
+}
+
 .table [role='row'] {
   border-bottom: 1px solid var(--border-color) !important;
 }
--- a/src/components/atoms/Table.module.css
+++ b/src/components/atoms/Table.module.css
@@ -23,8 +23,8 @@
   color: var(--color-secondary);
 }
 
-div [class*='rdt_TableCell'] {
-  min-width: 80px;
+.table [role='gridcell'] {
+  min-width: 0;
 }
 
 .table [role='row'] {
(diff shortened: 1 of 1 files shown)


KatunaNorbert:
The scrollbar is shown as a consequence of `padding` and `min-width` combination on the table cell. I think it's safer to change the min-width then the padding so we don't affect the design as much

KatunaNorbert:
From what I saw there no mush we can do about the scrollbar on Firefox, just made it thinner. As for other browsers, the scrollbar seems to be fine so I don't think we need some changes for those

claudiaHash (approved):
Just tested it against Chrome and Firefox, on Ubuntu. Looks good!

src/components/atoms/Table.module.css
```diff
@@ -23,6 +23,10 @@
   color: var(--color-secondary);
 }
 
+div [class*='rdt_TableCell'] {
+  min-width: 80px;
```
kremalicious, line 27: for me, only using `min-width: 0` here solves most scrollbar problems in Safari and Firefox. I think this was the intention of the basic reset done further up this file in line 5 & 6 where we have multiple selectors targeting the `role`. I would suggest adding the table cells selector there too, but also selected by `role` so in this case `.table [role='gridcell']`

mihaisc:
can't replicate it anymore.

kremalicious (approved):

PR #867, opened 2021-09-22
Fix/ Put back filters published page

bogdanfazakas:
Fixes # .

Changes proposed in this PR:

- added service type and access filters on published page

### DIFF
--- a/src/components/templates/Search/Filters.tsx
+++ b/src/components/templates/Search/Filters.tsx
@@ -27,33 +27,39 @@ export default function FilterPrice({
   serviceType,
   accessType,
   setServiceType,
-  setAccessType
+  setAccessType,
+  isSearch,
+  className
 }: {
   serviceType: string
   accessType: string
   setServiceType: React.Dispatch<React.SetStateAction<string>>
   setAccessType: React.Dispatch<React.SetStateAction<string>>
+  isSearch: boolean
+  className?: string
 }): ReactElement {
   const navigate = useNavigate()
   const [serviceSelections, setServiceSelections] = useState<string[]>([])
   const [accessSelections, setAccessSelections] = useState<string[]>([])
 
   async function applyFilter(filter: string, filterType: string) {
-    let urlLocation = ''
-    if (filterType.localeCompare('accessType') === 0) {
-      urlLocation = await addExistingParamsToUrl(location, ['accessType'])
-    } else {
-      urlLocation = await addExistingParamsToUrl(location, ['serviceType'])
-    }
-
-    if (filter && location.search.indexOf(filterType) === -1) {
-      filterType === 'accessType'
-        ? (urlLocation = `${urlLocation}&accessType=${filter}`)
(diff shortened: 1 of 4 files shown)


claudiaHash:
Looks good overall but I just caught what seems to be a bug. 

When selecting a service filter and both access filters, I get the correct results: 
Screenshot from 2021-09-22 14-38-40

When I deselect one of the access filters, I get the following result
Screenshot from 2021-09-22 14-38-29

That's the only case I encountered this problem. Couldn't reproduce it selecting both service filters and a single access filter.

bogdanfazakas:
yup, I was able to reproduce the issue. It's caused by the _page number_ we send to aqua and we kind of need to reset the page to 1 on every _filter change_. Sadly this happens to our existing _search_ also I'll create an issue for this.

jamiehewitt15:
The above issue appears to be fixed and it's working well.

jamiehewitt15 (approved):
Nice work

src/components/templates/Search/Filters.tsx
```diff
@@ -27,33 +27,39 @@ export default function FilterPrice({
   serviceType,
   accessType,
   setServiceType,
-  setAccessType
+  setAccessType,
+  isSearch,
```
KatunaNorbert, line 31: Functionally it's working well. 
Some suggestions for improvement would be to give a more generic name to this `isSearch` prop like `addFiltersToUrl` or `showInUrl`, I think this would be a better name. 
And we can also omit the prop when not necessary, so remove `isSearch = {false}` when not needed

bogdanfazakas: makes sense will update this 👍

claudiaHash (approved):
With the latest additions ad fixes everything seems to work properly. Looks good to me!

KatunaNorbert (approved):

kremalicious:
yup, all works and would be fine with doing it like this if we don't want to spend more time on it.

I still think there's a huge win if we would follow [the hint here](https://github.com/oceanprotocol/market/blob/main/src/providers/Profile.tsx#L193-L197) and "enhance" published DDO list with prices and store that in profile provider as it stores already the DDO list to get published number. Then the filters and the whole `PublishedList` can use that, and live-filter through it instead of sending remote queries all the time. But guess this would have required too much refactor?

bogdanfazakas:
> yup, all works and would be fine with doing it like this if we don't want to spend more time on it.
> 
> I still think there's a huge win if we would follow [the hint here](https://github.com/oceanprotocol/market/blob/main/src/providers/Profile.tsx#L193-L197) and "enhance" published DDO list with prices and store that in profile provider as it stores already the DDO list to get published number. Then the filters and the whole `PublishedList` can use that, and live-filter through it instead of sending remote queries all the time. But guess this would have required too much refactor?

Yep the idea with the PR was to quickly add the filters back to published tab, and leave the refactor for the v4 and split the view from fetching data etc

PR #868, opened 2021-09-22
Refine Pool Shares DDO fetching

### LINKED ISSUE #842: pools shares DDO fetching is ineffective

KatunaNorbert:
Fixes #842.

### DIFF
--- a/src/utils/aquarius.ts
+++ b/src/utils/aquarius.ts
@@ -80,8 +80,9 @@ export function transformChainIdsListToQuery(chainIds: number[]): string {
 
 export function transformDIDListToQuery(didList: string[] | DID[]): string {
   let chainQuery = ''
+  const regex = new RegExp('(:)', 'g')
   didList.forEach((did: any) => {
-    chainQuery += `id:${did.replace(/(:)/g, '\\:')} OR `
+    chainQuery += `id:${did.replace(regex, '\\:')} OR `
   })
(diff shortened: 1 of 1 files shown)


kremalicious:
Nice, that `retrieveDDOListByDIDs()` looks handy, and we can use it to solve #869 too to improve pool transactions.

For now, I get this error when using [this profile](https://market-git-fix-issue842-pool-shares-ddo-fetching-oceanprotocol.vercel.app/profile/0x903322C7E45A60d7c8C3EA236c5beA9Af86310c7):

Screen Shot 2021-09-23 at 11 30 44

KatunaNorbert:
That was due to `-isInPurgatory` filter, removed that and should be good now. But now I have some errors because some request are not triggered although locally it's working well. Here seems that the query for getting the ddo's is not triggered.

Screenshot 2021-09-23 at 13 37 53

bogdanfazakas (approved):
works well on my end, i also think we can ignore Code climate complains since the logic is not the same in the mentioned block

PR #873, opened 2021-09-24
Fix/ Issue872 search breaks page filter combo

### LINKED ISSUE #872: Search breaks on filters and page combinations
The search is displaying no result. This happens when the page selected is higher that. 

**To Reproduce**
1. Fill in the search text box 
2. Click on search
3. Go to page 2
4. Restrict filters so there should be displayed at max results for one page but you'll see no results

Screenshot 2021-09-23 at 16 40 56
results with page 1 selected

Screenshot 2021-09-23 at 16 33 42
results with page 2 selected

bogdanfazakas:
Fixes #872 .

Changes proposed in this PR:

- set page to 1 when filters are applied on search results

### DIFF
--- a/src/components/templates/Search/index.tsx
+++ b/src/components/templates/Search/index.tsx
@@ -35,20 +35,32 @@ export default function SearchPage({
     sortOrder as string
   )
 
+  async function fetchAssets() {
+    setLoading(true)
+    setTotalResults(undefined)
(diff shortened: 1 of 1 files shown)


mihaisc (approved):
LGTM

claudiaHash (approved):
Good to go!

Issue #870, opened 2021-09-23
Remove "Downloads" on the user profile page


jamiehewitt15:
I'm raising this here as a follow up to feedback on the new profile page from @w1kke at the [OceanDao town hall event](https://github.com/oceanprotocol/oceandao/wiki/OceanDAO-Town-Hall). 

### Issue:
The downloads counter is potentially confusing to new users who are looking at the profile pages. It could be interpreted as "the number of downloads of this user's published assets". If consumers misinterpret this, it could lead them to the false impression that the publisher is not popular or credible, which would be damaging. In many other digital marketplaces, the "downloads" counter is used to show the popularity of an asset or publisher, so it is understandable that users may interpret it that way here. 

![Downloads messaging](https://user-images.githubusercontent.com/20739535/134495106-72ec5c5b-214b-4789-b352-05369411fb9e.png)

### Proposed solution [rejected]:

Rename "Downloads" as "Assets Consumed"

![assets consumed](https://user-images.githubusercontent.com/20739535/134496220-4e783af0-491a-49eb-a51d-c694366ec2c6.png)

### Agreed solution solution:
Remove the downloads statistic on the user profile page to avoid confusion.

kremalicious:
yeah, we never find a proper word which is why we added a tooltip as a compromise. _Consume_ is also kinda confusing and to me just doesn't make it more clear. We do not use that word anywhere in the UI or our documentation so users have even less connection to what that is. Compare that with _Downloads_ which is used within the _Use_ tab, and here on the profile pages as a list and on the publish page as we communicate an asset can either be downloaded, or computed on. So no need to introduce another term here.

But fully agree on the misleading part as it's different user expectations based on checked out profile, and for a publisher pretty sure main user expectation would be to see metric about how many times that publisher's assets were downloaded.

As this seems to be a super controversial metric because unclear I would be in favor of simply removing it, ideally replacing it with some new more meaningful metric

jamiehewitt15:
Removing it seems like a reasonable compromise. I imagine most users who will visit this page will be interested in statistics relating to publishing rather than consuming.

mihaisc:
Agree with removing.

PR #875, opened 2021-09-24
Removing 'Downloads' statistic on profile page

jamiehewitt15:
Closes: #870

Changes proposed in this PR:

- Removing 'Downloads' statistic on the profile page to avoid confusion


kremalicious (approved):
Can't think of any quick to do metric, but also meaningful metric to replace it with so fine with just removing for now

bogdanfazakas (approved):

PR #878, opened 2021-09-27
Fix published assets

### LINKED ISSUE #877: No published assets on profile page
Getting _No results found_ on every profile, e.g. https://market.oceanprotocol.com/profile/dataunion.eth. Seems something with #867 wasn't right

Screen Shot 2021-09-27 at 10 59 03

mihaisc:
closes #877

### DIFF
--- a/src/utils/aquarius.ts
+++ b/src/utils/aquarius.ts
@@ -89,7 +89,6 @@ export async function queryMetadata(
       { cancelToken }
     )
     if (!response || response.status !== 200 || !response.data) return
-
     return transformQueryResult(response.data, query.from, query.size)
   } catch (error) {
     if (axios.isCancel(error)) {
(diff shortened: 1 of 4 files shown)


kremalicious (approved):

bogdanfazakas:
yep the assets are fetched now but there seems to be a problem on the pagination with this changes, trying to figure it out now

bogdanfazakas:
it looks like we can't rely on `queryResult?.page` since it has wrong values, for eg it's returned 1 for the first two pages and on the third page it's two and so on.

mihaisc:
i'll take a look

bogdanfazakas:
hmm seems to work with the update from aquarius helper, I'll revert my prev update

src/utils/aquarius.ts
```diff
@@ -64,7 +64,7 @@ export function transformQueryResult(
   )
   result.totalResults = queryResult.hits.total
   result.totalPages = Math.floor(result.totalResults / size)
```
bogdanfazakas, line 66: think i found another problem here. We use math.floor to get the page number and if for eg we have total result let's say 41 and size of the search 21 we get the total number of pages 1 and you can't switch to the second page and see the next 20 results. I think we use something like this
```suggestion
  result.totalPages = (result.totalResults / size < 1) ? Math.floor(result.totalResults / size) : Math.ceil(result.totalResults / size)
```

bogdanfazakas (approved):
All good now 🎉

PR #884, opened 2021-09-27
Add a suggestive message if no network selected

### LINKED ISSUE #883: 'No network selected' case not addressed on Profile history tabs
When no network is selected we should have a suggestive message instead of a loader or a _No results found_   message, on Profile history tabs.

claudiaHash:
Fixes #883 .

Changes proposed in this PR:

-  message displayed if no network selected

### DIFF
--- a/src/components/pages/Profile/History/PublishedList.tsx
+++ b/src/components/pages/Profile/History/PublishedList.tsx
@@ -61,9 +61,7 @@ export default function PublishedList({
     fetchPublishedAssets()
   }, [accountId, page, appConfig.metadataCacheUri, newCancelToken])
 
-  return chainIds.length === 0 ? (
-    <div>Please select a network.</div>
-  ) : accountId ? (
+  return accountId ? (
     <>
       <Filters
         serviceType={service}
--- a/src/components/pages/Profile/History/PublishedList.tsx
+++ b/src/components/pages/Profile/History/PublishedList.tsx
@@ -59,9 +59,11 @@ export default function PublishedList({
       await getPublished()
     }
     fetchPublishedAssets()
-  }, [accountId, page, appConfig.metadataCacheUri, chainIds, newCancelToken])
+  }, [accountId, page, appConfig.metadataCacheUri, newCancelToken])
 
-  return accountId ? (
+  return chainIds.length === 0 ? (
+    <div>Please select a network.</div>
+  ) : accountId ? (
     <>
       <Filters
         serviceType={service}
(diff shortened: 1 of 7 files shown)


jamiehewitt15:
This is good but it's also quite a bit of repetition so it might be worth putting it in a separate component and reusing it. That will also make it easier to update if we need to change it at any point in the future.

claudiaHash:
Good point, thanks. I'll change that right now

KatunaNorbert (approved):
Looks good

src/components/pages/Profile/History/PublishedList.tsx
```diff
@@ -59,7 +59,7 @@ export default function PublishedList({
       await getPublished()
     }
     fetchPublishedAssets()
-  }, [accountId, page, appConfig.metadataCacheUri, chainIds, newCancelToken])
```
kremalicious, line 62: Not sure if this change is responsible but for me the published list on profile does not work properly anymore. In this case in my screenshot there should be no assets shown as there are none on BSC. But whatever I select I always get the ETH mainnet asset shown

image

jamiehewitt15: The same thing seems to be happing on `main` so it doesn't seem to be anything to do with this PR. 

Network not updating


It seems to be an issue with the state not being updated. I found that if I refresh the page after selecting a different network, it starts showing the correct assets.

Network not updating 2

jamiehewitt15: I've created a [seperate issue](https://github.com/oceanprotocol/market/issues/887) for that bug

kremalicious: Nope, this is something else and unrelated to this PR. On `main` everything works as expected on the profile page so this specifically is caused by a change in here 
D4A137AC-1259-432E-B0D1-2979B02FF63C

claudiaHash: I just added back the chainIds to the dependency list. That should be the only change that affects the results on Published tabs.

KatunaNorbert: Looks like that was it. It's fixed now.

bogdanfazakas (commented):
functionality wise works great, but it would be nice if we can stick to the same spacing on the published tab if we display no results found or no network selected because the ui components are dancing around when deselecting all networks and displaying the new added message. If there is time maybe we can take a look also at the explorer urls and fix the same issue over there. I'll add a video to better understand the issues mentioned

https://user-images.githubusercontent.com/13447142/135407421-88dd94f1-f59d-4daf-8be3-2ed117a1f1af.mov

bogdanfazakas (approved):

PR #889, opened 2021-09-29
Data assets not updating properly when selected networks are changed in home page

### LINKED ISSUE #887: Data assets not updating when selected networks are changed.
When changing networks in the network selector, the data assets do not immediately update correctly. The assets from the Ethereum mainnet remain visible. In the image below, only assets from Ethereum Rineby should be shown.

image

It seems to be an issue with the state not being updated correctly. I found that if I refresh the page after selecting a different network, it starts showing the correct assets.

image

This issue was first reported [here](https://github.com/oceanprotocol/market/pull/884#discussion_r718377160)

KatunaNorbert:
Fixes #887.

How to reproduce:
have ETH and Rinkeby networks selected and then deselect ETH. You should still have assets from ETH network

### DIFF
--- a/src/components/pages/Home.tsx
+++ b/src/components/pages/Home.tsx
@@ -67,7 +67,6 @@ function SectionQueryResult({
   action?: ReactElement
   queryData?: string
 }) {
-  const { appConfig } = useSiteMetadata()
   const { chainIds } = useUserPreferences()
   const [result, setResult] = useState<any>()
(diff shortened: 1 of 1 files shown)


claudiaHash (approved):
tested for multiple combination of main/test networks. looks good!

jamiehewitt15 (approved):
Works great, thank you for fixing this so quickly

PR #890, opened 2021-09-29
Fix display 'No results found' message before displaying results

### LINKED ISSUE #876: 'No results found' shown before displaying results
_'No results found'_ message is displayed before the loader and correct results, in Profile's Pool Transactions and Compute Jobs sections.
Screenshot from 2021-09-24 17-13-40
Screenshot from 2021-09-24 17-26-16

claudiaHash:
Fixes #876  .

Changes proposed in this PR:

- fix 'No results found' message display on Profile history tabs

### DIFF
--- a/src/components/molecules/PoolTransactions/index.tsx
+++ b/src/components/molecules/PoolTransactions/index.tsx
@@ -14,9 +14,7 @@ import { CancelToken } from 'axios'
 import Title from './Title'
 import styles from './index.module.css'
 import { DDO, Logger } from '@oceanprotocol/lib'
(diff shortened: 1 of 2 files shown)


KatunaNorbert:
It seems to work well, but on the `PoolShares` tab the loading icon is not always displayed. For example when selecting and deselecting networks and when I switch to the tab with selected networks where I don't have pool shares.

kremalicious:
Does not work for me anymore in the Pool Transactions tab, which just shows same behavior as on live:

https://user-images.githubusercontent.com/90316/137107251-cfdaa09a-6feb-410c-9e48-2a0832d9fd1c.mov

PR #898, opened 2021-10-01
Use latest ocean.js

alexcos20:
Close #897

PR #900, opened 2021-10-04
add moonriver support

alexcos20:
Add Moonriver support

### DIFF
--- a/app.config.js
+++ b/app.config.js
@@ -10,10 +10,10 @@ module.exports = {
 
   // List of chainIds which metadata cache queries will return by default.
   // This preselects the Chains user preferences.
-  chainIds: [1, 137, 56],
+  chainIds: [1, 137, 56, 1285],
 
   // List of all supported chainIds. Used to populate the Chains user preferences list.
-  chainIdsSupported: [1, 3, 4, 137, 80001, 1287, 56, 2021000],
+  chainIdsSupported: [1, 3, 4, 137, 80001, 1287, 56, 2021000, 1285],
 
   rbacUrl: process.env.GATSBY_RBAC_URL,
 


alexcos20:
Depends on : 
- https://github.com/oceanprotocol/market/pull/898
- https://github.com/oceanprotocol/ocean.js/pull/1054

Just setting as a draft until all deps are fullfiled

kremalicious:
Just a note so we don't forget: since Moonriver uses Blockscout-based explorer (and also the other new networks) we have to adapt the logic for the `ExplorerLink` component as some URL slugs are different between Etherscan-based and Blockscout-based explorers.

kremalicious:
Great, so over at our database of networks Moonriver uses the `network` key wrong (supposed to be `mainnet` or `testnet`) https://github.com/ethereum-lists/chains/pull/345 which in the end results for our list in:

Screen Shot 2021-10-04 at 17 39 58

So I guess we need to hack in specific detection for moonriver, or we go over to [ethereum-lists/chains](https://github.com/ethereum-lists/chains) and fix the source but most likely will take a while

kremalicious:
There we go, this is what new users to Ocean Market will see:

Screen Shot 2021-10-04 at 17 58 06

But this what returning users will get even if they never touched the network selection, as we save all user preference on first visit in `localStorage` so they have the old default preference:

Screen Shot 2021-10-04 at 17 57 10

If we want to have it preselected in that case too we need to add its to the existing `localStorage` data so some small data migration so to speak

alexcos20:
Not sure if we need a "migration", but just a small picture:  how to activate moonriver

alexcos20:
all good, we can approve this

kremalicious (approved):
awesom! And another example that no matter how much we try to generalize and anticipate issues for new networks, we always end up having to fix specific problems but glad here it was all just about missing or wrong keys in our networks metadata source

KatunaNorbert (approved):
Working good, haven't find any issue

PR #902, opened 2021-10-04
Improve Pool Transactions page data fetching

### LINKED ISSUE #869: pool transactions if very slow
Well, the title says is all. If there are more that 3-4 pages it feel like an eternity ( exaggeration intended ) 
I'm sure (hope) that we can improve it.

KatunaNorbert:
Fixes #869.


bogdanfazakas (approved):
Loads much faster, good job here 👍 , I think we should wait for the release of the  #868 first and then merge this since it looks like this PR started from the changes in PR mentioned above

PR #907, opened 2021-10-05
Feature/energyweb

alexcos20:
https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-246.json looks fine, so no additional changes needed

### DIFF
--- a/src/utils/web3.ts
+++ b/src/utils/web3.ts
@@ -62,6 +62,9 @@ export function getNetworkDisplayName(
     case 137:
       displayName = 'Polygon'
       break
+    case 246:
+      displayName = 'Energy Web Chain'
+      break
     case 80001:
       displayName = 'Polygon Mumbai'
       break
--- a/app.config.js
+++ b/app.config.js
@@ -10,10 +10,10 @@ module.exports = {
 
   // List of chainIds which metadata cache queries will return by default.
   // This preselects the Chains user preferences.
-  chainIds: [1, 137, 56],
+  chainIds: [1, 137, 56, 246],
 
   // List of all supported chainIds. Used to populate the Chains user preferences list.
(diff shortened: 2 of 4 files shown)


src/utils/web3.ts
```diff
@@ -62,6 +62,9 @@ export function getNetworkDisplayName(
     case 137:
       displayName = 'Polygon'
       break
+    case 246:
+      displayName = 'Energy Web Chain'
```
kremalicious, line 66: this is the same as on the network metadata so we don't need to overwrite it here. This method is only for those cases where what we get from metadata is weird

kremalicious:
note for later: icon for `NetworkIcon` needed

alexcos20:
what about https://energyweb.org/wp-content/uploads/2020/02/EWF-EWBrandmark-Knockout.png ?  I think it's too large

kremalicious:
all good we have it here https://www.figma.com/file/bNr0pHAqvkZgSLuIcJeje7/logos?node-id=0%3A2

was just a reminder to get the logo mark over here

src/utils/web3.ts
alexcos20: yeah, I think "Energy Web Chain" sounds better than "energyweb"

KatunaNorbert:
Also can't start a compute job, says that the asset url is not available:

Screenshot 2021-10-07 at 16 32 20

alexcos20:
found the problem:
slags...

they are using "tokens", like :  https://explorer.energyweb.org/tokens/0x26E4991a72728b1a9B1044345e5bF9293E0A1434/token-transfers

we are using "token", like :https://explorer.energyweb.org/token/0x96856f1abC947322AcAeC3f3eBAb01Ef29B7E605   (token description for CO₂ Intensity of Electricity Generation in EU asset)

PR #924, opened 2021-10-15
Feature/issue 904 credential type dropdown

kianyee:
Fixes # . Credential type dropdown to 

Changes proposed in this PR:

- 
- 
-


kremalicious:
Revisit after v4, but advanced permissions outside of contract mechanisms might not be in v4 market codebase anymore

PR #926, opened 2021-10-18
Replace profile sold statistic value number with subgraph value

### LINKED ISSUE #906: Replace curent profile Sold statistic value with the subgraph value
Use subgraph newly added sold value (https://github.com/oceanprotocol/ocean-subgraph/pull/231) and delete curent complex calculation logic for the sold statistic on profile page

claudiaHash:
Fixes #906  .

Changes proposed in this PR:

- get subgraph sales value by user `accountId`
- add value in profile provider
- use new value in Profile Stats

### DIFF
--- a/src/utils/subgraph.ts
+++ b/src/utils/subgraph.ts
@@ -621,29 +621,6 @@ export async function getHighestLiquidityDIDs(
   return [searchDids, didList.length]
 }
 
-export async function getAccountNumberOfOrders(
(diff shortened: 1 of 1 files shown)


KatunaNorbert:
This looks good! The issue with number of sales not cleared when no network selected is also solved.
One question, do we still need the `getAccountNumberOfOrders` function?

Issue #862, opened 2021-09-17
Publish form


mihaisc:
Design/Spec talks can be found here https://github.com/oceanprotocol/market/issues/66
![publish](https://user-images.githubusercontent.com/26000280/133741644-3c1316e1-0036-4864-8d95-1da86ef01db9.png)

We will have only 2 steps in the wizzard

**Step 1 - define metadata related to nft/asset:**

Stored in metadata contract:
Title
Description
Author
Tags

Stored in NFT contract:
Name - autogenerated
Symbol - autogenerated

Stored in NFT Metadata
Name: autogenerated ( ideally we add the did here so there is a link to the asset)
Description : autogenerated
Image - autogenerated (research, maybe related to ocean)

**Step 2 - define metadata related to the service (only one service in the publish form, multiple services can be added after creation, in the asset details page) :**

Files (array )
Sample file
Access type
Timeout
Datatoken & Symbol and an option to add a current datatoken address instead of generating a new datatoken

This component (step 2) can be used in other components other than the publish form. It will be used to add/edit a service in the asset details page.

Suggested background flow (work in progress):

collect all data from form (nothing is created when moving from step 1 to step 2)
if no datatoken address is specified , create a datatoken
create service and ddo and add it to chain , did will be sha256(owner+title+description+networkId)
generate nft metadata and add it to chain

CatalinUXS:
Hi, the desktop version for the stepper component from publish section

![Publish_details](https://user-images.githubusercontent.com/87698146/135227664-2ee915c8-ea56-4401-ab7c-2080ffa3e3a4.jpg)
![Publish_services](https://user-images.githubusercontent.com/87698146/135227673-8c26a298-04da-40c3-9349-840cebbc2708.jpg)

CatalinUXS:
figma document: https://www.figma.com/file/CCLmYjkbdZ5GaiQsgUQ0Aj/Publish-v2-draft?node-id=1%3A162

kremalicious:
Nice start!

Some first notes:
- would make the step titles into: Metadata, Access, Pricing.
- "network" in page title is redundant and can be removed
- data type on metadata screen and access type should use same binary pattern we already have. Putting our tab style into the form feels out of place compared to other places, and we already removed one click for users in access type so should not fall back to worse solution
[...]

PR #933, opened 2021-10-21
new publish form: setup & data collection & live validation

### LINKED ISSUE #663: Design/Spec Publish component
The publish form(wizzard) will have 3 components 

- component A that will create the asset (nft), resulting an ERC721. The metadata structure will not change from the current one (or have minor changes) 
- component B that will create datatoken, service 
- component C that will create a pricing schema ( free, fixed, pool )

When creating an asset you will no longer need a datatoken like you currently do. An asset can have multiple services and each service has a datatoken.

Component B and C need to be decoupled because they can be used individually. 
You need to be able to create another service for an asset at any time and you need to be able to create another pricing schema for a service at any time.

Example : 
Asset with high quality data
- access service - fixed price 30k ocean 
- compute service with 2 cores maximum of 10 minutes - pool : datatoken/ocean   ~500 ocean
                                                                                          -  pool : datatoken/eth ~too lazy to do the math
                                                                                          -
- compute service with 4 cores maximum of 30 minutes - pool : datatoken/ocean ~1500 ocean
                                                                                          -  pool : datatoken/stable ocean ~too lazy to do the math

What to do in this issue ? 
- design screens for each step in flow
- spec what should happen in each screen

kremalicious:
Setup and refactoring loosely based on #862. This PR prepares most things but does not finalize the whole publish form yet. Closes #663

Best to check out the real thing for all the conditionals/interactivity:
https://market-git-v4-publish-oceanprotocol.vercel.app/publish

Changes proposed in this PR:

- new data structure, new layout 
- form split-up into multiple wizard steps
- collect all data we need in Formik context provider, no custom `onChange` handlers or local state managing anymore
- dynamic transformation of form data into DDO, which then feeds the preview
- lots of `AssetContent` and child components refactor to make it work for preview and fender it even with incomplete DDO
- prepare the final publish flow for actual DDO creation
- #941 → this destroys all page rendering except for `/publish` until Aquarius returns the new DDO structure
- #947
- PoC for solving creation of #944 
- visually highlight datatoken just the same to make connection NFT/metadata & datatoken/access more clear
- usability changes for all URL inputs, like `providerUrl`, `files`, `links` (samples), integrating their feedback into existing form system
- expose default compute environment options in form

Screen Shot 2021-11-24 at 15 34 28

Screen Shot 2021-11-24 at 15 35 23

Screen Shot 2021-11-24 at 15 37 51

Screen Shot 2021-11-24 at 15 38 01

Screen Shot 2021-11-29 at 16 51 35


jamiehewitt15:
Is the home page supposed to work on this branch or is it being fixed elsewhere?

jamiehewitt15:
A few initial bits of feedback based on my first impressions: 

- I'm not sure exactly how we will use the NFT image on the asset detail page? If it is going to be visible, it would be great to give an indication of this on the preview tab. 
- I would be inclined to change the order of the form on the first tab so that the Data NFT is lower down on the first tab, after tags. This way we would get them to input the metadata first and then we show them how that metadata will be represented on-chain. I think this might be a slightly more familiar flow, similar to how social media sites require you to create an account and then set the profile picture at the end. 
- Currently, the terms and conditions are missing, I guess this will be added in like it was before? Or will we just have a link to them? 
- For the provider input field, I think we should add some text saying "We recommend the default provider in most cases" so that people don't get confused by it or feel the need to research setting up their own. 
- I think the data token image and name could be moved down to the bottom of that tab for the same reasons as the data NFT image. 
- For "Add file" and "Add sample file" there are currently no toast pop-up messages on successful url validation. I would suggest we implement the same pop-up message that we have on the current publish form, so we let them know if there is an issue with their URL. 
- For each of the fields where a URL is added, I would suggest that we run the validation `onBlur`. This is how we do it on the publish page at the moment and I think it's a bit more seamless.

kremalicious:
> Is the home page supposed to work on this branch or is it being fixed elsewhere?

Can't be fixed on market, see my comment in issue description:

> #941 → this destroys all page rendering except for /publish until Aquarius returns the new DDO structure

kremalicious:
> I'm not sure exactly how we will use the NFT image on the asset detail page? If it is going to be visible, it would be great to give an indication of this on the preview tab.

agree, but preview === asset content now so whatever we decide later on when working on assets details page, will be reflected in preview too.

> I would be inclined to change the order of the form on the first tab so that the Data NFT is lower down on the first tab

Had this first too but found this too surprising and found piutting NFT/datatoken to the top gives each form a nice "header" explaining things without actually having to explain things. It also visually livens up that damn boring set of input fields, better inviting users to use the form.

> Currently, the terms and conditions are missing, I guess this will be added in like it was before? Or will we just have a link to them?

will be just a link to terms page.

> For "Add file" and "Add sample file" there are currently no toast pop-up messages

yes, why would there be as adding the file and getting file info IS the success? By adding those toasts for some things but not all user feedback we created a confusing and unreliable notification system for users where some feedback happens on form, and some in a completely different place on the screen. I removed the toasts so whenever you want to have user feedback I suggest do it in place on the actual field, just like our live validation feedback does. Ideally we either remove all toast messages, or use only toast messages to give users feedback. Having a mix is just confusing but not sure we'll get there time-wise. Until then would suggest to only leave some error toasts in place, but not for any success state.

> For each of the fields where a URL is added, I would suggest that we run the validation onBlur

yeah but it required us to overwrite all handlers on that custom form field, destroying all Formik behavior like keeping track of required/valid fields. So not sure that is worth it.

kremalicious:
changed a lot for all URL inputs, was a bit crazy what we had there with hitting endpoints on every keystroke creating a huge performance overhead, and duplicating entered values in multiple local states. It's a form field so it will always have the value so just using this now. Now everything only happens on actual user action, that's why we have the buttons.

So for provider URL this is now the error state, only triggered after button click:

Screen Shot 2021-11-24 at 12 05 44

And the success state which right now can't be reached as we use some `ocean.` method for that:

Screen Shot 2021-11-24 at 12 34 33

The problem we wanted to solve where users might add a URL but then not hit the button is solved in a different way where the form only gets valid if files, samples have a `valid: true` key.

jamiehewitt15:
> And the success state which right now can't be reached as we use some `ocean.` method for that:
> 
> Screen Shot 2021-11-24 at 12 34 33

The new success state looks great. 

A couple of small things:
- The error state message on the terms and conditions could be improved. I think changing it to just "Required" would be fine.

error message 1

- The timeout field is required but it doesn't seem to have an error state. Maybe we could change the default to "forever" instead of "---" so that the default value is at least valid input?
time out no error message

kremalicious:
Another round of refactor for the `providerUrl`, as implementation with a default value was a bit confusing and not helpful for us as we want most people to use the default one.

So think this is a better default state instead of adding different success state for a field which essentially behaves the same for users as the files/samples fields. Makes it also more clear that the field does not have to be changed:

Screen Shot 2021-11-25 at 14 06 34


Which after dismissing, simply gives you this:

Screen Shot 2021-11-25 at 14 06 45

Giving us a consistent success state for all of those fields in the end:

Screen Shot 2021-11-25 at 14 15 05

Lots of fighting with Formik's `meta.error` as all those fields have a nested data structure (`{ url: string, valid: boolean }`) yet Formik does not have any way to set nested errors via `helpers.setError()`

kremalicious:
> The timeout field is required but it doesn't seem to have an error state.

when exactly would you expect an error to show up? Cause we can't show errors by default before user has actually interacted with a field, so has been "touched". So in this case we could only show an error on some button click and trigger validation manually, like on continue button. But not sure we want to go that for for just this field as the success state in navigation might be already enough, and as soon as user interacts with dropdown and selects something there I no way to set something wrong anymore

jamiehewitt15:
The new provider URL input looks great, that's a big improvement. It should hopefully avoid people getting confused by that field.

Yeah, I agree with what you're saying about the timeout field. I don't think that there is a need for an error state, I was just thinking it might be better for the default value to be "forever" instead of "---" as that avoids the possibility of an error altogether.

mihaisc:
How do you validate the provider url (couldn't find it) ? we can use the the default get and check if it's a valid provider response, and maybe show the version like `URL confirmed. Provider version : 0.4.14` . Just checking for a valid url might not be enough in this case . If this is already done please ignore 😁

kremalicious:
> How do you validate the provider url 

I just left intact what we had in there but it's kinda not actually working, which was point of #948. Don't really understand why the check doesn't work as I'm using old ocean.js call here against existing provider. But check always fails.

For the default state where we now have the providerUrl added I simply assume that our hardcoded default works, so no validation at all for this. Here's the whole component: https://github.com/oceanprotocol/market/blob/v4-publish/src/components/%40shared/FormFields/Provider/index.tsx

kremalicious:
as for step success conditions I added them all in the Navigation component but later we might wanna move them up some components as most likely we also want to do something with the _Continue_ buttons for each step, like enabling them only when respective step has no errors: https://github.com/oceanprotocol/market/blob/v4-publish/src/components/Publish/Navigation/index.tsx#L19-L33

Conditions itself for each step might need tweaking too, e.g. for pricing screen I set success once user has touched the price field and that's it, but it's good start.

jamiehewitt15:
Creating a successful state for the pricing is tricky because the default is technically fine, no need for the user to actually touch anything. I think the way you've done it is good though, in pretty much every case the user will interact with the form.

Another thing I've noticed is that the free-pricing form has an error that shouldn't be there: 
Free pricing error 2021-11-26_14-22

kremalicious:
> Another thing I've noticed is that the free-pricing form has an error that shouldn't be there

good find, had everything in place so that validation schema for the price changes depending on value of it but seems it was not working. In any case, having input field there is confusing anyway so in a rather uninspired way switched to simply a "Free" text. In reality we only need `pricing.type` to figure out what pricing scheme was selected so no need to have `pricing.price` be `0` for free pricing.

Screen Shot 2021-11-26 at 13 06 04

jamiehewitt15:
Yeah, that's good. I think we just need to add a note saying "Gas fees still apply"

PR #956, opened 2021-12-06
Fix/948 remove is valid url dependency

### LINKED ISSUE #948: Remove ocean.provider.isValidProvider dependency
https://github.com/oceanprotocol/market/blob/v4-publish/src/components/@shared/FormFields/Provider/index.tsx#L17

- figure out if this is still in new ocean.js, if not see what it was doing in old ocean.js
- create @utils/provider method to hit some endpoint directly
- this can be already tested as is accessible from publish form
- currently fails because there's no `OceanProvider` around publish form anymore, which old ocean.js needed (although still unclear if we use something similar with new ocean.js too)

So that in the end we get a success state for this after hitting the button:
Screen Shot 2021-11-24 at 16 17 56

claudiaHash:
Fixes #948  .

Changes proposed in this PR:

- use new ocean.js validation method
- create @utils/provider method to hit it directly


src/@utils/provider.ts
```diff
@@ -56,3 +56,23 @@ export async function getFileInfo(
     }
   }
 }
+
+export async function fetchMethod(url: string): Promise<AxiosResponse> {
```
mihaisc, line 60: why do we need this? instead of just doing  the axios get?

PR #959, opened 2021-12-10
Fix/old lib dep

mihaisc:
Various changes so that the market runs with the local version of ocean.js, all tests will fail. The local install from package.json assumes that market and ocean.js are in the same folder, if not `npm install {relative path to ocean.js folder}`


jamiehewitt15:
Is this ready for review? All the tests are failing at the moment?

mihaisc:
Yes, no tests will pass on any v4 branch, this is just for local use with local version of ocean.js v4main branch

PR #960, opened 2021-12-13
Remove ocean.provider.isValidUrl dependency

### LINKED ISSUE #948: Remove ocean.provider.isValidProvider dependency

claudiaHash:
Fixes #948  .

Changes proposed in this PR:

- removed ocean.provider.isValidUrl() dependency
- used new isValidUrl provider function

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -108,6 +108,7 @@
       }
     },
     "../ocean.js": {
+      "name": "@oceanprotocol/lib",
       "version": "1.0.0-alpha",
       "license": "Apache-2.0",
       "dependencies": {
--- a/src/@utils/provider.ts
+++ b/src/@utils/provider.ts
@@ -1,5 +1,5 @@
 import axios, { CancelToken, AxiosResponse } from 'axios'
-import { DID, LoggerInstance } from '@oceanprotocol/lib'
+import { LoggerInstance, ProviderInstance } from '@oceanprotocol/lib'
 
 export interface FileMetadata {
   index: number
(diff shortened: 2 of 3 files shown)


claudiaHash:
None of the tests will pass on any of the v4 branches. The only way to review this is taking it locally along with local version of `ocean.js` v4main branch

src/@utils/provider.ts
```diff
@@ -54,22 +54,19 @@ export async function getFileInfo(
...
   return result
 }
 
 export async function isValidProvider(url: string): Promise<boolean> {
   try {
-    const response = await ProviderInstance.isValidProvider(
-      url,
-      fetchMethod(url)
-    )
+    const response = await ProviderInstance.isValidProvider(url, fetchMethod)
```
mihaisc, line 69: why not just `fetch` , the default one , no extra imports or other functions needed

claudiaHash: true, just removed `fetchMethod` and used default `fetch`. thanks!

mihaisc (approved):

package-lock.json
```diff
@@ -108,6 +108,7 @@
       }
     },
     "../ocean.js": {
```
kremalicious, line 110: the package-lock has references to symlinked ocean.js which might make things go weird

src/@utils/provider.ts
```diff
@@ -56,3 +53,13 @@ export async function getFileInfo(
     }
   }
 }
+
+export async function isValidProvider(url: string): Promise<boolean> {
```
kremalicious, line 57: this method seems overkill as `await ProviderInstance.isValidProvider(url, fetch)` can simply be called in the `FormFields/Provider` part. Likewise the PR title/final commit message is super confusing as original task and todo was meant to remove that dependency and hit provider API directly, but this PR simply switches methods for it, so no dependency is removed. All titles should reflect that

package-lock.json
mihaisc: this is irrelevant for the moment, we each work with local ocean.js. It will be sorted up once we have a release of lib

PR #965, opened 2021-12-17
Integrate provider fileinfo

### LINKED ISSUE #961: Integrate provider 'Encrypt' file info API

claudiaHash:
Fixes #961 .

Changes proposed in this PR:

- integrate provider `fileinfo` function


mihaisc:
don't bother with encrypt, already handled in the publish pr

src/@utils/provider.ts
```diff
@@ -15,44 +14,44 @@ export async function getEncryptedFileUrls(
...
-    const url = `${providerUrl}/api/v1/services/encrypt`
-    const response: AxiosResponse<{ encryptedDocument: string }> =
-      await axios.post(url, {
-        documentId: did,
-        signature: '', // TODO: add signature
-        publisherAddress: accountId,
-        document: files
-      })
-    return response?.data?.encryptedDocument
+    const response = await ProviderInstance.encrypt(
```
mihaisc, line 17: this whole function is not relevant anymore ,i mean `getEncryptedFileUrls`

src/@utils/provider.ts
```diff
@@ -15,44 +14,44 @@ export async function getEncryptedFileUrls(
...
-export async function getFileInfo(
-  url: string | DID,
-  providerUrl: string,
-  cancelToken: CancelToken
-): Promise<FileMetadata[]> {
-  let postBody
-  try {
-    if (url instanceof DID) postBody = { did: url.getDid() }
-    else postBody = { url }
+async function fetchMethod(provider: string, url: string): Promise<Response> {
```
mihaisc, line 31: No need to create a separate function, used like this https://github.com/oceanprotocol/market/blob/16475b55d13b474840a6567e346ab3e466f37b7b/src/components/Publish/index.tsx#L188

src/components/@shared/FormFields/FilesInput/index.tsx
```diff
@@ -4,29 +4,23 @@ import { toast } from 'react-toastify'
...
 
     async function validateUrl() {
       try {
         setIsLoading(true)
-        const checkedFile = await getFileInfo(
-          url,
-          config?.providerUri,
-          newCancelToken()
-        )
+        const checkedFile = await fileInfo(url, config?.providerUri)
```
mihaisc, line 23: this would look like 
```
  const checkedFile = await  ProviderInstance.fileinfo(
        url,
        config.providerUri,
        (url: string, body: string) => {
          return axios.post(url, body, {
            cancelToken: newCancelToken()
          })
        }
      )
```

mihaisc: Also the function will be renamed `fileinfo` - > `checkFileUrl` .
The `fileinfo` endpoint has multiple functionalities, so we will implement another function in ocean.js that will return the file info from a did (not relevant for this pr, just for context)

package.json
```diff
@@ -38,6 +38,7 @@
     "decimal.js": "^10.3.1",
     "dom-confetti": "^0.2.2",
     "dotenv": "^10.0.0",
+    "electron": "^16.0.5",
```
kremalicious, line 41: why add electron?

kremalicious:
Now after merging publish PR into `v4` what we have in there right now all works, including the file check addressed in here. From what I see it would be a matter of which pattern to choose if we want to continue on here. In `v4` for the file info we hit the provider API directly while for encrypt method we use `ProviderInstance.encrypt` so for consistency we also should use `ProviderInstance.checkFileUrl` for `getFileInfo()`. This should stay as individual method cause we need file info also in other places.

But I tried to implement the changes in here in my local `v4` branch but I can't get `ProviderInstance.checkFileUrl` to return an actual response. The network response works but that method never gives me the actual response. Were you able to make this work in this branch? Meaning were you actually able to add a file in the UI with this change?

claudiaHash:
> Were you able to make this work in this branch? Meaning were you actually able to add a file in the UI with this change?

I did, but using a local version of ocean.js which included a change  (discussed with @bogdanfazakas , bot not merged yet) that allowed market client to use an axios method, as a fetch method for ProviderInstance.checkUrlFile, in the idea that it would be more consistent to use only axios fetch methods. 
I just converted the fetchMethod to `fetch`, so now it works.
Screenshot from 2022-01-12 14-47-36

src/components/@shared/FormFields/FilesInput/index.tsx
```diff
@@ -4,28 +4,33 @@ import { toast } from 'react-toastify'
...
-  const newCancelToken = useCancelToken()
 
   function loadFileInfo(url: string) {
     const config = getOceanConfig(chainId || 1)
 
     async function validateUrl() {
       try {
         setIsLoading(true)
-        const checkedFile = await getFileInfo(
+        const checkedFile = await ProviderInstance.checkFileUrl(
```
kremalicious, line 24: we still have `getFileInfo()` in `src/@utils/provider.ts` so instead of adding another way of doing the same thing only here, would suggest rather modifying `getFileInfo()`. Cause we use `getFileInfo()` in other places so we should make sure when we get the file info we always do it the same way

src/components/@shared/FormFields/FilesInput/index.tsx
```diff
@@ -69,3 +74,6 @@ export default function FilesInput(props: InputProps): ReactElement {
     </>
   )
 }
+function newCancelToken(): import('axios').CancelToken {
```
kremalicious, line 77: this looks like something automatic was added by your editor?

src/components/Publish/_utils.ts
```diff
@@ -59,6 +59,7 @@ function transformTags(value: string): string[] {
 
 export async function transformPublishFormToDdo(
   values: FormPublishData,
+  cancelToken: CancelToken,
```
kremalicious, line 62: why is this `cancelToken` now passed here if it is never used in this method?

src/components/Asset/AssetActions/index.tsx
```diff
@@ -65,14 +65,11 @@ export default function AssetActions({
       setFileIsLoading(true)
 
       const asset = values?.services?.[0].files?.[0].url || ddo.id
-      const providerUrl =
```
kremalicious, line 68: the above comment was also the reason for this here, so as that the file info in the publish preview works

kremalicious (approved):
all perfect now and working!

Issue #950, opened 2021-11-30
Publish a pool dataset


kremalicious:
Make publishing an asset work in most pragmatic way, meaning activating the prepared publish flow in the `handleSubmit` method and go from there:

- https://github.com/oceanprotocol/market/blob/v4/src/components/Publish/index.tsx#L38-L81
- might require more tweaks to the `transformPublishFormToDdo()` DDO construction in case of errors because of metadata

### Prerequisites & gotchas

- needs to be made work against new `provider`, `ocean.js`, and in end to see it, new `aquarius`. Ideally this happens against Rinkeby, but using `barge` is an option too although time might get lost for trying to get market to work with `barge`
- `createNftWithErc()` needs to be imported from new ocean.js, figure out what params in which form it needs and construct and pass them
- figure out how to get `nftAddress` & `datatokenAddress` from this tx
- for final `txPublish` figure out what method to use from ocean.js, or create own utility method
- most likely we have to comment out the Pool widget on asset details page if we want to check out the published asset, as this would require integrating new subgraph structure first

PR #967, opened 2021-12-21
Publish flow changes

mihaisc:
Draft for publish. Whole flow still needs to be fixed but there will be minor changes

ToDo @me : 
- move all publish logic into a nice hook and some utils.
- figure out what to do with service id
- `No config found for given network 'undefined'` in preview page
- algo preview crashes
- compute preview crashes

🔥 ⚠️  Publishing is currently working on this branch , the golden path at least , for the following schemas : 
- free OK
- fixed rated OK
- pool OK

Closes:

https://github.com/oceanprotocol/market/issues/950
https://github.com/oceanprotocol/market/issues/961
https://github.com/oceanprotocol/market/issues/962


content/publish/form.json
```diff
@@ -128,7 +128,7 @@
         "label": "Access Type",
         "help": "Choose how you want your files to be accessible for the specified price.",
         "type": "boxSelection",
-        "options": ["Download", "Compute"],
+        "options": ["Access", "Compute"],
```
kremalicious, line 131: This should stay as "Download", or more precisely we need to make sure that in the final UI "Download" is displayed. Pretty sure you changed this cause somewhere along the data stack we probably use these values here for some detection

kremalicious: we do another title mapping in the form field definitions so all good

PR #985, opened 2022-01-12
Publish feedback

### LINKED ISSUE #951: Create submission feedback components
Last step in publish form is the Submission screen (working title) on which we have room to provide detailed feedback for the publish process.

First thoughts on the UI can be found over in https://github.com/oceanprotocol/market/issues/862#issuecomment-957452405

For consistency we should pick up the styles from the current wizard navigation, like for marking current, success, etc. All the different states can be already created so in end we only need to hook it up with more added state in `Publish/index.tsx`.

We need to figure out under what circumstances we want to show which feedback. E.g. currently implemented flow together with #950 will result in most likely only 2 transactions, but the first one is doing multiple things in background we should visualize (create NFT, create Datatoken, create pool, move OCEAN/datatoken in pool). Then think of the other possibilities, fixed & free price.

This addresses multiple issues users had with our publish feedback, like #931 or #946

### LINKED ISSUE #978: Continue button enabled on valid screen
The `Continue` button on each screen of Publish steps should be disabled until the screen is valid, as suggested in https://github.com/oceanprotocol/market/issues/862#issuecomment-984393445

kremalicious:
- closes #951
- closes #978

Feedback data flow: using the Formik state to store and modify feedback data mainly so we can continue using only the `Publish/index.tsx` for tweaking the flow. So in the `Publish/Steps` I sync the modified feedback with Formik so that other child components like `Publish/Submission` can get it from there without any more refactor. No clue yet if that creates too much of a performance impact somewhere in UI.

And started with some first copy trying to explain the steps which I already split up in `v4` into multiple `try/catch` so we get some control over each one's success or error.

Screen Shot 2022-01-12 at 16 24 37

Screen Shot 2022-01-12 at 17 19 16

Screen Shot 2022-01-12 at 17 19 55

Screen Shot 2022-01-12 at 17 47 48

### DIFF
--- a/content/publish/feedback.json
+++ b/content/publish/feedback.json
@@ -1,0 +1,20 @@
+{
+  "1": {
+    "name": "Create Tokens & Pricing",
+    "description": "The Data NFT representing your asset, the Datatokens defining access to it, and the pricing schema are all created in one transaction.",
+    "status": "pending",
+    "txCount": 1
+  },
+  "2": {
+    "name": "Construct & Encrypt DDO",
+    "description": "Entered metadata is transformed into a structured document (DDO) where the file URLs, and the whole DDO itself are encrypted.",
+    "status": "pending",
+    "txCount": 0
+  },
+  "3": {
+    "name": "Publish DDO",
+    "description": "The encrypted DDO is stored on-chain as part of the Data NFT. Indexers like Aquarius can decrypt the DDO for displaying purposes, but the file URLs can only be decrypted by exchanging the respective datatokens for this asset.",
+    "status": "pending",
+    "txCount": 1
+  }
+}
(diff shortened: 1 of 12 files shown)


kremalicious:
well then, the mechanism seems to be working well so far:

Untitled

When rejecting the first transaction right away, all further will actions will switch to fail right away and do nothing:

Screen Shot 2022-01-12 at 13 54 23

Also outputting more info into console when debug mode is activated prefixed with `[publish]`.

Now remaining things is all around UX, disabling buttons, outputting tx links in between the steps, and maybe have a big success screen with link to asset

kremalicious:
Copy changes, and finally transparency into all the tx created beforehand, all fully working already in this PR:

Screen Shot 2022-01-12 at 15 22 23

That tx badge then switches to explorer link when successful:

Screen Shot 2022-01-12 at 15 22 41

Cool feature for later would be then to also show gas estimation for those tx on that label

kremalicious:
good enough, already working well and way better than anything we had before. If I did everything right, there will always be a final link to the new published asset page at end of process.

PR #1006, opened 2022-01-18
Use asset's ddo.purgatory

### LINKED ISSUE #953: Migrate asset purgatory mechanism
New `Asset` returned by `aquarius` will have `asset.purgatory`: https://github.com/oceanprotocol/docs/blob/feature/ddo_v4/content/concepts/did-ddo.md#purgatory

```json
{
  "purgatory": {
    "state": true,
    "reason": "Copyright violation"
  }
}
```

So we do not have to hit our own purgatory proxy anymore and can get the purgatory state directly from the `Asset`

claudiaHash:
Fixes #953  .

Changes proposed in this PR:

- used `ddo.purgatory` data
- removed `setPurgatory` function


src/@types/apollo/AssetFrePrice.d.ts
```diff
@@ -0,0 +1,31 @@
+/* tslint:disable */
```
kremalicious, line 1: please remove all those auto-generated files which are now under `src/@types/subgraph/`

src/@context/Asset.tsx
```diff
@@ -83,19 +83,6 @@ function AssetProvider({
     setLoading(false)
   }
 
-  const setPurgatory = useCallback(async (did: string): Promise<void> => {
-    if (!did) return
-
-    try {
-      const result = await getAssetPurgatoryData(did)
```
kremalicious, line 90: as you see with above GitHub warnings, `getAssetPurgatoryData()` is now unused in this file, so:

1. the import needs to be removed
2. the method itself might not be needed anymore, if we don't use it anywhere else then it can be removed

src/components/Publish/index.tsx
```diff
@@ -30,6 +30,8 @@ import { validationSchema } from './_validation'
 // TODO: restore FormikPersist, add back clear form action
 const formName = 'ocean-publish-form'
 
+/// // did:op:e63d482178fd140fce6eb476b734488d7c95bf825257b5e251126214d93139c8
```
kremalicious, line 33: probably added for testing purposes?

src/components/Asset/AssetContent/index.tsx
```diff
@@ -55,8 +55,8 @@ export default function AssetContent({
             )}
 
             <MetaFull ddo={ddo} />
-            {price?.datatoken && <EditHistory />}
-            {price?.datatoken && debug === true && (
+            {price?.datatoken !== null && <EditHistory />}
```
kremalicious, line 58: this is interesting find, the fine differences between `undefined`, `null`, `0`, `false` in js are always fun to deal with. Just a note that if we have to do more of those checks we should rather implement better returns in our method which creates that `price` object

src/components/Publish/index.tsx
claudiaHash: busted, i'll remove it immediately

src/components/Asset/AssetContent/index.tsx
claudiaHash: yess, took me some time to understand from where the 00 came from:

Screenshot from 2022-01-19 14-42-09

kremalicious: right, so I guess we sometime in our refactoring switched to have our `price` object never return undefined but always `0` as a string. I have lots of those `price.datatoken` checks over in #1009 so I'll double check them.

In any case, now that I look at this line I do not know what the purpose of this check is to render `EditHistory` so would try removing that check altogether and just output EditHistory. If something crashes then most likely EditHistory needs to be adapted.

kremalicious (approved):
perfect, so nice to see all the code lines removed

PR #1009, opened 2022-01-19
Pool tab refactor

### LINKED ISSUE #894: Refresh user liquidity after removing liquidity
After you remove liquidity you need to do a page refresh so the data in `Your liquidly` updated. It should be done immediately after adding or removing liquidity
image

### LINKED ISSUE #971: Restore pool tab
Restore functionality of the pool tab using the new subgraph schema

Do some add/remove/trade from 2-3 wallets (multiple people during pr review) for better testing.

Done:
-  you can view all the pool statistics without connecting a wallet . View own stats after connecting wallet

**Blocked by #954**

kremalicious:
- Fixes #971
- Fixes #894

Refactor with the goal to make data fetching/manipulation more efficient and for developers making that massive `Pool/index.tsx` easier to understand. We always added stuff in there and it took me like 1 hour to figure what exactly is going on in there in terms of data dependencies and I initially created this 😀

- structure the effects more logically, split up data fetching & data checking/manipulation: 0 data fetching, 1 pool info, 2 creator info, 3 user info
- structure the state data the same way in nested state objects, unify them (`poolInfo`, `poolInfoOwner`, `poolInfoUser`)
- clean up and tweak all the effect dependencies
- rename swap fee to pool fee (that's what it is called in contracts/subgraph)
- remove that refetch hack thingy, use new structure for refetch after add/remove
- fix user transactions section styles
- fix user pool share display in case of no liquidity, just return and display `0` in that case
- more helpful logging mode of responses and transformed data:
  Screen Shot 2022-01-19 at 16 56 03

### DIFF
--- a/src/@utils/subgraph.ts
+++ b/src/@utils/subgraph.ts
@@ -25,6 +25,7 @@ import {
 } from '../@types/subgraph/PoolShares'
 import { OrdersData_orders as OrdersData } from '../@types/subgraph/OrdersData'
 import { UserSalesQuery as UsersSalesList } from '../@types/subgraph/UserSalesQuery'
+import { PoolLiquidity } from 'src/@types/subgraph/PoolLiquidity'
 
 export interface UserLiquidity {
   price: string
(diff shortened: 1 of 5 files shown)


kremalicious:
Good enough, as it's now a cool base for all the child components like add/remove which receive the data from this file. Further work is needed on presentation as we move forward, which is why I didn't touch the markup at all, which would be in some need of splitting up in components.

PR #1015, opened 2022-01-21
Pricing tab validation fix

### LINKED ISSUE #999: Pricing validation not working properly
If you don't change anything (or just edit the liquidity) in the pricing creation step it will not be set as valid 
image
image

claudiaHash:
Fixes #999 .

Changes proposed in this PR:

- keep pricing tab number green for valid pricing values, touched or untouched


src/components/Publish/Navigation/index.tsx
```diff
@@ -1,16 +1,12 @@
 import { FormikContextType, useFormikContext } from 'formik'
 import React, { ReactElement } from 'react'
-import { FormPublishData } from '../_types'
+import { FormPublishData, StepContent } from '../_types'
```
kremalicious, line 3: why is `StepContent` newly imported here?

Btw, are you using VSCode with our suggested plugins? Then you should see live warnings about ESlint failures. There really is no need to only solve unnecessary imports later during PR review

Screen Shot 2022-01-25 at 09 55 20

kremalicious:
As I can't find this defined anywhere, what's the exact goal here. What should be the condition for which the pricing step turns green?

claudiaHash:
> As I can't find this defined anywhere, what's the exact goal here. What should be the condition for which the pricing step turns green?

The `Pricing` tab should be green even if no change was done among its fields, if the two anterior steps are completed successfully, so the [initial success condition](https://github.com/oceanprotocol/market/pull/1015/files#diff-e61b7d000733a5332a53c11d5a4ff721e587c7e2702f57edef6976b4b5f1f17aL23) of this step  is wrong. But only removing it doesn't help because then `Pricing` will be green anytime, not only after you successfully pass the `Metadata` and `Access` steps. So, I believe the goal would be to have a green pricing tab, and to have it only after the two anterior steps are completed successfully

l.e. I thought it's enough to link the issue [in here](https://github.com/oceanprotocol/market/pull/1015#issue-1110553697) but seems not

src/components/Publish/Navigation/index.tsx
claudiaHash: it's my fault beacuse I forgot to npm run lint

kremalicious:
great, thanks for the reasoning!

> The Pricing tab should be green even if no change was done among its fields

That's what's bugging me a bit as this is all nice technically but from interaction standpoint with this we encourage people to just "skip" the screen.

A better interaction to encourage using the screen would be working with empty fields rather than default values which then makes this implemented logic here redundant. We should try making the initial price field empty so users have to interact with it, while the OCEAN value could be filled with some default already. To make this even more clear, upon mount of that screen we could trigger the validation warning for that empty & untouched field.

Issue #1014, opened 2022-01-21
"Side staking bot"


kremalicious:
The new funky thing preventing rug pulls, and not even sure where the name comes from but it's floating around in the team. Issue is for deciding if it needs some visualization with its own numbers on the Pool tab, and if so, how that would look like and then implement. Advantage of visualizing it is that UI helps in understanding what's special with the new pools, explaining it without a wall of text.

The facts:

- bot (another Ethereum account) for each pool auto-controls datatoken supply
- whenever someone stakes OCEAN into a pool, the pool mints datatokens
- just enough datatokens are minted to keep datatoken price stable (constant DT:OCEAN ratio)
- whenever OCEAN is unstaked, datatokens are burned
- bot will receive half of the liquidity provider's pool shares

That last point might mean we have to calculate/show things like creator/user liquidity differently and not base this all on the user's pool shares.

trentmc:
A reference: Bancor has "single-sided staking". They have different underlying mechanics than Ocean, but the effect is the same: you're a liquidity provider into a pool yet you only have to give tokens on one side.

Here's what it looks like.
https://app.bancor.network/pools/add-liquidity/0xb1CD6e4153B2a390Cf00A6556b0fC1458C4A5533

In app, go to "Earn -> Pools". For a given token (e.g. ETH), see the "+" button. Click it.
![bancor0](https://user-images.githubusercontent.com/5305452/150537752-6c53ca06-ae42-4e50-af9f-d159c921f4e8.png)

After you've clicked, you get the following screen.
![bancor1](https://user-images.githubusercontent.com/5305452/150537754-ce887e90-38a5-4c5f-bf73-767318ac6dfa.png)

trentmc:
Notes:

IMO we should _not_ call it "side staking bot" or "staking bot". We should call it "single sided staking" or "one-sided staking". It's much simpler for the user, because it's less about the "how" (robots? huh?) and more about the "what" (you only have to stake one side).

kremalicious:
the contracts call it side staking, which then also populates to ocean.js calling it like that https://github.com/oceanprotocol/contracts/blob/v4main/contracts/pools/ssContracts/SideStaking.sol

trentmc:
> the contracts call it "side staking", which then also populates to ocean.js calling it like that

"Side staking" is a terrible label for users; we definitely shouldn't use it in the GUI. 

[...]

Issue #468, opened 2021-04-01
Market code can support different base token


trentmc:
**What**

Ocean backend code supports different base tokens. Ocean frontend code has a stronger bias to use OCEAN as a base token. 

Goal of this issue: make it more straightforward for Ocean frontend code to use a different base token.

**Motivation**

This helps in two ways:
1. Helps several 3rd party marketplaces, which would like to use their own token at the frontend level, eg for pools staking/swapping and for FRE.
2. Helps Ocean Market switch from using OCEAN, to an OCEAN-backed stablecoin -- [market#447](https://github.com/oceanprotocol/market/issues/447)

kremalicious:
No, we had this in mind when refactoring things but we never tested a different base token, and there are still places which assume OCEAN alone. We always said to do this after v4

mihaisc:
Ok, closing this. Concrete implementation of multiple approved tokens will be tracked here https://github.com/oceanprotocol/market/issues/1335

PR #1017, opened 2022-01-24
Restore add/remove liquidity

### LINKED ISSUE #972: Restore Add/Remove liquidity
Not much more to explain.
Also take in consideration https://github.com/oceanprotocol/market/issues/894

Done:
- you can successfully add liquidity and the amount added is reflected in the pool stats without refreshing page
- you can successfully remove liquidity and the amount removed is reflected in the pool stats without refreshing page

**Blocked by #954**

kremalicious:
The just make it work part. As seen in #1014, there might be lots of other design & UI changes we need to do so numbers actually make sense.

* closes #972
* remove adding datatoken to pool, remove `CoinSelect` components and all logic for switching tokens
* remove the "Advanced" removal screen
* partially addresses #468 by reducing hardcoded OCEAN to `baseToken` in many places, which would allow reading out `baseToken` from some market config later on


src/components/@shared/TokenApproval/index.tsx
```diff
@@ -29,34 +30,39 @@ export default function TokenApproval({
...
-    //   accountId,
-    //   spender
-    // )
-    // amount &&
-    //   new Decimal(amount).greaterThan(new Decimal('0')) &&
-    //   setTokenApproved(
-    //     new Decimal(allowance).greaterThanOrEqualTo(new Decimal(amount))
-    //   )
-  }, [tokenAddress, spender, accountId, amount, isAssetNetwork])
+  // TODO: how to check if token is approved as .allowance does not exist?
```
kremalicious, line 33: Ideas for checking if token and amount have already been approved for user + pool are very welcome here.

src/components/Asset/AssetActions/Pool/Add/index.tsx
```diff
@@ -80,38 +81,58 @@ export default function Add({
...
+    dtAddress,
+    coin,
+    dtBalance,
+    balance.ocean
+  ])
 
   // Submit
   async function handleAddLiquidity(amount: number, resetForm: () => void) {
+    const poolInstance = new Pool(web3, LoggerInstance)
+    const minPoolAmountOut = '0' // ? TODO: how to get?
```
kremalicious, line 125: how to calculate this for `joinswapExternAmountIn()`? Doe we have ocean.js method or contract method for this or is this supposed to be calculated based on user input value combined with slippage value?

src/components/@shared/TokenApproval/index.tsx
mihaisc: fixed in new version of ocean.js

src/components/Asset/AssetActions/Pool/Add/index.tsx
mihaisc: this needs some thinking, i think fees should be taken in consideration as well

kremalicious:
Should work™ as defined in linked issue and we need this here to better test further functionality like #1020 with which the updating liquidity after adding/removing should also work more reliably.

One thing already found, think the `.allowance()` doesn't work as expected, always returns `0` for me despite former approval. Also addd more logging in the PR, so that's more obvious now:

Screen Shot 2022-01-26 at 12 07 51

src/components/Asset/AssetActions/Pool/Add/index.tsx
kremalicious: good point, moved for now to https://github.com/oceanprotocol/market/issues/1027

PR #1019, opened 2022-01-24
Feature/seo

corrie-sloot:
Fulfilment of [mPowered’s DAO, Round 13](https://port.oceanprotocol.com/t/round-13-improving-search-engine-discoverability-for-the-current-ocean-data-marketplace/1308) relating to search engine optimisations (SEO).

Changes:
- remove bug which adds noindex,nofollow meta tag to **ALL** pages, now search engines will crawl the market
- add a third party gatsby plugin which generates a `robot.txt` file on build with a sitemap
- add a meta description to every asset page, which comes from the first 150 characters of the description markdown
- only show the ocean twitter meta tag when in the ocean marketplace

Note:
This PR is dependant on a change to aquarius. The `robots.txt` file refers to a new endpoint which outputs an XML sitemap of all appropriate asset URLs. Please do not merge until the [feature/seo](https://github.com/oceanprotocol/aquarius/pull/677) branch is merged into aquarius and implemented in production.

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -85,7 +85,9 @@ module.exports = {
       options: {
         sitemap: `${
           appConfig.metadataCacheUri
-        }/sitemap?base=${encodeURIComponent(siteContent.site.siteUrl)}`,
+        }/api/v1/aquarius/assets/sitemap?base=${encodeURIComponent(
+          siteContent.site.siteUrl
+        )}`,
         env: {
           development: {
             policy: [{ userAgent: '*', disallow: ['/'] }]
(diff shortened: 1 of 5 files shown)


kremalicious:
sorry, we are not going to accept more features on the v3 code base which currently is `main`. Would suggest adding this against `v4` branch which does not use Gatsby anymore and where this change might not be needed.

Also not sure if a change to aquarius is needed for that, as you can generate the robots.txt or sitemap.xml during build process of market alone. All this then also depends on constant rebuilds of market, as assets published after last deployment would be not included. Likewise, I am not sure the intended title& description for every asset details page would work, as that's only available after some fetching on client-side.

To me, if we want to really make sure all published assets are properly indexed we need to pre-render all asset details pages based on querying Aquarius on build time, then on client-side re-hydrate what we show by querying Aquarius again. Both Gatbsy & Next.js allow this to be done but it's complex and locks us into one framework.

PR #1024, opened 2022-01-25
Pool creation fail fix

### LINKED ISSUE #998: Pool creation fails
If you change the ocean value from the default it fails, at least it did for me, needs more investigation and testing with various combinations of price and liquidity.
image

claudiaHash:
Fixes #998  .

Changes proposed in this PR:

- use the correct denomination for the parameters used in the publish process for fixed/dynamic priced assets


claudiaHash:
Using the latest version of ocean.js and correct parameters denomination, assets can be published successfully (changes interfere with #1013 but I'll resolve the conflicts after it will be merged in v4 branch). still investigating why prices can't be seen after asset is published, even if they are set correctly in the publish step

mihaisc:
probably graph was not is sync with new contracts, now it is

claudiaHash:
yes, now it works

PR #1031, opened 2022-01-26
XSS fix

kremalicious:
1. check for actual URL, fun thing is that javascript: is a valid URL, so also checking for that before enabling the validate button.
2. when transforming data, just remove `javascript:` from all file URLs just in case


mihaisc:
publish seems to not work 
image

PR #1035, opened 2022-01-28
Invalid transction link fix

claudiaHash:
Fixes # 1033

Changes proposed in this PR:

- get subgraph nftUpdates query fix


mihaisc:
We can also update the text based on the type of update 
image

src/components/Asset/AssetContent/EditHistory.tsx
```diff
@@ -10,20 +10,38 @@ import styles from './EditHistory.module.css'
...
 
+  function getUpdateType(type: string): string {
+    switch (type) {
+      case 'METADATA_CREATED':
+        return 'published'
+      case 'METADATA_UPDATED':
+        return 'updated'
+      case 'STATE_UPDATED':
+        return 'state updated'
+      case 'TOKEN_UPDATED':
```
mihaisc, line 39: this is `TOKENURI_UPDATED`, not token. It's when `tokenUri` is updated on the nft

src/components/Asset/AssetContent/EditHistory.tsx
```diff
@@ -10,20 +10,38 @@ import styles from './EditHistory.module.css'
...
+  function getUpdateType(type: string): string {
+    switch (type) {
+      case 'METADATA_CREATED':
+        return 'published'
+      case 'METADATA_UPDATED':
+        return 'updated'
+      case 'STATE_UPDATED':
+        return 'state updated'
+      case 'TOKENURI_UPDATED':
+        return 'token uri updated'
```
mihaisc, line 40: not sure if people would properly understand this @kremalicious any ideas how the label should look ? `nft metadata updated` and for `METADATA_UPDATED` have `asset metadata updated` ?

kremalicious: yup, think "NFT metadata updated" for this case is better, for asset updated we can leave as is with "updated"

kremalicious (approved):

kremalicious:
we can ignore Vercel, just had to switch it to Gatbsy to get a v3 deployment out, and now set it back to Next.js builds.

PR #1046, opened 2022-02-01
Refactor pricing and various components that are involved

### LINKED ISSUE #970: Consume asset
Restore functionality of the `Buy` button. 
Same steps: 
- check on subgraph what pricing schema the datatoken has
- buy from pool/fre/dispenser
- startOrder from erc20 contract

Subgraph url: https://subgraphv4.rinkeby.oceanprotocol.com/

Done:
- successfully consume from all 3 pricing schemas

mihaisc:
Does not fix #970 anymore.

Added `AccessDetails` instead of `BestPrice`. This hold information regarding pricing, ownership, order. 
Added `AssetExtended` where we should add various fields that make sense in multiple components
Changed `ddo` with `asset` or `assetExtended` we should not use `ddo` further on


src/@context/Asset.tsx
```diff
@@ -10,19 +10,19 @@ import React, {
...
+import { getConsumeDetails } from '@utils/consumeDetails'
 
 interface AssetProviderValue {
   isInPurgatory: boolean
   purgatoryData: Purgatory
   ddo: Asset
   title: string
   owner: string
-  price: BestPrice
+  consumeDetails: ConsumeDetails
```
kremalicious, line 25: in general I would suggest we stop using `consume` also in our code base, and replace it with `access` so in this case `accessDetails`.

Reason being that on the UI side we never say "Consume", I mean, who would speak like that, nobody says "I'm gonna consume that file.". "I'm gonna access that file" much more so. Using "access" as the umbrella term for download/compute/(streaming later on?) mirrors how we communicate to end users in our UI.

mihaisc: fair enough, more refactor yay

src/@context/Asset.tsx
```diff
@@ -58,110 +59,125 @@ function AssetProvider({
...
-      setError(undefined)
-    }
-    setLoading(false)
-    return ddo
-  }
+  const fetchAsset = useCallback(
+    async (token?: CancelToken) => {
+      LoggerInstance.log('[asset] Init asset, get assetExtended')
+      setLoading(true)
+      const assetExtended = await retrieveDDO(did, token)
```
kremalicious, line 66: slightly confusing naming in this method as this is getting `Asset`, but not `AssetExtended`

src/@context/Asset.tsx
```diff
@@ -58,110 +59,125 @@ function AssetProvider({
...
+        setError(undefined)
+      }
+      setLoading(false)
+      return assetExtended
+    },
+    [did]
+  )
 
-  const refreshDdo = async (token?: CancelToken) => {
+  const refreshAsset = async (token?: CancelToken) => {
```
kremalicious, line 81: do we need this method at all? Seems everything it does could be done above in `fetchAsset`, then we just call this

src/@context/Asset.tsx
mihaisc: There are probably still more ddos left, just trying to get this out of the way to fix the consume.

src/@context/Asset.tsx
mihaisc: no idea, was used in compute, didn't have time to check it out

Issue #1027, opened 2022-01-26
add liquidity: calculate minPoolAmountOut, calculate max


kremalicious:
Following up on #1017, calculating `minPoolAmountOut` passed to `joinswapExternAmountIn()` and calculating what we called "Max" before. For `minPoolAmountOut` fees should be taken in consideration as well

See `/AssetActions/Pool/Add/index.tsx`

mihaisc:
Will be handled in https://github.com/oceanprotocol/ocean.js/pull/1306

PR #1047, opened 2022-02-01
Restore trade tab

### LINKED ISSUE #973: Restore Trade tab
Restore trade using the new contracts.  this as well as https://github.com/oceanprotocol/market/issues/972 will probably need some changes in `ocean.js` as well.

Done: 
- price impact is calculated correct ( need to define what correct means)
- slippage is taken into account 
- you can successfully swap ocean to dt
- you can successfully swap dt to ocean

### LINKED ISSUE #1143: Profile image
image
The image says it all

claudiaHash:
* Fixes #973 and #1079 
* more refactor towards #468 
* Fixes #1143

Changes proposed in this PR:

- restore trade tab functionality


src/components/Asset/AssetActions/Trade/index.tsx
```diff
@@ -1,19 +1,106 @@
...
+    initialPoolInfo as PoolInfo
+  )
+  const [poolData, setPoolData] = useState<PoolDataPoolData>()
+  const [fetchInterval, setFetchInterval] = useState<NodeJS.Timeout>()
+
   const [maxDt, setMaxDt] = useState('0')
-  const [maxOcean, setMaxOcean] = useState('0')
+  const [maxBaseToken, setMaxBaseToken] = useState('0')
+
+  const fetchPoolData = useCallback(async () => {
```
kremalicious, line 42: duplicating fetching logic just to share data between 2 components should be avoided at all times and if that is needed moving logic either to parent component for both child components, or a fetching context provider usually is the best solution. So I did the latter over in #1057 so in here you can do after its merge:

```js
const {
  poolData,
  poolInfo,
  poolInfoUser,
  poolInfoOwner,
  poolSnapshots,
  hasUserAddedLiquidity,
  isRemoveDisabled,
  refreshInterval,
  fetchAllData
} = usePool()
```

And whatever other data you need you would then update the pool context provider.

src/components/Asset/AssetActions/Trade/Swap.tsx
```diff
@@ -56,148 +61,182 @@ export default function Swap({
...
-    // const spotPrice = await ocean.pool.getSpotPrice(
-    //   price.address,
-    //   tokenIn,
-    //   tokenOut
-    // )
-
-    // setSpotPrice(spotPrice)
+    if (name === 'baseToken') {
+      if (values.type === 'sell') {
+        newValue = await poolInstance.calcPoolOutGivenSingleIn(
```
mihaisc, line 171: the `calc....` functions are for liquidity add/remove calculation. We need to use `getAmountInExactOut ` and `getAmountOutExactIn` here

mihaisc:
look ok at a quick glimpse, didn't get to test it properly

mihaisc:
please merge v4 and use `appConfig.consumeMarketPoolSwapFee`

src/components/Asset/AssetActions/Trade/FormTrade.tsx
```diff
@@ -64,32 +73,70 @@ export default function FormTrade({
...
+      const tokenOutMarket: TokenInOutMarket = {
+        tokenIn: poolInfo.datatokenAddress,
+        tokenOut: poolInfo.baseTokenAddress,
+        marketFeeAddress: appConfig.consumeMarketPoolSwapFee
+      }
+
       const impact = new Decimal(
         new Decimal(100).sub(new Decimal(values.slippage))
       ).div(100)
       const precision = 15
```
kremalicious, line 105: oh boy, great leftover. Anybody knows a reason why we specifically were using `15` here? Cause in all other files we do something like `Decimal.set({ precision: 18 })` and not redefine that within the file

src/components/Asset/AssetActions/Trade/FormTrade.tsx
```diff
@@ -64,32 +73,70 @@ export default function FormTrade({
...
-      //           .toString()
-      //       )
-      // setTxId(tx?.transactionHash)
+
+      const amountsInOutMaxFee: AmountsInMaxFee = {
+        tokenAmountIn: new Decimal(values.datatoken)
+          .mul(impact)
+          .toFixed(precision)
+          .toString(),
+        minAmountOut: '2',
```
kremalicious, line 112: I guess we don't know yet how to calculate this hardcoded number?

src/components/Asset/AssetActions/Trade/FormTrade.tsx
```diff
@@ -64,32 +73,70 @@ export default function FormTrade({
...
+        swapMarketFee: swapMarketFeeIn
+      }
+
+      const amountsOutMaxFee: AmountsOutMaxFee = {
+        maxAmountIn: '50',
```
kremalicious, line 117: same here, guess we don't know yet how to calculate this hardcoded number?

src/components/Asset/AssetActions/Trade/Output.tsx
```diff
@@ -31,12 +34,14 @@ export default function Output({
     if (!poolAddress || !isAssetNetwork) return
 
     async function getSwapFee() {
-      // const swapFee = await ocean.pool.getSwapFee(poolAddress)
+      if (!web3) return
+      const poolInstance = new Pool(web3)
+      const swapFee = await poolInstance.getSwapFee(poolAddress)
```
kremalicious, line 39: mmmh, we should not rely on web3 calls to get this but use the subgraph instead, as this should always work, no matter if wallet is connected or not.

But if I remember correctly, fees are not fully implemented yet in subgraph so will always return 0 when queried so for time being, web3 calls are only way to make it work at least

src/components/Asset/AssetActions/Trade/FormTrade.tsx
claudiaHash: yes, not sure if i missed something but i think it's somehow related to [this issue](https://github.com/oceanprotocol/market/issues/1027)

src/components/Asset/AssetActions/Trade/FormTrade.tsx
claudiaHash: yup, exactly

src/components/Asset/AssetActions/Trade/Output.tsx
mihaisc: you don't even need this.

src/components/Asset/AssetActions/Trade/Output.tsx
```diff
@@ -28,28 +30,28 @@ export default function Output({
 
   // Get swap fee
   useEffect(() => {
-    if (!poolAddress || !isAssetNetwork) return
+    if (!poolInfo || !isAssetNetwork) return
 
     async function getSwapFee() {
```
kremalicious, line 35: 1. nothing `async` going on in here anymore so no need to wrap what's happening in this effect in a method.
2. `poolInfo` might change a lot but effect only makes use of `poolInfo.poolFee` so better use `poolInfo?.poolFee` in the dependency array, and at top `if (!poolInfo?.poolFee ...)...`
3. don't think we need `poolAddress` then in the dependency array as we are not using it anymore in this effect

src/components/Asset/AssetActions/Trade/Swap.tsx
```diff
@@ -56,148 +61,189 @@ export default function Swap({
...
-
-      // setDtItem((prevState) => ({
-      //   ...prevState,
-      //   amount: amountDataToken.toString(),
-      //   maxAmount: maximumDt.toString()
-      // }))
+          : new Decimal(maxBaseToken)
+
+      try {
+        const swapFee = await poolInstance.getSwapFee(
```
kremalicious, line 82: we need to make sure this returns same thing as when getting it from subgraph as right now we have 2 different ways of getting swap fee in this PR, output component and in here

src/components/Asset/AssetActions/Trade/Swap.tsx
```diff
@@ -56,148 +61,189 @@ export default function Swap({
   const [spotPrice, setSpotPrice] = useState<string>()
   const [totalValue, setTotalValue] = useState<string>()
   const [tokenAmount, setTokenAmount] = useState<string>()
+  const [swapFee, setSwapFee] = useState<string>()
 
   useEffect(() => {
-    if (!asset || !balance || !values?.type) return
+    if (!assetExtended || !balance || !values?.type) return
```
kremalicious, line 67: looking through effect I would assume it should `return` also when no `web3` available

src/components/Asset/AssetActions/Trade/Swap.tsx
```diff
@@ -56,148 +61,189 @@ export default function Swap({
...
+    web3,
+    dtItem.token,
+    baseTokenItem.token
   ])
 
   const switchTokens = () => {
     setFieldValue('type', values.type === 'buy' ? 'sell' : 'buy')
-    setCoin(values.type === 'sell' ? 'OCEAN' : asset.datatokens[0].symbol)
+    setCoin(
+      values.type === 'sell' ? 'OCEAN' : assetExtended.datatokens[0].symbol
```
kremalicious, line 155: last remaining hardcoded OCEAN in this PR, we should be able to replace all that with info from `usePool()` hook

src/components/Asset/AssetActions/Trade/Swap.tsx
```diff
@@ -7,44 +7,49 @@ import { FormikContextType, useFormikContext } from 'formik'
...
-import { Asset } from '@oceanprotocol/lib'
+import { LoggerInstance, Pool } from '@oceanprotocol/lib'
 import { AssetExtended } from 'src/@types/AssetExtended'
+import { usePool } from '@context/Pool'
 
 Decimal.set({ toExpNeg: -18, precision: 18, rounding: 1 })
 
 export default function Swap({
-  asset,
+  assetExtended,
```
kremalicious, line 21: the prop name can just stay as `asset` which is also consistent to all other components when using this prop

src/components/Asset/AssetActions/Trade/index.tsx
```diff
@@ -43,6 +47,13 @@ export default function Trade(): ReactElement {
...
   // Get maximum amount for either OCEAN or datatoken
   useEffect(() => {
+    if (
+      !web3 ||
+      !isAssetNetwork ||
+      !asset.accessDetails.addressOrId ||
+      asset.accessDetails.price === 0
+    )
+      return
     if (
```
mihaisc, line 57: this check is done twice, slightly different?

src/components/Asset/AssetActions/Trade/Swap.tsx
claudiaHash: right, i'll use the poolInfo data in here too

src/components/Asset/AssetActions/Trade/index.tsx
```diff
@@ -30,7 +34,7 @@ export default function Trade(): ReactElement {
     async function getTokenBalance() {
       const datatokenInstance = new Datatoken(web3)
       const dtBalance = await datatokenInstance.balance(
-        asset.services[0].datatokenAddress,
+        asset.accessDetails.datatoken.address,
```
kremalicious, line 37: just for consistency, throughout this PR, the datatoken info should always come from the same place as I found e.g. for address:

- `asset.services[0].datatokenAddress`
- `asset.accessDetails.datatoken.address`
- `poolInfo.datatokenAddress`

Considering that in future we might have multiple services and multiple datatokens, most future proof in this context here seems to be `poolInfo.datatokenAddress` as for swap we're interacting with a specific pool.

src/components/Asset/AssetActions/Trade/FormTrade.tsx
mihaisc: i'm not sure if we can set this globally, that would be great. Needs some testing, i'll create a new issue

src/components/Asset/AssetActions/Trade/Output.tsx
```diff
@@ -7,19 +7,21 @@ import styles from './Output.module.css'
...
 import Decimal from 'decimal.js'
 import { FormTradeData } from './_types'
+import { usePool } from '@context/Pool'
 
 Decimal.set({ toExpNeg: -18, precision: 18, rounding: 1 })
 
 export default function Output({
   dtSymbol,
-  oceanSymbol,
+  baseTokenSymbol,
```
kremalicious, line 16: with that new `usePool()` there would be no need to pass around all those props as you can get them from within this component with that hook under `poolInfo`

src/components/Asset/AssetActions/Trade/FormTrade.tsx
```diff
@@ -64,32 +71,70 @@ export default function FormTrade({
     .defined()
 
   async function handleTrade(values: FormTradeData) {
+    if (!web3 || !asset || !poolInfo || !values) return
+
     try {
+      const poolInstance = new Pool(web3)
+
+      const swapMarketFeeIn = await poolInstance.getMarketFees(
```
mihaisc, line 79: no need to get these 2 fees.

src/components/Asset/AssetActions/Trade/FormTrade.tsx
```diff
@@ -64,32 +71,70 @@ export default function FormTrade({
...
-      //       )
-      // setTxId(tx?.transactionHash)
+
+      const amountsInOutMaxFee: AmountsInMaxFee = {
+        tokenAmountIn: new Decimal(values.datatoken)
+          .mul(impact)
+          .toFixed(precision)
+          .toString(),
+        minAmountOut: '2',
+        swapMarketFee: swapMarketFeeIn
```
mihaisc, line 111: this is `appConfig.consumeMarketPoolSwapFee`

src/components/Asset/AssetActions/Trade/FormTrade.tsx
```diff
@@ -64,32 +71,70 @@ export default function FormTrade({
...
+      }
+
+      const amountsOutMaxFee: AmountsOutMaxFee = {
+        maxAmountIn: '50',
+        tokenAmountOut: new Decimal(values.baseToken)
+          .mul(impact)
+          .toFixed(precision)
+          .toString(),
+        swapMarketFee: swapMarketFeeOut
```
mihaisc, line 120: this is `appConfig.consumeMarketPoolSwapFee`

src/components/Asset/AssetActions/Trade/FormTrade.tsx
```diff
@@ -64,32 +71,70 @@ export default function FormTrade({
...
+      const tokenInOutMarket: TokenInOutMarket = {
+        tokenIn: poolInfo.baseTokenAddress,
+        tokenOut: poolInfo.datatokenAddress,
+        marketFeeAddress: appConfig.consumeMarketPoolSwapFee
+      }
+      const tokenOutMarket: TokenInOutMarket = {
+        tokenIn: poolInfo.datatokenAddress,
+        tokenOut: poolInfo.baseTokenAddress,
+        marketFeeAddress: appConfig.consumeMarketPoolSwapFee
+      }
```
mihaisc, line 98: this is an addres, meaning ` appConfig.marketFeeAddress`

src/components/Asset/AssetActions/Trade/FormTrade.tsx
```diff
@@ -64,32 +71,70 @@ export default function FormTrade({
...
+
+      const swapMarketFeeOut = await poolInstance.getMarketFees(
+        asset.accessDetails.addressOrId,
+        poolInfo.datatokenAddress
+      )
+
+      const tokenInOutMarket: TokenInOutMarket = {
+        tokenIn: poolInfo.baseTokenAddress,
+        tokenOut: poolInfo.datatokenAddress,
+        marketFeeAddress: appConfig.consumeMarketPoolSwapFee
```
mihaisc, line 92: this is an addres, meaning ` appConfig.marketFeeAddress`

mihaisc:
I need to do some tests when everything is back up and running, i have a feeling that we don't need both `swapExactAmountIn`  and also `swapExactAmountOut` . As for `maxAmountIn` and `minAmountOut` those should be the values inlcuding slippage. As i said, more details after i can actually do some tests

src/components/Asset/AssetActions/Trade/FormTrade.tsx
mihaisc: commented bellow, no actions needed yet

src/components/Asset/AssetActions/Trade/FormTrade.tsx
mihaisc: commented bellow, no actions needed yet

src/components/Asset/AssetActions/Trade/index.tsx
mihaisc: `Considering that in future we might have multiple services and multiple datatokens` for this we need to change everything 🤣 .But consistency is good  nonetheless.

mihaisc:
@claudiaHash after you finish the comments you can take a break on this pr. It's hard to review it without doing some proper tests regarding the core functionality so a lot of comments might be useless.

src/components/Asset/AssetActions/Trade/Swap.tsx
```diff
@@ -79,23 +79,24 @@ export default function Swap({
...
-          amountBaseToken.toString(),
-          swapFee
-        )
+        const maxBuyBaseToken: PoolPriceAndFees =
+          await poolInstance.getAmountOutExactIn(
+            asset.accessDetails.addressOrId,
+            poolInfo.datatokenAddress,
+            poolInfo.baseTokenAddress,
+            amountDataToken.toString(),
+            poolInfo.poolFee
```
mihaisc, line 90: this is `appConfig.consumeMarketPoolSwapFee`

src/components/Asset/AssetActions/Trade/Swap.tsx
```diff
@@ -79,23 +79,24 @@ export default function Swap({
...
+          )
+        const maxBuyDt: PoolPriceAndFees =
+          await poolInstance.getAmountOutExactIn(
+            asset.accessDetails?.addressOrId,
+            poolInfo.baseTokenAddress,
+            poolInfo.datatokenAddress,
+            amountBaseToken.toString(),
+            poolInfo.poolFee
```
mihaisc, line 98: `appConfig.consumeMarketPoolSwapFee` as above

mihaisc:
Good for testing.

EnzoVezzaro:
when approving the tokens, the message on the loader is confusing as it's not actually swapping the tokens, but approving them. 

Screenshot 2022-02-23 at 16 28 58

Screenshot 2022-02-23 at 16 28 50

The rest seems to be working

mihaisc:
The issue with decimals with leading 0 is still there. It is something related to fom/validation. It's weird

claudiaHash:
> Does it make sense to keep the price impact? It will only be red :)

Maybe it's a good idea to change it's color if we keep it. It may give user the impression he set trade parameters wrong.

mihaisc:
Fixed

src/components/Asset/AssetActions/Trade/Swap.tsx
```diff
@@ -93,6 +101,20 @@ export default function Swap({
             : new Decimal(balance.baseToken)
           : maxBaseTokenFromPool
 
+      const reserver = await poolInstance.getReserve(
+        asset.accessDetails.addressOrId,
+        poolInfo.datatokenAddress
+      )
+      console.log(
```
EnzoVezzaro: comment to remove

package.json
```diff
@@ -38,12 +38,14 @@
...
     "formik": "^2.2.9",
     "gray-matter": "^4.0.3",
+    "i": "^0.3.7",
     "is-url-superb": "^6.1.0",
     "js-cookie": "^3.0.1",
     "jwt-decode": "^3.1.2",
     "lodash.debounce": "^4.0.8",
     "lodash.omit": "^4.5.0",
     "next": "^12.1.0",
+    "npm": "^8.5.1",
```
EnzoVezzaro, line 48: do we really need npm install on the project?

mihaisc: i have no idea how that got there

src/@utils/nft.ts
```diff
@@ -113,8 +113,6 @@ export async function setNftMetadata(
     []
   )
 
-  console.log('[setNftMetadata]  est Gas set metadata --', estGasSetMetadata)
```
kremalicious, line 116: having this logged looks like it was intended, but could use the `LoggerInstance` instead

src/components/Asset/AssetActions/Trade/Output.tsx
```diff
@@ -4,93 +4,90 @@ import { useAsset } from '@context/Asset'
...
+          }
+          balance={outputWithSlippage}
         />
       </div>
       <div>
         <p>Swap fee</p>
         <Token
-          symbol={`${values.type === 'buy' ? oceanSymbol : dtSymbol} ${
-            swapFee ? `(${swapFee}%)` : ''
+          symbol={`${
```
kremalicious, line 81: consistency, don't need to stringify this and just output values like above

kremalicious:
right after swapping, we should make sure to update all shown numbers in UI, seems the maximum numbers take some time to adapt. I was first swapping OCEAN → DT, after success clicked arrow to turn around, but the numbers took a while to become correct.

First this:

Screen Shot 2022-03-02 at 15 24 30

Then most likely after our refetch interval for `poolInfo` numbers start to become correct:

Screen Shot 2022-03-02 at 15 24 38

kremalicious:
The max button does not make sense, and it might actually be a conceptional thing as I'm not sure the max here is understood by most users.

Currently, typed in 1 for DT:

Screen Shot 2022-03-02 at 16 23 00

then hit max button:

Screen Shot 2022-03-02 at 16 20 20

The max button should calculate the maximum amount I can receive considering the input of the first field. I think. Or maybe should update first field? Right now it just takes the maximum from the bottom field. 

Maybe we should just get rid of that button and make the `Maximum to ...` number lines clickable so their values just end up in respective input field and be done with it

src/@context/Pool/index.tsx
```diff
@@ -119,8 +119,6 @@ function PoolProvider({ children }: { children: ReactNode }): ReactElement {
     if (!poolData) return
 
     // Fees - this will be renamed again in subgraph
-    const poolFee = getFee(poolData.liquidityProviderFee)
-    const marketFee = getFee(poolData.marketSwapFee)
     const opfFee = getFee(poolData.opcFee)
```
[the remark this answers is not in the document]
mihaisc: i'm not sure i get what you are saying here

src/@utils/constants.ts
```diff
@@ -0,0 +1 @@
+export const MAX_DECIMALS = 5
```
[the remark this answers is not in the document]
mihaisc: well, this is not a variable, this is a const. It is a const in case we decide to change it (as a developer) but it's not a variable to be changed whenever by who deploys the market.

src/@utils/nft.ts
mihaisc: well, got used to removing all console.log , since i often forget them 😄

src/components/Asset/AssetActions/Trade/Output.tsx
mihaisc: we kinda do, we have 2 strings here, above it's just one logical condition

src/@context/Pool/index.tsx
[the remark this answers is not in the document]
mihaisc: got it

mihaisc:
Remove max button due to increase complexity of that implementation. This pr needs to be merged

kremalicious (approved):

PR #1055, opened 2022-02-03
Refactor Asset context provider

kremalicious:
- various renaming and method reordering
- make sure only relevant parts of asset state are updated
- remove redundant refetch method
- rename `assetExtended` param & props to `asset`


src/@context/Asset.tsx
```diff
@@ -48,136 +48,130 @@ function AssetProvider({
...
 
   // -----------------------------------
   // Check user network against asset network
   // -----------------------------------
   useEffect(() => {
-    if (!networkId || !assetExtended) return
+    if (!networkId || !asset) return
 
-    const isAssetNetwork = networkId === assetExtended?.chainId
+    const isAssetNetwork = networkId === asset?.chainId
```
mihaisc, line 140: should we stop using both `networkId` and `chainId` and stick with one? In theory there are differences, but we use it with the same meaning . Not sure if relevant to this exact pr

kremalicious: it kinda depends on what exactly Aquarius means with `chainId`. 

Just seeing that in our web3 context provider we catch both, `chainId` (`web3.eth.getChainId()`) & `networkId` (`web3.eth.net.getId()`)  so we in market codebase we can simply switch to use `chainId` instead. But we need to double check if it's the same for Aquarius.

kremalicious: switching to `chainId` in this context worked so just did it in here. Won't touch any other place for now

PR #1057, opened 2022-02-03
Pool stats updates

kremalicious:
Addressing #1014


kremalicious:
pool share percentage:
- can we just double it? 
- why is it not 50% here:
   152583361-3edce99c-87fc-41b4-a53c-689e7d02fc3a

mihaisc:
Will investigate the 50% part. Not sure minted/burned makes sense or if we can get it with the current subgraph (it is possible with some changes)

PR #1061, opened 2022-02-04
fix file info

### LINKED ISSUE #1059: fileinfo is broken on asset details
On consume/compute fileinfo does not show up, when looking at code it simply can't work as we are not getting the actual file URL:

https://github.com/oceanprotocol/market/blob/v4/src/components/Asset/AssetActions/index.tsx#L70-L72

We either use the preview files or pass sample URLs for whatever reason. Both does not help for actual file URLs

kremalicious:
closes #1059

Multiple things:
- use different methods from ocean.js depending on if file to check is a URL or a DID, this really should be just one method in ocean.js detecting either DID or URL but that's what it is right now
- grab the `providerUrl` from the DDO service instead of ocean config
- reverts #1029 where I couldn't figure out anymore why we added the sample files, as we never display them with the file icon

### DIFF
--- a/src/@utils/provider.ts
+++ b/src/@utils/provider.ts
@@ -18,7 +18,24 @@ export async function getEncryptedFiles(
   }
 }
 
-export async function getFileInfo(
+export async function getFileDidInfo(
+  did: string,
+  serviceId: string,
+  providerUrl: string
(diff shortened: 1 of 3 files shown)


claudiaHash (approved):
File info is now available for every asset, LGTM!

Issue #990, opened 2022-01-13
Pricing screen tabs


kremalicious:
As [noted](https://github.com/oceanprotocol/market/issues/862#issuecomment-984393445) misusing the tabs in our pricing screen for a selection state can be confusing:

<img width="537" alt="Screen Shot 2022-01-13 at 12 10 13" src="https://user-images.githubusercontent.com/90316/149327922-fdf920dc-a2b5-44ef-a233-cf0bdddc1ee9.png">

They are exclusive options so semantically they should be `radio` inputs and visually if there would be an element resembling a radio input inside of the tabs, it would make it already way more clear that selecting a tab also selects this as pricing option.

So we need to look into how much we can customize existing tabs of our `Tabs` component, then decide if we might create new custom elements only for this screen.

EnzoVezzaro:
Hi guys,

I wanted to make a brief summary to make sure we're in the same page. 

Based on the description, we need to make more clear the selection of the pricing model on the "Pricing" step of the Publishing wizard. For what I understand is just an UI addition to improve the UX. 

The route that I'd take is the follow: 

- Add prop to "Tabs" component (example: ShowRadio )
- This is an optional param that if passed:
-- Adds a selected Radio input in the correct active tab, while keeping disabled on the rest

This approached makes possible to add the Radio buttons only to the "Pricing" step on the Wizard. 

Looking at the code this could be done as follow: 

- Add prop to "Tabs" component (example: showRadio ): `src/components/Asset/AssetActions/index.tsx`
-- add param `showRadio` to Tabs component
- If the `showRadio` param is passed: `src/components/@shared/atoms/Tabs.tsx`
-- Add optional `showRadio` param to the the component
-- Add Radio Buttons to the `Tab` component inside the `TabList` map loop
-- Add logic to show/hide Radio Buttons

This is preview of the final result if what desired (is not yet styled).

![Screenshot 2022-02-01 at 13.27.34.png](https://images.zenhubusercontent.com/61f7b0225ea40b38a8684622/f56de447-2aa3-4a3c-b362-c18cc901a8d3)

PR #1067, opened 2022-02-07
added radio to pricing tabs on publish

EnzoVezzaro:
Enhancement [Pricing screen tabs #990](https://github.com/oceanprotocol/market/issues/990)

Changes proposed in this PR:

- Added radio inputs to pricing tabs on publish

### DIFF
--- a/src/components/@shared/FormInput/InputElement.module.css
+++ b/src/components/@shared/FormInput/InputElement.module.css
@@ -100,6 +100,10 @@
   padding-left: 0.5rem;
 }
 
+.radioLabelPricing {
+  margin-left: 10px;
+}
+
 .algorithmLabel {
   display: grid;
   gap: var(--spacer);
--- a/src/components/@shared/FormInput/InputElement.tsx
+++ b/src/components/@shared/FormInput/InputElement.tsx
@@ -87,6 +87,23 @@ export default function InputElement({
         />
       )
     case 'radio':
+      return (
+        <div className={styles.radioWrap}>
+          <input
+            className={styles[type]}
+            type={type}
+            name={name}
+            checked={props.defaultChecked}
+            {...props}
+            readOnly
+          />
+          <label
+            className={cx({ [styles.radioLabelPricing]: true, [size]: size })}
+          >
+            {label}
+          </label>
+        </div>
+      )
     case 'checkbox':
       return (
         <div className={styles.radioGroup}>
--- a/src/components/@shared/atoms/Tabs.tsx
+++ b/src/components/@shared/atoms/Tabs.tsx
@@ -1,5 +1,6 @@
 import React, { ReactElement, ReactNode } from 'react'
 import { Tab, Tabs as ReactTabs, TabList, TabPanel } from 'react-tabs'
+import InputElement from '@shared/FormInput/InputElement'
 import styles from './Tabs.module.css'
 
 export interface TabsItem {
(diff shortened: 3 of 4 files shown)


src/components/@shared/FormInput/InputElement.tsx
```diff
@@ -87,6 +87,23 @@ export default function InputElement({
         />
       )
     case 'radio':
+      return (
```
kremalicious, line 90: do we really need to introduce extra component for this as using same component for radio & checkbox worked out so far, In any case, if new markup is needed then you need to consider all places where radios are already used, right now e.g. the radio labels in the publish form don't come through correctly anymore with this change:

Screen Shot 2022-02-09 at 13 02 33

src/components/@shared/FormInput/InputElement.tsx
```diff
@@ -87,6 +87,23 @@ export default function InputElement({
...
+          <input
+            className={styles[type]}
+            type={type}
+            name={name}
+            checked={props.defaultChecked}
+            {...props}
+            readOnly
+          />
+          <label
+            className={cx({ [styles.radioLabelPricing]: true, [size]: size })}
```
kremalicious, line 101: this is introducing a generic component so should not be bound specific to the place it is being used. So should be just `radioLabel` but as mentioned above might not be needed to use extra markup here

src/components/@shared/FormInput/InputElement.module.css
```diff
@@ -100,6 +100,10 @@
   padding-left: 0.5rem;
 }
 
+.radioLabelPricing {
+  margin-left: 10px;
```
kremalicious, line 104: we try to limit magic numbers and absolute units. Ideally, calculate this with `var(--spacer)`, or use `rem` unit

kremalicious:
Style-wise, radios appear too big compared to its context. So instead of trying to make smaller radios, would try to increase top/bottom padding of the tabs which also makes them less look like our general tabs which is a good thing here. Used `padding: calc(var(--spacer) / 6) var(--spacer)` to get this, think `/ 4` would work too:

Screen Shot 2022-02-09 at 13 08 56

src/components/@shared/FormInput/InputElement.module.css
EnzoVezzaro: ok, got it. This has been remove.

src/components/@shared/FormInput/InputElement.tsx
EnzoVezzaro: you're completely right, sorry about that. This has been removed. Now this input has been corrected.

src/components/@shared/FormInput/InputElement.tsx
EnzoVezzaro: this class been removed with component. 👍🏽

src/components/Publish/Pricing/index.module.css
```diff
@@ -2,6 +2,10 @@
   text-align: center;
 }
 
+.pricing ul > li {
```
kremalicious, line 5: this seems like rather adventurous selector as it most likely matches way more than intended cause it matches all list items within the whole pricing screen. If there is no way to get a CSS module class onto what you want to select, then try to be as specific as possible, e.g. by using some attribute selector like we do in some places in this file like with `[class*='Tabs_tabContent']`

kremalicious (approved):
great, think that's good enough to make the selected state more clear

PR #1068, opened 2022-02-08
Restore order

### LINKED ISSUE #970: Consume asset

mihaisc:
Fixes #970 #1091


package.json
```diff
@@ -88,18 +88,28 @@
...
     "eslint-plugin-react": "^7.28.0",
     "eslint-plugin-react-hooks": "^4.3.0",
     "file-loader": "^6.2.0",
+    "https-browserify": "^1.0.0",
     "husky": "^7.0.4",
+    "os-browserify": "^0.3.0",
     "prettier": "^2.5.1",
     "pretty-quick": "^3.1.3",
+    "process": "^0.11.10",
+    "react-app-rewired": "^2.1.11",
```
kremalicious, line 107: do we really need all those new packages? Like, `assert`, `process`, `react-app-rewired`, `url` can all be removed from here until we really need them when used in next.config.js

src/@context/Pool/index.tsx
```diff
@@ -191,10 +194,11 @@ function PoolProvider({ children }: { children: ReactNode }): ReactElement {
       !poolData ||
       !poolInfo?.totalPoolTokens ||
       !asset?.chainId ||
-      !accountId
+      !accountId ||
+      !poolInfoUser
     )
       return
-
+    console.log('poolInfoUser', poolInfoUser)
```
kremalicious, line 201: missing console.log

src/@context/Pool/index.tsx
```diff
@@ -89,7 +92,7 @@ function PoolProvider({ children }: { children: ReactNode }): ReactElement {
       )
     }, refreshInterval)
     setFetchInterval(newInterval)
-  }, [fetchInterval, fetchAllData, accountId])
+  }, [fetchInterval, fetchAllData])
```
kremalicious, line 95: did you notice the code comment above this method and made sure switching accounts correctly switches the interval too? Cause I had issues where I switched accounts, numbers switched correctly, but then old interval kicked in and fetched old numbers.

src/@context/Pool/index.tsx
mihaisc: this is core feature !

src/@context/Pool/index.tsx
```diff
@@ -191,10 +194,11 @@ function PoolProvider({ children }: { children: ReactNode }): ReactElement {
       !poolData ||
       !poolInfo?.totalPoolTokens ||
       !asset?.chainId ||
-      !accountId
+      !accountId ||
+      !poolInfoUser
```
kremalicious, line 198: this `poolInfoUser` then also needs o be added to dependency array of this effect

package.json
mihaisc: `react-app-rewired` was due to copy paste frenzy. The rest are just devDependecies, i could try removing them but this whole mess was very time consuming

src/@utils/accessDetailsAndPricing.ts
```diff
@@ -235,6 +234,7 @@ export async function getAccessDetails(
   )
 
   const tokenPrice: TokenPrice = tokenQueryResult.data.token
+  console.log('acces token pri', tokenPrice, timeout)
```
kremalicious, line 237: missing console.log

src/@utils/feedback.ts
```diff
@@ -1,89 +1,13 @@
...
-  8: '5/5 Asset published succesfully'
-}
-
-// TODO: do something with this object,
-// consumeStep should probably return one of those strings
-// instead of just a number
-export const consumeFeedback: { [key in number]: string } = {
-  ...feedback,
-  3: '3/3 Access granted. Consuming file...'
+// TODO: can be better
```
kremalicious, line 1: think quick win would be to make it more generic as a method accepting `datatokenSymbol` & `baseTokenSymbol` and then switch them in the text

src/@context/Pool/index.tsx
mihaisc: i have to admit that i miss that, will investigate

src/@utils/feedback.ts
```diff
@@ -1,89 +1,13 @@
...
-
-export function getDispenseFeedback(dtSymbol: string): {
-  [key: number]: string
-} {
-  return {
-    1: `1/2 Requesting ${dtSymbol}...`,
-    2: `2/2 Received ${dtSymbol}.`
-  }
+  0: 'Ordering asset...',
+  1: 'Transfering data token.',
```
kremalicious, line 11: data token → datatoken. Or also pass `datatokenSymbol` here as mentioned above

src/@utils/order.ts
```diff
@@ -0,0 +1,100 @@
...
+import { getOceanConfig } from './ocean'
+import { TransactionReceipt } from 'web3-eth'
+import { getSiteMetadata } from './siteConfig'
+
+/**
+ * For pool you need to buy the datatoken beforehand, this always assumes you want to order the first service
+ * @param web3
+ * @param asset
+ * @param accountId
+ * @returns {TokenReceipt} receipt of the order
```
kremalicious, line 19: beautiful code comments, yay! This returns `TransactionReceipt`

src/@utils/order.ts
```diff
@@ -0,0 +1,100 @@
...
+ */
+export async function order(
+  web3: Web3,
+  asset: AssetExtended,
+  accountId: string
+): Promise<TransactionReceipt> {
+  const datatoken = new Datatoken(web3)
+  console.log('asset order', asset)
```
kremalicious, line 27: I guess some logging would be nice but maybe we leave the logging to the component initiating the order and leave the helper method without logging? Otherwise needs to use the `LoggerInstance` here

src/@utils/order.ts
```diff
@@ -0,0 +1,100 @@
...
+    consumer: accountId,
+    serviceIndex: 0,
+    _providerFees: initializeData.providerFee
+  } as OrderParams
+
+  switch (asset.accessDetails?.type) {
+    case 'fixed': {
+      // this approve implies that basetToken is the same as swap fee token
+      const totalCost =
+        asset.accessDetails.price + Number.parseFloat(appConfig.marketFee)
```
kremalicious, line 49: is this calculation safe? Still don't know if all our `Decimal` conversion and checks are all still needed in the pool tab, but if they are we should calculate with `Decimal` whenever we do calculations

mihaisc: it's safe market fee will be something like 0.2 so no problems there

src/@utils/pool.ts
```diff
@@ -0,0 +1,87 @@
...
+  accountId: string,
+  config: Config,
+  web3: Web3
+): Promise<TransactionReceipt> {
+  const pool = new Pool(web3)
+  const { appConfig } = getSiteMetadata()
+  await approve(
+    web3,
+    accountId,
+    config.oceanTokenAddress,
```
kremalicious, line 65: we should limit the usage of getting token info from the ocean config, as what we want to have here is the address of the base token from the pool the user wants to buy a datatoken from. So should rather be `accessDetails.baseToken.address` and then no need to pass `config`

src/@utils/order.ts
mihaisc: this is just a forgotten debug

src/@context/Pool/index.tsx
```diff
@@ -68,7 +69,9 @@ function PoolProvider({ children }: { children: ReactNode }): ReactElement {
     setPoolData(response.poolData)
     setPoolInfoUser((prevState) => ({
       ...prevState,
-      poolShares: response.poolDataUser?.shares[0]?.shares
+      poolShares: response.poolDataUser?.shares[0]
```
kremalicious, line 72: I think `poolShares: response?.poolDataUser?.shares[0]?.shares || '0'` does the same but shorter

src/@utils/order.ts
mihaisc: fixed just to be on the safe side

src/@utils/order.ts
```diff
@@ -0,0 +1,99 @@
...
+      const freParams = {
+        exchangeContract: config.fixedRateExchangeAddress,
+        exchangeId: asset.accessDetails.addressOrId,
+        maxBaseTokenAmount: web3.utils.toWei('1'),
+        swapMarketFee: web3.utils.toWei(appConfig.marketFee),
+        marketFeeAddress: appConfig.marketFeeAddress
+      } as FreOrderParams
+
+      const tx = await datatoken.buyFromFreAndOrder(
+        asset.services[0].datatokenAddress,
```
kremalicious, line 70: not sure if any implication but we should get datatoken address from one place only throughout this method and above we get it from `asset.accessDetails.datatoken.address`

src/@context/Pool/index.tsx
mihaisc: no, it will loop then because you set it in that effect. It's a good compromise for now, no time to refactor everything

src/@utils/pool.ts
```diff
@@ -0,0 +1,86 @@
...
+import { getDummyWeb3 } from './web3'
+import { TransactionReceipt } from 'web3-eth'
+
+export async function calculateBuyPrice(
+  accessDetails: AccessDetails,
+  web3?: Web3,
+  chainId?: number
+): Promise<PriceAndEstimation> {
+  if (web3 && chainId)
+    throw new Error("web3 and chainId can't be undefined at the same time!")
```
kremalicious, line 13: Logic and message are kinda mismatched, what is the intention here?

src/@utils/order.ts
mihaisc: no implications,the one in the service is 100% there because that is the source, but i'll change to keep it consistent

src/@utils/pool.ts
mihaisc: is it clearer now with the description?

mihaisc:
Still needs a bit more testing

src/components/Asset/AssetActions/Download.tsx
```diff
@@ -0,0 +1,167 @@
...
+}: {
+  asset: AssetExtended
+  file: FileMetadata
+  isBalanceSufficient: boolean
+  dtBalance: string
+  fileIsLoading?: boolean
+  isConsumable?: boolean
+  consumableFeedback?: string
+}): ReactElement {
+  const [accessDetails, setAccessDetails] = useState<AccessDetails>()
```
kremalicious, line 34: do we specifically need to store this in a new state? I mean it's under `asset.accessDetails` so can be grabbed from there or checked from there if it's there or not

src/components/Asset/AssetActions/Download.tsx
```diff
@@ -0,0 +1,167 @@
...
+  const [isLoading, setIsLoading] = useState(false)
+
+  useEffect(() => {
+    if (!asset || !asset.accessDetails) return
+
+    setAccessDetails(asset.accessDetails)
+  }, [asset])
+
+  useEffect(() => {
+    if (!accessDetails) return
```
kremalicious, line 49: if you don't store this in a new state in here, then this line could be simply `if (!asset?.accessDetails) return`

app.config.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
   marketFeeAddress:
     process.env.NEXT_PUBLIC_MARKET_FEE_ADDRESS ||
     '0x9984b2453eC7D99a73A5B3a46Da81f197B753C8d',
+  marketFee: process.env.NEXT_PUBLIC_MARKET_FEE || '0.2',
```
kremalicious, line 25: Let's add comment above to make it clear what value is needed here, along the lines of:

```
// Marketplace fee in % of each swap
```

src/@utils/pool.ts
kremalicious: So the error message should be `web3 and chainId can't be defined at the same time`?

mihaisc:
Due to contracts update and some other functionalities needing testing downgrading to draft

src/components/Asset/AssetActions/Download.tsx
```diff
@@ -0,0 +1,167 @@
...
+        setStatusText(
+          getOrderFeedback(
+            accessDetails.baseToken.symbol,
+            accessDetails.datatoken.symbol
+          )[1]
+        )
+        const orderTx = await order(web3, asset, accountId)
+
+        accessDetails.isOwned = true
+        accessDetails.validOrderTx = orderTx.transactionHash
```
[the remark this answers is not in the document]
mihaisc: when switching routes data will be refetched and all good. You need order for download

src/components/Publish/Preview/index.tsx
```diff
@@ -4,37 +4,35 @@ import { FormPublishData } from '../_types'
...
 
   useEffect(() => {
     async function makeDdo() {
-      const asset = await transformPublishFormToDdo(values)
-      setAsset(asset as Asset)
-
+      const asset = (await transformPublishFormToDdo(values)) as AssetExtended
       // dummy BestPrice to trigger certain AssetActions
-      const accessDetails: AccessDetails = {
+      asset.accessDetails = {
```
kremalicious, line 17: no direct manipulation of React state objects. This should be:


```
const accessDetails = { ... }

setAsset((prevState) => ({
  ...prevState,
  accessDetails
}))
```

mihaisc: this doesn't make sense i'm manipulating ` const asset`  and then setting it ` setAsset(asset)`

src/components/Asset/AssetActions/Download.tsx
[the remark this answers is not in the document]
kremalicious: so if we use that `asset.accessDetails` as a storage for values from child components then we need to use `setAsset` for this, also to be able to know from where that asset context provider state is modified. Not sure we should do this though, but I can see how this limits the amount of params we have to pass to things like `downloadFile` but the alternative would be using local state in the child component for `validOrderTx`.

And `accessDetails.isOwned` should be detected and set directly in the asset context provider and not from this child component

src/components/Publish/Preview/index.tsx
kremalicious: yup you,re right, all good. Was too quick when scanning all changes for direct React state manipulation

src/@utils/pool.ts
mihaisc: nope, undefined at same time, if web3 is there, chainId is ignored .

kremalicious: but the logic says, "if `web3` and `chainId` are defined at same time, return an error", that's why I'm a bit confused.

But already shows we should rather deal with that in our web3 context provider and initiate web3 at all times with `config.nodeUri` fallback, not just through our `connect` method cause also doing this in `useGraphSyncStatus`. Then we would not need hacks like this but oh boy, might be fun with web3modal and could become quickly out of scope

mihaisc:
stop testing significant refactors incoming  +  ocean.js update needed because contracts lib automatically updated to 18 and now nothing is working

src/@utils/pool.ts
mihaisc: > but the logic says, "if web3 and chainId are defined at same time, return an error", that's why I'm a bit confused.
my god.... 😆 , it was working properly that's why i didn't notice what i did.

Can't handle that globally because you don't know on what chain you will need it

src/@context/Pool/index.tsx
mihaisc: restored

src/@types/Price.d.ts
```diff
@@ -1,14 +1,53 @@
+import { ProviderFees } from '@oceanprotocol/lib'
+
+/**
+ * @interface OrdePriceAndFee
```
kremalicious, line 4:
```suggestion
 * @interface OrderPriceAndFee
```

Amazing documentation here though 😍

src/components/Asset/AssetActions/Download.tsx
```diff
@@ -0,0 +1,176 @@
...
+  const PurchaseButton = () => (
+    <ButtonBuy
+      action="download"
+      disabled={isDisabled}
+      hasPreviousOrder={isOwned}
+      hasDatatoken={hasDatatoken}
+      dtSymbol={asset?.datatokens[0]?.symbol}
+      dtBalance={dtBalance}
+      datasetLowPoolLiquidity={!asset.accessDetails?.isPurchasable}
+      onClick={handleConsume}
```
kremalicious, line 140:
```suggestion
      onClick={handleDownload}
```

kremalicious:
UX things:

On pool assets, first message I saw was "Buying MANJEL..." for which I first got an approve prompt in MetaMask for OCEAN.

On fixed assets, first message was correctly the "Approving OCEAN..." when MetaMask approve prompt came up

After that finalized I got this here with "Approving OCEAN...":

Screen Shot 2022-02-14 at 13 32 39

So think messages should be turned around, first "Approving OCEAN to buy MANJEL...", then "Buying MANJEL..."

At the very end ended up in this view with the Download button which is cool, we could improve on that to make all one flow and see if we can trigger the download button programmatically? Then we don't have to deal with the slight mental model mismatch with messages "....allowing you to use it without paying again." although user has not downloaded yet.

Screen Shot 2022-02-14 at 13 35 44

Seems to all work very well on pool assets. On fixed assets I always get failed transaction but maybe I use an old asset or something. Using `did:op:644a98f06b404a4c6ebe6f4a052d9cf23c978888fbf3bcfaa37da30c6a329545` and getting this:

`Fail with error 'FixedRateExchange: Too many base tokens'`

https://rinkeby.etherscan.io/tx/0x67a7a05b78bcc97db27d2087b3e06a9c1ce293ab9f57abb1b4c2f1b04438e715

src/@context/Pool/index.tsx
kremalicious: good find! To keep PR reviews sane, could you add a code comment like in this PR please? https://github.com/oceanprotocol/market/pull/1094/files

.env.example
```diff
@@ -1,6 +1,11 @@
 
 #NEXT_PUBLIC_INFURA_PROJECT_ID="xxx"
 #NEXT_PUBLIC_MARKET_FEE_ADDRESS="0xxx"
+#NEXT_PUBLIC_PUBLISHER_MARKET_FEE="1"
```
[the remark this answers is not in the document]
mihaisc: damn, got some corrupt github files at some point and had to restore everything, seems i missed this.

mihaisc:
Something is weird for the pool there for you. For fixed there are some old assets where ordering is not working.

mihaisc:
I would like to avoid doing any more refactoring that concerns UI/UX (if it's not major) because we need to merge this. We can always create more issues/prs

kremalicious (approved):
Merge, merge, merge! Also so satisfying this huge refactor removes more code than it adds

PR #1069, opened 2022-02-08
Restore compute functionality

### LINKED ISSUE #976: Restore compute functionality
Depends on https://github.com/oceanprotocol/market/issues/975

This is a bigger issue because i feel that it would be more efficient to tackle it in one go. It is probably a mixed bag between market and ocean.js

Done:
- easy, just have compute working like in v3.

bogdanfazakas:
Fixes #976  .

Changes proposed in this PR:

- restore order and start compute logic
- handle algorithms and pricing for start compute jobs
- add some helpers that previously were in oceanjs lib
- only private algorithms can be published (publish issues #1161 and #1160 )
- fixed your compute jobs section on asset details page
- fixed compute jobs status and download compute results
- Restore Profile compute jobs tab
- Updated Ui messages and button disable logic for starting compute jobs 
- Implemented reuse order  

To do: 

- [x] fix missing token symbol when asset to start job on is FREE
- [x] replace number conversions to decimals
- [x] fix job details modal that closes automatically (maybe related to a the following  Warning: Formik context is undefined, please verify you are calling useFormikContext() as child of a <Formik> component) the component has no timeout set and tried to force setting it to not close and we have the same result
- [x] new issue with pool price algorithms after getPricesAndFees changes (maybe refactor also for pool assets)
- [x] issue when trying to fetch prices and fees and not connected on the same network as the asset
- [x] UI/UX adjustments (center start job button, text message updates, alerts, etc)
- [x] Start a compute job immediately after starting one does not really work (recheck after the flow refactors using initializeCompute)
- [x] Integrate initializeCompute and refactor start compute flow and how we handle price display and token orders
- [x] Compute results not shown
- [x] Provider issue on mumbai (big response times, and 504 bad gateway on initalizeCompute)
- [x] Provider issue on reuse order 'AttributeDict' object has no attribute 'serviceIndex'
- [ ] Find a solution for trusted algorithm list checksum ( not sure if this is a problem on the market side at the moment, maybe the error handling can be better)
- [x] Publish algorithm's with custom docker not indexed
- [ ] With the current maxDuration (60 sec) value in computeEnv the provider fees can expire between the order call and the start compute job, especially when the testnets are congested
- [x] display separate provider fees when order is valid and only fees need to be paid, and proper message below the buy/order/start button
- [x] block algo selection until the started job is completed 
- [x] check unsuported networks
- [x] fix provider fees grather than 0


mihaisc:
Also take this in consideration https://github.com/oceanprotocol/market/issues/1101

src/@utils/accessDetailsAndPricing.ts
```diff
@@ -222,13 +222,21 @@ function getAccessDetailsFromTokenPrice(
...
+}
+
 /**
  * This will be used to get price including feed before ordering
  * @param {AssetExtended} asset
  * @return {Promise<OrdePriceAndFee>}
  */
 export async function getOrderPriceAndFees(
   asset: AssetExtended,
+  computeEnv: string = null,
```
mihaisc, line 238: put compute env after accountId , because we use accountId more often so we don't need to add `null, null` in params

src/@utils/ddo.ts
```diff
@@ -1,15 +1,27 @@
...
-  name: 'access' | 'compute'
+  name: 'access' | 'compute' | 'metadata'
 ): Service {
   if (!ddo) return
 
   const service = ddo.services.filter((service) => service.type === name)[0]
   return service
 }
 
+export function getServiceById(ddo: Asset | DDO, serviceId: string): Service {
```
mihaisc, line 18: This is a genuine question for basically everyone. Do we need one liner functions like this? Do they bring any benefit?

src/@utils/accessDetailsAndPricing.ts
bogdanfazakas: done my bad :) saw that `accountId?` was optional and didn't  want the linter and build to fail because we have optional values before required params, and for some reasons i forgot that initialised method params are considered optionals by default 🤦

src/@utils/ddo.ts
bogdanfazakas: from my perspective they have some benefits (small ones thats true 😃 ), like across the stack if you want to get a certain service by Id or name like in this particular example and you remember having this helpers and just call this method. Also the code readability  is better this way in my opinion if we have it like this 
`const service =  getSerivceById(asset, serviceId)
  doSomethingWithTheService(service)`  
rather than this 
`const service =  ddo.services.find((s) => s.id === serviceId)
  doSomethingWithTheService(service)`
  and maybe less codeclimate issues 😄 .
But totally got your point we can also get rid of them or just keep the ones we think bring some value and benefits to the stack 😃.

mihaisc: fair enough

mihaisc:
If selected network is different from the asset network then you will not see any compute jobs.This is because of reusing the compute jobs component.
image
image

Can't approve algorithm on asset, didn't see any errors.
image

src/@utils/aquarius.ts
```diff
@@ -40,7 +46,7 @@ export function generateBaseQuery(
         filter: [
           ...(baseQueryParams.filters || []),
           getFilterTerm('chainId', baseQueryParams.chainIds),
-          getFilterTerm('_index', 'aquarius'),
+          // getFilterTerm('_index', 'aquarius'),
```
mihaisc, line 49: why was this commented?

src/components/@shared/atoms/Blockies.tsx
```diff
@@ -10,6 +10,8 @@ export default function Blockies({
   className?: string
 }): ReactElement {
   if (!accountId) return null
+
+  // TODO: just put this back once finished
```
mihaisc, line 14: i think we can remove this comment

src/components/Asset/AssetActions/Compute/FormComputeDataset.tsx
```diff
@@ -79,42 +92,59 @@ export default function FormStartCompute({
...
-      hasPreviousOrder || hasDatatoken ? 0 : Number(asset.accessDetails.price)
+      hasPreviousOrder || hasDatatoken
+        ? 0
+        : Number(datasetOrderPriceAndFees?.price || asset.accessDetails.price)
     const priceAlgo =
       hasPreviousOrderSelectedComputeAsset || hasDatatokenSelectedComputeAsset
         ? 0
-        : Number(algorithmConsumeDetails.price)
+        : Number(
+            algoOrderPriceAndFees?.price ||
```
mihaisc, line 134: stop using number, only Decimal

src/components/Asset/AssetActions/Compute/FormComputeDataset.tsx
```diff
@@ -144,9 +174,11 @@ export default function FormStartCompute({
         hasDatatoken={hasDatatoken}
         selectedComputeAssetTimeout={selectedComputeAssetTimeout}
         hasDatatokenSelectedComputeAsset={hasDatatokenSelectedComputeAsset}
-        algorithmConsumeDetails={algorithmConsumeDetails}
+        algorithmConsumeDetails={selectedAlgorithmAsset?.accessDetails}
         symbol={oceanSymbol}
         totalPrice={Number.parseFloat(totalPrice)}
```
mihaisc, line 179: not ok, should be string or Decimal

EnzoVezzaro:
I'm experiencing these issues:

1) transaction history list doesn't finish loading
2) centered button instead of having it aligned left (I think it's a better position)

https://user-images.githubusercontent.com/13741335/163960607-97bba851-2979-4b43-b4a6-5ebbfe62cfe6.mp4

I'm not sure fees calculations are correct, I'm testing it now. So far I have found this issue:

1) If you select an algo option on a network that is not the assets, you get this error in console. Maybe we can avoid doing the call when on the wrong network

Screen Shot 2022-04-19 at 4 32 03 AM

claudiaHash:
We may hide the 'Your Compute Jobs' section in the Preview step, since it's show button it's disabled anyway:

imageedit_3_8317081433

src/components/Asset/AssetActions/Compute/FormComputeDataset.tsx
```diff
@@ -156,7 +188,7 @@ export default function FormStartCompute({
           !isValid ||
           !isBalanceSufficient ||
           !isAssetNetwork ||
-          algorithmConsumableStatus > 0
+          !selectedAlgorithmAsset?.accessDetails.isPurchasable
```
EnzoVezzaro, line 191: here we can add another chain op to avoid posible issues. change `!selectedAlgorithmAsset?.accessDetails.isPurchasable` to `!selectedAlgorithmAsset?.accessDetails**?**.isPurchasable`

bogdanfazakas:
> If selected network is different from the asset network then you will not see any compute jobs.This is because of reusing the compute jobs component. 
Hmm think it happens when you have no jobs on the selected network. In my case i have jobs on all and the jobs are here even on i'm not connected to the assets network.
Screenshot 2022-04-20 at 12 14 14

> Can't approve algorithm on asset, didn't see any errors. 
I was able to update the trusted algo list for assets see: (did:op:bb75cc003e60093601383eaa99760dee1eab6285ea29d55bd4714e03f9d70c36) but will check with free assets also saw that the asset from image is an asset with dispenser price schma

EnzoVezzaro:
when you are on this modal and the timer clicks, the modal closes.

Screen Shot 2022-04-20 at 8 20 13 AM

To reproduce:

- Go to [Asset](https://market-40rs5prfi-oceanprotocol.vercel.app/asset/did:op:86844e28945bfde0cc9cfa255ae10f7a8dcff2538f7a62190b2ada3710550ea5)
- Run an algo if not done yet
- Open Transaction History
- Click SHOW DETAILS

Just leave the modal opened, it'll close alone once the timer kicks in.

Another little thing I find

Screen Shot 2022-04-20 at 8 16 46 AM

Here we say "Watch the progress bellow or ...", I was actually looking for something below this text instead on opening the "Your Compute Jobs". I think is a bit confusing, we can change it to "Watch Your Transaction History down below or ..." to avoid any confusion.

AnaLoznianu:
> > If selected network is different from the asset network then you will not see any compute jobs.This is because of reusing the compute jobs component.
> 
> Hmm think it happens when you have no jobs on the selected network. In my case i have jobs on all and the jobs are here even on i'm not connected to the assets network. Screenshot 2022-04-20 at 12 14 14
> 
> > Can't approve algorithm on asset, didn't see any errors.
> 
> I was able to update the trusted algo list for assets see: (did:op:bb75cc003e60093601383eaa99760dee1eab6285ea29d55bd4714e03f9d70c36) but will check with free assets also saw that the asset from image is an asset with dispenser price schema. Later update it was working with free assets also (did:op:504d0038ae3038e3c6657803bf5cab5acb6c937fec0cbaf4dc31c5a6176c5820)

Yes, I noticed the same issue. When there are no compute jobs, the list is stuck in the loading state. https://market-git-feature-v4-c2d-oceanprotocol.vercel.app/asset/did:op:3979313203a1293b6d1c7260883561d6e35ae936ed95dd4a0e60c1f14cb5f677 
Screenshot 2022-04-20 at 15 47 40

AnaLoznianu:
On this [asset](https://market-40rs5prfi-oceanprotocol.vercel.app/asset/did:op:86844e28945bfde0cc9cfa255ae10f7a8dcff2538f7a62190b2ada3710550ea5), the amount to pay to start an algorithm is not very clear. I know it's OCEAN tokens, but we should display it

Screenshot 2022-04-20 at 16 03 27
Screenshot 2022-04-20 at 16 03 34

EnzoVezzaro:
> On this [asset](https://market-40rs5prfi-oceanprotocol.vercel.app/asset/did:op:86844e28945bfde0cc9cfa255ae10f7a8dcff2538f7a62190b2ada3710550ea5), the amount to pay to start an algorithm is not very clear. I know it's OCEAN tokens, but we should display it
> 
> Screenshot 2022-04-20 at 16 03 27
> 
> Screenshot 2022-04-20 at 16 03 34

Good point.

AnaLoznianu:
When editing an algorithm, we have the option to add "allowed algorithms" via the compute jobs settings. And you can add the same algorithm you are editing 
This seems wrong...

https://user-images.githubusercontent.com/84312885/164244375-873d7b56-c621-4e23-9ad3-d6cc6c5ec11e.mov

AnaLoznianu:
In the publishing process, in the preview step, the user has access to the "Select an algorithm to start a compute job" option. The functionality though is not working. Nothing can be clicked at this phase of the process

https://user-images.githubusercontent.com/84312885/164245148-94d7ad76-6cd8-4d87-a9b8-9ab23718091b.mov

EnzoVezzaro:
> In the publishing process, in the preview step, the user has access to the "Select an algorithm to start a compute job" option. The functionality though is not working. Nothing can be clicked at this phase of the process
> 
>  Screen.Recording.2022-04-20.at.16.45.41.mov

Yeah, we need to work on this. It's more an UX problem. We need to communicate better that this is a "Preview" of what you'll get (as listing in the market) after publishing, but it's not really a workable component. I get the confusion, we might just need to make more evident that the preview it's not a useful/workable component, but it's just a presentation of the end UI you'll get as listing.

AnaLoznianu:
> rk on this. It's more an UX problem. We need to communicate better that this is a "Preview" of what you'll get (as listing in the market) after publishing, but it's not really a workable component. I get the confusion, we might just need to make more evident that the preview it's not a useful/workable component, but it's just a presentation of the end UI you'll get as listing.

It's not the preview. It happens for Free assets

AnaLoznianu:
> > In the publishing process, in the preview step, the user has access to the "Select an algorithm to start a compute job" option. The functionality though is not working. Nothing can be clicked at this phase of the process
> 
> That's ok, it's a preview, it's intended. Improving the preview is veeeeeery low priority 😁

It's not the preview. It happens for Free assets

EnzoVezzaro:
> > > In the publishing process, in the preview step, the user has access to the "Select an algorithm to start a compute job" option. The functionality though is not working. Nothing can be clicked at this phase of the process
> > 
> > 
> > That's ok, it's a preview, it's intended. Improving the preview is veeeeeery low priority 😁
> 
> It's not the preview. It happens for Free assets

I couldn't replicate this on the Free assets, they seem to work properly. Maybe it has already been fixed.

LucaMilanese90:
Is it intended behaviour to see the "Set allowed algorithm" section when I'm in the process of editing the compute settings of an algorithm? Shouldn't this view show me the datasets instead? 
Screenshot 2022-04-28 at 10 18 08

bogdanfazakas:
> Is it intended behaviour to see the "Set allowed algorithm" section when I'm in the process of editing the compute settings of an algorithm? Shouldn't this view show me the datasets instead? !

No, we need to disable editing compute settings on algorithms

LucaMilanese90:
I see a weird behaviour when trying to edit the metadata of an asset: if I'm connected with another account and then switch to the account that published the asset, the "edit" button appears, but if I click, it says I don't have permission to edit. After a page refresh it starts working correctly again

https://user-images.githubusercontent.com/20192135/165713218-5e3c4b35-e051-45dd-a83b-641a34bf73d8.mov

.

LucaMilanese90:
When trying to publish an asset (both for download and ctd) I'm not allowed to select a fixed price for more than 1 ocean.

https://user-images.githubusercontent.com/20192135/165726762-dc2147f6-1fb5-4da5-b8ca-0aa7b5732ad8.mov

EnzoVezzaro:
> elect a fixed price for more than 1 ocean

this has been fixed in #1366

kremalicious:
Created new compute asset, without having to edit compute settings, algos are showing up to be run on. But when selecting any of them, nothing happens and I can't start a job:

Screen Shot 2022-04-28 at 14 40 35

For this asset:
https://market-git-feature-v4-c2d-oceanprotocol.vercel.app/asset/did:op:12db90f33672efb275e94277ce633a017d8e1d911f885230fff803ea11ca4319

By default it seems we set _Allow any published algorithm_, which we absolutely need to prevent. But can be done in another PR as most likely this is some bug in publish flow:

Screen Shot 2022-04-28 at 14 41 55

bogdanfazakas:
> Created new compute asset, without having to edit compute settings, algos are showing up to be run on. But when selecting any of them, nothing happens and I can't start a job:
> 

Probably appeared after latest merges with v4 and conflicts , checking now

src/components/Asset/AssetActions/Compute/PriceOutput.tsx
```diff
@@ -76,14 +80,20 @@ export default function PriceOutput({
             <Row
               hasPreviousOrder={hasPreviousOrder}
               hasDatatoken={hasDatatoken}
-              price={Number.parseFloat(asset?.accessDetails?.price)}
+              price={
+                datasetOrderPrice ||
+                Number.parseFloat(asset?.accessDetails?.price)
```
kremalicious, line 85: another candidate for using `Decimal` instead

src/components/Asset/AssetActions/Compute/PriceOutput.tsx
```diff
@@ -76,14 +80,20 @@ export default function PriceOutput({
...
               timeout={assetTimeout}
               symbol={symbol}
             />
             <Row
               hasPreviousOrder={hasPreviousOrderSelectedComputeAsset}
               hasDatatoken={hasDatatokenSelectedComputeAsset}
-              price={Number.parseFloat(algorithmConsumeDetails?.price)}
+              price={
+                algoOrderPrice ||
+                Number.parseFloat(algorithmConsumeDetails?.price)
```
kremalicious, line 95: another candidate for using `Decimal` instead

src/@utils/aquarius.ts
bogdanfazakas: think filter was not working at that time, uncommented it and all fine 😃

bogdanfazakas:
> When editing an algorithm, we have the option to add "allowed algorithms" via the compute jobs settings. And you can add the same algorithm you are editing This seems wrong...

Yes that was wrong, we've now disabled edit compute settings for algorithms

jamiehewitt15:
> I see a weird behaviour when trying to edit the metadata of an asset: if I'm connected with another account and then switch to the account that published the asset, the "edit" button appears, but if I click, it says I don't have permission to edit. After a page refresh it starts working correctly again

Yeah, there is a problem here (https://github.com/oceanprotocol/market/issues/1376). I'll fix that in a separate PR

EnzoVezzaro:
if you open the modal and wait until the timer kicks in, the modal closes itself.

https://user-images.githubusercontent.com/13741335/166223879-a6b9abd1-075f-4ec7-b1d3-65171e3f9816.mp4

Also, still believe the button should be centered. Like so:

Screen Shot 2022-05-02 at 6 45 00 AM
(Like in Pool tab when there's the Liquidity button alone)
Screen Shot 2022-05-02 at 7 01 25 AM

EnzoVezzaro:
I'm not sure why we wait so much to start updating the prices when changing algorithm. Can we do something to make it start immediately after selecting an algorithm?  

https://user-images.githubusercontent.com/13741335/166225225-fa395281-cb45-404b-b834-915ede66c546.mp4

LucaMilanese90:
After a compute job finishes to execute I don't see the results anymore
Screenshot 2022-05-16 at 10 47 41

LucaMilanese90:
Clicking on the "buy compute job" button too quickly after selecting an algorithm results in an error message. 

https://user-images.githubusercontent.com/20192135/168563266-2a6fceb9-6914-49fc-b68d-fb8234168198.mov

bogdanfazakas:
> After a compute job finishes to execute I don't see the results anymore 

We are trying to find a solution for this issue

bogdanfazakas:
> Clicking on the "buy compute job" button too quickly after selecting an algorithm results in an error message.

> EDIT: Also happening when quickly switching between 2 algos

> And if you start from an algo you already bought and select a different one, the transaction you are prompted to sign is the one to start the job with the algo you already own

> Maybe an idea could be to disable the buy button while fetching for the prices.

Yep we already do that but in a later stage, so i think moving this right after algorithm selection will solve all this issues

jamiehewitt15:
There is an unhelpful message telling me I don't have enough Ocean when I have plenty. A more helpful message here would be "Select an algorithm to see the total cost"
Not enough Ocean

claudiaHash:
Something's wrong with results filtering. Checked it on v4 main and it doesn't happen there.

Screenshot from 2022-05-19 16-09-28

EnzoVezzaro:
I'm not completely sure about this. If we have bought a `compute job` previously, do we need to pay again to rerun this same job? 

Screen Shot 2022-05-19 at 9 24 06 AM

For example: here I bought this job (you can see the result at the bottom), but to run it again, I need to pay. Is this correct?

bogdanfazakas:
> I'm not completely sure about this. If we have bought a `compute job` previously, do we need to pay again to rerun this same job? And also, why would somebody would do that?
> 
> For example: here I bought this job (you can see the result at the bottom), but to run it again, I need to pay. Is this correct?

Nope, fixing it now, there is an issue with assets with forever timeout in the market

EnzoVezzaro:
My question about the rerun stays though 😂 probably i'm missing a piece, but I've been thinking about it, why would somebody rerun a job? I mean, what's the goal of it? Update the metadata or something else? I'm not seeing the use case, probably as I said I might be missing something.

EnzoVezzaro:
> > My question about the rerun stays though 😂 probably i'm missing a piece, but I've been thinking about it, why would somebody rerun a job? I mean, what's the goal of it? Update the metadata or something else? I'm not seeing the use case, probably as I said I might be missing the point.
> 
> Possible scenarios: update algorithm code, update entries in asset file etc..

Yeah, that's actually what is confusing me. Because if we go to the edit page on the algorithm, we can change its metadata but not the actual code. The same for the asset, I believe once published, you can't change anything on the asset (unless you change the content of your file from where is hosted).

AnaLoznianu:
> > > My question about the rerun stays though 😂 probably i'm missing a piece, but I've been thinking about it, why would somebody rerun a job? I mean, what's the goal of it? Update the metadata or something else? I'm not seeing the use case, probably as I said I might be missing the point.
> > 
> > 
> > Possible scenarios: update algorithm code, update entries in asset file etc..
> 
> Yeah, that's actually what is confusing me. Because if we go to the edit page on the algorithm, we can change its metadata but not the actual code. The same for the asset. I believe once published, you can't change anything on the asset (unless you change the content on your hosting service).

Yes. Once you see the results, maybe you want to update your algorithm. By changing directly the file that you hosted

EnzoVezzaro:
Not sure we can do anything about it, but when you edit a compute the UI doesn't update immediately. It happens on any changes done to the asset (at least to me).

https://user-images.githubusercontent.com/13741335/169808403-fdcceb9c-d492-4997-845e-369107f1050d.mp4

bogdanfazakas:
> Not sure we can do anything about it, but when you edit a compute the UI doesn't update immediately. It happens on any changes done to the asset (at least to me).
> 

Nope unfortunately we can't to anything at the moment, because every update (including compute settings) is a metadata update that generates an event that needs to be indexed first in the aqua, after that the changes will be available in the market. Thats why also other fields that we edit need some time before the changes are visible in the market.

EnzoVezzaro:
> > Not sure we can do anything about it, but when you edit a compute the UI doesn't update immediately. It happens on any changes done to the asset (at least to me).
> 
> Nope unfortunately we can't to anything at the moment, because every update (including compute settings) is a metadata update that generates an event that needs to be indexed first in the aquarius, after that the changes will be available in the market. Thats why also other fields that we edit need some time before the changes are visible in the market.

I guess this is related to the same problem.

https://user-images.githubusercontent.com/13741335/169812532-7f8cd654-645a-4ee6-aeea-2783353f3e47.mp4

Also, I've been following our [testing guide](https://docs.google.com/document/d/1Zr64V0FfnamLqlgrIw0knX3gUWyulkd7msZOPr-fwYM/edit) to test the C2D. In here we need to use these 2 files:
- Compute: https://raw.githubusercontent.com/oceanprotocol/c2d-examples/main/branin_and_gpr/branin.arff
- Algorithm: https://raw.githubusercontent.com/oceanprotocol/c2d-examples/main/branin_and_gpr/gpr.py

Those files seems to have some sort of error during execution. Can you take at look please.

ALGORITHM LOGS:
`Traceback (most recent call last):
  File "/data/transformations/algorithm", line 10, in <module>
    import arff
ModuleNotFoundError: No module named 'arff'
`

CONFIGURATION LOGS:
`
Using address 0x3db8d2AEf9A1d1584adD1aDDfb7728fe64d18c3F as consumer if we connect to remote providers.
========== Fetching input assets ============
========
Processing input 0 (did:op:5ba767ae52938743b69d2aa7ff9cea3e529117911a7ff5050813c9adb5052583)
Downloading...
DDO saved to /data/ddos/did:op:5ba767ae52938743b69d2aa7ff9cea3e529117911a7ff5050813c9adb5052583
Downloading to /data/inputs/5ba767ae52938743b69d2aa7ff9cea3e529117911a7ff5050813c9adb5052583/0
Downloaded OK
========== Done with inputs, moving to algo ============
Downloading...
DDO saved to /data/ddos/did:op:580e6ad2396a2c8d6062962064d6e2acdc4dccad31c45e60b8fa203a551c3c65
https://v4.provider.rinkeby.oceanprotocol.com//api/services/fileinfo
[
  {
    contentLength: '1122',
    contentType: 'text/plain',
    index: 0,
    valid: true
  }
]
Trying to download https://v4.provider.rinkeby.oceanprotocol.com/api/services/download?fileIndex=0&documentId=did:op:580e6ad2396a2c8d6062962064d6e2acdc4dccad31c45e60b8fa203a551c3c65&transferTxId=0xae65a36a25c802666087b4e42aeccdf27fcb1146c6cf6b813c01de98e32fef57&serviceId=49b41bd2a3cd02fe40f85b531c1545287756b8b918bd5d7c7b65e8bd351e90af&consumerAddress=0x3db8d2AEf9A1d1584adD1aDDfb7728fe64d18c3F&nonce=1653305945923&signature=0x8fec25966c2ceea015869a7b62038eb4eb40ac0932685fe9b861bb86ec27611a5745dbe306b8cd412da819e339f0f3c34ea5550ee177cfc25b2758ff8b6c51b11bto /data/transformations/algorithm
Downloading to /data/transformations/algorithm
Downloaded OK
============ Done , setting the status =============
Updated 1a08c019c86e41a5a0823c4e598a05ef with status 30
Finished!
`

PUBLISH LOGS: 
`null`

claudiaHash:
Compute functionality doesn't work as expected on Moonbase: 

Screenshot from 2022-05-23 15-56-05

bogdanfazakas:
> There is an unhelpful message telling me I don't have enough Ocean when I have plenty. A more helpful message here would be "Select an algorithm to see the total cost" 

@jamiehewitt15  tried various assets and not able to reproduce even tho I remembered seeing this before. Do you have some hints to reproduce the issue ?

bogdanfazakas:
> I guess this is related to the same problem.

Nope not really, here we need to trigger a refetch of the compute jobs if the section is open when starting a job. I will handle this asap, otherwise the jobs are fetched every time the sections is made visible.

> Also, I've been following our [testing guide](https://docs.google.com/document/d/1Zr64V0FfnamLqlgrIw0knX3gUWyulkd7msZOPr-fwYM/edit) to test the C2D. In here we need to use these 2 files:
> 
> * Compute: https://raw.githubusercontent.com/oceanprotocol/c2d-examples/main/branin_and_gpr/branin.arff
> * Algorithm: https://raw.githubusercontent.com/oceanprotocol/c2d-examples/main/branin_and_gpr/gpr.py
> 
> Those files seems to have issues during execution. Can you take at look please.
> 
> ALGORITHM LOGS: `Traceback (most recent call last): File "/data/transformations/algorithm", line 10, in <module> import arff ModuleNotFoundError: No module named 'arff' `
> 

This seems to be an issue with the algo itself or the container it's running within, probably you are using our default py container

jamiehewitt15:
> @jamiehewitt15 tried various assets and not able to reproduce even tho I remembered seeing this before. Do you have some hints to reproduce the issue ?

It doesn't seem to be showing that message anymore... I'll let you know if I see it again.


Meanwhile, I'm finding that the spinner is stuck loading whenever I select an algorithm. Don't see any errors in the console or network tab.
Spinner is stuck

EnzoVezzaro:
> > @jamiehewitt15 tried various assets and not able to reproduce even tho I remembered seeing this before. Do you have some hints to reproduce the issue ?
> 
> It doesn't seem to be showing that message anymore... I'll let you know if I see it again.
> 
> Meanwhile, I'm finding that the spinner is stuck loading whenever I select an algorithm. Don't see any errors in the console or network tab. Spinner is stuck

There's something wrong on the price too.
Screen Shot 2022-05-30 at 6 11 20 AM


Screen Shot 2022-05-30 at 6 11 45 AM

jamiehewitt15:
For example, if i go to [this asset](https://market-hnuxz2h4z-oceanprotocol.vercel.app/asset/did:op:5ba767ae52938743b69d2aa7ff9cea3e529117911a7ff5050813c9adb5052583) and select any algorithm I just see the loading spinner going round perpetually

EnzoVezzaro:
yeah, I have this issue on all compute assets.

Screen Shot 2022-05-30 at 10 59 54 AM

[example](https://market-hnuxz2h4z-oceanprotocol.vercel.app/asset/did:op:9587852075396e57385943ac61fe26e6f6ee18f36dc6c5aa899afa1e4772af02)

EnzoVezzaro:
I'm not it's related to your changes but I'm getting this error now. 🤔

Screen Shot 2022-06-01 at 10 12 22 AM

EnzoVezzaro:
Now, I'm getting stuck on the selection.

https://user-images.githubusercontent.com/13741335/171426346-8f9488dd-87a7-4613-b089-2a729e1e2974.mp4

EnzoVezzaro:
This latest changes has introduced a problem on the price calculation. The price on the second algo (on this [asset](https://market-m9i69kd9d-oceanprotocol.vercel.app/asset/did:op:ef6c63db09b0eb0bae369df6ecc5367e8bcb9fed713b183c6212804727c370a1)) doesn't get updated.

https://user-images.githubusercontent.com/13741335/171431217-145e67a0-4b50-4ccc-a0cd-703abb1446fe.mp4

bogdanfazakas:
> This latest changes has introduced a problem on the price calculation. The price on the second algo (on this [asset](https://market-m9i69kd9d-oceanprotocol.vercel.app/asset/did:op:ef6c63db09b0eb0bae369df6ecc5367e8bcb9fed713b183c6212804727c370a1)) doesn't get updated correctly.

Cannot reproduce on the lastest preview, but also pls make sure you don't have a valid order for the `Dataset Algorithm` ?

bogdanfazakas:
> I'm not sure it's related to your changes but I'm getting this error now. 🤔
> **Update**: Now, I'm getting stuck on the selection.

This is a mumbai asset, the provider on this deployment has some issue (big response time, 504 bad gateway responses on some endpoints) so not relevant for testing at this time

jamiehewitt15:
The app crashes when trying to select an algorithm while on a network that we don't support. A quick fix here would be some error handling and a toast message saying "Please change network"

Crashing on invalid network - 2022-06-02_09-18

EnzoVezzaro:
> The app crashes when trying to select an algorithm while on a network that we don't support. A quick fix here would be some error handling and a toast message saying "Please change network"
> 
> Crashing on invalid network - 2022-06-02_09-18

after #1482 this case should solved itself. On unsupported networks, the selection should stay disabled.

Screen Shot 2022-06-02 at 11 44 39 AM

EnzoVezzaro:
Found a new bug on this [asset](https://market-7vbxdqq7u-oceanprotocol.vercel.app/asset/did:op:98fe77267bc2628999b735cc0f34c51e34190b7434665d6b67842ae7273076a4)
(run it locally to see the error message)

https://user-images.githubusercontent.com/13741335/171670664-ea9698a0-be05-4853-9331-76de16feaa23.mp4

bogdanfazakas:
> Found a new bug on this [asset](https://market-7vbxdqq7u-oceanprotocol.vercel.app/asset/did:op:98fe77267bc2628999b735cc0f34c51e34190b7434665d6b67842ae7273076a4) (run it locally to see the error message)

okk interesting, will take a look 😃

src/components/Asset/AssetContent/index.tsx
```diff
@@ -45,6 +45,13 @@ export default function AssetContent({
     setIsOwner(isOwner)
   }, [accountId, owner, asset])
 
+  useEffect(() => {
```
kremalicious, line 48: why is this effect here doubled? There is an effect doing the same thing just above it

kremalicious:
@bogdanfazakas could you please verify that my deleting things commit didn't remove anything I shouldn't have please? Mostly removed unused imports, variables and such but in `Compute/index.tsx` component was a bunch of state which was never re-set in the component so seemed unused https://github.com/oceanprotocol/market/pull/1069/commits/db23ed6ada378105fcfe320032db273f8fb6ef15#diff-76db2f64e7d91a7c125ccc2151035c211746e95e5040b95ae101c5bc0de266d9

src/components/Asset/AssetActions/Compute/index.tsx
```diff
@@ -1,232 +1,286 @@
...
   useEffect(() => {
-    if (!algorithmConsumeDetails) return
+    if (!asset?.accessDetails || !accountId) return
 
-    setIsAlgoConsumablePrice(algorithmConsumeDetails.isPurchasable)
-  }, [algorithmConsumeDetails])
+    setIsConsumablePrice(asset?.accessDetails?.isPurchasable)
+    // setIsOwned(asset?.accessDetails?.isOwned)
+    setValidOrderTx(asset?.accessDetails?.validOrderTx)
+  }, [asset?.accessDetails])
```
kremalicious, line 250: this should get `accountId` as dependency. Unless we are sure if user connects wallet, or switches accounts, this effect actually changes

src/components/Asset/AssetActions/Compute/index.tsx
```diff
@@ -1,232 +1,286 @@
...
+      await initPriceAndFees()
+      setIsRequestingAlgoOrderPrice(false)
+    }
 
-  useEffect(() => {
-    if (!accountId) return
-    checkPreviousOrders(ddo)
-  }, [ddo, accountId])
+    initSelectedAlgo()
+  }, [selectedAlgorithmAsset])
```
kremalicious, line 270: same here with `accountId`, effect uses it but not in dependency array.

src/@utils/ddo.ts
```diff
@@ -2,14 +2,21 @@ import { Asset, DDO, Service } from '@oceanprotocol/lib'
 
 export function getServiceByName(
   ddo: Asset | DDO,
-  name: 'access' | 'compute'
+  name: 'access' | 'compute' | 'metadata'
```
kremalicious, line 5: There is no metadata service in v4 anymore, so this should stay as:

```suggestion
  name: 'access' | 'compute'
```

src/components/Asset/AssetActions/Compute/index.tsx
```diff
@@ -1,232 +1,286 @@
...
   dtBalance: string
   file: FileInfo
   fileIsLoading?: boolean
-  isConsumable?: boolean
   consumableFeedback?: string
 }): ReactElement {
-  const { appConfig } = useMarketMetadata()
-  const { accountId } = useWeb3()
+  const { accountId, web3 } = useWeb3()
+  const { getOpcFeeForToken } = useMarketMetadata()
```
kremalicious, line 66: why do we have 2 hooks doing exactly the same thing, but named differently, and both used in this file?

kremalicious: wrong line, I mean the `useAbortController()` and `useCancelToken()` hooks which are both in use in here. axios `CancelToken` is deprecated so only `useAbortController()` should be used in new code

src/components/Asset/AssetActions/Compute/index.tsx
```diff
@@ -1,232 +1,286 @@
...
   const [isJobStarting, setIsJobStarting] = useState(false)
   const [error, setError] = useState<string>()
 
   const [algorithmList, setAlgorithmList] = useState<AssetSelectionAsset[]>()
   const [ddoAlgorithmList, setDdoAlgorithmList] = useState<Asset[]>()
-  const [selectedAlgorithmAsset, setSelectedAlgorithmAsset] = useState<Asset>()
+  const [selectedAlgorithmAsset, setSelectedAlgorithmAsset] =
+    useState<AssetExtended>()
   const [hasAlgoAssetDatatoken, setHasAlgoAssetDatatoken] = useState<boolean>()
   const [isPublished, setIsPublished] = useState(false)
```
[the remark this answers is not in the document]
mihaisc: 'ButtonBuy' is becoming more and more a ticking time bomb 😆

src/components/Asset/AssetActions/Compute/index.tsx
kremalicious: sorry, didn't see the chain of events ends up with `queryMetadata()` requiring `CancelToken` so changing this there to `AbortSignal` has far reaching consequences in whole codebase which we should not do in here

kremalicious:
Now, my actual first try:

Screen Shot 2022-06-22 at 14 16 28

Just picked first assets I could find:

- dataset: `did:op:f6a9d9a7cb59bac6e1819e4321e3d4c8f19fe307f3dec12eb938ef6fd4284c0e`
- algo: `did:op:f965d5a61e5d73c5a9c91f67926bdd9e97a15f20f1e76467fca4f6e59b401ce1`

error payload:
Screen Shot 2022-06-22 at 14 25 39

bogdanfazakas:
> Now, my actual first try:
> 
> Just picked first assets I could find:
> 
> * dataset: `did:op:f6a9d9a7cb59bac6e1819e4321e3d4c8f19fe307f3dec12eb938ef6fd4284c0e`
> * algo: `did:op:f965d5a61e5d73c5a9c91f67926bdd9e97a15f20f1e76467fca4f6e59b401ce1`
> 
> error payload: Screen Shot 2022-06-22 at 14 25 39

yep we are trying to fix this now in provider 😃 , hopefully we'll get rid of this soon

src/components/Asset/AssetContent/index.tsx
bogdanfazakas: ohh think this duplication appeared after a v4 merge or rebase 😢 , but good point of moving this out going to add in the `AssetProvider` something similar with what we have for `isAssetNetwork`.

kremalicious:
Got the complexity of that `handleComputeOrder()` method down a bit by removing some nested if/else/elseif and just returning early if needed, hopefully correctly following the intended logic. Thanks for the great code comments there on top outlining the logic! 

We are left with only 1 level of if statements making this way easier to grasp. On top of that made sure that each transaction gets its own error messaging so we always know exactly where stuff fails in debug mode.

As a general refactor direction for next works on compute, we would greatly benefiting from moving most logic out of the components, abstract them into utilities or a dedicated hook, so that the actual `Compute` component can orchestrate better. Cause to make the UX in case of errors really great we would need to go the route of publish form, saving tx hashes in between in local state so that when users fail or decline 1 tx in between, we don't start all again. But not for this PR

src/@utils/order.ts
```diff
@@ -182,60 +231,66 @@ export async function handleComputeOrder(
...
-    if (!txApproveWei) {
-      toast.error('Failed to approve provider fees!')
-      return
+  try {
+    // Return early when valid order is found, and no provider fees
+    // are to be paid
+    if (
+      initializeData?.validOrder &&
+      (!initializeData.providerFee ||
+        initializeData?.providerFee?.providerFeeAmount === '0')
```
bogdanfazakas, line 242: think we do not need this check here `initializeData?.providerFee?.providerFeeAmount === '0')`, because for the algo for eg if the order is valid and you have to pay the provider fees those will show up in the dataset field and the providerFee?.providerFeeAmount for the algorithm will be '0'. Posting an example below for wich we should reuse orders on both datasets and algorithm:
`{
  "algorithm": {
    "datatoken": "0x37E7aeC8f9056ACf95C893a9B883c6b41e7651a8",
    "providerFee": {
      "providerData": "0x7b22656e7669726f6e6d656e74223a22636f6d707574652d76342d656e67696e652d6465762d656e7632227d",
      "providerFeeAddress": "0x00c6A0BC5cD0078d6Cd0b659E8061B404cfa5704",
      "providerFeeAmount": "0",
      "providerFeeToken": "0x5e8DCB2AfA23844bcc311B00Ad1A0C30025aADE9",
      "r": "0x502634912c556526f842d4092b7811ccea873ea0e509bfe8a723a4e0b6317aeb",
      "s": "0x72536c7586fe132d7f76377f897c85a8a245a4cf00fecf2aa26531b8ee7cba94",
      "v": 27,
      "validUntil": 1655980103
    },
    "validOrder": "0x2f17111b6d6e509c98afbf2740510b2789e4c20c3f1ea7b53a9c3146447504cc"
  },
  "datasets": [
    {
      "datatoken": "0x497E4bF570e8cB61D080a891D0C644D1d3001fb2",
      "providerFee": {
        "providerData": "0x7b22656e7669726f6e6d656e74223a22636f6d707574652d76342d656e67696e652d6465762d656e7632227d",
        "providerFeeAddress": "0x00c6A0BC5cD0078d6Cd0b659E8061B404cfa5704",
        "providerFeeAmount": "1983333333333333400",
        "providerFeeToken": "0x5e8DCB2AfA23844bcc311B00Ad1A0C30025aADE9",
        "r": "0x3c76a4970963ca1b37e254da23e10969a034094a1a884fccd7a91e0b70e01c47",
        "s": "0x54c274035b5531cd8615b679642e0c90bc44893226026579229861c86f47e843",
        "v": 27,
        "validUntil": 1655980103
      },
      "validOrder": "0x0248852caa92e34a9deae4607cd12ba030c39d9830eb8348d45e99fe646da7bd"
    }
  ]
}`

kremalicious: cool, thanks, reverted that change. But in general it's ok to return early with existing valid order? Cause wasn't sure under which circumstances `!initializeData.providerFee` is true.

bogdanfazakas: yes should be fine, but just realised a possible issue after the changes made on provider yesterday for that fix start compute job fix.

kremalicious:
assuming that provider 400 issue is fixed after latest deployments, all is working out in this PR with the final things before merging it and continuing in separate PRs:

- that buy button is disabled by default when no former order happened so far. Behaves as expected and is active for a second compute run
- fixes for new provider version

kremalicious (approved):
giphy-5

Not the full feature flows but this PR. Amazing job holding this all together for so long

kremalicious:
> On the [preview](https://market-hdukd9y2g-oceanprotocol.vercel.app/) the buy button stays disabled.

ah dang, you're right, I thought it was actually fixed cause I ended up with one of the accounts which had a previous order when verifying this. So guess [both points here](https://github.com/oceanprotocol/market/pull/1069#issuecomment-1164347346) are still to be solved before we can merge

EnzoVezzaro:
the issue on the button button is gone 🦾. The network is super slow, but there's nothing we can do about it. Once the tx finished, I got this error. 

Screen Shot 2022-06-23 at 10 35 34 AM

Issue #1075, opened 2022-02-10
Mobile publish flow (layout)


EnzoVezzaro:
Splitted from [Issue 1001](https://github.com/oceanprotocol/market/issues/1001)

Pragmatic design of new publish form in #928 was pretty much desktop-centric but we mostly made use of existing components and patterns so there shouldn't be any major visual design problems.

on mobile:

- check all publish screens and make sure everything is readable and usable, no viewport overflow anywhere
- adapt problematic component layouts as needed
- check all form navigation for user surprises (e.g. using Continue scrolls to top of viewport)

PR #1076, opened 2022-02-10
Mobile publish fixes

EnzoVezzaro:
Closes [Mobile publish flow (layout)#1075](https://github.com/oceanprotocol/market/issues/1075).

Changes proposed in this PR:

- check all publish screens on mobile
- adapt problematic component layouts as needed (not really needed, everything looks ok)
- check all form navigation for user surprises (e.g. using Continue scrolls to top of viewport) - didn't find any issue on the forms

### DIFF
--- a/src/components/Publish/Pricing/Fees.module.css
+++ b/src/components/Publish/Pricing/Fees.module.css
@@ -1,8 +1,8 @@
 .fees {
   display: grid;
   gap: var(--spacer);
-  grid-template-columns: repeat(auto-fit, minmax(8rem, 1fr));
-  padding: var(--spacer);
+  grid-template-columns: repeat(auto-fit, minmax(7rem, 1fr));
+  padding: var(--spacer) 0;
   text-align: center;
 }
 
@@ -19,8 +19,9 @@
   max-width: 5rem;
 }
 
-@media screen and (min-width: 40rem) {
-  .fees > div {
-    justify-self: center;
+@media screen and (min-width: 25rem) {
+  .fees {
+    grid-template-columns: repeat(auto-fit, minmax(8rem, 1fr));
+    padding: var(--spacer);
   }
 }
--- a/src/components/Publish/Pricing/Fees.module.css
+++ b/src/components/Publish/Pricing/Fees.module.css
@@ -8,7 +8,7 @@
 
 .fees > div {
   margin-bottom: 0;
-  justify-self: center;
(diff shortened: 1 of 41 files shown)


src/components/Publish/Pricing/Fees.module.css
```diff
@@ -18,3 +18,9 @@
 .fees input {
   max-width: 5rem;
 }
+
+@media screen and (max-width: 41rem) {
```
kremalicious, line 22: we do mobile first in our CSS codebase so only work with `min-width` and adapt elements as you scale up viewports. So the default for all elements is their mobile view and as you scale up the viewport they get adapted if needed

kremalicious:
would suggest looking at the NFT & Datatoken form fields which to me look like they should be stacked by default and then get left/right placement as viewport gets bigger.

src/components/Publish/Pricing/Fees.module.css
EnzoVezzaro: Sorry for this weird correction. For some reason my local github wasn't able to push to the correct branch. It got stuck on the re-named old branch.

EnzoVezzaro:
Correction for mobile merged to PR

EnzoVezzaro:
@kremalicious thank you, you're right. Fixed in the last commits

kremalicious:
when switching between _Metadata_ & _Access_, the Data NFT and Datatoken fields should look the same. Right now the NFT is stacked, but Datatoken is still left/right placement on mobile. Although Datatoken field kinda fits better, being consistent is more important here.

Also, would just left-align image/text on mobile, centering that NFT image looks unbalanced when input label above and content below is left-aligned.

kremalicious:
Found 2 more things we can improve:

Using really narrow viewport here but still seems 2 fee fields at bottom would still fit beside each other, we could also make them slightly smaller by default to gain a bit more space. On same screen the tokens in the pool box are misaligned, should both be left aligned when they are stacked like that, otherwise looks chaotic and unbalanced in current state.

Screen Shot 2022-02-15 at 17 11 19

On submission screen, number & title should always be on one line, and the transaction label should be then in new line, left aligned with the edge of the text beneath it

Screen Shot 2022-02-15 at 17 12 17

EnzoVezzaro:
pricing and submit tabs fixed for small viewports 🤙🏽

kremalicious (approved):

PR #1077, opened 2022-02-10
Restore highest liquidity tab

### LINKED ISSUE #1074: Restore highest liquidity tab
Adapt the subgraph query. Not a lot of changes needed .
- get highest liquidity pools from all selected chains
- get assets based on pool datatoken (we can do this since we have 1 service - 1 datatoken - 1 asset)
- ignore price fetching , it will be handled in [restore order pr ](https://github.com/oceanprotocol/market/blob/e11fe49bef44c285f12a844fb6e4a728e0473503/src/%40utils/accessDetailsAndPricing.ts#L241)

claudiaHash:
Fixes #1074  .

Changes proposed in this PR:

- restore highest liquidity tab

### DIFF
--- a/src/@utils/subgraph.ts
+++ b/src/@utils/subgraph.ts
@@ -59,9 +59,9 @@ const HighestLiquidityAssets = gql`
   query HighestLiquidityAssets {
     pools(
       where: { datatokenLiquidity_gte: 1 }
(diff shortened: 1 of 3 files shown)


kremalicious (approved):
awesome, working well although we will only fully know once we have multiple assets from multiple networks

Issue #1060, opened 2022-02-04
Free pricing not validated properly


mihaisc:
![image](https://user-images.githubusercontent.com/26000280/152553021-cb37051b-a4a5-4c2e-9d50-f3613b8b92c9.png)
If i change the free price for example then switch to free again and move to the next step i get 
![image](https://user-images.githubusercontent.com/26000280/152553166-b4daf391-4b29-49a9-a910-a79a7c5148ae.png)

kremalicious:
more issues with free pricing screen:

- "Continue" button stays disabled, you're not able to continue in the "FREE" tier
- either we add some user intention like a checkbox ("I want this asset to be free", or "I understand network fees are still to be paid"), or automatically make pricing valid when user hits the free screen

EnzoVezzaro:
if it's not a problem, I can take this.

EnzoVezzaro:
issue on the layout is done, now working on the issue 2 (pricing don't show up)

mihaisc:
No, that is not an issue here , is already handed in the restore order pr.

PR #1086, opened 2022-02-11
added confirm checkbox and connect logic to publish

EnzoVezzaro:
Fixes [Free pricing not validated properly #1060](https://github.com/oceanprotocol/market/issues/1060).

Changes proposed in this PR:

- Added "confirm" checkbox to validate "free" option
- Fixed validation error on "fixed" option

### DIFF
--- a/src/components/Publish/Preview/index.tsx
+++ b/src/components/Publish/Preview/index.tsx
@@ -8,7 +8,6 @@ import { Asset } from '@oceanprotocol/lib'
 
 export default function Preview(): ReactElement {
   const [asset, setAsset] = useState<Asset>()
-  const [accessDetails, setAccessDetails] = useState<AccessDetails>()
   const { values } = useFormikContext<FormPublishData>()
 
   useEffect(() => {
@@ -34,8 +33,10 @@ export default function Preview(): ReactElement {
         owned: false,
         validOrderTx: ''
       }
-      setAccessDetails(accessDetails)
+      // TODO: add Property 'accessDetails' to type 'DDO' on ocean.js. see <Asset>
+      asset.accessDetails = accessDetails
     }
+
     makeDdo()
   }, [values])
 
--- a/src/components/Publish/Pricing/index.tsx
+++ b/src/components/Publish/Pricing/index.tsx
@@ -23,7 +23,10 @@ export default function PricingFields(): ReactElement {
   function handleTabChange(tabName: string) {
     const type = tabName.toLowerCase()
     setFieldValue('pricing.type', type)
+    setFieldValue('pricing.price', 0)
+    setFieldValue('pricing.freeAgreement', false)
     type === 'dynamic' && setFieldValue('pricing.amountDataToken', 1000)
+    type === 'fixed' && setFieldValue('pricing.amountDataToken', 1000)
   }
 
   // Always update everything when price value changes
--- a/content/price.json
+++ b/content/price.json
@@ -28,7 +28,18 @@
     },
     "free": {
       "title": "Free",
-      "info": "Set your data set as free. The datatoken for this data set will be given for free via creating a faucet."
+      "info": "Set your data set as free. The datatoken for this data set will be given for free via creating a faucet.",
+      "fields": [
+        {
+          "name": "freeAgreement",
+          "label": "Confirm",
+          "type": "checkbox",
+          "options": [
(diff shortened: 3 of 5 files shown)


src/components/Publish/Preview/index.tsx
```diff
@@ -34,8 +33,10 @@ export default function Preview(): ReactElement {
         owned: false,
         validOrderTx: ''
       }
-      setAccessDetails(accessDetails)
+      // TODO: add Property 'accessDetails' to type 'DDO' on ocean.js. see <Asset>
```
mihaisc, line 36: no, `accessDetails` is only on market. it is part of `AssetExtended`

EnzoVezzaro: ok, restored 👍🏽

src/components/Publish/Pricing/index.tsx
```diff
@@ -23,7 +23,10 @@ export default function PricingFields(): ReactElement {
   function handleTabChange(tabName: string) {
     const type = tabName.toLowerCase()
     setFieldValue('pricing.type', type)
+    setFieldValue('pricing.price', 0)
+    setFieldValue('pricing.freeAgreement', false)
     type === 'dynamic' && setFieldValue('pricing.amountDataToken', 1000)
+    type === 'fixed' && setFieldValue('pricing.amountDataToken', 1000)
```
kremalicious, line 29: not entirely sure whey we do this but might be a good reason. In any case, it's unneeded duplication if we just turn around the logic with `type !== 'free' && setFieldValue(...)`

kremalicious:
And this still feels like we just tack on things onto that screen:

Screen Shot 2022-02-14 at 11 25 12

Put that checkbox where currently `Free` is shown in the grey box instead of as another field

src/components/Publish/Pricing/index.tsx
EnzoVezzaro: this is need it to avoid having the `FIXED` tab disabled like this: 

Screenshot 2022-02-14 at 14 56 19

maybe i misunderstood how this is being handle. But in anyway, your logic is more correct: `type !== 'free' && setFieldValue(...)` should work as well.

src/components/Publish/Pricing/Price.tsx
```diff
@@ -20,7 +23,13 @@ export default function Price({
   return (
     <div className={styles.price}>
       {values.pricing.type === 'free' ? (
-        <h4 className={styles.free}>Free</h4>
+        <h4 className={styles.free}>
```
kremalicious, line 26: are we able to add the `className` onto the `Field`? Cause we output quite a bit of markup in the DOM in the end and semantically it shouldn't be wrapped in an `h4`

EnzoVezzaro: yeah, that's right. I made some changes: change the h4 to div since it's worthless as such, and also swith the Field component with an Input one

src/components/Publish/Pricing/Price.tsx
```diff
@@ -1,26 +1,40 @@
...
 
   return (
     <div className={styles.price}>
       {values.pricing.type === 'free' ? (
-        <h4 className={styles.free}>Free</h4>
+        <div className={styles.free}>
+          <Input
+            {...getFieldContent('freeAgreement', content.fields)}
+            name="pricing.freeAgreement"
+            onChange={() => {
```
kremalicious, line 30: not fully sure we need that, we have bad experiences in pst with trying to overwrite Formik's default behavior. Are you sure that using `name="pricing.freeAgreement"` isn't enough? Usually this connects any input element with Formik when within a certain Formik provider which is the case here. So let's double check fi we can remove that `onChange`

EnzoVezzaro: Yeah. You're right. the `onChange` method is not the right approach, the name property is not enough though so I changed the approach and use the `useField` hook

kremalicious: alright, got it, think these are the rules:

- when using Formik's `Field` component only `name` is required to connect with form state
- when using any other input component directly, has to be connected with spreading `field`:
   ```
   const [field] = useField('pricing.freeAgreement')
  
   <Input {...field} />
   ```

In any case couldn't find DOM differences anymore between the 2 possibilities so went back to using `Field` again as this is how we do most other input fields in publish form.

kremalicious (approved):

Issue #1025, opened 2022-01-25
Show proper errors in submit step of form


mihaisc:
![image](https://user-images.githubusercontent.com/26000280/151020984-1386b045-e41c-4949-ad90-bbcec5974133.png)

We need a proper error message, not what we currently have.

kremalicious:
now that we have errors shown suddenly they are too much 😀 I think we have to do some string manipulation here as that red block shown is the actual `error.message`

mihaisc:
I'm not saying that it's a problem with the font/color or whatever but with the fact that we are just dumping the error. It means nothing for the common user, and we need some text wrapping

EnzoVezzaro:
Guys, any of you have any idea how to replicate this issue? I can't figure out if this has been fixed.

mihaisc:
It was not fixed, you can just go in the publish form and change some params to mess it up, then you will get an error

EnzoVezzaro:
found it. If shows up when you reject the first transaction, but accept the second one.

EnzoVezzaro:
now, it gets weirder. I don't have 2 transactions anymore. On the "Create Tokens & Pricing" step at the end of publish, it used to make 2 transactions, now makes only one transaction. is that ok @mihaisc ?

PR #1105, opened 2022-02-15
show proper errors on submit publish

EnzoVezzaro:
Fixes [Show proper errors in submit step of form #1025](https://github.com/oceanprotocol/market/issues/1025).

Changes proposed in this PR:

- in the case the message is longer than one row, then apply ellipses
- if the message is too long, return a generic label
- show error message when user rejects the approval transaction


kremalicious (approved):
hard to test, but code change is nicely isolated towards error messages only and should work as intended so let's run with it

PR #1120, opened 2022-02-17
Collect tokens by publisher

### LINKED ISSUE #940: Collect tokens by publisher for fixed rate exchange
We need a section for the publisher to collect tokens earned by selling assets using fixed rate exchange. 
The component will:
- display how many base tokens the owner can collect . Should be fetched from the subgraph, basically all the base token supply in the fixed rate exchange is for the owner :[example](https://subgraphv4.rinkeby.oceanprotocol.com/subgraphs/name/oceanprotocol/ocean-subgraph/graphql?query=%7B%0A%20%20fixedRateExchanges%7B%0A%20%20%20%20id%0A%20%20%20%20baseTokenBalance%0A%20%20%7D%0A%7D%0A)
- button to collect tokens https://github.com/oceanprotocol/ocean.js/blob/9b71c853f7cd1f596b5fbd451f63e9cdb5af3b72/src/pools/fixedRate/FixedRateExchange.ts#L801

claudiaHash:
Fixes #940 .

Changes proposed in this PR:

- collect tokens by  publisher for fixed rate asset


claudiaHash:
Not sure if I placed this button in the right place and how should we style it. What do you guys think? @kremalicious @mihaisc

mihaisc:
look ok, but it seems to not work and we need some loading states and success . The tx was ok, but i got no tokens (on mumbai) didn't test on rinkeby. It says i should be able to collect 4
image

mihaisc:
Subgraph fixed and deployed

claudiaHash:
> Subgraph fixed and deployed

Verified again the transactions and everything looks fine now. Not sure if we should display the collect button if there are no tokens to be collected. What do you think?

mihaisc:
I'd say disable it , this way the user also sees that there are 0 tokens to be collected

src/components/Asset/AssetActions/Download.tsx
```diff
@@ -80,6 +103,30 @@ export default function Download({
...
+      let baseTokenSum = 0
+      const variables = {
+        user: accountId.toLowerCase()
+      }
+      const result: OperationResult<FixedRateExchanges> = await fetchData(
+        FixedRateExchangesQuery,
+        variables,
+        queryContext
+      )
+      for (let i = 0; i < result.data.fixedRateExchanges.length; i++) {
```
mihaisc, line 120: this is not right, you are collecting tokens from just one fixed rate exchange, so why are you adding from all fixed rate exchanges?

claudiaHash: my bad, I've made a confusion. corrected it by querying by user and exchangeId

mihaisc:
image
should be collect OCEAN

mihaisc (approved):

PR #1123, opened 2022-02-17
Restore global stats

### LINKED ISSUE #1096: Restore global stats
Restore existing global market stats in the footer, and look into what other useful data we could show/replace based on new subgraph.

Can only be reliably tested once we have test assets on multiple test networks.

**Blocked by https://github.com/oceanprotocol/ocean-subgraph/issues/316**

kremalicious:
closes #1096

* split up into multiple files
* split up data fetching and manipulation into multiple effects
* adapt for new subgraph

Locally switched to querying all test networks as we have value locked there to check the calculations:

Screen Shot 2022-02-18 at 15 28 33

But as we are not deployed on any mainnet yet, this is what this PR will look like in the end, until we have assets in those networks:

Screen Shot 2022-02-18 at 15 29 50


kremalicious:
Current state with querying all supported testnets so we see something:

Screen Shot 2022-02-18 at 13 08 00

Bunch of assumptions embedded in the code:

- like with pool stats, doubling the returned OCEAN liquidity to calculate total OCEAN and TVL
- if users can only add/remove OCEAN seems showing datatokens is useless here too
- effectively total OCEAN & TVL use same underlying value, TVL just converts it to fiat

Open to suggestions if current approach makes sense, or for more useful stat values we could add

mihaisc:
agree on the TVL = ( 2 x ocean liquidity ) x (fiat rate)
why do you double OCEAN ? Shouldn't that be just OCEAN liquidity? So if the pool has 100 ocean liquidity 
TVL = ~ 100 $
OCEAN = 100

kremalicious:
wasn't sure if being technically correct, or conceptional correct makes more sense so used the conceptional path we do on the single pool stats. To stay in your example with just outputting returned OCEAN liquidity:

pool has 100 ocean liquidity, fiat price is 0.50 EUR
TVL = ~ 200 $
OCEAN = 100

which I guess would be fine if I add back the "plus datatokens for each pool" line

mihaisc (approved):
Looks cool, maybe in the future we have a separate page to include the other data from the global statistics (and add more to those in the subgraph)

PR #1125, opened 2022-02-17
Feat: Display NFT in asset details

### LINKED ISSUE #995: Display NFT metadata somewhere in asset details
Output at least the NFT image assuming our current approach in #944, and the NFT address with link to explorer somewhere.

### LINKED ISSUE #1131: Publisher line: deal with NFT owner/publisher mismatch
By default show "Owned by ...", but display both in case they are different. So always check NFT owner against actual publisher from the initial metadata created(?) event, and if different output both, "Owned by 0xxx, published by 0xxx"

Taken from #1010

moritzkirstein:
Fixes #995 
Fixes #1131

Changes proposed in this PR:

- Breakdown of `MetaMain` asset details header into multiple components
- Display of nft image & addition of nft tooltip
- Adjustments of publish `_utils` to show nft data in publish preview

grafik

grafik

Something worth discussing could be how to display assets that were not generated from within the market. Currently the `alt` text gets displayed, but maybe a visual fallback would be a preferred option?

grafik


mihaisc:
We should handle the case where the data is not how we want it :)  [example](https://market-git-feat-display-nft-asset-details-oceanprotocol.vercel.app/asset/did:op:ae281edc7884d5a36ceec9f1ed9a72b5cad6eb40c73dbd6d0767fa915076c27c)
image

moritzkirstein:
Thanks for the feedback. What would be a preferred way to handle NFTs with different tokenURIs than those generated from the market? Should we use a placeholder graphic there?

Regarding OpenSea: currently the Ocean NFTs still are not displayed correctly on OpenSea. I commented this in the code, but I can also comment out the code itself to not display OpenSea links for now until it is fixed. I believe this means getting in contact with the OpenSea team themselves. I believe we spent quite some time on trying to figure out what is going wrong and could not find any major differences that could cause the issue here.

mihaisc:
> Thanks for the feedback. What would be a preferred way to handle NFTs with different tokenURIs than those generated from the market? Should we use a placeholder graphic there?

Good idea with the placeholder

> Regarding OpenSea: currently the Ocean NFTs still are not displayed correctly on OpenSea. I commented this in the code, but I can also comment out the code itself to not display OpenSea links for now until it is fixed. I believe this means getting in contact with the OpenSea team themselves. I believe we spent quite some time on trying to figure out what is going wrong and could not find any major differences that could cause the issue here.

Didn't see that, it's ok then, leave it as it is for now.

src/@utils/nft.ts
```diff
@@ -54,6 +55,8 @@ export function generateNftMetadata(): NftMetadata {
   return newNft
 }
 
+const tokenUriPrefix = 'data:application/json;base64,'
+
 export function generateNftCreateData(nftMetadata: NftMetadata): any {
   // TODO: figure out if Buffer.from method is working in browser in final build
```
mihaisc, line 61: i think we can remove this todo

src/components/Asset/AssetContent/MetaMain/MetaInfo.tsx
```diff
@@ -0,0 +1,34 @@
+import { Asset } from '@oceanprotocol/lib'
+import AssetType from '@shared/AssetType'
+import Time from '@shared/atoms/Time'
+import { getServiceByName } from '@utils/ddo'
+import React, { ReactElement } from 'react'
+import styles from './MetaInfo.module.css'
+
+export default function MetaInfo({ ddo }: { ddo: Asset }): ReactElement {
```
mihaisc, line 8: should be `asset` not `ddo`

src/components/Asset/AssetContent/MetaMain/NftTooltip.tsx
```diff
@@ -0,0 +1,78 @@
...
+  address: string
+  chainId: number
+  isBlockscoutExplorer: boolean
+}): ReactElement {
+  // Currently Ocean NFTs are not displayed correctly on OpenSea
+  // Code prepared to easily integrate this feature once this is fixed
+  //
+  // Supported OpeanSea networks:
+  // https://support.opensea.io/hc/en-us/articles/4404027708051-Which-blockchains-does-OpenSea-support-
+  const openseaNetworks = [1, 137, 8217]
```
mihaisc, line 25: we should display this link just for the supported networks (eth, polygon , rinkeby)

src/components/Asset/AssetContent/MetaMain/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+import MetaAsset from './MetaAsset'
+import MetaInfo from './MetaInfo'
+import Tooltip from '@shared/atoms/Tooltip'
+import NftTooltip from './NftTooltip'
+import Logo from '@shared/atoms/Logo'
+
+export default function MetaMain({ ddo }: { ddo: Asset }): ReactElement {
+  const nftMetadata = decodeTokenURI(ddo?.nft?.tokenURI)
+
+  const blockscoutNetworks = [1287, 2021000, 2021001, 44787, 246, 1285]
```
mihaisc, line 14: why do we have this check?

src/components/Asset/AssetContent/MetaMain/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const isBlockscoutExplorer = blockscoutNetworks.includes(ddo?.chainId)
```
mihaisc, line 15: why do we have this check?

moritzkirstein: I copied this implementation from the previous `MetaMain` code. It basically helps to decide whether to use `{explorer}/tokens` or `{explorer}/token` to link to the nft / datatoken respectively.
See line 52 of src/components/Asset/AssetContent/MetaMain/NftTooltip.tsx for example
```ts
isBlockscoutExplorer ? `tokens/${address}` : `token/${address}`
```

mihaisc: i see, thanks

mihaisc:
for :  https://market-git-feat-display-nft-asset-details-oceanprotocol.vercel.app/asset/did:op:bdde998387fc886c3fb7220731b5165298f49bd4cc7edd4c85e1016820720905
image

kremalicious (changes requested):
Serious things:

seems to mostly work locally but on the preview deployments not able to open any asset because of this error in console:

Screen Shot 2022-03-16 at 13 02 33

And assets can't be opened at all by all non-web3-capable browsers, both in remote deployment and locally:

Screen Shot 2022-03-16 at 13 06 48

Small tweaks

- don't output long NFT address, concatenate it like we do with account addresses

Issue #1004, opened 2022-01-17
v3 plan


kremalicious:
Most likely includes lots of devops work but handling this here on `market`. All still WIP.

Main Goals:

- keep the currently published assets and their liquidity alive as much as possible
- keep current `market.oceanprotocol.com` environment until v4 launch day
- provide a user flow for publishers/liquidity providers to migrate from v3 to v4

## Endpoint Changes

Unless stated otherwise, `→` means the underlying instance is available under the new given URL, while the initial URLs become the new default v4 ones with new v4 instances under the hood. Undecided on the actual URL scheme, we could also prefix everything with `v3`, like `v3.market.ocean.oceanprotocol.com`?

#### public testing phase

  - [x] setup `v4` → `v4.market.oceanprotocol.com`
- [x] in UI, both the v3 market and v4 market get a dropdown for the version number beside the app title for easy switching between above domains
- [x] split up deployments by creating new projects `market-v3` in Vercel, make it deploy the `v3` branch and PRs against it
- [x] create `v3.market.oceanprotocol.com`, deploy `v3` branch to it
- [x] switch default branch to `v4`
- [x] ~~ocean.js v3 track `v0.xx.x` needs a release with the below mentioned endpoints and v3 contract addresses for everything in `ConfigHelper`~~
- [x] sync decided on networks below with the network selector list in `app.config.js` in `v3` market branch
- [x] implement v4 migration into `v3` branch, handled in #1023
- [x] remove all publishing functionality from v3 code base, handled in  #1451
- [x] release all the v3 codebase changes to existing live market, handled in #1456

More tasks regarding v3 codebase are in the milestone https://github.com/oceanprotocol/market/milestone/10

#### live production phase

  - [x] setup `v3` → `v3.market.oceanprotocol.com`
  - [x] `v4` → `market.oceanprotocol.com`
- [x] implement redirects for all v3 assets in v4 code base, handled in #1446
- [x] rename `main` → `v3`, rename `v4`→ `main`

### Purgatory

- do we need to do something with https://github.com/oceanprotocol/market-purgatory and `market-purgatory.oceanprotocol.com` endpoint?

### Backend

We will keep all existing URLs as the v3 URLs and introduce a new versioning schema for all newly deployed v4 components.

#### v3

[...]

PR #1129, opened 2022-02-21
Restore profile header and downloads tab

### LINKED ISSUE #1039: Restore profile header
It refers to this 
image

claudiaHash:
Fixes #1039 and #1044 .

Changes proposed in this PR:

- restore `Sales` number in profile header stats
- restore `Downloads` tab

### DIFF
--- a/app.config.js
+++ b/app.config.js
@@ -7,7 +7,7 @@ module.exports = {
   // return appConfig.metadataCacheUri
   metadataCacheUri:
     process.env.NEXT_PUBLIC_METADATACACHE_URI ||
-    'https://v4.aquarius.oceanprotocol.com',
+    'https://aquariusv4.oceanprotocol.com',
 
   // List of chainIds which metadata cache queries will return by default.
   // This preselects the Chains user preferences.
--- a/app.config.js
+++ b/app.config.js
@@ -7,7 +7,7 @@ module.exports = {
   // return appConfig.metadataCacheUri
   metadataCacheUri:
     process.env.NEXT_PUBLIC_METADATACACHE_URI ||
-    'https://aquariusv4.oceanprotocol.com',
+    'https://v4.aquarius.oceanprotocol.com',
 
   // List of chainIds which metadata cache queries will return by default.
   // This preselects the Chains user preferences.
--- a/src/@context/Profile.tsx
+++ b/src/@context/Profile.tsx
@@ -177,8 +177,6 @@ function ProfileProvider({
     if (!accountId || !isEthAddress) return
 
     const cancelTokenSource = axios.CancelToken.source()
-    console.log('ACC ID: ', accountId)
-    console.log('CHAINIDS: ', chainIds)
 
     async function getAllPublished() {
       try {
@@ -187,7 +185,6 @@ function ProfileProvider({
           chainIds,
           cancelTokenSource.token
         )
-        console.log('RESULTS: ', result)
         setAssets(result.results)
         setAssetsTotal(result.results.length)
         LoggerInstance.log(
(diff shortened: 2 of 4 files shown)


app.config.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
   // return appConfig.metadataCacheUri
   metadataCacheUri:
     process.env.NEXT_PUBLIC_METADATACACHE_URI ||
-    'https://v4.aquarius.oceanprotocol.com',
+    'https://aquariusv4.oceanprotocol.com',
```
kremalicious, line 10: why are we changing this URL again? `v4.aquarius.oceanprotocol.com` is the final URL and it's up, see #1004

src/@context/Profile.tsx
```diff
@@ -186,7 +186,7 @@ function ProfileProvider({
           cancelTokenSource.token
         )
         setAssets(result.results)
-        setAssetsTotal(result.totalResults)
+        setAssetsTotal(result.results.length)
         LoggerInstance.log(
           `[profile] Fetched ${result.totalResults} assets.`,
```
kremalicious, line 191: this needs to be changed too, based on above change, `result.totalResults` won't work here

src/@context/Profile.tsx
```diff
@@ -224,12 +224,8 @@ function ProfileProvider({
       const tokenOrders = await getUserTokenOrders(accountId, chainIds)
 
       for (let i = 0; i < tokenOrders?.length; i++) {
-        const did = web3.utils
-          .toChecksumAddress(tokenOrders[i].datatoken.address)
-          .replace('0x', 'did:op:')
-        didList.push(did)
+        didList.push(tokenOrders[i].datatoken.address)
```
kremalicious, line 227: as far as I see, this is not a DID list anymore but a list of Datatoken addresses, so the variable should be named differently

src/@utils/subgraph.ts
```diff
@@ -157,12 +164,10 @@ const UserTokenOrders = gql`
 
 // TODO: counting orders might be enough here to get sales for a user
```
kremalicious, line 165: if it's working, remove the TODO

kremalicious:
Checked out some profiles and all numbers check out, great!

One final small thing: when there are no sales the number goes from a displayed `0` to just empty space. Should just stay as `0` in that case, you can see that on your own profile:

Screen Shot 2022-02-21 at 14 27 29

claudiaHash:
I'll investigate this a little, because in my profile I can see this: 

Screenshot from 2022-02-21 16-43-34

claudiaHash:
I'll investigate this a little because I can see the sales number on my profile, which seems to be a correct one [according to the subgraph ](https://v4.subgraph.rinkeby.oceanprotocol.com/subgraphs/name/oceanprotocol/ocean-subgraph/graphql?query=%7B%0A%20%20user(subgraphError%3A%20deny%20id%3A%220x134b4edd7de8ea003e721e7670800ba97d9af1fd%22)%20%7B%0A%20%20%20%20%20%20id%0A%20%20%20%20%20%20totalSales%0A%20%20%20%20%7D%0A%7D%0A)
Screenshot from 2022-02-21 16-43-34

kremalicious:
very interesting!

So in Firefox this is my setup, without wallet connected or with wallet connected doesn't make difference for me. But activate Polygon Mumbai and you will see the _Sales_ number disappear:

Screen Shot 2022-02-21 at 14 53 38

So I think 2 things:

- double check why some Polygon Mumbai call does not succeed
- make sure calculations and numbers stay even if one subgraph errors out

claudiaHash:
Think i found the problem. You're right, it came out when I used Polygon Mumbai. Just checked with Ropsten after I made the fix and seems to be fine now for my profile and other users's profiles

Issue #1093, opened 2022-02-14
Show gas cost estimation for publish


kremalicious:
For each method we call during publish we have a corresponding method to estimate its gas usage in ocean.js. So make use of those in the submission screen and display estimated cost (in ETH + user's currency, or just user currency) for each step beside the transaction count labels

EnzoVezzaro:
are these the methods i need to call, right?

file: node_modules/@oceanprotocol/lib/dist/src/factories/NFTFactory.d.ts 
methods: 
- Step 1 (Create Tokens & Pricing)
- - estGasCreateNFT
- Step 2 (Construct & Encrypt DDO)
- - none
- Step 3 (Publish DDO)
- - estGasCreateNftWithErc20

EnzoVezzaro:
I've been trying to debug these params but for what i can see all params are being populated, I'm not sure if the information is actually right thought. Here's the full payload for each call.

Free Pricing - method: estGasCreateNftErc20WithDispenser

`accountId: 0xC50ce98B13Aa9782cAc4aF6809007B2b035E599e,
nftCreateData : {
    "name": "Ocean Asset NFT",
    "symbol": "OCEAN-NFT",
    "templateIndex": 1,
[...]

PR #1132, opened 2022-02-21
add gas cost estimation on publish 

EnzoVezzaro:
Fixes [Show gas cost estimation for publish](https://github.com/oceanprotocol/market/issues/1093).

Changes proposed in this PR:

- add gas estimation costs on publish


kremalicious:
already a conflict here so best to merge or rebase & force push against current `v4` branch

kremalicious:
however you like, if you use GitHub UI then it's also just a normal merge in the end so all good

kremalicious:
Let's save some space, and put "Gas fee estimation" into a standard HTML `title` on the price, and make transaction count and their cost visually the same element:

Screen Shot 2022-02-21 at 14 39 44

We could go fancy and extend this later on to have a more sophisticated tooltip with more info, but for now we could just add the gas price in Gwen to that HTML `title` like "Gas fee estimation with 40 gwei"

kremalicious:
this reminds me, we probably want some interval fetching here, when submission screen is mounted can be valid to refetch every 5 sec. or so?

EnzoVezzaro:
> this reminds me, we probably want some interval fetching here, when submission screen is mounted can be valid to refetch every 5 sec. or so? But then stop the interval when user starts submission

Actually, I'm getting the same amount of gwei on every interval call. Do you know if this call gets cached?

EnzoVezzaro:
> Let's save some space, and put "Gas fee estimation" into a standard HTML `title` on the price, and make transaction count and their cost visually the same element:
> 
> Screen Shot 2022-02-21 at 14 39 44
> 
> We could go fancy and extend this later on to have a more sophisticated tooltip with more info, but for now we could just add the gas price in Gwei we use for calculation to that HTML `title` like "Gas fee estimation with 40 gwei"

Do we keep the "Gas fee estimation with 40 gwei" message? It might be a problem as this is only related to the etherium network. 

Screenshot 2022-02-22 at 15 11 42

If we change it, what can we use as global unit for all network?

EnzoVezzaro:
@mihaisc do you think is acceptable that we update the gasFees on the fly?

I added the dummy data to get the `estGasSetMetadata` alongside the other estimates:  [see here](https://github.com/oceanprotocol/market/blob/8530f7ead5e55334eb5cb63ce7fd2288e91eb104/src/components/Publish/_utils.ts#L544)

The difference between the "dummy" estimate and the "real" estimate is of about 0.05

We can keep the "dummy" estimate (with the 0.05 difference on price) or show the "dummy" price at first and then update it after the first transaction is done. What would be the best approach here?

mihaisc:
not working without a wallet (there is a dummy web3 `getDummyWeb3` ) . If i connect a wallet in the preview page it just crashes 
image

There is no need to update the estimation. Calculate them when reaching the step. After the publish flow starts there is no point in estimating again, the user has the actual gas cost in the tx popup

EnzoVezzaro:
> not working without a wallet (there is a dummy web3 `getDummyWeb3` ) . If i connect a wallet in the preview page it just crashes image
> 
> There is no need to update the estimation. Calculate them when reaching the step. After the publish flow starts there is no point in estimating again, the user has the actual gas cost in the tx popup
> 
> Looks good.
> 
> the price seems off image 2951726.592 gwei should be around 0.002955067.917 matic (1.46$ ) , way off 7.79 $

Great catch. I've added the dummyWeb3. Now, it's not breaking. 

I'm looking into the pricing issue now 👍🏽

EnzoVezzaro:
the price seems to be thrown off by the conversion to ocean [`ethToOceanConversionRate`](https://github.com/oceanprotocol/market/blob/feature/issue-1093-gas-cost-estimation-publish-/src/@context/Prices.tsx#L64). When we calculate the gas fees, we convert those [fees to ocean](https://github.com/oceanprotocol/market/blob/feature/issue-1093-gas-cost-estimation-publish-/src/components/Publish/_utils.ts#L498) to then pass the value to the [Conversion](https://github.com/oceanprotocol/market/blob/feature/issue-1093-gas-cost-estimation-publish-/src/components/@shared/Price/Conversion.tsx#L41) component which return the value in EUR. 

This flow doesn't seems to care if it's etherium, polygon or any other network. I'm still investigating, but I guess this issue is present also on the [Data NFT preview](https://github.com/oceanprotocol/market/blob/feature/issue-1093-gas-cost-estimation-publish-/src/components/@shared/FormFields/Nft/TxFee.tsx#L27) in the metadata step.

mihaisc:
Indeed , we need to update that to support all the chains

EnzoVezzaro:
this might be an issue as I don't see MATIC on the [supported currency ](https://api.coingecko.com/api/v3/simple/supported_vs_currencies). I'm investigating to find out more about it

mihaisc:
Check this out https://api.coingecko.com/api/v3/coins/matic-network
https://www.coingecko.com/en/api/documentation

kremalicious:
Just UI testing, this is what I see when pointing wallet against ETH mainnet:

Screen Shot 2022-03-09 at 14 18 03

1. If the first step produces 2 tx and the last step produces 1 tx displayed cost on both steps can't be the same.
2. Gwei seems not right, at the time of writing it should say something like 30-40 Gwei.

kremalicious:
closing this, as:

- 400 lines of code are too much for this functionality
- nobody uses ETH mainnet in our market, so usefulness is questionable

Issue #1112, opened 2022-02-17
show "insuficient funds" and disabled the "submit" button on publish if the user has low balance


EnzoVezzaro:
**Is your feature request related to a problem? Please describe.**
when the user has low balance, the UI doesn't reflect that and allows users to continue on their interactions with the contracts. 

It should be better to check if the user has enough balance on his wallet, if not we should block the submit button (like Uniswap for example). This is will prevent weird messaging from the EVM if the balance is not enough. 

<img width="2557" alt="Screenshot 2022-02-17 at 11 06 51" src="https://user-images.githubusercontent.com/13741335/154460695-d039188a-17e1-49cf-b259-2a82913e4a31.png">

mihaisc:
This can be based on the estimation of transaction + a delta

EnzoVezzaro:
this is being worked out on [Show gas cost estimation for publish #1093](https://github.com/oceanprotocol/market/issues/1093)

mihaisc:
Please link prs, not issues.

PR #1133, opened 2022-02-21
added "insuficient funds" label and disabled the "submit" button

EnzoVezzaro:
Fixes [show "insuficient funds" and disabled the "submit" button on publish if the user has low balance](https://github.com/oceanprotocol/market/issues/1112).

Changes proposed in this PR:

- added "insuficient funds" label and disabled "submit" button when the user has no enough balance


src/components/Publish/Submission/GasFees.tsx
```diff
@@ -1,9 +1,28 @@
...
+            status: 'error'
+          }
+        }
+      })
+    }
+  }, [gasFees])
+
+  return gasFees === 'insufficient-funds' ? (
+    <span className={`${styles.gasFees} ${styles.noFunds}`}>
+      Insufficient ETH balance
```
kremalicious, line 23: This check seems to be exclusively for the gas fees, so we should say that, like "Insufficient ETH balance to cover gas fees".

Then, we only have ETH on the ETH mainnet and testnets, but not on all the other production/test networks for which we need to detect the native token of that chain and display it here

src/components/Publish/_utils.ts
```diff
@@ -248,17 +248,6 @@ export async function createTokensAndPricing(
         poolParams
       )
 
-      // the spender in this case is the erc721Factory because we are delegating
-      const txApprove = await approve(
```
kremalicious, line 252: just double-checking, are we sure we don't need an approve transaction here before creating the pool?

src/components/Publish/_utils.ts
```diff
@@ -479,11 +468,16 @@ export async function getFeesTokensAndPricing(
   }
 
   const gasPrice = await web3.eth.getGasPrice()
+  const userBalance = await web3.eth.getBalance(accountId)
 
   const gasFeeEth = web3.utils.fromWei(
     (+gasPrice * +gasEstimate).toString(),
     'ether'
   )
+  const balanceEth = web3.utils.fromWei(userBalance.toString(), 'ether')
```
kremalicious, line 477: same here with different native tokens on different chains, most likely we have to adapt this here too

src/components/Publish/_utils.ts
```diff
@@ -517,11 +511,16 @@ export async function getFeesPublishDDO(
   )
 
   const gasPrice = await web3.eth.getGasPrice()
+  const userBalance = await web3.eth.getBalance(accountId)
 
   const gasFeeEth = web3.utils.fromWei(
     (+gasPrice * +gasEstimate).toString(),
     'ether'
   )
+  const balanceEth = web3.utils.fromWei(userBalance.toString(), 'ether')
```
kremalicious, line 520: same here with different native tokens on different chains, most likely we have to adapt this here too

src/components/Publish/_utils.ts
EnzoVezzaro: sorry, it has been restore. I probably remove it while testing

src/components/Publish/Submission/GasFees.tsx
mihaisc: no, on the pricing screen we just check if the user has enough ocean to add as liquidity, it does not handle the actual gas cost. I think we have the chain token when we get various details regarding the chain.

EnzoVezzaro: ok, now I see it. The pricing screen seems to be checking the user's balance, but only for the "Dynamic" pricing tab. 

Screenshot 2022-02-21 at 15 02 19
Screenshot 2022-02-21 at 15 02 25


It seems there's a bug on it, because I can go to the next step anyway. I think is better to do this on the pricing screen as planned, I'll just remove my calculations and fix what's not working. is that ok?

kremalicious: > on the pricing screen we just check if the user has enough ocean to add as liquidity, it does not handle the actual gas cost

yup, that's totally fine but yes, the continue button on there should be disabled if there's not enough baseToken in user's wallet.

Fyi, this only concerns the dynamic tab because with that pricing method user adds liquidity when creating the pool so the baseToken amount entered in that dynamic screen will e sent out of user's wallet (hence the approve tx before creating pool). This all doesn't apply to the other pricing methods

EnzoVezzaro: > no, on the pricing screen we just check if the user has enough ocean to add as liquidity, it does not handle the actual gas cost. I think we have the chain token when we get various details regarding the chain.

got it. still, shouldn't this check on OCEAN balance be also done on the "Fixed" and "Free" tabs?

src/components/Publish/_utils.ts
kremalicious: one idea apart from contract calls would be to use our `useNetworkMetadata()` as have info about all chains within our codebase. The native token for each chain should be in there somewhere

EnzoVezzaro: can this be achieve with `web3.utils.fromWei(await web3.eth.getBalance(accountId, 'latest'))`, right? like the `getUserBalance` helper in `Web3.tsx`

EnzoVezzaro: > 

got it. I'll try `useNetworkMetadata`. 

I don't know what's going on with the comments, but i'm not getting them in real time 😂 I'm seen your answers always late

kremalicious: yup, just checked, `await web3.eth.getBalance(accountId, 'latest')` works and will return native token balance for each chain. So would be enough unless we require the native token symbol somewhere, then the `useNetworkMetadata()` would be handy

kremalicious: also, just seeing that our Web3 context provider holds and refetches user balance and under `balance.eth` will always put the native token balance. So within components we can always use `const { balance } = useWeb3()` to limit excessive balance checks

PR #1134, opened 2022-02-22
Use subgraph enums

### LINKED ISSUE #1040: Use subgraph enums when available
We should use the enums defined in the subgraph.
For example [here ](https://github.com/oceanprotocol/market/blob/2fb7ed35167abd9bc836d7a2857a4b95be63b1a0/src/components/Asset/AssetContent/EditHistory.tsx#L33) we can use 
```
export enum NftUpdateType {
  METADATA_CREATED = "METADATA_CREATED",
  METADATA_UPDATED = "METADATA_UPDATED",
  STATE_UPDATED = "STATE_UPDATED",
  TOKENURI_UPDATED = "TOKENURI_UPDATED",
}
```

KatunaNorbert:
Fixes #1040.

### DIFF
--- a/src/components/@shared/PoolTransactions/Title.tsx
+++ b/src/components/@shared/PoolTransactions/Title.tsx
@@ -4,12 +4,13 @@ import { useUserPreferences } from '@context/UserPreferences'
 import ExplorerLink from '@shared/ExplorerLink'
 import { formatPrice } from '@shared/Price/PriceUnit'
 import styles from './Title.module.css'
+import { PoolTransactionType } from '../../../@types/subgraph/globalTypes'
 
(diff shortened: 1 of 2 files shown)


KatunaNorbert:
I get the following error when importing the enums from `globalTypes.d.ts`. Any ideas why it say that it can't be found?
Screenshot 2022-02-22 at 14 40 22

mihaisc:
it seems there are some issue when generating the types, some queries are outdated, maybe because of that. Does it generate that type locally?

KatunaNorbert:
Yes, it generates it, I removed the file and it got regenerated.

mihaisc:
It seems to be another webpack issue, took a quick look didn't figure it out.

PR #1139, opened 2022-02-23
Restore Pool Shares section

### LINKED ISSUE #1042: Restore pool shares section

### LINKED ISSUE #1144: Highest liquidity section fails if one subgraph is down
the whole section fails if one subgraph is down. It should be just ignore and fetch data from other ones

### LINKED ISSUE #1145: Profile stats fail if one subgraph is down
Screenshot from 2022-02-25 14-51-42

### LINKED ISSUE #1157: Pool transaction tab in profile shows incorrect infos
**Describe the bug**
In the pool transaction tab in the profile section only title (but the link it points to is broken) and time seem to be correct, data set field is empty, network is always unknown.

**Screenshots**
Screenshot 2022-03-03 at 16 04 29

### LINKED ISSUE #1166: Unified representation of value in pools/profile
Currently we still have a mix of liquidity and tvl 
image
image

We need to be consistent and the data to make sense . Also when you add/remove liquidity you see liquidity

image

So i have a tvl of 1090 but when i remove i can remove 67% ~ 579 ocean , 100% ~ 874 ocean (rough calculation, not accurate)

Also we need to validate these numbers
This is more of a discussion issue. I'll handle the changes in the pool shares pr

claudiaHash:
- fixes #1042
- fixes #1144 
- fixes #1145
- fixes #1166
- fixes #1157 

Changes proposed in this PR:

- restore pool shares section tab in profile history, highest liquidity section and profile stats, fixes and refactors all around


mihaisc:
If you have selected network where the subgraph is offline the tab never loads. If i select just the network with online subgraph and refresh it works.
image

src/@context/Profile.tsx
```diff
@@ -141,6 +141,7 @@ function ProfileProvider({
           poolShares
         )
       } catch (error) {
+        setIsPoolSharesLoading(false)
```
kremalicious, line 144: theoretically™, the `finally` is reached on both, success and `catch`ed error so `setIsPoolSharesLoading(false)` should fire in any case. But I guess this is one of those weirdnesses you encountered? But if not we could remove that extra `setIsPoolSharesLoading(false)` here

src/@utils/accessDetailsAndPricing.ts
```diff
@@ -308,22 +308,26 @@ export async function getAccessDetails(
...
+    )
 
-  const tokenPrice: TokenPrice = tokenQueryResult.data.token
-  const accessDetails = getAccessDetailsFromTokenPrice(tokenPrice, timeout)
-  return accessDetails
+    const tokenPrice: TokenPrice = tokenQueryResult.data.token
+    const accessDetails = getAccessDetailsFromTokenPrice(tokenPrice, timeout)
+    return accessDetails
+  } catch (error) {
+    LoggerInstance.log('Error getting access details: ', error.message)
```
kremalicious, line 329:
```suggestion
    LoggerInstance.error('Error getting access details: ', error.message)
```

src/@utils/accessDetailsAndPricing.ts
```diff
@@ -333,43 +337,48 @@ export async function getAccessDetailsForAssets(
...
+        )
 
-      currentAsset.accessDetails = accessDetails
-    })
+        currentAsset.accessDetails = accessDetails
+      })
+    }
+    return assetsExtended
+  } catch (error) {
+    LoggerInstance.log('Error getting access details: ', error.message)
```
kremalicious, line 382:
```suggestion
    LoggerInstance.error('Error getting access details: ', error.message)
```

src/@utils/aquarius.ts
```diff
@@ -170,6 +170,28 @@ export async function getAssetsFromDidList(
   }
 }
 
+export async function getAssetsFromDtList(
+  dtList: string[],
+  chainIds: number[],
+  cancelToken: CancelToken
+): Promise<any> {
```
kremalicious, line 177: we should be able to be more precise here on the return type instead of `any`

src/@utils/ocean.ts
```diff
@@ -4,22 +4,27 @@ import { AbiItem } from 'web3-utils/types'
...
+        network === 'moonbeamalpha' ||
+        network === 1287 ||
+        network === 'bsc' ||
+        network === 56 ||
+        network === 'gaiaxtestnet' ||
+        network === 2021000
+        ? undefined
+        : process.env.NEXT_PUBLIC_INFURA_PROJECT_ID
+    ) as Config
+    // TODO: remove hack once address is fixed
```
kremalicious, line 20: puh, can't remember why this hack is there. Most likely can be removed as the `oceanTokenAddress` should come from ocean.js now?

src/@utils/ocean.ts
```diff
@@ -4,22 +4,27 @@ import { AbiItem } from 'web3-utils/types'
...
+    if (network === 'rinkeby' || network === 4)
+      config.oceanTokenAddress = '0x8967bcf84170c91b0d24d4302c2376283b0b3a07'
+    return config as Config
+  } catch (error) {
+    LoggerInstance.log('Error getting ocean config: ', error.message)
```
kremalicious, line 25:
```suggestion
    LoggerInstance.error('Error getting ocean config: ', error.message)
```

src/@utils/ocean.ts
mihaisc: this is my hack from like 2 months ago :))

src/@context/Profile.tsx
claudiaHash: yes, u're right. I'l fix it in a moment

src/@utils/subgraph.ts
```diff
@@ -388,17 +415,21 @@ export async function getPoolSharesData(
...
+      chainIds
+    )
+    for (let i = 0; i < result.length; i++) {
+      result[i].poolShares.forEach((poolShare: PoolShare) => {
+        data.push(poolShare)
+      })
+    }
+    return data
+  } catch (error) {
+    console.error('Error fetchData: ', error.message)
```
kremalicious, line 431:
```suggestion
    LoggerInstance.error('Error fetchData: ', error.message)
```

src/@utils/subgraph.ts
```diff
@@ -223,10 +244,15 @@ export async function fetchData(
 ): Promise<any> {
   try {
     const client = getUrqlClientInstance()
+    const checkUrl = await verifyUrl(context.url)
```
kremalicious: not sure about this although it's a nice simple trick. Just concerned that this doubles the requests we make.

When `client.query` does a request, and that errors, that is essentially a URL check so might be worth rather looking into what exactly to react to from the urql response. They have [whole page on error handling](https://formidable.com/open-source/urql/docs/basics/errors/) and somehow they will return a bunch of errors when encountered: https://formidable.com/open-source/urql/docs/api/core/#combinederror. Which should be part of the `OperationResult` we are receiving: https://formidable.com/open-source/urql/docs/api/core/#operationresult

So this might be a more reliable way of detecting subgraph errors. There are also ways to add more what they call [Exchanges](https://formidable.com/open-source/urql/docs/architecture/#the-exchanges), one of them is [errorExchange](https://formidable.com/open-source/urql/docs/api/core/#errorexchange).

That all might take longer to check out and figure out, so if we're in a hurry currently seems to work, but it comes with a heavy price of double the requests.

src/@utils/ocean.ts
kremalicious: or, that whole `try/catch` in here might be overkill as we are simple reading values from a json and if we don't find anything method just returns `null`. So almost sure the `try/catch` here would never be used

src/@utils/subgraph.ts
kremalicious: nice, fixed with [76db8b9](https://github.com/oceanprotocol/market/pull/1139/commits/76db8b9464dfd6fa31f82308e68fc41c87b44dce) while I was writing this

src/@utils/subgraph.ts
```diff
@@ -200,20 +201,40 @@ const OpcFeesQuery = gql`
   }
 `
 
+async function verifyUrl(url: string): Promise<number> {
```
kremalicious, line 204: we can now remove this method

src/@utils/subgraph.ts
```diff
@@ -200,20 +201,40 @@ const OpcFeesQuery = gql`
...
+}
+
 export function getSubgraphUri(chainId: number): string {
-  const config = getOceanConfig(chainId)
-  return config.subgraphUri
+  try {
+    const config = getOceanConfig(chainId)
+    return config.subgraphUri
+  } catch (error) {
+    LoggerInstance.log('Get ocean config error: ', error.message)
```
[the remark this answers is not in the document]
kremalicious: same here [as above](https://github.com/oceanprotocol/market/pull/1139/files#r818687915)

src/@utils/subgraph.ts
```diff
@@ -200,20 +201,40 @@ const OpcFeesQuery = gql`
...
+    }
+  }
+}
+
 export function getSubgraphUri(chainId: number): string {
-  const config = getOceanConfig(chainId)
-  return config.subgraphUri
+  try {
+    const config = getOceanConfig(chainId)
+    return config.subgraphUri
```
kremalicious, line 220: if we do this here as `return config?.subgraphUri` then again no need for `try/catch` as the only error here could be that `subgraphUri` can't be accessed from undefined `config`, and the `?` prevents that from erroring

src/@utils/subgraph.ts
```diff
@@ -237,21 +259,22 @@ export async function fetchDataForMultipleChains(
...
-    const context: OperationContext = {
-      url: `${getSubgraphUri(
-        chainId
-      )}/subgraphs/name/oceanprotocol/ocean-subgraph`,
-      requestPolicy: 'network-only'
-    }
-    try {
+  try {
+    for (const chainId of chainIds) {
+      const context: OperationContext = {
```
kremalicious, line 264: shouldn't we use `getQueryContext()` helper from above here too?

src/components/Profile/History/PoolShares.tsx
```diff
@@ -174,6 +170,7 @@ export default function PoolShares({
         )
         setAssets(assets)
       } catch (error) {
+        setLoading(false)
```
kremalicious, line 173: same as above: https://github.com/oceanprotocol/market/pull/1139/files#r818637638

src/components/Profile/History/PoolShares.tsx
```diff
@@ -174,6 +170,7 @@ export default function PoolShares({
...
         console.error('Error fetching pool shares: ', error.message)
```
kremalicious, line 174: while we're at it:

```suggestion
        LoggerInstance.error('Error fetching pool shares: ', error.message)
```

src/@utils/aquarius.ts
mihaisc: not sure we can, this is the result from elasticsearch. It's kinda complicated and not worth mapping now.  I created an issue at some point to see if we can handle some light mapping of es types but it gets complicated and not worth it.

claudiaHash: I mapped it to Asset[ ]. since we don't use accessDetails in this case I think it's enough

kremalicious:
Now UI, output of liquidity should be exactly like we do it on the pool tab where we do not output the datatoken. This is what I get on my pool shares:

Screen Shot 2022-03-03 at 14 45 40

src/@utils/subgraph.ts
```diff
@@ -237,21 +242,17 @@ export async function fetchDataForMultipleChains(
...
       const response = await fetchData(query, variables, context)
-      datas = datas.concat(response.data)
-    } catch (error) {
-      console.error('Error fetchData: ', error.message)
+      if (!response || response.error) continue
+      datas = datas.concat(response?.data)
     }
+    return datas
+  } catch (error) {
+    LoggerInstance.error('Error fetchData: ', error.message)
```
kremalicious, line 254: for debugging really helps to be precise in those messages for these generic methods, doesn't help if all of them say they come from `fetchData`. So this here should say `Error fetchDataForMultipleChains`.

src/@utils/subgraph.ts
```diff
@@ -268,7 +269,7 @@ export async function getOpcFees(chainId: number) {
     )
     opcFees = response?.data?.opc
   } catch (error) {
-    console.error('Error fetchData: ', error.message)
+    LoggerInstance.error('Error fetchData: ', error.message)
```
kremalicious, line 272: and this one `Error getOpcFees`

src/@utils/subgraph.ts
```diff
@@ -388,17 +389,21 @@ export async function getPoolSharesData(
...
+      chainIds
+    )
+    for (let i = 0; i < result.length; i++) {
+      result[i].poolShares.forEach((poolShare: PoolShare) => {
+        data.push(poolShare)
+      })
+    }
+    return data
+  } catch (error) {
+    LoggerInstance.error('Error fetchData: ', error.message)
```
kremalicious, line 405: and this one `Error getPoolSharesData`

claudiaHash:
> Each time i swap tabs/change networks i get different numbers 

Yes, I encountered the same issue, still on it

kremalicious:
somewhere we have way too many refetches, like when I deselect a network while pool shares tab is open, it looks like 2 rounds of fetches happen shortly after another. This is also visible in the above posted screen recording

mihaisc:
I'll hop on this pr, the whole `Profile` component is acting weird, hope we won't need big refactor

kremalicious:
Current state, picked up on what was in here, think all the errors are gone and I also made sure to unify the calculations between what we do on asset's pool tab and here in the pool shares table. So far, so good but one major problem I can't find the exact reason yet. Basically the numbers are correct, but they are on the wrong asset, here the below asset should display 15 OCEAN and the top one 11 OCEAN

Screen Shot 2022-03-07 at 14 07 11

Somewhere in `getAssetsFromPoolShares()` matching `ddoList` with `assetList` seem to not work out correctly, investigating...

claudiaHash:
I'm checking it out too

mihaisc:
The numbers look good. We can do a lot of optimizations at a later time. Currently if i uncheck/check another network we refetch everything and it's slow.
image

kremalicious:
> Currently if i uncheck/check another network we refetch everything and it's slow.

yes, whenever `chainIds` changes all subgraphs are hit again to get the new `poolShares` in the `Profile` provider. Which then triggers various effects in subcomponents to recalculate. Only using some delta between previous selected `chainIds` and newly selected ones, and then only hit those subgraphs could be a first effective optimization

kremalicious (approved):
good enough™ and has lots of changes already so we should run with it and test in daily development. Thanks @claudiaHash & @mihaisc, there was something for everybody in here :-)

Issue #1130, opened 2022-02-21
Radio inputs have wrong color in Firefox


kremalicious:
The radio inputs we use in the Publish form in Pricing tabs have wrong color and styling in Firefox:

<img width="520" alt="Screen Shot 2022-02-21 at 11 27 28" src="https://user-images.githubusercontent.com/90316/154946793-a5da0abe-a78e-4259-abb2-09562ec135b2.png">

This is in contrast to other radio inputs where in Firefox they show the correct color, like in Compute Environment so this is what we want:

<img width="376" alt="Screen Shot 2022-02-21 at 11 27 39" src="https://user-images.githubusercontent.com/90316/154946879-a0dbd814-3e62-47d9-b447-089e676fb245.png">

PR #1142, opened 2022-02-24
remove unnecessary appearance on radio

EnzoVezzaro:
Fixes [Radio inputs have wrong color in Firefox#1130](https://github.com/oceanprotocol/market/issues/1130).

Changes proposed in this PR:

- remove unnecessary appearance CSS property


EnzoVezzaro:
this CSS property was causing the issue. I couldn't trace why this has been applied, for that reason I decided to remove it. If there's the need to restore it, please let me know where this is been used.

mihaisc (approved):

PR #1146, opened 2022-02-25
Issue 1023 v4 migration next

md0x:
Fixes # .

Changes proposed in this PR:

- 
- 
-


mihaisc:
I am a bit confused, why is this against v4?

md0x:
> I am a bit confused, why is this against v4?

My bad, its against v3

kremalicious:
Is this in addition to #1036 or is this PR here superseding #1036? If the latter we can close #1036. 

Needs another push because of branch change to see how deployment goes.

jamiehewitt15:
I've tested that flow a couple of times and it works well for assets with a pool. Couple of small UI issues to work on:
- The `Loader` component doesn't seem to be working on any of the migration components
- The info message is showing the number of locked pool shares with the wrong number of decimals "You have locked 100000000000000000000 shares" when it should show 100
- We should make use of the `refreshMigrationStatus` function to refresh the message and action button on the asset detail page after the user has completed an action.

app.config.js
```diff
@@ -74,5 +74,29 @@ module.exports = {
...
+  rinkebyMigrationContractAddresss:
+    process.env.GATSBY_MIGRATION_CONTRACT_ADDRESS_RINKEBY ||
+    '0x0e4806e14bd02A809f49aD37F8C749f117D84239',
+  ropstenMigrationContractAddresss:
+    process.env.GATSBY_MIGRATION_CONTRACT_ADDRESS_RINKEBY ||
+    '0x003323388790872320fbf0152a49E2de4046E0b3',
+  moonbaseMigrationContractAddresss:
+    process.env.GATSBY_MIGRATION_CONTRACT_ADDRESS_RINKEBY ||
+    '0x7874E8bA0f96fA3cD883453a596BF97aEd5c7a32',
+  mumbaiMigrationContractAddresss:
```
kremalicious, line 99: this is full of networks we never intended to do a "migration" for so please clean up

kremalicious:
please stop touching that package-lock.json, fixed it in [316940a](https://github.com/oceanprotocol/market/pull/1146/commits/316940afc0eda6202023bc710a99963d5579cbdf) but can't do this after every commit to get the test runs green again

jamiehewitt15:
> please stop touching that package-lock.json, fixed it in [316940a](https://github.com/oceanprotocol/market/pull/1146/commits/316940afc0eda6202023bc710a99963d5579cbdf) but can't do this after every commit to get the test runs green again. Not sure why your npm always adds `ssh+git` to dependencies, which then fails when run in CI. They all should be `git+https`

Sorry, not sure why that's happing. I'll avoid committing any changes to it

mihaisc:
I know, i hate that ssh thing so much, tried at some point to find a permanent fix but couldn't . In theory it shouldn't happen but...npm

jamiehewitt15:
> most of the added code in here deal with some migration started by user. I thought everything happens manually so why is all this logic in here?

I've changed the logic so that it's only checking if the deadline has been met & if the pool shares are already locked. It's now implicitly assumed that the migration period has already started when this branch is deployed

kremalicious (approved):

PR #1167, opened 2022-03-07
Feat: add final did to tokenURI in NFT metadata

### LINKED ISSUE #996: Add final DID to NFT metadata
As defined in #944, we have in the NFT metadata json `external_url`, which should be populated with the full link to the respective asset in the live market. Additionally, the `description` could include the link too.

Both things show up e.g. on opensea within their UI.

Right now we generate the NFT metadata before doing anything else and only in 2nd step we get the DID. So most likely we have to inject the DID at the end of the process when we do `nft.setMetadata`.

LucaMilanese90:
Resolves #996

Changes proposed in this PR:

- add final asset did in tokenURI
- add implementation for setMetaDataAndTokenURI function
- remove tokenURI value from first transaction and encode it in the second one

### DIFF
--- a/src/@utils/nft.ts
+++ b/src/@utils/nft.ts
@@ -5,7 +5,8 @@ import {
   getHash,
   Nft,
   ProviderInstance,
-  DDO
+  DDO,
+  MetadataAndTokenURI
 } from '@oceanprotocol/lib'
 import Web3 from 'web3'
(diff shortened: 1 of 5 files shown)


kremalicious (approved):
should work™, needs to go through daily testing in v4 branch so just yolo merging

Issue #1106, opened 2022-02-15
if an user rejects the approval of the spender, the transaction doesn't get restore again


EnzoVezzaro:
**Describe the bug**
if you reject the transaction to approve the spender, it doesn't ask you again the next time.

**To Reproduce**
1. Go to publish
2. Fill out everything with Dynamic pricing
2. Click on "Submit"
3. Reject the first transaction (reject approval spender)
4. You should see the error from metamask. As you can see now you have `1 transaction` which is wrong as it should continue to have `2 transactions` and also if you retry to submit, all the transactions fail (i'm guessing because is missing the approval).

**Expected behavior**
I'd expect to be asked again for approval of the spender.

**Screenshots**

https://user-images.githubusercontent.com/13741335/154083799-a3f03d40-e171-4266-9405-83b20597d84a.mp4

PR #1173, opened 2022-03-09
stop execution on publish if user rejects approval tokens

EnzoVezzaro:
Fixes [if an user rejects the approval of the spender, the transaction doesn't get restore again #1106](https://github.com/oceanprotocol/market/issues/1106).

Changes proposed in this PR:

- catch user rejection on token approval


kremalicious (approved):
Finally some nice small changes PRs!

Issue #1161, opened 2022-03-03
Publishing an algorithm requires to interact with the checkbox "Keep my algorithm private"


LucaMilanese90:
**Describe the bug**
When publishing an algorithm, in the second step, the continue button is disable until the user selects the "Keep my algorithm private" checkbox, even if that's not the desired choice

**To Reproduce**
1. Start algo publish
2. On step 2 fill all the fields without touching the "Keep my algorithm private" checkbox
3. See that "continue" is disabled

**Expected behavior**
If the user don't want to keep the algo private it should not be forced to check and uncheck the option

**Screenshots**
![discoverable and catch the interest of data consumers](https://user-images.githubusercontent.com/20192135/156613280-cc8e9cfd-1d64-4de9-a287-adf405debe90.png)

PR #1174, opened 2022-03-10
Keep algorithm private checkbox fixes

### LINKED ISSUE #1160: "Keep my algorithm private" checkbox seems to have no effect when publishing
**Describe the bug**
When publishing an algorithm for download, checking or un-checking the "Keep my algorithm private" option still lead to the message This algorithm has been set to private by the publisher and can't be downloaded. You can run it against any allowed data sets though!

**To Reproduce**
1. Publish algo
2. Don't select the option "Keep my algorithm private"
3. Complete publish
4. See message in asset details "This algorithm has been set to private by the publisher and can't be downloaded. You can run it against any allowed data sets though!"

**Screenshots**
Screenshot 2022-03-03 at 17 52 27

KatunaNorbert:
Fixes #1160 && #1161


kremalicious:
Checkbox now does something else unintended but looks like that's also happening on `v4` branch so no clue what happened there:

Untitled

That Compute Environment field should not be connected with the privacy checkbox in any way. But not even sure this should be present at all for algos.

@bogdanfazakas do we set compute environments for compute datasets AND algos, or just compute data sets?

bogdanfazakas:
> Checkbox now does something else unintended but looks like that's also happening on `v4` branch so no clue what happened there:
> 
> Untitled
> 
> That Compute Environment field should not be connected with the privacy checkbox in any way. But not even sure this should be present at all for algos.
> 
> @bogdanfazakas do we set compute environments for compute data sets AND algos, or just compute data sets?

true,I think we should have had them set only on datasets, and as we handle things now we are setting compute environments by querying  the provider and getting the values from there. But i think we can keep them so the publisher would be able to set some values there like recommended compute env settings or something like so when you as a consumer when reserving the c2d environment have an idea on what to you would need

kremalicious:
I missed something conceptionally, and yes because the environments are now set on the provider we do not set them anymore during publish. So that form field and all logic/code attached to it can be removed.

Later on we might want to still find some way to show those environments, like when you publish an algo we can show to publisher which resources are available from the provider but should not be made in a way which suggests those are set during publish, so no normal form field

KatunaNorbert:
Ok, I will remove the field

kremalicious:
I discovered more where we still have the compute environments set during publish as current implementation, including `ServiceComputeOptions` coming from ocean.js does not reflect the current state of the spec: https://deploy-preview-819--docs-oceanprotocol.netlify.app/concepts/did-ddo/#compute-options

This applies to all compute assets, so most likely we want to do this in another PR. Or did @bogdanfazakas update this in #1069 somewhere?

bogdanfazakas:
> I discovered more where we still have the compute environments set during publish as current implementation, including `ServiceComputeOptions` coming from ocean.js does not reflect the current state of the spec: https://deploy-preview-819--docs-oceanprotocol.netlify.app/concepts/did-ddo/#compute-options
> 
> This applies to all compute assets, so most likely we want to do this in another PR. Or did @bogdanfazakas update this in #1069 somewhere?

Good findings we need to update oceanjs typings first (remove namespace, cpu, gpu, gpuType, memory, and volumeSize from ServiceComputeOptions) and then update the market publish in #1069

kremalicious:
On first tests, fixes original issues. But published algos would be now public by default (cause of `access` type default), and we want the other way around, checkbox should be ticked by default.

KatunaNorbert:
I see, I'm changing it now

kremalicious:
also seeing I left comments so this would be alternative as the form data and the data in final DDO do not need to be the same, so we could do the transformation also in our utility method. But this has lots of logic already so probably best to keep more logic out of it

https://github.com/oceanprotocol/market/pull/1174/files#diff-ccc0182ce5c4d89f431f804248ab537210f99b7a800587f30ba6344d28321b3eR43-R44

kremalicious (approved):
both linked issues are fixed with this based on some quick testing

PR #1178, opened 2022-03-10
Replace collect base tokens button

### LINKED ISSUE #1170: Confusing collection button all publishers see
What does it collect? Why is it 0? Publishers will have questions so we can't just dump a button there and hope people will use it to trigger possible costly transactions.

Screen Shot 2022-03-09 at 13 14 48

1. deprioritize this button in visual hierarchy, most likely by using smaller/secondary button, or just a text link.
2. add help text explaining why this button is there and what it does.
3. most likely should live in `AssetActions` and not in `Download` if it also applies to Compute
4. it actually has nothing to do with the Use tab. So we might want to look into moving this outside of the box below it. Conceptionally we already have publisher actions though, and they are in the main content area

claudiaHash:
Fixes #1170  .

Changes proposed in this PR:

- replace collect base tokens button into `AssetActions` component


src/components/Asset/AssetContent/CollectTokens.tsx
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  return (
+    asset &&
+    isOnwer &&
+    baseTokenBalance > 0 &&
+    (!isCollectLoading ? (
+      <a onClick={handleCollectTokens}>
+        Collect {baseTokenBalance} {asset?.accessDetails?.baseToken.symbol}
+        <Tooltip content="As an asset owner you can collect base tokens in the contract." />
```
claudiaHash, line 102: Not sure about the tooltip content, if it should be more explicit about the number of tokens you can collect and what do those tokens mean

src/components/Asset/AssetContent/CollectTokens.tsx
```diff
@@ -0,0 +1,108 @@
...
+      LoggerInstance.log(error.message)
+    } finally {
+      setIsCollectLoading(false)
+    }
+  }
+
+  return (
+    asset &&
+    isOnwer &&
+    baseTokenBalance > 0 &&
```
claudiaHash, line 98: Should we even show this if there are no tokens to be collected?

src/components/Asset/AssetContent/CollectTokens.tsx
mihaisc: 'Collect the tokens that you earned by selling this asset'

src/components/Asset/AssetContent/CollectTokens.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const [collectStatusText, setCollectStatusText] = useState('')
+
+  const { accountId, web3 } = useWeb3()
+
+  useEffect(() => {
+    if (!asset || error) {
+      return
+    }
+    setIsOwner(owner === accountId)
+  }, [asset, error])
```
kremalicious, line 40: should have `accountId` as dependency so this fires when wallet account is changed. Should then also just `return` if `!accountId` as `accountId` is `undefined` when no wallet connected

src/components/Asset/AssetContent/CollectTokens.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  useEffect(() => {
+    if (!accountId || asset.nft.owner !== accountId) return
```
kremalicious, line 43: we should get the `owner` consistently throughout this file, but then in first effect we have this check done already  so should use here in the end:

```suggestion
    if (!accountId || !isOwner) return
```

src/components/Asset/AssetContent/CollectTokens.tsx
```diff
@@ -0,0 +1,108 @@
...
+      }
+      exchangeId
+      baseTokenBalance
+    }
+  }
+`
+
+export default function CollectTokens(): ReactElement {
+  const { asset, error, owner } = useAsset()
+  const [isOnwer, setIsOwner] = useState(false)
```
kremalicious, line 28:
```suggestion
  const [isOwner, setIsOwner] = useState(false)
```

src/components/Asset/AssetContent/CollectTokens.tsx
```diff
@@ -0,0 +1,108 @@
...
+          )
+        : setBaseTokenBalance(0)
+    }
+    getBaseTokenBalance()
+  }, [accountId, asset?.accessDetails?.addressOrId, asset.chainId, asset.nft])
+
+  async function handleCollectTokens() {
+    if (baseTokenBalance === 0) return
+    setIsCollectLoading(true)
+    const config = getOceanConfig(asset?.chainId)
```
kremalicious, line 68: rather use `oceanConfig` from `useAsset()` as getting correct config is already being done there

src/components/Asset/AssetContent/index.tsx
```diff
@@ -74,11 +75,14 @@ export default function AssetContent({
         <div className={styles.actions}>
           <AssetActions asset={asset} />
           {isOwner && isAssetNetwork && (
-            <div className={styles.ownerActions}>
-              <Link href={`/asset/${asset?.id}/edit`}>
-                <a>Edit</a>
-              </Link>
-            </div>
+            <>
```
kremalicious, line 78: the `<>...</>` are unnecessary here and can be removed

src/components/Asset/AssetContent/CollectTokens.tsx
```diff
@@ -0,0 +1,108 @@
...
+    }
+  }
+`
+
+export default function CollectTokens(): ReactElement {
+  const { asset, error, owner } = useAsset()
+  const [isOnwer, setIsOwner] = useState(false)
+  const [baseTokenBalance, setBaseTokenBalance] = useState(0)
+  const [isCollectLoading, setIsCollectLoading] = useState(false)
+  const [collectStatusText, setCollectStatusText] = useState('')
```
kremalicious, line 31: we are not outputting `collectStatusText` anywhere in this file

src/components/Asset/AssetContent/CollectTokens.tsx
```diff
@@ -0,0 +1,108 @@
...
+      setIsCollectLoading(false)
+    }
+  }
+
+  return (
+    asset &&
+    isOnwer &&
+    baseTokenBalance > 0 &&
+    (!isCollectLoading ? (
+      <a onClick={handleCollectTokens}>
```
kremalicious, line 100: semantically this is not a link but should be a `button`. We should be able to use our `Button` component here as the baseline with `style="text`, as this component deals with switching between `a` & `button` already

mihaisc:
collectButton
The button needs some work 😁

claudiaHash:
> The button needs some work grin

yes, ofc. I'm on it but forgot to make this pr a draft again

EnzoVezzaro:
As for the UI, I'm might not be the right person here 😹 but if I may, I'd suggest something like this for the collect button. Of course the text should be adapted, but I think this layout could represent better this component. I think this can be discussed further of course.

Group 4

PS: you need to take into consideration when the user is on network different than the assets', as the message to collect should show up only on the correct network. Do you agree @kremalicious ?

claudiaHash:
> As for the UI, I'm might not be the right person here joy_cat but if I may, I'd suggest something like this for the collect button. Of course the text should be adapted, but I think this layout could represent better this component. I think this can be discussed further of course.
> 
> Group 4
> 
> PS: you need to take into consideration when the user is on network different than the assets', as the message to collect should show up only on the correct network. Do you agree @kremalicious ?

good catch!

claudiaHash:
> On each order there will be an automatic collect on contract level.

Ok then should we close it?

kremalicious:
See #1222

PR #1185, opened 2022-03-15
Add market version dropdown

### LINKED ISSUE #1186: disable publish
As discribed in #1004, we need to:

- in UI, both the v3 market and v4 market get a dropdown for the version number beside the app title for easy switching between above domains
- remove all publishing functionality from v3 code base

md0x:
Signed-off-by: Pablo Maldonado <pablo@oceanprotocol.com>

- addresses #1450

Changes proposed in this PR:

- Add market v3/v4 dropdown 

Screen Shot 2022-05-23 at 16 56 09

Screen Shot 2022-05-23 at 16 55 32


mihaisc:
@kremalicious any idea why the vercel build are mixed up?

kremalicious:
> any idea why the vercel build are mixed up?

it's complex, explained in Slack

md0x:
> Think we should go a bit more minimal with that as users won't constantly interact with this dropdown, so it should not be on same visual hierarchy as all the other menubar actions.
> 
> So something like this:
> 
> Screen Shot 2022-03-16 at 12 38 21

@kremalicious I have updated the dropdown design and updated the screenshots accordingly in the PR description. Was the offset of the dropdown intended? 
image

jamiehewitt15:
Just want to confirm, @kremalicious are you taking this one now?

src/components/molecules/Menu.tsx
```diff
@@ -9,6 +9,13 @@ import Badge from '../atoms/Badge'
 import Logo from '../atoms/Logo'
 import Networks from './UserPreferences/Networks'
 import SearchBar from './SearchBar'
+import Tooltip from '../atoms/Tooltip'
+import { ReactComponent as Caret } from '../../images/caret.svg'
+import appConfig from '../../../app.config'
+
+const redirectToMarketV4 = (): void => {
+  window.location.href = appConfig.marketV4Url
```
kremalicious, line 17: TODO: just use a plain link with `href`

src/components/molecules/Menu.tsx
```diff
@@ -39,10 +46,34 @@ export default function Menu(): ReactElement {
...
       <Link to="/" className={styles.logo}>
         <Logo noWordmark />
-        <h1 className={styles.title}>
-          {siteTitle} <Badge label="v3" />
-        </h1>
+        <h1 className={styles.title}>{siteTitle}</h1>
       </Link>
+      <Tooltip
+        content={
+          <ul className={styles.preferencesDetails}>
```
kremalicious, line 53: TODO: way too much markup for what we want to do, simplify

kremalicious:
TODO: there is no version badge on mobile, so also no way to switch versions on mobile

mihaisc:
instead of a dropdown wouldn't it better to have a switch?
https://codepen.io/alvarotrigo/pen/wvPRrZW
toggle

kremalicious:
> instead of a dropdown wouldn't it better to have a switch?

was also thinking that but our transition between both versions is not so smooth as we are switching domains. So switch then feels a bit weird as user would never really see it move

kremalicious:
ok, currently pushed state works because menu only has 1 item left. For the v4 versions we need to figure out some other placement in small viewport view.

Regardless, we have to touch this again as we have all those different market URLs which will then change and switch in the future, so every time we do that, we have to update future `v3` branch always together with future `main` branch

mihaisc:
Ah, didn't think this through

Issue #1184, opened 2022-03-14
Transactions number in publish submit step


claudiaHash:
When publishing an asset with dynamic price the transactions number and message  for "Create Tokens & Pricing" is correctly displayed until hitting the submit button:

![Screenshot from 2022-03-14 17-27-01](https://user-images.githubusercontent.com/49017601/158206351-84e8c214-5957-4de8-8ab9-6ac893a4b14e.png)

![Screenshot from 2022-03-14 17-27-04](https://user-images.githubusercontent.com/49017601/158206376-68fdc43d-a847-4b7e-803f-28c17c00f9cd.png)

PR #1187, opened 2022-03-15
fix txCount on 'Create tokens & pricing' in publish

EnzoVezzaro:
Fixes [Transactions number in publish submit step #1184](https://github.com/oceanprotocol/market/issues/1184).

Changes proposed in this PR:

- fix txCount on 'Create tokens & pricing' in publish

### DIFF
--- a/src/components/Publish/Steps.tsx
+++ b/src/components/Publish/Steps.tsx
@@ -1,6 +1,6 @@
 import { ReactElement, useEffect } from 'react'
 import { useFormikContext } from 'formik'
-import { wizardSteps } from './_constants'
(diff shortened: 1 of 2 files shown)


claudiaHash (commented):
The transactions number is fixed now, but I think we should do the same with the message which still changes after submit.
Screenshot from 2022-03-16 12-57-52

EnzoVezzaro:
> The transactions number is fixed now, but I think we should do the same with the message which still changes after submit. Screenshot from 2022-03-16 12-57-52

Great, thanks. Just push this fix 👍🏽

kremalicious (approved):
confirmed numbers/description stay as they should now

PR #1198, opened 2022-03-17
disable price calculation for free price

### LINKED ISSUE #1168: Unnecessary "Calculating price incl. fees"
This text and loading flow should only appear on pool assets, as there is nothing to calculate for fixed or free pricing. Fixed/Free assets should get the action button right away.

KatunaNorbert:
Fixes #1168.

### DIFF
--- a/src/components/Asset/AssetActions/Download.tsx
+++ b/src/components/Asset/AssetActions/Download.tsx
@@ -75,7 +75,7 @@ export default function Download({
     async function init() {
       if (
         asset?.accessDetails?.addressOrId === ZERO_ADDRESS ||
-        asset?.accessDetails.type === 'free'
+        asset?.accessDetails?.type === 'free'
       )
         return
       setIsLoading(true)
--- a/src/components/Asset/AssetActions/Download.tsx
+++ b/src/components/Asset/AssetActions/Download.tsx
@@ -73,7 +73,11 @@ export default function Download({
     setValidOrderTx(asset?.accessDetails?.validOrderTx)
     // get full price and fees
     async function init() {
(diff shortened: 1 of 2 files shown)


claudiaHash (approved):
LGTM

src/components/Asset/AssetActions/Download.tsx
```diff
@@ -73,7 +73,11 @@ export default function Download({
     setValidOrderTx(asset?.accessDetails?.validOrderTx)
     // get full price and fees
     async function init() {
-      if (asset?.accessDetails?.addressOrId === ZERO_ADDRESS) return
+      if (
+        asset?.accessDetails?.addressOrId === ZERO_ADDRESS ||
```
EnzoVezzaro, line 77: I couldn't verify the need for this. In which case does the address equal to ZERO_ADDRESS? is this really need it?

KatunaNorbert: I don't know why that check is in there, that's why I didn't touch it.

mihaisc: that check is for the publish preview

EnzoVezzaro (approved):

src/components/Asset/AssetActions/Download.tsx
```diff
@@ -73,7 +73,11 @@ export default function Download({
...
+        asset?.accessDetails.type === 'free'
```
kremalicious, line 78: if there's `asset?.accessDetails?` everywhere else, the `?` should be set here too

EnzoVezzaro (approved):

Issue #1213, opened 2022-03-18
missing baseToken symbol breaks Download component


EnzoVezzaro:
**Describe the bug**
In Free asset: did:op:0acc800639868d472a6faffcb599c5e0a54e059d9f425dacc06a623c3a381ec4
Page breaks with error: Cannot read properties of undefined (reading 'symbol')

**To Reproduce**
1. Go to Asset with Free pricing
2. Open asset

**Expected behavior**
Should display this asset's information

**Screenshots**
DEV env
<img width="1440" alt="Screen Shot 2022-03-18 at 6 33 53 AM" src="https://user-images.githubusercontent.com/13741335/158988023-b664c154-63c5-4313-885f-1978c7c5923c.png">

Live env
<img width="1428" alt="Screenshot 2022-03-18 at 10 21 20" src="https://user-images.githubusercontent.com/13741335/158988158-7750d83f-cc5e-4a99-bf4d-3895176fc3ee.png">

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Brave

kremalicious:
this has not been fixed and is the actual reason for #1241 so this issue here should take precedence as #1241 has no info on the actual reason.

EnzoVezzaro:
Maybe I got it wrong, but I understood that the "Collect" button was going [to be removed](https://github.com/oceanprotocol/market/issues/1222). This issue is related to a parameter on this ["Collect" button](https://github.com/oceanprotocol/market/blob/v4/src/components/Asset/AssetActions/Download.tsx#L244:L244).

kremalicious:
well even better if #1222 solves 3 other issues but as of right now the open bugs are correct as nothing is fixed

PR #1214, opened 2022-03-18
render CollectTokensButton only if all required params are available

EnzoVezzaro:
Fixes missing baseToken symbol breaks Download component #1213 .

Changes proposed in this PR:

- render CollectTokensButton only if all required params are available to solve breaking issue

### DIFF
--- a/src/components/Asset/AssetActions/Download.tsx
+++ b/src/components/Asset/AssetActions/Download.tsx
@@ -236,7 +236,7 @@ export default function Download({
       disabled={baseTokenBalance === 0 || !baseTokenBalance}
       hasPreviousOrder={false}
       hasDatatoken={false}
-      dtSymbol={asset?.accessDetails?.baseToken.symbol}
+      dtSymbol={asset.accessDetails.datatoken.symbol}
       dtBalance={baseTokenBalance.toString()}
       datasetLowPoolLiquidity={false}
       assetType=""
@@ -267,7 +267,8 @@ export default function Download({
       </div>
 
       {asset?.accessDetails?.datatoken?.name !== '' &&
-        asset?.nft.owner === accountId && (
+        asset?.nft.owner === accountId &&
+        asset?.accessDetails?.datatoken?.symbol && (
           <div className={styles.collect}>
             <CollectTokensButton />
           </div>


src/components/Asset/AssetActions/Download.tsx
```diff
@@ -236,7 +236,7 @@ export default function Download({
       disabled={baseTokenBalance === 0 || !baseTokenBalance}
       hasPreviousOrder={false}
       hasDatatoken={false}
-      dtSymbol={asset?.accessDetails?.baseToken.symbol}
+      dtSymbol={asset.accessDetails.datatoken.symbol}
```
kremalicious, line 239: and now a missing datatoken symbol will break the component. The solution for all those `Cannot read properties of undefined` errors is to make correct use of optional chaining operator so in this case `asset?.accessDetails?.datatoken?.symbol` which is the same as `asset && asset.accessDetails && asset.accessDetails.datatoken && asset.accessDetails.datatoken.symbol`

This is required cause both `asset` and then `asset.accessDetails` take some time to be fetched and constructed

EnzoVezzaro: but it will never reach that point if it doesn't have the datatoken symbol. On line 271, I'm checking also that the symbol is valued, otherwise it doesn't render this component (<CollectTokensButton/>). Since this parameter (dtSymbol) is required for the correct functioning of the component, I though to check if the symbol is valued before rendering it. This will prevent the component to break as it'll never render without symbol. Maybe is not the right approach?

src/components/Asset/AssetActions/Download.tsx
```diff
@@ -236,7 +236,7 @@ export default function Download({
       disabled={baseTokenBalance === 0 || !baseTokenBalance}
       hasPreviousOrder={false}
       hasDatatoken={false}
-      dtSymbol={asset?.accessDetails?.baseToken.symbol}
```
claudiaHash, line 239: I'm pretty sure we need to use `.baseToken` here because we need the symbol of the base token to be collected by the publisher, not the datatoken, so the needed change would be: `dtSymbol={asset?.accessDetails?.baseToken?.symbol}`, as Matthias suggested

src/components/Asset/AssetActions/Download.tsx
```diff
@@ -267,7 +267,8 @@ export default function Download({
       </div>
 
       {asset?.accessDetails?.datatoken?.name !== '' &&
-        asset?.nft.owner === accountId && (
+        asset?.nft.owner === accountId &&
+        asset?.accessDetails?.datatoken?.symbol && (
```
EnzoVezzaro, line 271: @kremalicious @claudiaHash here I'm checking `asset?.accessDetails?.datatoken?.symbol` is valued otherwise it won't render the component (this avoids breaking the component as it will never missed this .symbol). This should be enough to avoid rendering the component without it necessary values. 

I can add these same optional chaining operator in line 239, but it will be useless as it'll never reach the component without having `.symbol` valued. I can change this implementation to add the optional chain op `asset?.accessDetails?.datatoken?.symbol` directly to the component as it won't change much, but i think it's safer to avoid rendering the component without this value as if the value is `underfined` (because of the optional chaining), then the component will show "collect 0 undefined". We can have a fallback for this, but i don't think matters as it should always render the token's symbol. Does make sense or should we stick with the optional chaining as param on the component?

kremalicious:
See #1222

kremalicious:
title & linked issue mismatch, if this is still issue after #1222 feel free to reopen this PR and change its title

PR #1221, opened 2022-03-21
Fix/issue1182 no error message

### LINKED ISSUE #1182: No error message is displayed for invalid url
No error sign is displayed for file and sample file fields when populated with wrong url inputs

claudiaHash:
Fixes #1182 .

Changes proposed in this PR:

- show url validation for file and sample file fields on publish form

### DIFF
--- a/src/components/@shared/FormFields/FilesInput/index.tsx
+++ b/src/components/@shared/FormFields/FilesInput/index.tsx
@@ -10,6 +10,7 @@ import { FormPublishData } from 'src/components/Publish/_types'
 
 export default function FilesInput(props: InputProps): ReactElement {
   const [field, meta, helpers] = useField(props.name)
+  const [urlHasError, setUrlHasError] = useState(false)
   const [isLoading, setIsLoading] = useState(false)
   const { values } = useFormikContext<FormPublishData>()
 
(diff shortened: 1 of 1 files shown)


kremalicious:
Some visual regressions found in here:

### invalid URL

When I type in like this and then unfocus by clicking outside of field, I get this with this PR's preview:

Screen Shot 2022-03-21 at 17 23 58

While on `v4` I get this, which is a more complete error state because of red border so this is what we want in this case:

Screen Shot 2022-03-21 at 17 24 54

### invalid file

Getting in both versions the same thing after hitting _Validate_, so no message is there, meaning PR seems to not fully solve linked issue:

Screen Shot 2022-03-21 at 17 27 06

Ideally we would show a message within our existing form input error states, so when `/fileinfo` response is `valid: false`, something like: "No valid file detected. Check your URL and try again."

claudiaHash:
@kremalicious you're right, thanks for explaining this. I added an auxiliary message and different background and border color for wrong url inputs

kremalicious:
That one UI regression for _invalid URL_ case is still in here.

On live `v4`, which is what we want in this case after clicking outside of input:

Screen Shot 2022-03-29 at 13 35 57

this PR preview:

Screen Shot 2022-03-29 at 13 37 01

But for _invalid file_ use case this is really great now!

Screen Shot 2022-03-29 at 13 39 32

kremalicious (approved):
all working perfectly now

PR #1225, opened 2022-03-21
Price length and decimals number validation

### LINKED ISSUE #1220: No form validation on price length
**Describe the bug**
The pricing form let's you enter any number of any size but then if it's too big the transactions won't go through. It would be better to catch this on the pricing form with an error message that pops up saying the max number size. Same issue happens when there are too many decimals. 

**To Reproduce**
1. Enter information to publish asset
2. Set fixed price as 100000000000000000000000000000000000000000000000000000000000000000000000000000000

**Screenshots**
image

claudiaHash:
Fixes #1220  .

Changes proposed in this PR:

- price length validation (max 1000000) and number of decimal digits (max 6 ) validation for price in publish form


src/components/Publish/_validation.ts
```diff
@@ -54,13 +55,23 @@ const validationService = {
...
   // https://github.com/jquense/yup#mixedwhenkeys-string--arraystring-builder-object--value-schema-schema-schema
   price: Yup.number()
     .min(1, (param: { min: number }) => `Must be more or equal to ${param.min}`)
+    .max(
+      1000000,
+      (param: { max: number }) => `Must be less than or equal to ${param.max}`
+    )
+    .test(
+      'maxDigitsAfterDecimal',
+      'Must have less than 7 decimal digits',
```
kremalicious, line 72: there's `MAX_DECIMALS` in `src/@utils/constants.ts` where this number should come from

src/components/Publish/_validation.ts
```diff
@@ -54,13 +55,23 @@ const validationService = {
   })
 }
 
+const decimalDigitsValidation = new RegExp(/^\d+(\.\d{1,6})?$/)
```
kremalicious, line 58: hard to come up with clear variable name so we don't have to explain regex in code comment. Is `maxDecimalsValidation` more clear?

src/components/Publish/_validation.ts
```diff
@@ -54,13 +56,23 @@ const validationService = {
   })
 }
 
+const maxDecimalsValidation = new RegExp(/^\d+(\.\d{1,5})?$/)
```
kremalicious, line 59: uuuh fun, does interpolation work in regex? Cause this also needs to be based on `MAX_DECIMALS`

claudiaHash: I tried that and it seems not, because of the non-literal argument found in the regex. looking for alternatives

mihaisc:
image

src/components/Publish/_validation.ts
```diff
@@ -54,25 +56,56 @@ const validationService = {
...
+    )
     .required('Required'),
   weightOnDataToken: Yup.string().required('Required'),
   weightOnOcean: Yup.string().required('Required'),
   swapFee: Yup.number()
     .min(0.1, (param) => `Must be more or equal to ${param.min}`)
     .max(10, 'Maximum is 10%')
+    .test(
+      'maxDigitsAfterDecimal',
+      `Must have maximum ${MAX_DECIMALS - 1} decimal digits`,
```
claudiaHash, line 106: Got an  error on publish for 5 decimals on swap fee. It worked fine with 4 decimals

mihaisc: `MAX_DECIMALS`  is 5 so, 5-1 =4  . i think  `MAX_DECIMALS`   should be 6, no idea why i set it to 5

claudiaHash: should i change it in the constants.ts file?

mihaisc: yes please

src/components/Publish/_validation.ts
```diff
@@ -60,7 +60,7 @@ const maxDecimalsValidation = new RegExp(
   '^\\d+(\\.\\d{1,' + MAX_DECIMALS + '})?$'
 )
 const swapFeeMaxDecimalsValidation = new RegExp(
-  '^\\d+(\\.\\d{1,' + (MAX_DECIMALS - 1) + '})?$'
+  '^\\d+(\\.\\d{1,' + (MAX_DECIMALS - 2) + '})?$'
```
mihaisc, line 63: why is it -2?

src/components/Publish/_validation.ts
```diff
@@ -103,7 +103,7 @@ const validationPricing = {
     .max(10, 'Maximum is 10%')
     .test(
       'maxDigitsAfterDecimal',
-      `Must have maximum ${MAX_DECIMALS - 1} decimal digits`,
+      `Must have maximum ${MAX_DECIMALS - 2} decimal digits`,
```
mihaisc, line 106: same as above

claudiaHash: Publishing asset doesn't work for more than 4 decimals in case of the swap fee field

mihaisc: ah, you are right, there is a min value

mihaisc (approved):

PR #1226, opened 2022-03-21
fix preview nft on step preview in publish

### LINKED ISSUE #1215: Publish preview shows Ocean logo instead of NFT preview
During publish on Preview screen the Ocean logo is shown where the NFT image should be:

Screen Shot 2022-03-18 at 13 32 10

The NFT preview shown on first Metadata screen should be shown on the Preview screen too.

EnzoVezzaro:
Fixes #1215 

Changes proposed in this PR:

- added preview nft on step preview in publish


kremalicious:
works great for intended issue, but breaks all asset details views. 

On top of that we have to be care fully with using `const { values } = useFormikContext<FormPublishData>()` within any component which will be rendered outside the publish page. That one case we do it already was meant to be replaced somehow as this will produce console warnings when those components are rendered within asset details, as there the publish form context is not present. If there's no easy other way it can be fine introducing another occurrence of this as it "just" produces warnings, no errors, but should be documented like in the other occurrence with code comments at least.

EnzoVezzaro:
I changed my implementation with something more "dumb" to avoid any rendering issue. 

I have some doubt tough as I'm adding the Data NFT image information into the asset.accessDetails (this to avoid having to change asset.metadata<Metadata> interface as it would require adding a new field for the NFT image, in my opinion this should be optimal result) object to then simply showing the image. This should solve any issues on the rendering sides, but maybe we should open another issue in ocean.js to change the `Metadata` interface to add this image, do you agree @kremalicious ?

kremalicious:
great, thanks for looking into alternatives. I guess using `asset.nft.tokenURI` for preview too is a bit overkill to construct from initial `image_data` (or maybe not?), so would say using this dummy `asset.accessDetails` only in the preview is kinda fine. We should then be more transparent to the next developer and in `MetaMain/index.tsx` explain in code comment why we try to get `asset?.accessDetails?.dataImage`.

On other hand, the more we hack around to not having to use `useFormikContext()`, it seems that would be the most clean way in terms of code we write here, like if `values` is defined we are on publish preview. So far I have not seen any more disadvantages of using context hook within component which misses the context, and there is demand for conditional context checking, see https://github.com/jaredpalmer/formik/issues/2338

So maybe another lazy way of solving this would be to figure out suppressing this `Warning: Formik context is undefined, please verify you are calling useFormikContext() as child of a <Formik> component`

kremalicious:
haha sorry, you see that I'm also undecided but yes, using `useFormikContext()` seems most clean as we do not have to introduce any workarounds, and only downside seems to be that warning messages.

Basically we need a pattern for when we want to use in a component conditionally either a value from the publish form context, or some passed value. At the linked issue there is an example of creating a Higher Order Component to do just that but didn't look into that further yet

EnzoVezzaro:
working on suppressing `Warning: Formik context is undefined, please verify you are calling useFormikContext() as child of a <Formik> component`

kremalicious (approved):
works great so let's just run with it

Issue #1219, opened 2022-03-21
No error message for swapping more than max tokens


jamiehewitt15:
### Describe the bug
There is no error message shown in the UI when trying to swap more than the maximum allowed. There is just an error in the console. Ideally there would be some form validation and a pop-up error message if people try to swap more than the maximum available datatokens.

### To Reproduce
Go to [asset](https://v4.market.oceanprotocol.com/asset/did:op:cf94d27ae9e0cc7a420f19884609afd76a26c5c9d0fbc1fc877235eed396f5f6) trade tab. Try to swap 100 Ocean tokens. An error will be shown in the console but nothing shown to the user in the UI

![image](https://user-images.githubusercontent.com/20739535/159240768-5976dc00-4644-416f-a813-b4d2c0401606.png)

EnzoVezzaro:
Probably here there's more a communication issue. When you enter a value over the allow to spend, the UI (aka Swap button) will get disabled. It turn enabled when you enter a valid value below "maximun to spend" value.

https://user-images.githubusercontent.com/13741335/159274929-5bcc623d-739c-4b9c-b4a7-f34f1f56355f.mp4

 It happens the same if you try to over spend the datatoken. Basically, it a dynamic validation that checks every time you enter a value, but if the value is not valid it will just disabled the "Swap" button.

If the button is enabled with a value over the "Maximun to spend" this is an issue, and we need to look more into it as I couldn't replicate this. Instead, if it's more a communication issue (the user doesn 't know why the button is disabled), then we can change a little bit the UI to reflect an issue on the value entered.

kremalicious:
just thought dumping, shouldn't a `max` attribute on that input field we fill with that max value trigger some formik error in this case?

EnzoVezzaro:
yes, it does. It's actually validating the values correctly. When the input is over the `maximun to spend` (is also set as `max` attribute in the input field), the "Swap" button get disabled correctly (the value is getting validated). The problem is that disabling the "Swap" button doesn't seem to be enough to let know the user that his value is not allowed. 

Maybe what we can do is, besides disabling the "Swap" button, marking the input field as error field to let the user know there's something wrong with the value he entered. As an exaggeration (in my opinion is too much), we could also show an Alert banner.

kremalicious:
> yes, it does. It's actually validating the values correctly.

ah cool, then that was my point, the bug then is that the validation rules are not output properly in the established error state, which is this as an example:

[...]

PR #1230, opened 2022-03-22
Fix No error message for swapping more than max tokens

EnzoVezzaro:
Fixes No error message for swapping more than max tokens #1219.

Changes proposed in this PR:

- added Error component with field validation
- moved Error (field) component in shared components


src/components/Asset/AssetActions/Trade/TradeInput.tsx
```diff
@@ -61,11 +62,15 @@ export default function TradeInput({
             form={form}
             value={`${field.value}`}
             onChange={(e: ChangeEvent<HTMLInputElement>) => {
-              handleValueChange(name, Number(e.target.value))
-              validateForm()
               handleChange(e)
+              handleValueChange(name, Number(e.target.value))
+              // timeout needed to avoid validating the wrong (pass) value
```
kremalicious, line 67: cool trick, remember we have some `lodash.debounce` usage in our codebase so search for that and mot likely this would be better package to do what you're trying to do here. Cause then we do not have to worry to clear that timeout anywhere

kremalicious (approved):
Confirmed this solves linked issue nicely. Up to you deciding if `setTimeout` or `lodash.debounce` works better, result should be the same

src/components/Asset/AssetActions/Trade/TradeInput.tsx
EnzoVezzaro: great. Will do. 👍🏽

EnzoVezzaro:
> 

Nice 👽 changed setTimeout with debounce to be aligned with codebase 👌🏽

Issue #1210, opened 2022-03-17
broken links in the Pool transactions history list


EnzoVezzaro:
**Describe the bug**
Issues in “Your Pool Transaction” history dropdown:
- the link of the swaps or adding transactions sometimes are broken (see video)
- the history list doesn’t refresh if the user has no actions associated (if you change from an account with actions to one without actions, you can still see the previous account’s history instead of clearing the dropdown)

**To Reproduce**
- Go to: https://market-git-v4-oceanprotocol.vercel.app/asset/did:op:958e879d468d42349bb3998e7e1a877bf0343ab2c8673862187a4a36929f592e
- Add liquidity
- Swap tokens
- Open “Your Pool Transactions” dropdown
- Click on the links (Sometimes the links for adding liquidity are broken, other times the swap links are)
Asset: https://market-git-v4-oceanprotocol.vercel.app/asset/did:op:958e879d468d42349bb3998e7e1a877bf0343ab2c8673862187a4a36929f592e

**Screenshots**

https://user-images.githubusercontent.com/13741335/158831299-e7ed6c06-a21d-4eea-bee3-2999f5a1a906.mp4

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Brave
 - Version [1.36.109](https://brave.com/latest/)

PR #1251, opened 2022-03-24
Fix missing pool creation tx

### LINKED ISSUE #1209: pool creation tx are missing from pool transactions lists
Consider this pool which was created by currently connected user:

Screen Shot 2022-03-17 at 14 09 34

And 50 OCEAN were sent into that pool upon creation:
https://rinkeby.etherscan.io/tx/0x9465e0f040323faedfa94ea0e938cab382c9a9ea60c7ba831ef2e868c6c3ee5e

So the pool transactions list at very end should include an entry `Created pool with 50 OCEAN`. This applies to pool transactions list on profile page too. Worth double checking if pool exit tx are missing too

KatunaNorbert:
Fixes #1209 .

### DIFF
--- a/src/components/@shared/PoolTransactions/index.tsx
+++ b/src/components/@shared/PoolTransactions/index.tsx
@@ -22,7 +22,7 @@ const txHistoryQueryByPool = gql`
     poolTransactions(
       orderBy: timestamp
       orderDirection: desc
-      where: { user: $user, pool: $pool }
+      where: { pool: $pool }
       first: 1000
     ) {
       baseToken {


KatunaNorbert:
Is there a case where datatoken address differs from one transaction to another?
From what I see it's the same, and we expect to get a different asset for each tx datatoken but instead we get only one:
https://github.com/oceanprotocol/market/blob/v4/src/components/%40shared/PoolTransactions/index.tsx#L183
So there is no `ddoList[1], ddoList[2]` and that's why networkId is set as undefined resulting transaction link to fail.

mihaisc:
In that case it means `getAssetsFromDtList` is failing for some reason.

KatunaNorbert:
The filter it's working, the issue is that we are filtering by the the same id that's we are getting only 1 result. 
But it's not directly related to this PR and saw that it was handled over #1210

mihaisc:
So from what i can see this is working , i have pool create.
image
Is there an issue left regarding this?

KatunaNorbert:
No, all good from my side

mihaisc (approved):

PR #1256, opened 2022-03-25
Move auto-changing providerUrl

### LINKED ISSUE #1254: Provider endpoint sometimes doesn't update in the publish form
Went through steps 1-4 on rinkeby and on step 5 before hitting submit switched to mumbai. We still have the rinkeby provider.
image
[Sub link](https://v4.subgraph.mumbai.oceanprotocol.com/subgraphs/name/oceanprotocol/ocean-subgraph/graphql?query=%7B%0A%20%20nft(id%3A%220xe55b57c4894d3fea69ad6af39be8eac70379ac77%22)%7B%0A%20%20%20%20id%2C%0A%20%20%20%20providerUrl%2C%0A%20%20%20%20%20symbol%0A%20%20%7D%0A%7D)  [tx](https://mumbai.polygonscan.com/tx/0x49e9a5aaa5e42a3c7c058ef19a3da6ce5f5a733d95bfbda89225bd6ceeb0e164)
@jamiehewitt15  also got this but he didn't change the network during publish  : [sub link](https://v4.subgraph.mumbai.oceanprotocol.com/subgraphs/name/oceanprotocol/ocean-subgraph/graphql?query=%7B%0A%20%20nft(id%3A%220xa86328d8e330fc6f5499364e830540a8622e1d32%22)%7B%0A%20%20%20%20id%2C%0A%20%20%20%20providerUrl%2C%0A%20%20%20%20%20symbol%0A%20%20%7D%0A%7D)
 [nft tx](https://mumbai.polygonscan.com/tx/0x7f427eb3f33b3618e42d213405e25ede33ad84f1918d3d1998adb21344ca5441)

When this happens `Aquarius` will obviously not index it, and you will be unable to order anyway

kremalicious:
- fixes #1254

Was only active when _Access_ screen was rendered, so moved into `Steps.tsx` which holds all our auto effects across all publish screens.

Additional tweaks:

- when user dismisses existing default provider, then changes network: nothing will happen so no custom provider URL is being overwritten
- when clicking Default Provider action, now grabbing the default provider URL from ocean config based on user wallet network, and if no wallet is connected grabbing this from initialValues as before
- additionally check if fetched provider is compatible with user's wallet network

### DIFF
--- a/src/components/@shared/FormFields/Provider/index.tsx
+++ b/src/components/@shared/FormFields/Provider/index.tsx
@@ -44,7 +44,7 @@ export default function CustomProvider(props: InputProps): ReactElement {
   }
 
   function handleFileInfoClose() {
-    helpers.setValue({ url: '', valid: false })
+    helpers.setValue({ url: '', valid: false, custom: true })
     helpers.setTouched(false)
   }
 
@@ -55,8 +55,7 @@ export default function CustomProvider(props: InputProps): ReactElement {
     const providerUrl =
       oceanConfig?.providerUri || initialValues.services[0].providerUrl.url
 
-    console.log(providerUrl)
-    helpers.setValue({ url: providerUrl, valid: true })
+    helpers.setValue({ url: providerUrl, valid: true, custom: false })
   }
 
   return field?.value?.valid === true ? (
--- a/src/components/@shared/FormFields/Provider/index.tsx
+++ b/src/components/@shared/FormFields/Provider/index.tsx
@@ -8,8 +8,11 @@ import Button from '@shared/atoms/Button'
 import { initialValues } from 'src/components/Publish/_constants'
 import { LoggerInstance, ProviderInstance } from '@oceanprotocol/lib'
 import { FormPublishData } from 'src/components/Publish/_types'
+import { getOceanConfig } from '@utils/ocean'
+import { useWeb3 } from '@context/Web3'
 
 export default function CustomProvider(props: InputProps): ReactElement {
+  const { chainId } = useWeb3()
   const [field, meta, helpers] = useField(props.name)
   const [isLoading, setIsLoading] = useState(false)
   const { setFieldError } = useFormikContext<FormPublishData>()
@@ -47,7 +50,13 @@ export default function CustomProvider(props: InputProps): ReactElement {
 
   function handleDefault(e: React.SyntheticEvent) {
     e.preventDefault()
-    helpers.setValue(initialValues.services[0].providerUrl)
+
+    const oceanConfig = getOceanConfig(chainId)
+    const providerUrl =
+      oceanConfig?.providerUri || initialValues.services[0].providerUrl.url
+
+    console.log(providerUrl)
+    helpers.setValue({ url: providerUrl, valid: true })
   }
 
   return field?.value?.valid === true ? (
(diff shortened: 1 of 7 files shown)


EnzoVezzaro:
Now when changing the network, the default provider never kicks in. Is this the intended result? I think is quite useful, maybe we can check if the provider is custom before blocking this functionality, what you think?

https://user-images.githubusercontent.com/13741335/160140777-437f6747-6625-4da5-a38f-55ea80101e0c.mp4

mihaisc:
Can't seem to replicate the issue Enzo is running into

kremalicious:
also getting mixed results but golden path works way better with this PR. Specifically there seems to be a mismatch between what's in the form context data (which is always correct, and default provider kicks in for user journey custom provider → change network), and what we display in UI (upon custom provider → change network UI does not pick up new provider value from form context data).

So yes, to cover all this we could disable any reaction to user network change when in form context data custom provider is used for which easiest way to detect that would be to add some new key to the form data

EnzoVezzaro:
> Now when changing the network, the default provider never kicks in. Is this the intended result? I think is quite useful, maybe we can check if the provider is custom before blocking this functionality, what you think?
> 
>  Publish.-.Ocean.Market.-.25.March.2022.mp4

I can't replicate this either, probably it has been fixed. Now in [V4](https://v4.market.oceanprotocol.com/publish) if you change network, the provider link stays the same. I found another issue though; if you change network and then press "Use Default Provider" link, it always use mainnet endpoint. I don't know if it's intended.

https://user-images.githubusercontent.com/13741335/160614159-96117aae-8cfd-44eb-b791-9ac159cbfbc2.mp4

kremalicious:
> if you change network and then press "Use Default Provider" link, it always use mainnet endpoint

good point, I'll change that. On that button click we simply grab the `providerUri` from the form's `initialValues`

kremalicious:
alright, think I watched all use cases. So now when user dismisses existing default provider, and then change network nothing will happen so no custom provider URL is being overwritten.

When then clicking _Default Provider_ action, now grabbing the default provider URL from ocean config based on user wallet network.

mihaisc:
Check out this gem
providerBug

Before i just switched between networks and removed default provider

kremalicious:
yeah, should have #1279 in here first which might already solve some things so making this a draft for now

kremalicious:
bump for review

EnzoVezzaro:
I'm still experiencing the issue with the 'Use Default Provider' button. If I change network to Rinbeky let's say, I'd expect the default provider to be Rinbeky's default provider instead of Mainnet's (https://v4.provider.mainnet.oceanprotocol.com). 

Basically, I get always `https://v4.provider.mainnet.oceanprotocol.com` if I click on 'Use Default Provider', even when I'm on testnet chains.

mihaisc:
@EnzoVezzaro your wallet is not connected 
image

mihaisc:
Seems to be ok, one extra step, when we validate the provider we can also check that the provider has the same chainId as the network. When you do a http get to the provider url you get a json with various data , the first one is `"chainId":1287`.  Doing this check we can prevent a user using a rinkeby provider on polygon , for example.

EnzoVezzaro:
> @EnzoVezzaro your wallet is not connected image

you're right. Ignore my comment. It's working as expected.

kremalicious:
Like so, preventing provider from being added to form state altogether if mismatch between provider and user chainId:

Screen Shot 2022-04-07 at 13 54 15

kremalicious:
Technically correct™ as user's `chainId` is undefined but valid point from user side as now no provider can't be fully checked without wallet connected which is now side effect of that chainId check. In order to not introduce more logic or blocking provider check without wallet (and then having to show a respective message) would opt for simply falling back to our fallback network `1`. So with that adding any mainnet provider would work without wallet

EnzoVezzaro (approved):
LGTM 👍🏽 great work!

PR #1261, opened 2022-03-28
Fix text allowed on numeric input

### LINKED ISSUE #1238: Text input allowed for trade inputs
Text input is allowed for numeric inputs in trade tab, on **Firefox** . The error message, 'Required',  is not relevant.

Screenshot from 2022-03-23 12-48-53

KatunaNorbert:
Fixes #1238

### DIFF
--- a/src/components/@shared/FormInput/InputElement.tsx
+++ b/src/components/@shared/FormInput/InputElement.tsx
@@ -159,6 +159,15 @@ export default function InputElement({
             disabled={disabled}
             {...field}
             {...props}
+            onKeyPress={
+              type === 'number'
+                ? (event) => {
+                    if (!/[0-9.]/.test(event.key)) {
+                      event.preventDefault()
+                    }
+                  }
+                : props.onKeyPress
+            }
           />
           {postfix && (
             <div className={cx({ postfix: true, [size]: size })}>{postfix}</div>
@@ -172,6 +181,15 @@ export default function InputElement({
           disabled={disabled}
           {...field}
           {...props}
+          onKeyPress={
+            type === 'number'
+              ? (event) => {
+                  if (!/[0-9.]/.test(event.key)) {
+                    event.preventDefault()
+                  }
+                }
+              : props.onKeyPress
+          }
         />
       )
   }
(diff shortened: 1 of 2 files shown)


kremalicious:
Fascinating, could not believe browser vendors have still not agreed on how to interpret [the spec](https://html.spec.whatwg.org/multipage/input.html#number-state-(type=number)) yet: https://bugzilla.mozilla.org/show_bug.cgi?id=1398528#c3

Current way in this PR could be done right on the `DefaultInput`, my only concern would be using our own `onKeyPress` conflicts with some Formik validation when that input is used in a Formik form so that would be something to look out for. If that's the case we would rather have to do the validation with Yup on specific input components

KatunaNorbert:
Don't think Yup validation fixes it because the value it's not changing when you are introducing text, but it still appears inside the input

src/components/@shared/FormInput/InputElement.tsx
```diff
@@ -159,6 +159,15 @@ export default function InputElement({
             disabled={disabled}
             {...field}
             {...props}
+            onKeyPress={
+              type === 'number'
+                ? (event) => {
+                    if (!/[0-9.]/.test(event.key)) {
```
claudiaHash, line 165: We might consider putting the regex in a constant, something like  `numberInputValidation` and use it in both places

EnzoVezzaro: Also, do we really need to use onKeyPress? wouldn't a `pattern` attribute on input work in this case?

kremalicious: also could not believe it but yes, `pattern` does not prevent users from entering letters, it just triggers the `:valid/:invalid` pseudo classes on the input.

In any case, Claudia is right with reducing duplication, but [as said before](https://github.com/oceanprotocol/market/pull/1261#issuecomment-1080499602), if we decide to put this on every input, then it should be set on `DefaultInput`.

Also with "number" in this case we mean all numbers users can enter and this regex is only allowing integers, or decimals in US notation. We want all the notations

kremalicious:
closing, as we discovered more complex than initially expected, see https://github.com/oceanprotocol/market/issues/1238#issuecomment-1090389840

PR #1258, opened 2022-03-25
fix remove liquidity issues

EnzoVezzaro:
- Fixes #1236 
- Fixes #1203 

Changes proposed in this PR:

- reset slider after transaction
- fix NaN error on max button
- fix issue on remove button hover

PR #1265, opened 2022-03-28
Disable pool fetching interval

### LINKED ISSUE #1152: Pool tab not updating when changing account
When you are in add/remove liquidity screen and you change your account the values don't update. When you go back to main pool tab  the data is still from the old account.
Also, because of this, when you remove all your liquidity you can still see the Remove button

### LINKED ISSUE #1248: Pool Tab: The "Your Locked Value" is wrongly set to 0 after a refresh
**Describe the bug**
As you can see in this video: https://drive.google.com/file/d/1eUrzKJqsZU7PanTPvo46HRbZwIkhGtHx/view?usp=sharing, if you refresh the pool liquidity page the value will go to 0 after a few seconds and the "Remove" button will be removed as well.

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Chrome
 - Version: 99.0.4844.74

kremalicious:
Sledgehammer temporary fix. Numbers should still update correctly after add/remove with this change.

- remove interval fetching
- refetch specifically on every back button click on add/remove screen, theoretically should be unnecessary as adding/removing liquidity methods refetch at their end but did not always work for whatever reason
- switch `urql`'s `requestPolicy` to `network-only`
- fixes #1248
- fixes #1152 
- should help with #1258

Adding back the functionality tracked in #1266


mihaisc:
Better but there still semes to be some caching involved

mihaisc:
We need to make sure it's not a subgraph issue

kremalicious:
Just added a dumb manual refresh action for now which also helps a bit in debugging. Se we can move forward with this PR and handle adding back interval fetching in new PR with #1266

Screen Shot 2022-03-29 at 13 14 10

mihaisc (approved):

PR #1273, opened 2022-03-29
created HOC to avoid rendering formit context when is not necessary

### LINKED ISSUE #1272: fix Formik context warning
**Is your feature request related to a problem? Please describe.**
Following up on this [thread](https://github.com/oceanprotocol/market/pull/1226#issuecomment-1080727267): when you are inside an asset, formit throws this warning as there's no formit context from which calling the `useFormitContext` hook.

Message:
`Warning: Formik context is undefined, please verify you are calling useFormikContext() as child of a <Formik> component`

We need to check dynamically and call the hook only if necessary (on preview page). The result should cleared any warning in console.

Screen Shot 2022-03-29 at 9 36 54 AM

EnzoVezzaro:
Fixes #1272.

Changes proposed in this PR:

- Added HOC for dynamic context

### DIFF
--- a/src/components/Asset/AssetActions/index.tsx
+++ b/src/components/Asset/AssetActions/index.tsx
@@ -14,24 +14,20 @@ import { getOceanConfig } from '@utils/ocean'
 import { useCancelToken } from '@hooks/useCancelToken'
 import { useIsMounted } from '@hooks/useIsMounted'
 import styles from './index.module.css'
-import { useFormikContext } from 'formik'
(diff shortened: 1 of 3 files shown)


EnzoVezzaro:
closed as makes no sense to keep it opened at this stage

kremalicious:
but cool, thanks for looking into it as we now know how much effort it would be to use this HOC in place of current mechanism. Just in case we have to introduce it a third time which would be then a good time to revisit this here

PR #1279, opened 2022-03-29
Warn about publishing html files, unified file error states

### LINKED ISSUE #1246: The download fails for assets published that are not publicly available
The publish form allows asset publication even if the asset is not publicly reachable. This creates an issue when downloading the file. 

https://v4.market.oceanprotocol.com/asset/did:op:f66b072ec293aa76a29668df9ba7dbaf2cb0bb610f3475f8ebeebcfb731749b8

Video with the issue: https://drive.google.com/file/d/1I9lOBTEgI4iLDXh72W9rCEDQGGHgNDXJ/view?usp=sharing

kremalicious:
- closes #1246
- if file is valid and content type ends up as `html`, show a warning.
- cleanup for less repetitive markup
- cleanup for less code doing the same thing with less logic in file check 

Screen Shot 2022-03-30 at 14 34 19
 
- cleaned up and visually unified error states, resulting in file only being "added" when `valid: true`
  - while each error state (toast, input error, file added but show as error) was kinda nice, using 3 different error UI for 3 different errors on the same thing seemed a bit overkill
  - so left the only intended UI for all errors in the publish form:
  - provider not reachable/no network etc. (was shown in toast before with no error state on input at all):
  Screen Shot 2022-03-29 at 22 37 54
  
  - file response is `valid: false` (was a new error state before, added in #1221):
  Screen Shot 2022-03-29 at 23 00 21

### DIFF
--- a/src/components/@shared/FormFields/FilesInput/index.tsx
+++ b/src/components/@shared/FormFields/FilesInput/index.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, useCallback, useEffect, useState } from 'react'
+import React, { ReactElement, useState } from 'react'
 import { useField, useFormikContext } from 'formik'
 import { toast } from 'react-toastify'
 import FileInfo from './Info'
@@ -7,6 +7,7 @@ import { InputProps } from '@shared/FormInput'
 import { initialValues } from 'src/components/Publish/_constants'
 import { getFileUrlInfo } from '@utils/provider'
 import { FormPublishData } from 'src/components/Publish/_types'
+import { LoggerInstance } from '@oceanprotocol/lib'
 
 export default function FilesInput(props: InputProps): ReactElement {
   const [field, meta, helpers] = useField(props.name)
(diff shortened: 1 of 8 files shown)


src/components/@shared/FormFields/FilesInput/index.tsx
```diff
@@ -7,55 +7,30 @@ import { InputProps } from '@shared/FormInput'
...
-        }
-      }
-
-      validateUrl()
-    },
-    [helpers, values.services]
-  )
-
-  useEffect(() => {
-    // try load from initial values, kinda hacky but it works
```
kremalicious, line 45: if I got this right, we only had this effect for refilling this field from saved `localStorage` form values but we don't have this right now so removed it all. If anybody remembers another reason we had this, let me know

EnzoVezzaro:
Found potencial regression on Provider URL validation (not sure if it's related, worth mention it)

Tested Links: 
- https://bafybeifrwhcjy2zp4ckif2ent5os27lm3bcijrqeihh6mjhiv2kzol7z6a.ipfs
- https://drive.google.com/file/d/1r0P_HRu8ZRTlxs-AyJJOzW9F6r3ID

https://user-images.githubusercontent.com/13741335/160809881-dcd41f54-c2e4-4ba3-bf74-3c07fa37554d.mp4

The second thing is more related to the use of Google Drive (in this case).

If I have a PDF (or Excel or any format really) in drive and use the standard shared link as "File" url, the UI actually detects it as HTML
For example: 
Screen Shot 2022-03-30 at 6 27 58 AM

I imagine there are limitations on what we can do about it, as I guess it's related to combination browser / drive. But the message it's a bit confusing for a regular user (that maybe doesn't even know what's html ) in this cases. Can we do anything about it?

mihaisc:
Well this is a compromise, we can't cover all the drives/shares. You are talking now just about google drive. Realistically we can't cover all cases and we shouldn't.  Maybe at some point we can have some better integrations but i feel it's way out of scope.

kremalicious:
> If I have a PDF (or Excel or any format really) in drive and use the standard shared link as "File" url, the UI actually detects it as HTML

well yes, this is the preview page of Google Drive and the file download with this will fail. Open that link in a browser and you will get something like this:

Screen Shot 2022-03-30 at 11 50 10

Using the link behind this icon might work better:

Screen Shot 2022-03-30 at 11 50 40

> Found potencial regression on Provider URL validation (not sure if it's related, worth mention it)

yes, thanks for noticing, indeed related so I'll take a look at it

mihaisc:
Asking the user to fish for obscure links from gdrive i fell we are complicating things more than needed. We have some mechanisms in place to make sure the data is valid but it is not our responsibility. This is decentralized, it's the publishers responsibility to have a working url. Investing more time seems like a waste. The more i think even this message is kinda weird. So i paste the link from a drive and i get this message, what do i do? Am i sure? Will i have issues? Why am i getting this warning?

kremalicious:
same error check & passthrough now in provider too. And modified copy making it more actionable:

Screen Shot 2022-03-30 at 14 34 19

> Investing more time seems like a waste.

Agree, the warning is right now best we can do, only proper way would be to integrate file pickers for various hosting and communicate with their APIs directly to e.g. get signed links, like from Azure or AWS.

src/components/@shared/FormFields/Provider/index.tsx
```diff
@@ -1,41 +1,47 @@
...
-      helpers.setValue({ url, valid: isValid })
-      helpers.setError(undefined)
+      setIsLoading(true)
+      const isValid = await ProviderInstance.isValidProvider(field.value.url)
+
+      // error if something's not right from response
+      // TODO: Figure out way to detect failed response, and throw that before
+      // this check.
+      // No way to detect a failed response with ProviderInstance.isValidProvider,
+      // making this error show up for multiple cases it shouldn't, namely network
```
mihaisc, line 28: This is not accurate and overcomplicating things. If a provider is down then it is not a valid provider.  This is not a debug software for your web services/infrastructure.

mihaisc:
Like Enzo said the error states are a bit off.

image
Way better than v4 main

image

src/components/@shared/FormFields/Provider/index.tsx
kremalicious: It's the very basic of web requests to know why they fail and we destroyed that here. If user's network is down, saying the entered provider is not valid is simply wrong but yes, considering custom provider is advanced use case anyway we don't need to bother

PR #1280, opened 2022-03-30
Display token approval message on loading

### LINKED ISSUE #1253: Pool action displayed instead of approval
When doing one of the following actions: add, remove liquidity or trade inside a pool, when approving the ocean token it says that the action is taking place instead of the approval.

For example I'm doing an approval but the following message is displayed:
Screenshot 2022-03-25 at 11 18 27

KatunaNorbert:
Fixes #1253.

### DIFF
--- a/src/components/@shared/TokenApproval/index.tsx
+++ b/src/components/@shared/TokenApproval/index.tsx
@@ -11,7 +11,8 @@ export default function TokenApproval({
   amount,
   tokenAddress,
   tokenSymbol,
-  setSubmitting
+  setSubmitting,
+  setIsTokenApproved
 }: {
   actionButton: JSX.Element
   disabled: boolean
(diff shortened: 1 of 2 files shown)


KatunaNorbert:
Not the best approach since now we have 2 states holding the `isTokenApproved` values but I don't see another way without refactoring.
Other option could be as we discussed to just remove the messages and only show a spinner.

claudiaHash (approved):
Yes, seems to be a fair compromise

src/components/@shared/TokenApproval/index.tsx
```diff
@@ -44,6 +46,9 @@ export default function TokenApproval({
       setTokenApproved(
         new Decimal(allowanceValue).greaterThanOrEqualTo(new Decimal(amount))
       )
+    setIsTokenApproved(
```
kremalicious, line 49: What's the difference to setTokenApproved from above?

kremalicious: Saw the comment now, so far down on mobile compared to the code

EnzoVezzaro (approved):

PR #1283, opened 2022-03-30
Undefined graph labels fix

### LINKED ISSUE #1277: Graph labels are undefined
image

A picture is worth a thousand words

claudiaHash:
Fixes #1277  .

Changes proposed in this PR:

-  removed `,` character form price string so forming price doesn't return `NaN`

### DIFF
--- a/src/components/@shared/Price/PriceUnit.tsx
+++ b/src/components/@shared/Price/PriceUnit.tsx
@@ -6,7 +6,7 @@ import { useUserPreferences } from '@context/UserPreferences'
 import Badge from '@shared/atoms/Badge'
 
 export function formatPrice(price: string, locale: string): string {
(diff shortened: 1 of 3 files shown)


EnzoVezzaro (approved):
works great! 👏🏽

kremalicious:
As this touches a component responsible for displaying every number throughout our app let's double check.

I'm concerned about different locales, where some users will use a `,` as the decimal point where others use this as thousands separator. Are we sure this change does not make e.g. in de-DE a `3,5` (which is `3.5` in en-US) into `35`?

From that perspective a more safe way would be to convert that price string input with something else than `Number()` which considers above concerns.

Did not check out which numbers this component receives and from where so might be not a concern at all when numbers passed are always coming from subgraph or so

claudiaHash:
works for me either, on Firefox & Chrome. thanks for fixing this

PR #1284, opened 2022-03-30
c2d show price with fee, exclude provider fee

soonhuat:
Fixes # .

Changes proposed in this PR:

- Show dataset price with fee
- Show total [dataset + algo] price with fee (not sure if should show price with fee for each algorithm in list)
- NOTE that in `FormComputeDataset` does check if has DT and exclude price, however now there is swap and order fee, therefore if has DT should only exclude swap fee, but not order fee, I guess more to be enhance on `accessDetailsAndPricing.getOrderPriceAndFees`, so of course didn't include part of this PR
- NOTE that provider fee also excluded for both dataset and algo


bogdanfazakas:
Still needs some polish, added an image below where we must add the price with the fees also on the algorithm and dataset so the math from the tooltip is still valid. Also some spinners would be nice when fetching the price with fees.
Screenshot 2022-03-31 at 09 37 52

bogdanfazakas (approved):

PR #1292, opened 2022-03-31
add custom codeclimate config

kremalicious:
Making Codeclimate useful again.

Changes from former defaults:

- `argument-count`: `4` → `6`
- `complex-logic`: `4` → `10`
- `file-lines`: `250` → `500`
- `method-complexity`: `5` → `10`
- `method-lines`: `25` → `100`
- `similiar-code`: `45` → `50`
- `identical-code`: `45` → `50`

for the last 2 items, see https://docs.codeclimate.com/docs/default-analysis-configuration#per-language-mass-threshold-defaults

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -20,7 +20,7 @@ checks:
       threshold: 20
   method-lines:
     config:
-      threshold: 100
+      threshold: 250
   nested-control-flow:
     config:
       threshold: 4
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,3 +1,4 @@
+# https://docs.codeclimate.com/docs/default-analysis-configuration
 # https://docs.codeclimate.com/docs/advanced-configuration
 
 version: '2'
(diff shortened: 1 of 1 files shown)


kremalicious:
great, when using a fresh `v4` branch analysis on CodeClimate with the defaults via UI, I get 173 issues, but with this override it suddenly produces 2702 issues although the changed thresholds should have reduced issues by a lot.

iu

kremalicious:
solved after conversation and great explanation from CodeClimate support, the `similar-code` & `identical-code` thresholds refer to a "mass threshold" while we were thinking they refer to the duplication count:  https://docs.codeclimate.com/docs/default-analysis-configuration#per-language-mass-threshold-defaults

Since the default for JS is `45` for both, we got hit with that many errors when we lowered this to `4`

Slightly increased this threshold and now we're taking:

Screen Shot 2022-05-16 at 11 38 08

EnzoVezzaro (approved):
👍🏽

mihaisc (approved):

PR #1294, opened 2022-03-31
Trade input accepts  decimals with leading 0

### LINKED ISSUE #1003: Decimals with leading 0 can't be typed in Trade
**Describe the bug**
 I want to buy 1.0 datatokens for use later. In trying to do this, encountered two issues:
- I can't specify a "minimum datatokens to receive" in TRADE
- I can't specify decimals in amounts of OCEAN or datatoken, which would have been a workaround
- I did find a workaround, by fiddling with number of OCEAN. But it shouldn't be so hard

**To Reproduce**
Step 1. Go to https://market.oceanprotocol.com/asset/did:op:7Bce67697eD2858d0683c631DdE7Af823b7eea38, and connect wallet. See image below for what you see.

Step 2. I want to buy 1.0 datatokens for use later. If I go to "USE" button, it's for immediate usage, but I don't want to use the datatoken immediately. So I go to "TRADE" button. I can trade OCEAN for a datatoken (eg QUICRA-0). But it only gives me a "maximum to receive", not a "minimum to receive". For my use case, I want a "minimum to receive" of 1.0 datatokens.  (I know that this is possible in Balancer pools, via `function swapExactAmountOut` where `tokenAmountOut` gets set to 1.0 datatokens. I guess this functionality wasn't exposed to the market yet.)

Step 2. I specified the field "maximum to receive" of 1 QUICRA-0, it tells me that "Minimum received" is 0.95. OK, that's close. So I try to change it from "1" to "1.05", but it auto-changes it back to 1. It won't let me enter any decimals for the number of datatokens. That's frustrating; it seems like a bug to me.

Step 3. I thought maybe I can tweak the number of OCEAN. So go into the OCEAN field above, and change it from 212.24 to other numbers, fiddling until I arrive at 224.1 OCEAN. OK, that works, it doesn't kick out the number of decimals there. It gives me about 1.05 QUICRA. So this is the workaround for now. But that only works for datasets that are relatively expensive 

**Expected behavior**
- [ ] I should be able to specify decimal values in the field for the # datatokens that I want. This should be a quick win, since it's already supported for # OCEAN.
- [ ] I should be able to specify a _minimum_ number of datatokens to receive, where under the hood it uses `swapExactAmountOut`.  This will take a bit more UI work to do a good job on.

Wrt priorities, given V4 I see that this can happen after V4 is shipped. 

screenshot

claudiaHash:
Fixes #1003  .

Changes proposed in this PR:

- fix trade input so it accepts decimals leading 0

### DIFF
--- a/src/components/@shared/FormInput/index.tsx
+++ b/src/components/@shared/FormInput/index.tsx
@@ -29,7 +29,7 @@ export interface InputProps {
   options?: string[]
   sortOptions?: boolean
   additionalComponent?: ReactElement
-  value?: string
+  value?: string | number
   onChange?(
     e:
       | FormEvent<HTMLInputElement>
--- a/src/components/Asset/AssetActions/Trade/TradeInput.tsx
+++ b/src/components/Asset/AssetActions/Trade/TradeInput.tsx
@@ -59,18 +59,18 @@ export default function TradeInput({
             min="0"
             prefix={item?.token}
             placeholder="0"
-            // field={field}
+            field={field}
             form={form}
+            disabled={!accountId || disabled}
+            additionalComponent={<Error meta={meta} />}
             value={`${field.value}`}
+            {...field}
             onChange={(e: ChangeEvent<HTMLInputElement>) => {
               handleChange(e)
               handleValueChange(name, Number(e.target.value))
               // debounce needed to avoid validating the wrong (pass) value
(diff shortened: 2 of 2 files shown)


kremalicious (approved):
surprisingly simple change, works flawlessly!

PR #1295, opened 2022-03-31
Pool components UI refactor

### LINKED ISSUE #1293: Move pool TVL number
Make it a new section under `Owner Value Locked` so it does not appear as being part of the chart

kremalicious:
Massive PR as all the markup within `components/AssetActions/Pool` was changed,m made sure to not touch any calculations.

- closes #1293
- new `PoolSection` component only dealing with layout and spacing
- highly reduce passing around props, replace with `usePool` hook in almost all components
- ...


mihaisc:
it this how it's supposed to look? aren't we wasting a lot of space and making the graph sooo small?
image

EnzoVezzaro:
Found small issue. If you're not connected, "Your Locked Value" percentage is `undefined`.

Screen Shot 2022-04-05 at 6 34 53 AM

if you logged in, it works.

EnzoVezzaro:
Now I see a difference between logged in / logged out user.

Not logged in (no percentage):
Screen Shot 2022-04-05 at 7 13 51 AM

Logged in (with percentage):
Screen Shot 2022-04-05 at 7 13 37 AM

I don't know if hiding the percentage was intended, if so, we need to do the same for the logged in user.

kremalicious:
> I don't know if hiding the percentage was intended

yeah was intended, not really opinionated if we should show nothing or 0, just decided quickly that without wallet no need to even show 0 as there's no way for it to change, so removed it. When wallet is connected it becomes active and shows 0.

So it's the distinction between `undefined` & `0` for something which only works when wallet is connected. This is the difference to the other sections

mihaisc:
Something is wrong in remove liquidity 
image
in v4: 
image

EnzoVezzaro (approved):

EnzoVezzaro (changes requested):
found issue on remove too

kremalicious:
thanks for noticing! Indeed was passing the wrong value in that `getMax` which then only calculated with the `totalPoolTokens`  instead of `totalPoolTokens`  & user pool shares. Should behave the same now as on `v4`

src/components/@shared/Token/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+        <figure
+          className={`${styles.icon} ${symbol} ${noIcon ? styles.noIcon : ''}`}
+        >
+          <Logo noWordmark />
+        </figure>
+        <PriceUnit price={balance} symbol={symbol} size={size} />
+      </div>
+      {conversion?.greaterThan(0) && (
+        <Conversion
+          price={conversion.toString()}
```
mihaisc, line 33: i think we should have a `.toDecimalPlaces` here

mihaisc (approved):
good to go, with that minor comment

mihaisc:
Any more comments/changes to this PR? i'd like to merge it

EnzoVezzaro (approved):
LGTM

PR #1302, opened 2022-04-04
Dynamic price publish input fields fixes

### LINKED ISSUE #1176: Confusing Publish "Dynamic" price experience
When Publishing an asset with "Dynamic" price, the user needs to provide 2 values: 
* the Ocean price
* the datatoken liquidity pool (Ocean token) value
The report between these two values is directly correlated to the "Precipitate Manatee Token" value that is automatically generated. 
If the proportion is not correct, an error is shown on the 3rd field and the "Continue" button is disabled. Is not very clear the values the user needs to add in order to be able to proceed. 
Maybe we can add an instruction message or display a better error message

### LINKED ISSUE #1314: Dynamic price publish input fields fixes
- Remove confusing message from datatoken input as it is disabled and user can't change it, it shouldn't be any messages displayed around it.
- Automatically set the ocean amount based on the price to the minimum value required to create the pool.
- Add min value to ocean amount input so user can't set an invalid amount reducing error messages.

### LINKED ISSUE #1340: Increase minimum datatokens for pools
When creating a pool we currently require a minimum of 50 datatokens to be created. I suggest to setting a new limit to 100 or 150. Thoughts?

KatunaNorbert:
- Fixes #1314
- Fixes #1340
- Addressing #1176

### DIFF
--- a/src/components/Publish/Pricing/Dynamic.tsx
+++ b/src/components/Publish/Pricing/Dynamic.tsx
@@ -78,7 +78,9 @@ export default function Dynamic({ content }: { content: any }): ReactElement {
           name="amountOcean"
           datatokenOptions={{ symbol: 'OCEAN', name: 'Ocean Token' }}
           weight={`${Number(weightOnOcean) * 10}%`}
-          min={(price * Number(weightOnOcean) * 10).toString()}
+          min={
+            price > 0 ? (price * Number(weightOnOcean) * 10).toString() : '50'
+          }
         />
         <Coin
           name="amountDataToken"
--- a/src/components/Publish/Pricing/Dynamic.tsx
+++ b/src/components/Publish/Pricing/Dynamic.tsx
@@ -10,6 +10,8 @@ import Price from './Price'
 import Decimal from 'decimal.js'
 import { useWeb3 } from '@context/Web3'
 import { FormPublishData } from '../_types'
+import { minAmountOcean } from '../_constants'
+import { MAX_DECIMALS } from '../../../@utils/constants'
 
 export default function Dynamic({ content }: { content: any }): ReactElement {
   const { networkId, accountId, balance } = useWeb3()
@@ -79,7 +81,13 @@ export default function Dynamic({ content }: { content: any }): ReactElement {
           datatokenOptions={{ symbol: 'OCEAN', name: 'Ocean Token' }}
           weight={`${Number(weightOnOcean) * 10}%`}
           min={
-            price > 0 ? (price * Number(weightOnOcean) * 10).toString() : '50'
+            price > 0
+              ? new Decimal(price)
+                  .mul(weightOnOcean)
+                  .mul(10)
+                  .toDecimalPlaces(MAX_DECIMALS)
+                  .toString()
+              : minAmountOcean.toString()
           }
         />
         <Coin
--- a/src/components/Publish/Pricing/Dynamic.tsx
+++ b/src/components/Publish/Pricing/Dynamic.tsx
@@ -78,6 +78,7 @@ export default function Dynamic({ content }: { content: any }): ReactElement {
           name="amountOcean"
           datatokenOptions={{ symbol: 'OCEAN', name: 'Ocean Token' }}
           weight={`${Number(weightOnOcean) * 10}%`}
+          min={(price * Number(weightOnOcean) * 10).toString()}
         />
         <Coin
           name="amountDataToken"
--- a/src/components/Publish/Pricing/Dynamic.tsx
+++ b/src/components/Publish/Pricing/Dynamic.tsx
@@ -10,7 +10,7 @@ import Price from './Price'
 import Decimal from 'decimal.js'
 import { useWeb3 } from '@context/Web3'
 import { FormPublishData } from '../_types'
-import { minAmountOcean } from '../_constants'
+import { initialValues } from '../_constants'
 import { MAX_DECIMALS } from '../../../@utils/constants'
 
 export default function Dynamic({ content }: { content: any }): ReactElement {
@@ -87,7 +87,7 @@ export default function Dynamic({ content }: { content: any }): ReactElement {
                   .mul(10)
                   .toDecimalPlaces(MAX_DECIMALS)
                   .toString()
-              : minAmountOcean.toString()
+              : initialValues.pricing.amountOcean.toString()
           }
         />
         <Coin
--- a/src/components/Publish/Pricing/Dynamic.tsx
+++ b/src/components/Publish/Pricing/Dynamic.tsx
@@ -10,7 +10,6 @@ import Price from './Price'
 import Decimal from 'decimal.js'
 import { useWeb3 } from '@context/Web3'
 import { FormPublishData } from '../_types'
-import { initialValues } from '../_constants'
 
 export default function Dynamic({ content }: { content: any }): ReactElement {
   const { networkId, accountId, balance } = useWeb3()
@@ -21,7 +20,6 @@ export default function Dynamic({ content }: { content: any }): ReactElement {
   const { dataTokenOptions } = values.services[0]
 
   const {
-    price,
     weightOnDataToken,
     weightOnOcean,
     swapFee,
--- a/src/components/Publish/Pricing/Dynamic.tsx
+++ b/src/components/Publish/Pricing/Dynamic.tsx
@@ -11,7 +11,6 @@ import Decimal from 'decimal.js'
 import { useWeb3 } from '@context/Web3'
 import { FormPublishData } from '../_types'
 import { initialValues } from '../_constants'
-import { MAX_DECIMALS } from '../../../@utils/constants'
 
 export default function Dynamic({ content }: { content: any }): ReactElement {
   const { networkId, accountId, balance } = useWeb3()
@@ -80,15 +79,6 @@ export default function Dynamic({ content }: { content: any }): ReactElement {
           name="amountOcean"
           datatokenOptions={{ symbol: 'OCEAN', name: 'Ocean Token' }}
           weight={`${Number(weightOnOcean) * 10}%`}
-          min={
-            price > 0
-              ? new Decimal(price)
-                  .mul(weightOnOcean)
-                  .mul(10)
-                  .toDecimalPlaces(MAX_DECIMALS)
-                  .toString()
-              : initialValues.pricing.amountOcean.toString()
-          }
         />
         <Coin
           name="amountDataToken"
--- a/src/components/Publish/Pricing/index.tsx
+++ b/src/components/Publish/Pricing/index.tsx
@@ -5,6 +5,7 @@ import Tabs from '@shared/atoms/Tabs'
 import { isValidNumber } from '@utils/numbers'
 import Decimal from 'decimal.js'
 import { FormPublishData } from '../_types'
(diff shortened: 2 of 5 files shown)


src/components/Publish/Pricing/Dynamic.tsx
```diff
@@ -78,6 +78,9 @@ export default function Dynamic({ content }: { content: any }): ReactElement {
           name="amountOcean"
           datatokenOptions={{ symbol: 'OCEAN', name: 'Ocean Token' }}
           weight={`${Number(weightOnOcean) * 10}%`}
+          min={
+            price > 0 ? (price * Number(weightOnOcean) * 10).toString() : '50'
```
kremalicious, line 82: no magical numbers please, where is the `50` coming from? Would assume this refers to the defaults we set, and they are set in `Publish/_constants.ts`. You can get the defaults easily via `initialValues` from the Formik context.

src/components/Publish/Pricing/index.tsx
```diff
@@ -28,6 +28,21 @@ export default function PricingFields(): ReactElement {
...
+  useEffect(() => {
+    if (type === 'fixed' || type === 'free') return
+
+    const amountOcean =
+      isValidNumber(weightOnOcean) &&
+      isValidNumber(price) &&
+      isValidNumber(weightOnDataToken) &&
+      price > 0
+        ? new Decimal(price).mul(new Decimal(weightOnOcean).mul(10))
+        : '50'
```
kremalicious, line 41: same here, see https://github.com/oceanprotocol/market/pull/1302/files#r841730936

src/components/Publish/Pricing/index.tsx
```diff
@@ -36,22 +51,16 @@ export default function PricingFields(): ReactElement {
...
       isValidNumber(price) &&
-      isValidNumber(weightOnDataToken)
+      isValidNumber(weightOnDataToken) &&
+      price > 0
         ? new Decimal(amountOcean)
             .dividedBy(new Decimal(weightOnOcean))
             .dividedBy(new Decimal(price))
             .mul(new Decimal(weightOnDataToken))
-        : 0
+        : '50'
```
kremalicious, line 60: same here, see https://github.com/oceanprotocol/market/pull/1302/files#r841730936

kremalicious:
I second Enzo in that the issue this PR claims to fix has lots more to it than what's currently in here. Can be fine to keep this PR small, but then just `addressing #1176`.

But raises good point as #1176 is becoming more of a brainstorming issue so nothing immediately actionable to actually close it right now. So for future better process just for this could be extracting small ideas from it into new small issues and then work on them.

src/components/Publish/Pricing/Dynamic.tsx
mihaisc: also please only use Decimal function to do the math . Also use `.toDecimalPlaces(MAX_DECIMALS)` at the end.

mihaisc:
The only other thing that makes sense to do in this PR is to move the validation from datatoken (it's a readonly field) to ocean 
image
In this image the error should be on ocean liquidity something like `Must be more or equal to 1000 so the pool has a reasonable minimum amount of datatokens` . And no validation should be on datatoken

src/components/Publish/Pricing/Dynamic.tsx
KatunaNorbert: I took the `50` from the yup validation schema where is also magically set, so I guess there should be a constant added for min values inside `_constants.ts` and that to be used in multiple places

src/components/Publish/_constants.tsx
```diff
@@ -118,3 +118,6 @@ export const algorithmContainerPresets: MetadataAlgorithmContainer[] = [
 ]
 
 export const initialPublishFeedback: PublishFeedback = contentFeedback
+
+export const minAmountOcean = 50
```
kremalicious, line 122: we declare them further down in the `initialValues`, no need to declare them here again. Just use the `initialValues` from the Formik context in any component you want

src/components/Publish/Pricing/index.tsx
```diff
@@ -28,6 +29,21 @@ export default function PricingFields(): ReactElement {
...
+      isValidNumber(price) &&
+      isValidNumber(weightOnDataToken) &&
+      price > 0
+        ? new Decimal(price).mul(new Decimal(weightOnOcean).mul(10))
+        : new Decimal(initialValues.pricing.amountOcean)
+
+    setFieldValue('pricing.amountOcean', amountOcean)
+  }, [price, weightOnDataToken, weightOnOcean, type, setFieldValue])
+
   // Always update everything when price value changes
```
kremalicious, line 47: `price` was removed from dependency array so this comment should be adapted. As there are now 2 effects, most likely "update everything" is not true anymore as both effects update different things. Which should be made clear in comments instead of duplicating the comments

src/components/Publish/Pricing/index.tsx
```diff
@@ -28,6 +29,21 @@ export default function PricingFields(): ReactElement {
...
   }
 
+  // Always update everything when price value changes
+  useEffect(() => {
+    if (type === 'fixed' || type === 'free') return
+
+    const amountOcean =
+      isValidNumber(weightOnOcean) &&
+      isValidNumber(price) &&
+      isValidNumber(weightOnDataToken) &&
```
kremalicious, line 39: why is `weightOnDataToken` checked for in this effect, and also in dependency array? I don't see it used in it at all

src/components/Publish/Pricing/Dynamic.tsx
```diff
@@ -78,6 +80,15 @@ export default function Dynamic({ content }: { content: any }): ReactElement {
           name="amountOcean"
           datatokenOptions={{ symbol: 'OCEAN', name: 'Ocean Token' }}
           weight={`${Number(weightOnOcean) * 10}%`}
+          min={
```
kremalicious, line 83: what is the goal of this? Do we really need this logic here when we additionally validate with Yup?

src/components/Publish/Pricing/index.tsx
KatunaNorbert: good point, haven't seen that

src/components/Publish/Pricing/index.tsx
KatunaNorbert: right, it should be removed

src/components/Publish/Pricing/Dynamic.tsx
KatunaNorbert: the purpose is to not let user entry a lower amount. The Yup validation would have shown an error but allow user to enter invalid ocean amount. 
Also if the ocean amount is invalid, the dataToken amount would become invalid too now that they are connected. So to prevent invalid dataToken amount too which will also need some messages I thought it would be better to limit the input to the min val.

kremalicious: good thinking, but `min` and `max` only prevents going into those number ranges with the stepper arrows, not actual typing, they are there for HTML-only form validation which will never kick in with Formik form fields. 

When typing number outside of range, I get this:

Screen Shot 2022-04-11 at 15 05 09

Which to me looks sufficient and it's clear faulty values in other fields are based on this errored field, so that's already great with this PR!
 
So the only implication of using `min` here would be that stepper respects them, and the input `:valid/:invalid` pseudo classes might get set which could be used in CSS but we don't, as we use js validation and use those classes then. So having some logic in codebase, which duplicates a calculation we do somewhere else, and which in end we don't really use seems unnecessary. So would opt to remove this logic again

kremalicious: came here to write that we need to move the Yup validation definition into the component to get access to dynamic values but I see you figured out a much better way already, great find with that custom validation method!

src/components/Publish/_validation.ts
```diff
@@ -78,11 +80,23 @@ const validationPricing = {
       (param) => maxDecimalsValidation.test(param?.toString())
     )
     .required('Required'),
-  amountDataToken: Yup.number()
-    .min(50, (param) => `Must be more or equal to ${param.min}`)
-    .required('Required'),
+  amountDataToken: Yup.number().required('Required'),
   amountOcean: Yup.number()
-    .min(50, (param) => `Must be more or equal to ${param.min}`)
+    .test('validator-custom-name', '', function (value) {
```
kremalicious, line 85: no clue what naming this method does but for the sake of completeness we should give it a fitting name, like `validator-min-amountOcean` or something

src/components/Publish/Pricing/Dynamic.tsx
```diff
@@ -10,6 +10,7 @@ import Price from './Price'
 import Decimal from 'decimal.js'
 import { useWeb3 } from '@context/Web3'
 import { FormPublishData } from '../_types'
+import { initialValues } from '../_constants'
```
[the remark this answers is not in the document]
KatunaNorbert: price also not used so will be a change in file :)

kremalicious:
Only getting the datatoken amount to update sometimes and then looks like math is not correct. Like, when only changing price, the amount of datatokens does not update. Here I just typed in a new price and got this:

Screen Shot 2022-04-13 at 15 04 22

Only works in user flow change price → change amount ocean. But when I then change price again, datatokens goes back to default 50

KatunaNorbert:
I agree, I still had to double check to make sure it's right... people might have the question why datatoken amount only changes on ocean amount and resets on price change without knowing what happens in the background. We should also make the min datatoken amount visible somehow, might make more sense why it's always reset to that.

kremalicious:
gonna sneak in #1340 and setting default ocean/datatoken to 100

src/components/Publish/_validation.ts
```diff
@@ -78,11 +80,23 @@ const validationPricing = {
...
+        this.parent.price > 0
+          ? new Decimal(this.parent.price)
+              .mul(this.parent.weightOnOcean)
+              .mul(10)
+              .toDecimalPlaces(MAX_DECIMALS)
+              .toString()
+          : initialValues.pricing.amountOcean.toString()
+      return value < parseInt(minValue)
+        ? this.createError({
+            message: `Must be more or equal to ${minValue}`
```
kremalicious, line 96: Changing this message might be already high impact without starting to show errors on datatoken field, which we wanted to prevent as it invites editing of a field is not editable.

Not sure how feasible but something like this:

`Must be more or equal to ${minValue}, as at least ${minValueDatatoken} ${datatokenSymbol} are required for this pool to work properly.`

kremalicious, line 96: Changing this message might be already high impact without starting to show errors on datatoken field, which we wanted to prevent as it invites editing of a field is not editable.

Not sure how feasible but something like this:

`Must be more or equal to ${minValue}, as at least ${minValueDatatoken} ${datatokenSymbol} are required for this pool to work properly.`

KatunaNorbert: Alright, yeah good idea, this way we could show min datatoken amount and give some explanation without adding messages around datatoken input.

KatunaNorbert: Was looking into getting the `datatokenSymbol` inside the message but didn't find a feasible way for it, so could we just say 'data tokens' instead?

kremalicious: yup, using "datatokens" in place of symbol is totally fine here

Issue #695, opened 2021-06-22
Port from old 3box to new IDX


AnaLoznianu:
https://blog.ceramic.network/migrating-from-3box-to-ceramic-and-idx/amp/

"3Box Labs is sunsetting 3Box products in favor of the new and more powerful combination of Ceramic Network and IDX. In this post we provide a timeline and various upgrade paths for projects currently using 3Box"

trentmc:
https://blog.ceramic.network/migrating-from-3box-to-ceramic-and-idx/amp/

3Box Labs is sunsetting 3Box products in favor of the new and more powerful combination of Ceramic Network and IDX. In this post we provide a timeline and various upgrade paths for projects currently using 3Box

kremalicious:
That issue most likely can't be solved right now as that blog post is from February and not much has changed. We should only start working on this once there is an actual public Ceramic mainnet, which at this point in time there is not. Only test network, and a waitlist for mainnet. Until then this all stays in prototype mode so not really worth starting it

trentmc:
I think it's best to wait as long as possible so that the thing can be maximum stable. But not so long that we get deprecated of course :)

mihaisc:
Tested a bit 
- 2 libraries needed (or more, not sure, didn't get stuff working)
![image](https://user-images.githubusercontent.com/26000280/122927651-f00f9f00-d31d-11eb-87ad-4718692c0dc1.png)
- documentation is not good
- getting errors on basic stuff described in examples 
- few issues on the github repo

mihaisc:
To migrate your account you need to go here and follow the steps https://ceramicstudio.github.io/web-playground/ . First try failed for me, will update.

mihaisc:
Currently the main net is still in closed beta and there is no way to update your idx profile. They say they will migrate the current edit tool from 3box to idx. Until that time i don't think we should migrate since at this point we don't plan on creating a profile page in our market.

mihaisc:
Alternatives to this are welcomed

trentmc:
> Alternatives to this are welcomed

Some  related identity plays, with DIDs at the core.
- [BrightID](https://www.brightid.org). We use it for Sybil tolerance in OceanDAO. Gitcoin does too, and many others. Seems pretty mature.
- [WaltID](https://walt.id/). The deltaDAO folks have integrated it in the Minimum Viable Gaia-X fork of Ocean Market, and work closely with them. WaltID received OceanDAO grants on this.
[...]

PR #1303, opened 2022-04-04
fix "owned by" label on preview

### LINKED ISSUE #1299: owner doesn't change in Preview
**Describe the bug**
If you are in the Preview step in publish and decides to change account, the owner wallet's information doesn't change in the UI. If you exit and enter back to preview, you'll see the change taking place.

**To Reproduce**
1. Go to Publish
2. Fill out until Preview
3. Change wallet's account
4. See old "Owner" account and "Owned by <old account>"

**Expected behavior**
To change in realtime

**Screenshots**
Screen Shot 2022-04-01 at 7 19 50 AM

EnzoVezzaro:
- Fixes #1299 

Changes proposed in this PR:

- fix "owner by" label on preview
- cleanup - remove unused code

### DIFF
--- a/src/components/@shared/Publisher/index.tsx
+++ b/src/components/@shared/Publisher/index.tsx
@@ -2,11 +2,8 @@ import React, { ReactElement, useEffect, useState } from 'react'
 import styles from './index.module.css'
 import classNames from 'classnames/bind'
 import Link from 'next/link'
-import get3BoxProfile from '@utils/profile'
 import { accountTruncate } from '@utils/web3'
 import axios from 'axios'
-import Add from './Add'
-import { useWeb3 } from '@context/Web3'
 import { getEnsName } from '@utils/ens'
 import { useIsMounted } from '@hooks/useIsMounted'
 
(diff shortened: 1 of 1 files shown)


src/components/@shared/Publisher/index.tsx
```diff
@@ -41,13 +38,6 @@ export default function Publisher({
         setAccountEns(accountEns)
         setName(accountEns)
       }
-
-      // 3box
-      const profile = await get3BoxProfile(account, source.token)
```
kremalicious, line 46: removing this effectively removes all 3box support for the publisher line, also for those for which it still works. When looking at current live (v3) market, 3box is still working for all the accounts which have highest liquidity assets:

Screen Shot 2022-04-04 at 13 41 14

Migrating 3box to its successor, or something else, should be left for #695 and for now we leave this in place also for publisher line here

EnzoVezzaro: you're right, 3box restored 👍🏽 thanks for that

mihaisc (approved):

PR #1309, opened 2022-04-04
fix transferable NFTs

kremalicious:
- send along `transferable` param when creating NFT
- new optional `transferable` param in our `generateNftCreateData()` helper, default to `true`. So we don't need to bother adding this to the publish tx fee component
- capture new key in form data, `metadata.transferable`, set to `true` by default. This prepares this to be added as a UI option later on

### DIFF
--- a/src/components/Publish/_constants.tsx
+++ b/src/components/Publish/_constants.tsx
@@ -53,6 +53,7 @@ export const initialValues: FormPublishData = {
   },
   metadata: {
     nft: { name: '', symbol: '', description: '', image_data: '' },
+    transferable: true,
     type: 'dataset',
     name: '',
     author: '',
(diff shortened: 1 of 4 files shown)


src/components/Publish/_constants.tsx
```diff
@@ -53,6 +53,7 @@ export const initialValues: FormPublishData = {
   },
   metadata: {
     nft: { name: '', symbol: '', description: '', image_data: '' },
+    transferable: true,
```
kremalicious, line 56: would seem logical to put it under `metadata.nft` but this would have had some more refactoring implications as what we mean with `metadata.nft` is capturing the metadata of the NFT alone which we then put in place for OpenSea and the like, and adding and then removing `transferable` from that object during the process seemed a bit overkill so I just dropped it on here

mihaisc: it doesn't matter, this is not a library

mihaisc (approved):

Issue #1118, opened 2022-02-17
Think of a way to handle slow network


mihaisc:
On rinkeby we are getting pretty often the following `Transaction was not mined within 50 blocks, please make sure your transaction was properly sent... ` . This is because the network is slow .  What can we do to improve the experience?

EnzoVezzaro:
I'm my opinion it'd be much better if we could show a "Pending Transaction" (button / alert) instead of just restoring the UI. It's very confusing when this happens. I think that showing a "Pending Transaction" somewhere (with an alert or on the buttons themselves) to notify the user that there's a pending transaction and he is basically "Stuck" for a bit could be already a better user experience. Because when this happens, the user is tempted to retry (queuing more transactions) because he's never gotten any feedback. But if we show something, at least the user will know there's being a problem and he must wait a bit. I've seen this in other dApps. This doesn't solve the initial problem, but it might be a little improvement.

mihaisc:
This could work for all actions except publish where you have 2 chained tx.

EnzoVezzaro:
> This could work for all actions except publish where you have 2 chained tx.

Yeah, exactly. Unfortunately, the publish flow requires another solutions even thought with this we would notify the user there's something going on, at least the user has the change to see what's happening and he can decide if retry (causing more tx to get queued) or wait until the one pending is done.

kremalicious:
We need to figure out where this "Transaction was not mined within 50 blocks" threshold is coming from, and then as a start increase this so we simply wait longer for tx to finish

mihaisc:
we can do that in ocean.js config. The main thing we need to to is to check if the ocean.js config regarding gas is actually working. Also at the end of the day if the network is busted what can you do?

kremalicious:
> Also at the end of the day if the network is busted what can you do?

yeah exactly. So just more transparency would be sufficient solution too, meaning somehow detecting slow network, then on asset detail screen put prominent warning at very top that network is slow or congested and things might fail. Not sure we find a reliable check which does not trigger this warning all the time.

mihaisc:
no concrete action for this

PR #1310, opened 2022-04-04
Incorrect wallet balance on trade tab fix

### LINKED ISSUE #1301: incorrect wallet balance on trade tab
**Describe the bug**
The balance shown in the trade tap doesn't reflect what actually on the wallet. 

These assets on mumbai are not correct: 
- https://v4.market.oceanprotocol.com/asset/did:op:c2a31770d4e6a8fde05a8a90e30fa98b35443b21f7ef155baedd2cd0a5fb3fe8
- https://v4.market.oceanprotocol.com/asset/did:op:84577212e4b8c1ce4a46e43360e45c6b9904e0eb366cacbad25a8a09cb66e2c2

These assets on rinbeky have the correct amount:
- https://v4.market.oceanprotocol.com/asset/did:op:bfa03dc8d250da7786f596b6098532007d3b25f5869027bf528a347984f105cc
- https://v4.market.oceanprotocol.com/asset/did:op:f66b072ec293aa76a29668df9ba7dbaf2cb0bb610f3475f8ebeebcfb731749b8

~Maybe it's related just to mumbai, but I'm still investigating~

**To Reproduce**
1. Go to Asset
2. Click on Trade tab
3. See amount balance (& other numbers)

**Screenshots**

Wrong:
Screen Shot 2022-04-01 at 9 30 03 AM

Correct:
Screen Shot 2022-04-01 at 9 39 07 AM

Some issue on rinkeby too:

- https://v4.market.oceanprotocol.com/asset/did:op:f66b072ec293aa76a29668df9ba7dbaf2cb0bb610f3475f8ebeebcfb731749b8

Screen Shot 2022-04-01 at 9 42 15 AM

claudiaHash:
Fixes #1301  .

Changes proposed in this PR:

- display correct  token balance  from wallet


src/components/Asset/AssetActions/Trade/Swap.tsx
```diff
@@ -144,7 +144,9 @@ export default function Swap({
 
         setBaseTokenItem((prevState) => ({
           ...prevState,
-          amount: amountBaseToken.toDecimalPlaces(MAX_DECIMALS).toString(),
+          amount: new Decimal(balance.baseToken)
```
claudiaHash, line 147: Not sure if correct but the change translates in: I have this amount of base tokens from my wallet, but I ca only spend the maximum amount of token, so the token balance displayed is the one from my wallet and doesn't depend on any other calculations based on pool data.

Screenshot from 2022-04-04 23-37-42

mihaisc:
image
I have 0 dt in wallet

EnzoVezzaro:
Found some issues. 

1) Maximun to spend DT is not reset to 0 if the user has 0 DT

Screen Shot 2022-04-06 at 4 57 01 AM

2) Maximun to spend Ocean seems incorrect. I'm not sure how much we should show, but this number is not consistent between In or Out 

If you're trading ocean for dt (IN) the value for maximun to spend is 859.994 but if you trade DT to Ocean the Maximun is 580.143 OCEAN

IN - 
Screen Shot 2022-04-06 at 5 06 17 AM

OUT -
Screen Shot 2022-04-06 at 5 06 23 AM


3) Maximun to spend DT seems off 

When you trade IN the amount of maximun to spend DT is different than when you trade OUT (see above)

4) If you change from IN to OUT, it kicks in the validation anyway, I'm not completely sure but I think it should not validate the fields if it's all still 0

https://user-images.githubusercontent.com/13741335/161939920-9282e9cd-39af-4304-9617-28e61a586c71.mp4

5) if you start a tx we should disable all the inputs to avoid issues

https://user-images.githubusercontent.com/13741335/161940808-21673aff-bee3-4f42-bfde-282d9e14c0f5.mp4

6) After the tx, the trade tab doesn't work anymore. (see error in console)

https://user-images.githubusercontent.com/13741335/161941438-c5523167-c953-467e-a2eb-4882372667cd.mp4

Sorry for the long post 😅 I'm sure some of this are not related to wallet's balance itself, but it'd be helpful to take a look at it. Maybe some of this has been fix in the meanwhile, so try to update from v4 branch to see if there's any fix for any of this issues.

claudiaHash:
> Sorry for the long post sweat_smile I'm sure some of this are not related to wallet's balance itself, but it'd be helpful to take a look at it. Maybe some of this has been fix in the meanwhile, so try to update from v4 branch to see if there's any fix for any of this issues.

No problem really, thanks for testing it thoroughly. I'll try to reproduce the issues mentioned one by one and see what could be done

mihaisc:
It seems this is a bit more complicated, not sure yet the extent of the issue. I'll jump on this PR.

claudiaHash:
Given that @mihaisc made various fixes which include the issues in here, I think we can close this pr after all the fixes are merged.

mihaisc:
This pr is not yet done

mihaisc:
All math is good now. There is still one weird exception. After the swap i call `fetchAllData` from pool context but it seems to cache the data, when it is called on the interval it still fetches old data. Only when i click on refresh it fetched new data from subgraph. All these events call the same function. It is very strange or i am too tired and missing something. Still debugging this

EnzoVezzaro:
I'm still not sure these numbers are actually correct as I keep getting weird differences between `balance` and `max to spend`. For example:

The balance on [this asset](https://market-5kqcg6lai-oceanprotocol.vercel.app/asset/did:op:42db932411ab2b30b2779a0bdd85d62d66c140d93cae861ced9031943a06ebbd) seems correct: 
Screen Shot 2022-04-26 at 10 31 57 AM

But [this asset](https://market-5kqcg6lai-oceanprotocol.vercel.app/asset/did:op:63920bb64ad1e79bed73214e5fd574fc68c0977cdba5b7031aa98a3a83e28197) instead doesn't seem like it:
Screen Shot 2022-04-26 at 10 38 06 AM

It might be correct, but I'm having difficulties to completely get [these calculations](https://github.com/oceanprotocol/market/blob/fix/issue-1356-loader-stuck-button-asset/src/components/Asset/AssetActions/Trade/Swap.tsx#L122:L122). They seem correct though. Any thoughts on this?

mihaisc:
Maximum to spend is a mix between your balance and how much you can actually spend in a pool. There is a limit to how much you can trade based on the pool liquidity.

mihaisc:
I fetch the pool liquidity from the contracts now so we don't depend on subgraph refreshing fast enough. This should be good to go

src/components/Asset/AssetActions/Trade/Swap.tsx
```diff
@@ -73,88 +75,78 @@ export default function Swap({
...
         setDtItem((prevState) => ({
           ...prevState,
-          amount: amountDataToken.toDecimalPlaces(MAX_DECIMALS).toString(),
+          amount: datatokenLiquidity,
           maxAmount: maximumDt
         }))
-      } catch (error) {
-        LoggerInstance.error(error.message)
+      } else {
+        const maxDtFromPool = calcMaxExactIn(datatokenLiquidity)
```
kremalicious, line 120: this looks like almost everything in this `else` statement is doing exactly the same as in above `if (values.type === 'buy')` clause, just with different const names. So CodeClimate warning for this is valid

mihaisc: No, there are multiple differences. I prefer it this way that having 3-4 chained conditional operators like we had before and nobody understood anything, most of the conditions were wrong.

kremalicious:
accessing the trade tab with this PR preview just crashes app for me:
https://market-git-fix-issue1301-incorrect-wallet-balance-oceanprotocol.vercel.app/asset/did:op:86838d23c988aa91ed225ed7df04cbfbe9daf542d9230c4c4af032c0c286db78

Screen Shot 2022-04-30 at 11 01 11

mihaisc:
Yes, this is a new feature. Looking into it 😁

mihaisc:
Done.

EnzoVezzaro:
I found a couple of issue on my end. 

1) the balance is not updating after the tx

https://user-images.githubusercontent.com/13741335/166219758-1f2ad082-2ab1-4295-bb50-50c3f15bb61d.mp4

2) "Your Liquidity" is not updating as expected, also if you refresh, most of the time this value doesn't get updated

https://user-images.githubusercontent.com/13741335/166220177-6f4a1f5b-8608-45ac-9c47-f290551d0dca.mp4

mihaisc:
We discussed many times this issue, it is still open. But how is this related to this pr?

EnzoVezzaro:
> We discussed many times this issue, it is still open. But how is this related to this pr?

Yeah, you're right. Never mind, I'll remove this comment as not related.

EnzoVezzaro (approved):

mihaisc:
@kremalicious can you take another look?

src/components/Asset/AssetActions/Trade/Swap.tsx
```diff
@@ -167,12 +160,15 @@ export default function Swap({
     asset,
     web3,
     dtItem.token,
+    dtItem.amount,
     baseTokenItem.token,
+    baseTokenItem.amount,
     poolInfo.liquidityProviderSwapFee,
```
kremalicious, line 166: as we added the full `poolInfo` object as dependency further down, we can clean up this array a bit and remove these 3 `poolInfo.` sub-values.

src/components/Asset/AssetActions/Trade/Swap.tsx
```diff
@@ -167,12 +160,15 @@ export default function Swap({
...
     poolInfo.datatokenAddress,
     poolInfo.baseTokenAddress,
     appConfig,
-    appConfig.consumeMarketPoolSwapFee
+    appConfig.consumeMarketPoolSwapFee,
```
kremalicious, line 170: same here, `appConfig` alone should be enough as dependency

kremalicious:
### Functionality

- in general, all numbers are top-notch now, could trade in both directions, would trade again.
- continue to run into long tx issue where we simply error out and stop everything although tx is still running in background, although this might require more generic solution in #1118
  Screen Shot 2022-05-05 at 12 33 39
- our site-suggested gas fee is still dramatically lower than all the auto-values and would assume this is reason sometimes tx not get through within 50 blocks, just noting here as PR is not supposed to solve this
  Screen Shot 2022-05-05 at 12 30 32

Some UI findings, can also be moved to new enhancement issue as it's present on `v4` too:

### Labeling 

Would suggest instead of `Balance` & `Available from pool` to rather use `Your Balance` & `Pool Balance` for the top labels. Simple change so we can sneak this into this PR as final commit.

### Switching tokens interaction

So users start with this screen, all nice:

Screen Shot 2022-05-05 at 12 22 46

Then they hit the switch arrow and are presented with errors right away, giving them impression they did something wrong:

Screen Shot 2022-05-05 at 12 24 00

After hitting the arrow, expectation would be to get a blank screen, not a prefilled screen with wrong values filled in triggering errors.

mihaisc:
The gas stuff is not related to this
Changed the labels
Agree with the validation stuff but would leave it for another pr.

EnzoVezzaro (approved):

kremalicious (approved):
so perfect

Screen Shot 2022-05-05 at 13 03 36

PR #1325, opened 2022-04-05
width changes

### LINKED ISSUE #1324: Content size too small for text pages
Pages like t&c , privacy and whatever else we have don't look ok
image

claudiaHash:
Fixes #1324  .

### DIFF
--- a/src/components/@shared/atoms/Container.module.css
+++ b/src/components/@shared/atoms/Container.module.css
@@ -7,7 +7,7 @@
   width: 100%;
 }
 
-@media (min-width: 40rem) {
+@media (min-width: 55rem) {
   .container {
     padding-left: calc(var(--spacer) / 1.2);
     padding-right: calc(var(--spacer) / 1.2);
@@ -15,5 +15,5 @@
 }
 
 .container.narrow {
-  max-width: 42rem;
+  max-width: 62rem;
 }


EnzoVezzaro (approved):
LGTM

src/components/@shared/atoms/Container.module.css
```diff
@@ -7,13 +7,13 @@
   width: 100%;
 }
 
-@media (min-width: 40rem) {
+@media (min-width: 55rem) {
```
kremalicious, line 10: what's the goal of this change? Why change this spacing switch for each and every page we have, as this component here is used in many places?

This change e.g. on iPad 9.7" gives less consistent spacing to the left/right of content.

This is v4, which we want, where the space left/right if exactly half of the space between the elements:

Screen Shot 2022-04-06 at 12 52 17

And this is this PR reducing all left/right spacing making it feel cramped:

Screen Shot 2022-04-06 at 12 53 31

src/components/@shared/atoms/Container.module.css
```diff
@@ -7,13 +7,13 @@
...
   .container {
     padding-left: calc(var(--spacer) / 1.2);
     padding-right: calc(var(--spacer) / 1.2);
   }
 }
 
 .container.narrow {
```
kremalicious, line 17: it's not narrower than the standard `.container` anymore so this class name is really misleading. Same goes for the prop `narrow` which with this change also became misleading, so prop and class name need to be changed to something describing what we do here

src/components/@shared/atoms/Container.module.css
claudiaHash: yes, you're right, I just removed it

kremalicious (approved):
great, now it's a nice minimal change

PR #1326, opened 2022-04-06
Making it more obvious that something needs to be entered in the timeout field

### LINKED ISSUE #1313: Confusing experience with the 'timeout' field in the publish form
When publishing an asset, is not very visible that the timeout field is a mandatory field. 
After adding the file URL, is a bit confusing why the user cannot move to the next steps using the continue button. It gives the impression that he needs to submit a sample file, although that field is not mandatory.
Maybe we can highlight what the missing field is.

jamiehewitt15:
Fixes #1313 

Changes proposed in this PR:

- Removing dashes to make it more obvious input is needed

### DIFF
--- a/src/components/@shared/FormInput/InputElement.tsx
+++ b/src/components/@shared/FormInput/InputElement.tsx
@@ -65,9 +65,7 @@ export default function InputElement({
           disabled={disabled}
           multiple={multiple}
         >
-          {field !== undefined && field.value === '' && (
-            <option value="">---</option>
-          )}
+          {field !== undefined && field.value === '' && <option value="" />}
           {sortedOptions &&
             (sortedOptions as string[]).map((option: string, index: number) => (
               <option key={index} value={option}>


EnzoVezzaro:
instead of having the field empty, can we add a placeholder like the other 2 fields? It might not be super easy considering it's a dropdown, not an input, but maybe we could style the default value differently to make it look like a placeholder (like in the other fields).

Screen Shot 2022-04-06 at 6 17 43 AM

kremalicious:
> instead of having the field empty, can we add a placeholder like the other 2 fields?

I think it would have opposite effect again and at same time would give impression the `select` is a text `input` with the way our inputs are designed so wouldn't bother with this. So leaving this empty indeed creates a better affordance nudging users into doing something with it even when just scanning the page

mihaisc (approved):

EnzoVezzaro (approved):

PR #1327, opened 2022-04-06
Various fixes in the pool component

### LINKED ISSUE #1164: Trying to remove liquidity using the "max" button sometimes gives an error
**Describe the bug**
When trying to remove liquidity from this asset [did:op:ca285f858304b20e4e319e9c1b17fcc60b8a2e08af8e52d88fd2b9427f9d5649](https://market-git-v4-oceanprotocol.vercel.app/asset/did:op:ca285f858304b20e4e319e9c1b17fcc60b8a2e08af8e52d88fd2b9427f9d5649) using the "max" button, the value is too high and an error is shown

**Screenshots**
Screenshot 2022-03-03 at 16 48 59

### LINKED ISSUE #1266: Update numbers on pool tab automatically
With #1265 we disabled interval fetching as it was creating way too many issues but the principle feature is still something we want so users do not have to reload the page to get updated numbers.

For this, check out urql and how its architecture could have a replacement for simple interval re-fetching which in turn might be better solution than adding and managing intervals. Or make use of subscribing feature of GraphQL.

Like:

https://formidable.com/open-source/urql/docs/advanced/retry-operations/
https://formidable.com/open-source/urql/docs/basics/ui-patterns/#reacting-to-focus-and-stale-time
https://formidable.com/open-source/urql/docs/advanced/subscriptions/

### LINKED ISSUE #1316: Pool numbers don't add up (part 2)
I added 130 ocean to this pool . I mostly don't understand "your value locked"
image
image

mihaisc:
- closes #1316 (wasn't really an issue)
- should also fix #1164
- proper validation in add liquidity
- added "Expected output" in remove liquidity
image
- fixed percentage calculations in remove liquidity
- update all UI after removing liquidity. Didn't do anything regarding this. Sometimes the subgraph updates before i get the confirmation popup from metamask, sometimes it takes like a second. At this moment i feel like solving this would be too complicated and it's not worth it.
- small tweaks to `Token` regarding conversion
- Refactors in pool context

`Your liquidity` - in base token
`Owner liquidity` - in base token
`Total value locked` - is just the base token liquidity 

- price for pool is calculated locally (not calling contracts) . Price calculation still takes some time because we need to fetch pool data from subgraph and fee from provider.
- pool liquidity calculation for user and owner are done locally (not calling contracts) 
- restored pool interval fetch, closes #1266

- added a global context. Now it is used only for opc fee, we can use this for all various data 
- use and display correct opc based on chain and approved token

In profile :
- `TVL in Own Assets` is a sum of `Your liquidity` in own assets
- `TVL` is a sum of `Your liquidity` in all assets
- in pool shares table `Your value locked` becomes `Your liquidity`  and `Total value locked` is just the base token liquidity converted

- `getAccountTVLInOwnAssets` renamed to `getAccountLiquidityInOwnAssets` and now is calculated locally.
- fixed `getPoolSharesLiquidity`
- fixed `getAssetsFromPoolShares`
- fixed  `Liquidity`


kremalicious:
> WIP : update all UI after removing liquidity

nice. Fun thing was that I was not even able to get fresh numbers with the `Refresh Data` button, was always fetching me old pool shares. Cause it's super weird, we refetch in multiple places automatically but the response simply is old stuff and wasn't able to figure out yet why and where this comes from

EnzoVezzaro:
> * Pool numbers don't add up (part 2)  #1316

I had a similar issue the `refresh data` button, but while adding liquidity. I add some liquidity and after the tx success, the UI changed (with new balance) but the Pool numbers stayed the same for like 30sec before I was able to update the data with the `refresh data` button. 

Also, I found a weird bug: after removing all my liquidity, the UI still reflected the old expected output. I couldn't replicate this consistently as it doesn't happens always.

Screen Shot 2022-04-06 at 10 05 36 AM

kremalicious:
> after removing all my liquidity, the UI still reflected the old expected output.

in this case we should set it to 0 manually, we do the same for all other displayed values in the `finally` clause for handling removing liquidity

mihaisc:
Delaying this a bit because the math in the remove part is WRONG 😭 , not sure yet how much needs to be refactored.

EnzoVezzaro (approved):
found no issues while testing. Tested: 

- add liquidity
- remove liquidity (with / without max button)
- trade in & out

kremalicious:
adding/removing and values shown on those screens seem fine in my tests.

On pool screen, problem with my liquidity, added 10 OCEAN but shows about 20 OCEAN:

Screen Shot 2022-04-08 at 13 10 29

Also double checked and after adding liquidity the fetched subgraph data indeed gives back no pool shares, only after hitting _refresh data_ multiple times I got updated numbers after like 20 sec. So on frontend we seem to be doing everything right and that fetching of subgraph multiple times is theoretically unnecessary

mihaisc:
@kremalicious soo, is there any issue? 😆  Also can in the future when you see that the subgraph updates so slow can you write the time when this happened? Extra step can you do a direct query to the subgraph?

kremalicious:
was just basically my testing session report, sorry 😀 But first part is independent of allegedly delayed subgraph update, added 10 OCEAN but shows about 20 OCEAN. So that's an issue, which is also present with `v4`.

Above screenshot is on this dataset:
https://market-git-fix-1316-oceanprotocol.vercel.app/asset/did:op:63920bb64ad1e79bed73214e5fd574fc68c0977cdba5b7031aa98a3a83e28197

Doing only this tx:
https://rinkeby.etherscan.io/tx/0x2307dda4f5ca0e7816545f172faec0459352321ae4d872e49353661c9ed91984

kremalicious:
On [this asset](https://market-git-fix-1316-oceanprotocol.vercel.app/asset/did:op:42d405447ee257a8033121e15e37b81c9bb894c2d4047736abf94e66e4842e49) I also added only 10 OCEAN, but shows I have about 33 OCEAN. Maybe this is expected and reflects liquidity provider rewards like swaps happening on that pool with in last 3 days?

Screen Shot 2022-04-08 at 20 01 32

mihaisc:
> was just basically my testing session report, sorry 😀 But first part is independent of allegedly delayed subgraph update, added 10 OCEAN but shows about 20 OCEAN in Your Value Locked. So that's an issue, which is also present with v4.

Yes, because that is TVL  = liquidity x2. Once again tvl proves to be more confusing than liquidity, not sure it was a good idea to change everything to tvl.

The second issue i don't understand . From what i can see you have `932,904,168,293,271,000` pool tokens. That means you should get `6.359891681121760024` OCEAN
image

mihaisc:
Drop 'locked' from names. The amount will be just the amount of ocean the user is able to remove.

mihaisc:
We have been using in code both pool tokens and pool shares . While they mean the same thing, for some people it might be confusing so we should stick to one. So pool tokens or pool shares ? opinions...

src/@context/GlobalData/_types.ts
```diff
@@ -0,0 +1,11 @@
+export interface OpcValue {
```
kremalicious, line 1: this all is so generic in naming, I wouldn't know what "OceanProtocolCommunityValue" written out would exactly get, and "GlobalData" is kinda what all our context providers hold.

As for file organization, full refactor way would be to merge `useSiteMetadata` hook and this context in this PR into a new e.g. `MarketMetadata` provider, where all values added in here could be put under a `fees` object

src/@context/Pool/index.tsx
```diff
@@ -97,137 +89,120 @@ function PoolProvider({ children }: { children: ReactNode }): ReactElement {
...
+        newPoolOwnerInfo
+      )
     }
-    setPoolInfoOwner(newPoolOwnerInfo)
-    LoggerInstance.log('[pool] Created new owner pool info:', newPoolOwnerInfo)
+
+    calculatePoolOwnerInfo()
   }, [
-    ownerPoolShares,
+    asset.chainId,
```
kremalicious, line 149: throughout this file we seem to assume that `asset` can be undefined as we are doing `asset?....` a lot so I guess we should keep it for every `asset` access in this file, so in this case `asset?.chainId`

src/@context/Pool/index.tsx
```diff
@@ -97,137 +89,120 @@ function PoolProvider({ children }: { children: ReactNode }): ReactElement {
...
+      const poolSharePercentage = new Decimal(poolInfoUser.poolShares)
+        .dividedBy(new Decimal(poolInfo.totalPoolTokens))
+        .mul(100)
+        .toFixed(2)
+
+      setUserHasAddedLiquidity(Number(poolSharePercentage) > 0)
+
+      const newPoolInfoUser = {
+        liquidity: userLiquidity,
+        poolshares: poolInfoUser.poolShares,
```
kremalicious, line 185: it's `poolShares` in the typings, the new `const` should be assigned that typing above so this small typo should get visible in type checking then

src/@types/Utils.d.ts
```diff
@@ -0,0 +1,9 @@
+interface CalcInGivenOutParams {
+  tokenInLiquidity: string
+  tokenOutLiqudity: string
```
kremalicious, line 3: small typo:

```suggestion
  tokenOutLiquidity: string
```

src/components/Asset/AssetActions/Download.tsx
```diff
@@ -55,19 +59,35 @@ export default function Download({
...
+
+      const params = {
+        tokenInLiquidity: poolData?.baseTokenLiquidity,
+        tokenOutLiqudity: poolData?.datatokenLiquidity,
+        tokenOutAmount: '1',
+        opcFee: getOpcForToken(poolData.baseToken.address, asset?.chainId),
+        lpSwapFee: poolData?.liquidityProviderSwapFee,
+        publishMarketSwapFee: poolData?.publishMarketSwapFee,
+        consumeMarketSwapFee: '0'
+      } as CalcInGivenOutParams
```
kremalicious, line 77: this converts from one type to the other so errors in above object creation would not be checked. To get that, we want the variable to be assigned a type, so `const params: CalcInGivenOutParams`

src/components/Asset/AssetActions/Pool/Add/index.tsx
```diff
@@ -40,15 +41,25 @@ export default function Add({
   const [newPoolShare, setNewPoolShare] = useState('0')
   const [isWarningAccepted, setIsWarningAccepted] = useState(false)
 
+  // eslint-disable-next-line security/detect-non-literal-regexp
+  const maxDecimalsValidation = new RegExp(
```
kremalicious, line 45: we also have this defined in `Publish/_validations.ts` so we should abstract this to somewhere, guess if in doubt drop into generic `@utils/constants.ts` :-)

src/components/Asset/AssetActions/Pool/Remove/_utils.ts
```diff
@@ -1,19 +1,42 @@
-import { calcMaxExactOut } from '@oceanprotocol/lib'
+import { PoolInfo, PoolInfoUser } from '@context/Pool/_types'
+import { calcMaxExactOut, Pool } from '@oceanprotocol/lib'
 import Decimal from 'decimal.js'
+// eslint-disable-next-line camelcase
+import { PoolData_poolData } from 'src/@types/subgraph/PoolData'
```
kremalicious, line 5: in other places when using those subgraph typings we convert this to a camelcase typename which then ESLint seem to like again. So in this case maybe:

```suggestion
import { PoolData_poolData as PoolData } from 'src/@types/subgraph/PoolData'
```

src/components/Asset/AssetActions/Pool/Remove/index.module.css
```diff
@@ -50,25 +50,30 @@
...
+.output p {
+  font-weight: var(--font-weight-bold);
+  margin-bottom: calc(var(--spacer) / 8);
+  font-size: var(--font-size-small);
 }
 
-.output [class*='token'] > figure {
-  display: inline-block;
+.output [class*='token'] {
+  white-space: normal;
```
kremalicious, line 70: Was introducing wrapping intended here? This would make token image + value + symbol wrap to next line, whereas original intention was for them to be always on one line. Wrapping after token image would be fine in some instances I guess but see we are hiding it anyway in this case. Effectively making value and symbol wrap which should be prevented at all times

src/components/Asset/AssetActions/Pool/Remove/index.tsx
```diff
@@ -137,20 +134,18 @@ export default function Remove({
       .dividedBy(100)
       .mul(new Decimal(poolInfoUser.poolShares))
       .toString()
-
-    setAmountPoolShares(`${amountPoolShares.slice(0, 18)}`)
+    setAmountPoolShares(amountPoolShares)
   }
 
-  function handleMaxButton(e: ChangeEvent<HTMLInputElement>) {
+  async function handleMaxButton(e: ChangeEvent<HTMLInputElement>) {
```
kremalicious, line 140: nothing `async` going on here but guess some state setting works more reliably with it here?

src/components/Asset/AssetActions/Pool/Remove/index.module.css
```diff
@@ -50,25 +50,30 @@
...
 }
 
-.output figure[class*='pool shares'] {
+.output [class*='token'] > figure {
```
kremalicious, line 73: do we actually need this? As further down we put `noIcon` onto `Token` component (in code-review-only phase, did not check out visually yet)

src/components/Profile/History/PoolShares/Liquidity.tsx
```diff
@@ -14,38 +13,22 @@ export function Liquidity({
   row: AssetPoolShare
   type: string
 }) {
-  let price = '0'
   let liquidity = '0'
```
kremalicious, line 16: guess it works as it was done quickly originally but proper React way would be to put this all into `useState` and `useEffect` to minimize recalculations

src/components/Asset/AssetActions/Pool/Remove/index.module.css
```diff
@@ -50,25 +50,30 @@
...
 
 .output {
-  composes: output from '../Add/Output.module.css';
+  border-top: 1px solid var(--border-color);
+  padding: var(--spacer);
+  padding-bottom: calc(var(--spacer) / 1.5);
+  display: grid;
+  gap: var(--spacer);
+  grid-template-columns: 1fr 1fr;
+  margin-left: -2rem;
```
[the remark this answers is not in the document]
kremalicious: there it is, dealing with `margin` & `padding` already

```css
composes: container from '../Section/index.module.css';
```

src/@context/GlobalData/_types.ts
mihaisc: Agree on all the naming stuff. Added that context like this so that we can fix/evolve in another pr so that this one doesn't get any bigger.

src/@context/Pool/index.tsx
```diff
@@ -97,137 +89,120 @@ function PoolProvider({ children }: { children: ReactNode }): ReactElement {
...
-    const poolShare = new Decimal(ownerPoolShares)
-      .dividedBy(poolInfo.totalLiquidityInOcean)
-      .mul(100)
-      .toFixed(2)
-
-    const newPoolOwnerInfo = {
-      liquidity: new Decimal(ownerPoolShares), // liquidity in base token, values from from `calcSingleOutGivenPoolIn` method
-      poolShares: ownerPoolShares,
-      poolShare
+    async function calculatePoolOwnerInfo() {
```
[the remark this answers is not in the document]
mihaisc: Yes, i tried to keep refactoring to a minimum. Basically make it work , don't add bugs. If we go this route we can refactor soo much . The async part is just a remnant that i forgot to remove, i'll take a look.

src/components/Asset/AssetActions/Pool/Remove/index.module.css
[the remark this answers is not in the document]
mihaisc: This answer is valid for all the css related comments: 
I just copy pasted stuff so it doesn't break and looks reasonable. It's always under the assumption that it will be fixed/made prettier in a later pr 🤣

src/components/Profile/History/PoolShares/Liquidity.tsx
mihaisc: true 😞

mihaisc:
Ok, fixed all comments. Found some other "less optimal" stuff but we can't refactor everything here. Ain't touching the css 😁

README.md
```diff
@@ -134,11 +134,11 @@ const queryLatest = {
 }
 
 function Component() {
-  const { appConfig } = useSiteMetadata()
+  const { siteMetadata } = useMarketMetadata()
   const [result, setResult] = useState<QueryResult>()
 
   useEffect(() => {
-    if (!appConfig.metadataCacheUri) return
+    if (!siteMetadata?.appConfig.metadataCacheUri) return
```
kremalicious, line 141: so nested :-) Awesome, thanks for doing this context merge in here. As we are accessing `appConfig` a lot think would make sense moving it up one level into `MarketMetadataProviderValue` data structure so `SiteMetadata` is then left to be only the content from json files. But that's like, super bonus. Would have advantage that `appConfig` would be always present from the beginning as we can directly import the file and add it to the context provider value so never a need for optional chaining then

kremalicious: also remembered dealing with importing data and merging in extra method is a remnant from Gatsby where we had to get this data with graphql and then pass it through. With Next.js we could ditch all this when we deal with import of json & js files which are present at build time, and just directly import in `@context/MarketMetadata/index.tsx` and add to context value

mihaisc: ok, this should be the final version

kremalicious:
ah well, I was wrong, some internals of hooks seem to make even direct imports being not available for small amount of time. This is what I get when starting up locally:

Screen Shot 2022-04-21 at 16 41 07

Just removing `appConfig.chainIdsSupported` from dependency array makes it work again for me

mihaisc:
A small null check did the trick. I missed it there, it should be all over the place.

kremalicious:
oh, now I know why, cause we start `appConfig` with empty state. What I meant is that we can directly put the import on the output value, without any local state in between. I"ll quickly test this and will push in here if that works so we end up without that `getSiteMetadata()` at all

kremalicious (approved):
Excellent, catches so many issues and numbers make way more sense than before and should be continued to be tested in `v4`. And with 65 changed files one of the more lighter refactors we had, but also excellent add/remove line ratio

PR #1329, opened 2022-04-06
Showing connect wallet button on submit page when no wallet is connected

### LINKED ISSUE #1320: Show connect wallet button on submit page
If a user arrives at the submit page in the publish wizard without a wallet connected, it would be nice to present them with a "connect wallet" button rather than the greyed out submit button. 

So for example, it could look like this if you arrived at this page without a wallet connected:

Connect wallet Button on submit page 2 - 2022-04-05_16-21

This is the same sort of UI that you'll see on Uniswap and other Dapps. This is still a connect wallet button at the top but also one in place of the swap button:

Connect wallet Button on submit page - Uniswap example - 2022-04-05_16-21

jamiehewitt15:
Fixes #1320 

Changes proposed in this PR:

- Showing connect wallet button on submit page when no wallet is connected

### DIFF
--- a/src/components/Publish/Actions/index.tsx
+++ b/src/components/Publish/Actions/index.tsx
@@ -21,8 +21,7 @@ export default function Actions({
     isSubmitting,
     setFieldValue
   }: FormikContextType<FormPublishData> = useFormikContext()
-
-  const { connect } = useWeb3()
+  const { connect, accountId } = useWeb3()
 
   async function handleActivation(e: FormEvent<HTMLButtonElement>) {
     // prevent accidentially submitting a form the button might be in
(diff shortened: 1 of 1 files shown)


kremalicious:
for me, the _Connect Wallet_ is always present, no matter if I have wallet connected or not. Also without wallet connected, then clicking button connecting my wallet, the button also doesn't change

jamiehewitt15:
Ok, I see the issue. That was happening when a form field wasn't valid. Should be working ok now

kremalicious (approved):

EnzoVezzaro (approved):

PR #1331, opened 2022-04-06
Unsupported Network Tooltip

### LINKED ISSUE #1278: Unsupported networks should be more obvious in publish screens
In this use case the small red square in header is just not enough to prevent people from trying to publish into it, as the form UI suggests that this is possible:

Screen Shot 2022-03-29 at 16 15 33

We could just block whole publish screen with error message in that case.

jamiehewitt15:
Fixes #1278 

## Changes proposed in this PR:

- Name of network is changed to red if that network is unsupported 
- If network is unsuported the new tooltip is shown
- Tooltip contains buttons the user can click to immediately change their wallet network
- The same tooltip has been added to the submit page
- Additional fix: the logic now uses `chainIdsSupported` from `app.config.js` as ocean.js was not giving reliable information on which networks are supported. Previously the red square wasn't shown for mainnets that we supported in V3 (even though we don't support them yet). It's fairly important we communicate that these networks aren't supported as someone could easily of the V4 link and try publish to mainnet etc.  

## Screenshots
- In page title:
Screenshot from 2022-04-08 13-49-08

- In place of the submit button:
Screenshot from 2022-04-08 13-49-22


kremalicious:
I don't think there's a need to introduce a specific unsupported networks tooltip. Did you notice how I used the existing tooltip in my draft and just added a list of supported networks? https://github.com/oceanprotocol/market/issues/1278#issuecomment-1090509431

Showing "Unsupported network" beside the red network name is enough, no need to add even more custom components and copy for this use case.

src/components/Publish/Actions/index.tsx
```diff
@@ -78,6 +81,15 @@ export default function Actions({
             <Button type="submit" style="primary" onClick={handleActivation}>
               Connect Wallet
             </Button>
+          ) : !isSupportedOceanNetwork ? (
+            <Tooltip
+              content={<UnsuportedNetwork />}
+              className={styles.errorTooltip}
+            >
```
jamiehewitt15, line 88: > I don't think there's a need to introduce a specific unsupported networks tooltip.

I'm still using the same `Tooltip` component from `'@shared/atoms/Tooltip'` but I'm passing in a new component as the content:

kremalicious:
> Not sure how we would do that if we just pass in text?

Sorry, was unclear, I mean that the supported networks should be shown at all times in the existing tooltip. Then no need to create a dedicated one only for the unsupported network case

jamiehewitt15:
> > Not sure how we would do that if we just pass in text?
> 
> Sorry, was unclear, I mean that the supported networks should be shown at all times in the existing tooltip. Then no need to create a dedicated one only for the unsupported network case

Ok sure, sounds good

jamiehewitt15:
I've updated to tooltip message to change it to a generic message about the available networks. The same tooltip will be always be shown regardless of which network the user is on. 

Screenshot from 2022-04-11 14-38-19

src/components/Publish/Actions/index.tsx
```diff
@@ -78,6 +81,15 @@ export default function Actions({
             <Button type="submit" style="primary" onClick={handleActivation}>
               Connect Wallet
             </Button>
+          ) : !isSupportedOceanNetwork ? (
+            <Tooltip
+              content={<AvailableNetworks />}
+              className={styles.errorTooltip}
+            >
+              <Button type="submit" style="primary" disabled>
+                Change Network
```
kremalicious, line 90: Would go more info style on this one and rather show "Unsupported Network", and then add a tooltip indicator after it so it's less surprising when a tooltip opens on that button. Also to have an affordance on touch devices

src/components/@shared/AvailableNetworks/index.module.css
```diff
@@ -0,0 +1,12 @@
+.alert {
+  margin-bottom: 2rem;
+}
+.networks {
+  column-count: 2;
```
kremalicious, line 5: you are assuming that this component will always be a direct children of an element which has `display: grid` on it which goes against the isolated components approach all components under `@shared` should follow. And hardcoded message also assumes component is only used within publish.

On first quick look, it seems the network list alone should be the `<AvailableNetworks />` component, with `display: grid` set on its root element, detached from the message which then would be added whenever component is used, or pass `message` as an optional prop

kremalicious:
UI-wise, we need to do something with the spacing as when we fill our tooltip component with more than few lines of text, it looks mashed together. To counter that we set more padding on the first child of it in such cases, like in our menubar tooltips so they can be used as spacing guideline for this specific tooltip too:

Screen Shot 2022-04-11 at 14 45 45

src/components/@shared/AvailableNetworks/index.tsx
```diff
@@ -0,0 +1,51 @@
+import React, { ReactElement } from 'react'
+import styles from './index.module.css'
+import NetworkOptions from './networkOptions'
```
kremalicious, line 3: consistency, `networkOptions.tsx` should be named `NetworkOptions.tsx` file name needs to be capitalized. Be aware this is not that simple to do on case-insensitive systems so most likely you have to use [`git mv`](https://www.hanselman.com/blog/git-is-casesensitive-and-your-filesystem-may-not-be-weird-folder-merging-on-windows)

src/components/@shared/AvailableNetworks/networkOptions.tsx
```diff
@@ -0,0 +1,35 @@
...
+  const { networksList } = useNetworkMetadata()
+  const { web3Provider } = useWeb3()
+
+  function changeNetwork(chainId: number) {
+    const networkNode = networksList.find((data) => data.chainId === chainId)
+    addCustomNetwork(web3Provider, networkNode)
+  }
+
+  return (
+    <div key={chainId}>
```
kremalicious, line 23: that's a weird markup for a list, and gets even more wrapping divs when rendered with `AvailableNetworks/index.tsx`. Semantically, it's a list and not a form so no form elements like `label`. Should be a `ul` with `button` inside of `li` and that's it, no further divs

src/components/Publish/Actions/index.module.css
```diff
@@ -8,3 +8,13 @@
 .actions button {
   margin: 0 calc(var(--spacer) / 2);
 }
+
+.infoIcon {
+  width: 0.85em;
+  height: 0.85em;
+  cursor: help;
```
kremalicious, line 15: with the button having `cursor: pointer`, the disabled button having `cursor: no-allowed`, I am pretty sure this cursor is impossible to trigger for a user

src/components/Publish/Actions/index.module.css
```diff
@@ -8,3 +8,13 @@
...
+  display: inline-block;
+  margin-bottom: -0.1em;
+  margin-left: calc(var(--spacer) / 6);
+  fill: var(--brand-white);
```
kremalicious, line 19: Using this is more universal as we would not have to change colors in case we would change button

```suggestion
  fill: currentColor;
```

src/components/Publish/Actions/index.module.css
jamiehewitt15: One option could be to do it by overriding the style that's already been added to disabled buttons. e.g. passing in a new class to the button like this:

```css
.infoButton {
  cursor: help !important;
  pointer-events: all !important;
}
```

src/components/Publish/Title/index.module.css
```diff
@@ -14,3 +17,17 @@
   width: 0.5em;
   height: 0.5em;
 }
+.errorTooltip {
+  width: 0.5em;
+  height: 0.5em;
+  fill: var(--brand-alert-red);
+}
+.errorIcon {
+  width: 0.85em;
```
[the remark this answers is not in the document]
jamiehewitt15: The styling is needed as this is a different instance of the Info component:

```Typescript
            <Tooltip
              content={<AvailableNetworks />}
              className={styles.errorTooltip}
            >
              <Info className={styles.errorIcon} />
            </Tooltip>
```

If we user the default trigger `className={styles.icon}` changes the fill colour to grey at a lower level of specifity so the info icon won't be red:

```Typescript
const DefaultTrigger = React.forwardRef((props, ref: any) => {
  return <Info className={styles.icon} ref={ref} />
})
```

src/components/@shared/AvailableNetworks/index.module.css
jamiehewitt15: I'm not sure I follow what you're saying here? This isn't a grid, the parent component is just the default `display: block`

src/components/@shared/AvailableNetworks/Network.tsx
```diff
@@ -0,0 +1,35 @@
...
+  const { networksList } = useNetworkMetadata()
+  const { web3Provider } = useWeb3()
+
+  function changeNetwork(chainId: number) {
+    const networkNode = networksList.find((data) => data.chainId === chainId)
+    addCustomNetwork(web3Provider, networkNode)
+  }
+
+  return (
+    <ul key={chainId}>
```
kremalicious, line 23: this markup now creates a `ul` for each item which seems unintended

src/components/@shared/AvailableNetworks/Network.tsx
```diff
@@ -0,0 +1,35 @@
...
+      <li className={styles.radioLabel}>
```
kremalicious, line 24: where is this class defined? I don't see it in the CSS

src/components/@shared/AvailableNetworks/index.module.css
kremalicious: yeah, my bad, `column-count` indeed has nothing to do with `display: grid` and we cleaned up markup a lot already. Still, I don't see this component being used outside of `src/components/Publish` so tending towards moving it there as it is simply not a shared component, and its messaging is only for publishing. As said, we could move the message out of component and pass it as prop but that all feels like premature optimization which we can do once there's a need using this component outside of publish.

content/publish/index.json
```diff
@@ -1,6 +1,5 @@
 {
   "title": "Publish",
   "description": "Highlight the important features of your data set or algorithm to make it more discoverable and catch the interest of data consumers.",
-  "warning": "Publishing into a test network first is strongly recommended. Please familiarize yourself with [the market](https://oceanprotocol.com/technology/marketplaces), [the risks](https://blog.oceanprotocol.com/on-staking-on-data-in-ocean-market-3d8e09eb0a13), and the [Terms of Use](/terms).",
-  "tooltipNetwork": "Assets are published into the network your wallet is connected to. Switch your wallet's network to publish into another one."
```
kremalicious, line 5: if we move the component under `components/Publish` then keeping the actual message in this content file would make sense again.

src/components/Publish/Title/index.module.css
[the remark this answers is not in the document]
kremalicious: maybe some more specific selector would help then? In any case still lots of duplication going on in this css file and a better approach would be to use more utility classes. Like, `margin-left` is duplicated on `.network` & `.wrongNetwork`, `width` & `height` is on both, `.tooltip` & `.errorTooltip`.

So for both those elements simply use a base class and then add additional class e.g. for error state and with that only overwrite what you need, like so:

```css
.network {
  margin-left: xxx;
}
.network.error {
  color: red;
}
```

Everything else just makes us duplicate the same CSS over and over again.

src/components/@shared/AvailableNetworks/index.module.css
```diff
@@ -0,0 +1,15 @@
+.alert {
```
kremalicious, line 1: this seems not used

src/components/@shared/AvailableNetworks/index.module.css
jamiehewitt15: Ok cool, I'll move it into `src/components/Publish` for now

src/components/@shared/AvailableNetworks/index.module.css
jamiehewitt15: Yeah, good point. I've got rid of it

EnzoVezzaro (approved):
Great job 👏🏽

src/components/Publish/Title/index.module.css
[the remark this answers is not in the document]
kremalicious: 👆 this is still valid. And also seeing we are duplicating `infoIcon` styles, with 2 places in this PR alone defining the same thing. Now that we are back within `components/Publish` you could easily `compose` one of them like:

```css
.infoIcon {
  composes: infoIcon from '../Actions/index.module.css';
}
```

Or the other way around. But most likely you could simply grab all those definitions from the original info tooltip icon component, which then results in 2 full duplications removed.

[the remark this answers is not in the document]
jamiehewitt15: I've used `compose` for the `infoIcon` styling from `'../Actions/index.module.css';` like you suggested. 

Regarding your comments about the repetition in `errorTooltip` and `wrongNetwork` I have removed both of those classes and am just adding the red styling with a single new `error` class

src/@context/Web3.tsx
```diff
@@ -106,6 +109,8 @@ function Web3Provider({ children }: { children: ReactNode }): ReactElement {
     eth: '0',
     ocean: '0'
   })
+  const [isSupportedOceanNetwork, setIsSupportedOceanNetwork] =
+    useState<boolean>(true)
```
kremalicious, line 113: type can be inferred from usage here so no need to add `<boolean>`

src/@context/Web3.tsx
```diff
@@ -304,6 +309,17 @@ function Web3Provider({ children }: { children: ReactNode }): ReactElement {
     }
     await web3Modal.clearCachedProvider()
   }
+  // -----------------------------------
+  // Get valid Networks and set isSupportedOceanNetwork
+  // -----------------------------------
+
+  useEffect(() => {
+    if (appConfig.chainIdsSupported.includes(networkId)) {
```
kremalicious, line 317: let's clean up this logic as it's so short, it shouldn't be confusing at all. When next developer reads this logic, it says "when current user's networkId is among appConfig.chainIdsSupported, then network is supported, but when accountId is present, network is not supported". See how hard this logic is to get here?

This effect should only do what the comment says it does, check current user's network against supported networks, set to `false` if not. That's it. All the rest tries to accommodate to use case somewhere deep down in a component so this should be handled there.

jamiehewitt15: Ok yeah, I get your point here. I was thinking that we don't want the title to be red when the user doesn't have a wallet connected but I've moved that logic into the component now and this effect is simpler. 

```Typescript
<NetworkName
            networkId={networkId}
            className={
              isSupportedOceanNetwork || !accountId
                ? styles.network
                : `${styles.network} ${styles.error}`
            }
          />
```

kremalicious: let's also update the effect dependencies as we only want this to fire when `networkId` or `appConfig.chainIdsSupported` is changed

kremalicious (approved):
great, with new ui/interaction design + new HTML + new CSS + new React behavior all in one PR was a lot to cover

PR #1334, opened 2022-04-07
Display generic info message for file inputs

### LINKED ISSUE #1319: Market wrongly detects xlsx file as HTML
When publishing an asset, the market wrongly detects file as HTML format even if the file is a valid xlsx.
File URL: https://data.gov.ro/dataset/b86a78a3-7f88-4b53-a94f-015082592466/resource/2a84fe08-6502-4df6-b973-1bfa0303a352/download/transparenta_aprilie_2022.xlsx

Result:

claudiaHash:
Fixes #1319  .


src/components/@shared/FormFields/FilesInput/index.tsx
```diff
@@ -6,12 +6,19 @@ import { InputProps } from '@shared/FormInput'
...
+import content from '../../../../../content/publish/form.json'
+import FormHelp from '@shared/FormInput/Help'
 
 export default function FilesInput(props: InputProps): ReactElement {
   const [field, meta, helpers] = useField(props.name)
   const [isLoading, setIsLoading] = useState(false)
   const { values, setFieldError } = useFormikContext<FormPublishData>()
 
+  const helperMessage = content.services.fields[2].help.replace(
+    'Please enter the URL to your data set file and click "ADD FILE" to validate the data.',
```
kremalicious, line 18: what are you trying to do here? Why don't you change the actual content instead of clearing it out here?

There should be only 1 help text, not a help tooltip and a help text. So we need a mechanism to have a defined help text either as a tooltip (default) or output directly in field as `FormHelp`

src/components/@shared/FormFields/FilesInput/Info.tsx
```diff
@@ -19,7 +19,7 @@ export default function FileInfo({
   // Prevent accidential publishing of error pages (e.g. 404) for
   // popular file hosting services by warning about it.
   // See https://github.com/oceanprotocol/market/issues/1246
-  const shouldWarnAboutFile = file.valid && contentTypeCleaned === 'html'
```
kremalicious, line 22: if none of this is used, all the comments and commented out code should be removed

kremalicious:
> There should be only 1 help text, not a help tooltip and a help text. So we need a mechanism to have a defined help text either as a tooltip (default) or output directly in field as FormHelp.

Because this is just too much help at once:

Screen Shot 2022-04-11 at 15 42 52

Screen Shot 2022-04-11 at 15 42 58

Also the compute part is not relevant for people doing a data set so ideally this text should not be there in this case but we should stop with those string replacements regardless, so quick fix would be to change order of sentences to have what's important first. So message flow encrypted → file accessible → compute file

kremalicious:
looks like we are back to only having tooltips when looking at PR preview. Posting the draft from #1319 on here as this is what we want in terms of UI (exact copy is fine in this PR):

162206947-12d85670-7a4e-40e8-9614-3ccbcedd9f46

There is:
1. help text under the input field
2. no tooltip 

Solution:

> we need a mechanism to have a defined help text either as a tooltip (default) or output directly in field as FormHelp.

All our form fields use the tooltip (default) but in this case we want the text output directly in field as `FormHelp`. So most likely this has to be done in our input components and needs to be usable for all form fields. One way would be to add a new key in the form data for that file field like `prominentHelp: true` and then conditionally switch tooltip & FormHelp but I'm sure there are more ways to get to the above draft UI

src/components/Asset/Edit/FormEditMetadata.tsx
```diff
@@ -103,6 +103,7 @@ export default function FormEditMetadata({
                   ? timeoutOptionsArray
                   : field.options
               }
+              prominentHelp={field.name === 'links'}
```
claudiaHash, line 106: should we apply the same mechanism on the edit form?

kremalicious: would say yes. But looking at mechanism, conceptionally, it should be up to content author to decide about tooltip help or prominent help on the respective field content, so would put the `prominentHelp` key in the content file and work from there. 

Because otherwise we always would have to edit specific components when we decide to use prominent help on another field too. Once we get this key from the content file, we don't need to think about adding this as a prop to some components or not

src/components/@shared/FormInput/index.tsx
```diff
@@ -118,16 +121,17 @@ export default function Input(props: Partial<InputProps>): ReactElement {
...
           </span>
         )}
-        {help && <Tooltip content={<Markdown text={help} />} />}
+        {help && !prominentHelp && (
+          <Tooltip content={<Markdown text={help} />} />
+        )}
       </Label>
       <InputElement size={size} {...field} {...props} />
-
+      {help && prominentHelp && <FormHelp>{help}</FormHelp>}
```
kremalicious, line 129: this now pushes down all error messages:

Screen Shot 2022-04-20 at 15 28 21

As help text can be long or short we are at the limit of hacking it in place with `position: absolute` like I did here with the default provider link: https://github.com/oceanprotocol/market/blob/v4/src/components/%40shared/FormFields/Provider/index.module.css#L10-L11

So we need to figure out a new layout so that error messages are always attached to input field, but components added before or after it won't push it down

src/components/@shared/FormFields/URLInput/index.module.css
```diff
@@ -9,4 +9,6 @@
 
 .error {
   composes: error from '@shared/FormInput/index.module.css';
+  position: absolute;
```
kremalicious, line 12: please no more `absolute` positioning hacks in this place, it is impossible to predict the exact size of the help text block as there are too many viewport + text size + text length combinations. I only used it in the formerly linked place because it was pretty much guaranteed that "Default provider" winkle stay on one text line in almost all viewports. If you can't guarantee this, never start using `position: absolute` to take elements out of box model flow.

As written before, we need a new layout here with more elements but don't know yet what exactly. Goal is to have another `position: relative` wrapper the error is attached to, but not the help text.

Case in point, with this change, this is what I get on mobile, again, impossible to deal with this in absolute positioning and percentage values:

Screen Shot 2022-04-20 at 21 14 41

kremalicious (approved):

PR #1339, opened 2022-04-11
show gas estimation on buy

### LINKED ISSUE #1066: Show gas cost estimation for buy
We can display how much the tx will cost to order ( or buy dt + order for pool).  So somewhere in the use tab we can display this.

EnzoVezzaro:
- Fixes #1066 

Changes proposed in this PR:

- added gas estimation to buy

### DIFF
--- a/src/@utils/ocean.ts
+++ b/src/@utils/ocean.ts
@@ -2,6 +2,8 @@ import { ConfigHelper, LoggerInstance, Config } from '@oceanprotocol/lib'
 // import contractAddresses from '@oceanprotocol/contracts/artifacts/address.json'
 import { AbiItem } from 'web3-utils/types'
 import Web3 from 'web3'
+import { getDummyWeb3 } from './web3'
+import Decimal from 'decimal.js'
 
 export function getOceanConfig(network: string | number): Config {
   const config = new ConfigHelper().getConfig(
@@ -71,4 +73,24 @@ export async function getOceanBalance(
   } catch (e) {
     LoggerInstance.error(`ERROR: Failed to get the balance: ${e.message}`)
   }
+}
+
+export async function convertGasFeesToOcean(
+  amount: Decimal,
+  ethToOceanConversionRate: number,
+  chainId: number,
+  web3?: Web3
+): Promise<string> {
+  if (!web3) {
+    web3 = await getDummyWeb3(chainId)
+  }
+  const gasPrice = await web3.eth.getGasPrice()
+
+  const gasFee = web3.utils.fromWei((+gasPrice * +amount).toString(), 'ether')
+
+  const gasFeeOcean =
+    ethToOceanConversionRate > 0
+      ? (+gasFee / +ethToOceanConversionRate).toString()
+      : ''
+  return gasFeeOcean
 }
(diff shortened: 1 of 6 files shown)


src/@utils/order.ts
```diff
@@ -48,57 +47,153 @@ export async function order(
...
 
-  // TODO: we need to approve provider fee
+  return {
+    initializeData,
+    orderParams
+  }
+}
+
+/**
+ * Before stating the transactions, we need to initialized our provider's instance
```
claudiaHash, line 57: Here are some typos which we should correct before merging

src/@utils/ocean.ts
```diff
@@ -72,3 +74,23 @@ export async function getOceanBalance(
     LoggerInstance.error(`ERROR: Failed to get the balance: ${e.message}`)
   }
 }
+
+export async function convertGasFeesToOcean(
```
kremalicious, line 78: not sure we have to do all this conversion as making our `usePrices()` hook also deliver ETH prices is super trivial as the Coingecko API allows passing multiple token IDs, so would stay just one request. But guess we would have to refactor some places then

src/@utils/order.ts
```diff
@@ -48,57 +47,153 @@ export async function order(
...
+/**
+ * Get and execute transaction based on asset type
+ * @param web3
+ * @param asset
+ * @param accountId
+ * @param action
+ * @param orderPriceAndFees
+ * @returns {TransactionReceipt} receipt of the order / gas estimate
+ */
+async function getTransaction(
```
kremalicious, line 65: good direction centralizing this, just a note that we need to test all possible download options with this change as PR touches all of them.

Just logically, the estimations are no transactions so not sure this is the best place to hold them too

src/@utils/ocean.ts
EnzoVezzaro: Ok, I will look into it. At this level also, we'll need to make some adjustments based on changes to [`Prices.tsx`](https://github.com/oceanprotocol/market/pull/1132/files). We'll need to add the fix [here](https://github.com/oceanprotocol/market/blob/feature/issue-1066-gas-estimation-buy/src/components/Asset/AssetActions/Download.tsx#L72:L72) to make sure these calculations are done for matic too.

src/components/Asset/AssetActions/Download.tsx
```diff
@@ -53,16 +63,42 @@ export default function Download({
     setValidOrderTx(asset?.accessDetails?.validOrderTx)
     // get full price and fees
     async function init() {
+      // we only show the gas estimation if the user is connected
```
kremalicious, line 66: this sounds like it should rather be handled in its own effect after this one, which also would make this init method a bit less complex and we nicely separate concerns

src/@utils/ocean.ts
kremalicious: yup, trivial, with this in `@context/Prices`:

```js
const tokenIds = 'ocean-protocol,ethereum,matic-network'
```

we get this back:

Screen Shot 2022-04-15 at 17 03 52

Requires changing a bit data structure in `usePrices()` as it looks like though

kremalicious:
closing this, as:

- 230 lines of code are too much for this functionality
- nobody uses ETH mainnet in our market, so usefulness is questionable

PR #1342, opened 2022-04-13
add steps back history to publish

### LINKED ISSUE #1243: Using the "Back" browser functionality takes the user out of the publish flow
**Describe the bug**
Using the “back” browser functionality in step 2 of the Asset publishing goes back to the list of assets, instead of going back to the 1st step

EnzoVezzaro:
- Fixes #1243 

Changes proposed in this PR:

- handle back button with steps
- load step state on refresh


src/components/Publish/Navigation/index.tsx
```diff
@@ -1,21 +1,53 @@
...
+  const router = useRouter()
+
   const {
     values,
     errors,
     touched,
     setFieldValue
   }: FormikContextType<FormPublishData> = useFormikContext()
 
+  const queryString = window.location.search
```
kremalicious, line 18: seems to all work but just wondering if there's a way to do same thing without resorting to `window` (as this might conflict with some server-side rendering) and without having to sync more values between different places. 

Just thinking out loud, we already have a routing system going on with Next.js so we should be able to end up with URLs like `/publish/1` by using nested routes within `src/pages/`: https://nextjs.org/docs/routing/introduction

With that in place, the whole screen switching mechanism could be switched to then getting the screen page from `router.pathname` instead of using `values.user.stepCurrent` and we would not have to sync this form state with the URL query params.

claudiaHash (commented):
Not sure if this is intended but once you access the publish form, you can't go back to the home page (for instance), using the browser 'Back' functionality

EnzoVezzaro:
> Not sure if this is intended but once you access the publish form, you can't go back to the home page (for instance), using the browser 'Back' functionality

is not intended. I will check on it

claudiaHash (approved):

src/components/Publish/Navigation/index.tsx
```diff
@@ -13,7 +15,8 @@ export default function Navigation(): ReactElement {
   }: FormikContextType<FormPublishData> = useFormikContext()
 
   function handleStepClick(step: number) {
-    setFieldValue('user.stepCurrent', step)
+    // Change route to include steps
+    router.push({ pathname: `${router.pathname}`, query: { step: step } })
```
kremalicious, line 19: why do we need to set both, a pathname and a query as part of options object? As this now looks like we have 2 systems to get `step` across. Was thinking that with `publish/[step]` system we can simply do `router.push(`/publish/${step}`)` so no need to keep track of additional `query.step`, no?

EnzoVezzaro: yeah, sure. I'm not sure there's any difference as I was catching router.pathname `publish/[step]` and changing his [step] prop. But it's alright, your solution seems cleaner 👍🏽

kremalicious (approved):
awesome, works great!

PR #1351, opened 2022-04-18
Sample file validation in edit form fix

### LINKED ISSUE #1350: Validating file on edit form doesn't work
Screenshot from 2022-04-18 11-14-16

claudiaHash:
Fixes #1350  .

### DIFF
--- a/src/components/@shared/FormFields/FilesInput/index.tsx
+++ b/src/components/@shared/FormFields/FilesInput/index.tsx
@@ -19,7 +19,6 @@ export default function FilesInput(props: InputProps): ReactElement {
     e.preventDefault()
 
     try {
-      console.log('FILES URL: ', url, values, asset)
       const providerUrl = values?.services
         ? values?.services[0].providerUrl.url
         : asset.services[0].serviceEndpoint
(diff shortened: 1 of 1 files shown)


src/components/@shared/FormFields/FilesInput/index.tsx
```diff
@@ -6,18 +6,22 @@ import { InputProps } from '@shared/FormInput'
...
 
   async function handleValidation(e: React.SyntheticEvent, url: string) {
     // File example 'https://oceanprotocol.com/tech-whitepaper.pdf'
     e.preventDefault()
 
     try {
-      const providerUrl = values?.services[0].providerUrl.url
+      const providerUrl = values?.services
+        ? values?.services[0].providerUrl.url
+        : asset.services[0].serviceEndpoint
```
mihaisc, line 24: Not sure what to think about this. Considering this component will not be used in other places ... i think it's ok. Refactoring this will take too much time.

mihaisc (approved):

src/components/@shared/FormFields/FilesInput/index.tsx
bogdanfazakas, line 24: Why don't we keep just `asset?.services?[0]?.serviceEndpoint` since the providerUrl can't be edited ?

claudiaHash: not sure bout the implications since this component is used in publish form as well

bogdanfazakas: yep nvm them makes sense 😃

bogdanfazakas (approved):

Issue #1148, opened 2022-03-01
Test walletconnect


mihaisc:
Test if portis/walletconnect(+ ledger) are working in v4 for all the basic flows: publish (all 3 pricing schemas) , order, trade,  add/remove liquidity 

Previous issues with walletconnect : https://github.com/oceanprotocol/market/issues/592

LucaMilanese90:
I can't seem to be able to connect to the market with Portis, is it because https://v4.market.oceanprotocol.com/ is a preview link? 
![Screenshot 2022-03-17 at 16 38 09](https://user-images.githubusercontent.com/20192135/158838570-e1fbca8b-cd2b-419f-b1b8-2d41ea4f1bed.png)

AnaLoznianu:
@kremalicious Can we add the Portis key?

kremalicious:
Fun fact, 229 accounts seem to have used Portis in Ocean Market so far:

<img width="1043" alt="Screen Shot 2022-03-29 at 11 42 40" src="https://user-images.githubusercontent.com/90316/160594342-f7febe4f-a604-4a65-8d58-f86d2d4aef32.png">

kremalicious:
Alright, set the allow list domains to:

```
https://market.oceanprotocol.com,https://*.oceanprotocol.vercel.app,https://*.market.oceanprotocol.com
```

kremalicious:
Regression of combination walletconnect + Ledger Live in v4 preview:

On iOS and iPadOS getting only the QR code after selecting walletconnect when it should bring up the mobile app screen to use it on same device, like on balancer:

![image](https://user-images.githubusercontent.com/90316/160603383-ba178c67-8e5f-4267-902e-939506a3ae74.png)

We want this:
![image](https://user-images.githubusercontent.com/90316/160603418-e4e58e0c-3bf2-49f1-8ce3-12949fb625f9.png)

mihaisc:
Interesting last time i tested i explicitly remember that cool walletconnect select popup, i was kinda impressed. Not sure what happened , i don't think we touched the integration since then

LucaMilanese90:
I have been able to successfully publish an asset from a wallet connected through WalletConnect scanning the QR through Metamask Mobile. Correctly received the transaction notifications and sent the confirmation from my smartphone (iOS).
https://v4.market.oceanprotocol.com/asset/did:op:6137dd7a9d068708df3b9527ebb3a82420c21c49eed36e28ffb067053839fa54

[...]

PR #1355, opened 2022-04-25
package updates

kremalicious:
* bump to ocean.js v1.0.0-next.37
* bump all web3 & wallet related packages
* enable all custom portis components again, ref #1148


EnzoVezzaro:
I was unable to connect, but i think it's been worked out in #1148 as it has been already reported in there. 
Screen Shot 2022-04-25 at 7 25 36 AM

EnzoVezzaro (approved):

kremalicious:
> I was unable to connect

works via whitelisting domains and we have setup `https://market.oceanprotocol.com,https://*.oceanprotocol.vercel.app,https://*.market.oceanprotocol.com` on Portis so maybe you used `localhost`? Although I was actually able to login when using `localhost` which I think should not be possible

EnzoVezzaro:
> > I was unable to connect
> 
> works via whitelisting domains and we have setup `https://market.oceanprotocol.com,https://*.oceanprotocol.vercel.app,https://*.market.oceanprotocol.com` on Portis so maybe you used `localhost`? Although I was actually able to login when using `localhost` which I think should not be possible

Mmm.... weird, I'm actually testing what's in [preview](https://market-hp7vp4gg5-oceanprotocol.vercel.app/) so It should work then, but I'm still facing this issue. I'll keep on eye on #1148

PR #1358, opened 2022-04-25
add dataset search 

corrie-sloot:
This pull request implements first-class integration of the market with [Google Dataset Search](https://datasetsearch.research.google.com/). Dataset Search is a search engine for datasets. By adding a little standardised [structured data](https://developers.google.com/search/docs/advanced/structured-data/dataset) (JSON-LD) to the market HTML source, it will allow users outside of the Ocean ecosystem to more easily find and purchase datasets.

Certain properties already existed for a DID, such as name, description, creator, license, etc. Others may be added in the future when datasets are published such as temporalCoverage, spatialCoverage, Tabular datasets, etc. [Complete list of machine-readable properties for a dataset](https://schema.org/Dataset).

This project was proposed during the [Round 16 DAO](https://port.oceanprotocol.com/t/mpowered-first-class-integration-of-market-with-google-dataset-search-round-16/1679) and subsequently approved by @trentmc


src/components/Asset/AssetActions/index.tsx
```diff
@@ -81,6 +81,18 @@ export default function AssetActions({
             )
           : await getFileDidInfo(asset?.id, asset?.services[0]?.id, providerUrl)
         fileInfoResponse && setFileMetadata(fileInfoResponse[0])
+
+        // set the content type in the Dataset Schema
+        const datasetSchema = document.getElementById('DatasetSchema')
```
kremalicious, line 86: Direct DOM manipulation is quite adventurous in a React app. For changing anything in the `head` of any page, we'll have to use [`next/head`](https://nextjs.org/docs/api-reference/next/head). Next.js devs also suggest to use [`next/script`](https://nextjs.org/docs/basic-features/script) when adding `script` to `head` but we might not use this in this case as we are not loading external scripts with it.

mihaisc (approved):

Issue #1357, opened 2022-04-25
Download button refreshes to show buy button 


jamiehewitt15:
During the process of purchasing download access to a dataset, the button constantly keeps refreshing to update the price. This continues to happen after all of the transactions have been signed. When the download button refreshes, it shows the "buy" button again which is very confusing for a user who has already bought the dataset. 

## Video

https://user-images.githubusercontent.com/20739535/165107608-3497facf-b13e-4241-b564-966bc60ac6c7.mp4

## To recreate

1. Buy the dataset and complete the transactions
2. After the "Download" button shows wait a few seconds until the next time that it refreshes.
3. You'll now see the "Buy" button is visible instead of the "Download" button

kremalicious:
If there is a simple fix to close issue we should go for it but the underlying issue is the `ButtonBuy` trying to do way too many things at the same time. This is also reflected in CodeClimate where this component show up as 2nd worse file we maintain:

<img width="815" alt="Screen Shot 2022-05-17 at 12 55 53" src="https://user-images.githubusercontent.com/90316/168805368-7a19e5ac-7296-468f-9d4c-1a05619877f0.png">

And if we have to add that many props to a component although we mostly try to use hooks, then it's clear something is wrong:

<img width="606" alt="Screen Shot 2022-05-17 at 12 58 04" src="https://user-images.githubusercontent.com/90316/168805702-07b157d5-1d1f-467a-ba70-21a63a9daed3.png">

So if we ant to solve this properly and prevent possible further bugs, this component needs to be split up based on our current usage, e.g. dedicated order button & download button components to split the complexity and such

PR #1360, opened 2022-04-26
restore order for assets with fixed pricing

### LINKED ISSUE #1356: Price is never calculated - Cannot read properties of undefined (reading 'baseToken')
**Describe the bug**
There are a couple of assets where the price isn't calculated. It is just endlessly loading....

**To Reproduce**
1. Go to [this asset](https://v4.market.oceanprotocol.com/asset/did:op:8edc467bce44355de5b6ee7a860f6d463cf494bbba3420a64e4f3038522d3ccb) and see that the price is never loaded.
2. Go to [this asset](https://v4.market.oceanprotocol.com/asset/did:op:b0d9ea2a46172d9425228d4bcef696536d44583af98f44e7610ebe3c0d02be81) and see that the price is never loaded.

**Expected behavior**
There should be a price and it should be possible to buy the dataset.

**Screenshots**

Price never loads - 2022-04-25_13-27

**Logs**

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'baseToken')
    at 629-8534c0ba3d0887c7.js:1:32271
    at s (main-914fbfab4f90b52f.js:1:89133)
    at Generator._invoke (main-914fbfab4f90b52f.js:1:88921)
    at Generator.P.forEach.e.<computed> [as next] (main-914fbfab4f90b52f.js:1:89556)
    at xt (629-8534c0ba3d0887c7.js:1:29419)
    at i (629-8534c0ba3d0887c7.js:1:29623)
    at 629-8534c0ba3d0887c7.js:1:29684
    at new Promise (<anonymous>)
    at 629-8534c0ba3d0887c7.js:1:29563
    at o (629-8534c0ba3d0887c7.js:1:32593)
```

### LINKED ISSUE #1363: Buy doesn't work for assets published with a fixed price of 1 Ocean
Steps to reproduce:
1. Publish an asset using a fixed price schema and a price of 1 Ocean
2. Try to buy it
3. A "Failed to buy datatoken from pool" error shows up

https://v4.market.oceanprotocol.com/asset/did:op:3958da6ccc5439d48888d8b98c960444f0d1832b95e5ef20af899f03b5758be4 

Screenshot 2022-04-27 at 12 49 00

### LINKED ISSUE #1364: Download not working for Free assets
**Describe the bug**
There is an issue when trying to purchase Free assets. After clicking the "Get" button and signing the transactions, the asset isn't downloaded. The "Get" button remains in place and isn't replaced by the download button

**To Reproduce**
1. Purchase an asset
2. Notice that it isn;t downloaded

**Screen recording**

https://user-images.githubusercontent.com/20739535/165502878-f31f4056-c2e2-4b20-83de-059cb3aa0d42.mp4

EnzoVezzaro:
- Fixes #1356 .
- Fixes #1364 
- Fixes #1363 

Changes proposed in this PR:

- restore order for fixed pricing
- restore download button after purchase
- disable download/buy button if the user is on network different as the asset's

### DIFF
--- a/src/@utils/accessDetailsAndPricing.ts
+++ b/src/@utils/accessDetailsAndPricing.ts
@@ -256,6 +256,7 @@ export async function getOrderPriceAndFees(
     asset?.services[0].serviceEndpoint
   )
   orderPriceAndFee.providerFee = initializeData.providerFee
+  if (!orderPriceAndFee.providerFee) return
 
   // fetch price and swap fees
   switch (asset?.accessDetails?.type) {
--- a/src/components/Asset/AssetActions/Compute/index.tsx
+++ b/src/components/Asset/AssetActions/Compute/index.tsx
@@ -420,11 +420,7 @@ export default function Compute({
             assetTimeout={datasetTimeout}
             hasPreviousOrderSelectedComputeAsset={hasPreviousAlgorithmOrder}
             hasDatatokenSelectedComputeAsset={hasAlgoAssetDatatoken}
-            oceanSymbol={
-              accessDetails?.baseToken?.symbol
-                ? accessDetails.baseToken.symbol
-                : ''
-            }
+            oceanSymbol={accessDetails?.baseToken?.symbol || ''}
             dtSymbolSelectedComputeAsset={
               selectedAlgorithmAsset?.datatokens[0]?.symbol
             }
--- a/src/components/Asset/AssetActions/Compute/index.tsx
+++ b/src/components/Asset/AssetActions/Compute/index.tsx
@@ -420,7 +420,11 @@ export default function Compute({
             assetTimeout={datasetTimeout}
             hasPreviousOrderSelectedComputeAsset={hasPreviousAlgorithmOrder}
             hasDatatokenSelectedComputeAsset={hasAlgoAssetDatatoken}
-            oceanSymbol={accessDetails ? accessDetails.baseToken.symbol : ''}
+            oceanSymbol={
+              accessDetails?.baseToken?.symbol
+                ? accessDetails.baseToken.symbol
+                : ''
+            }
             dtSymbolSelectedComputeAsset={
               selectedAlgorithmAsset?.datatokens[0]?.symbol
             }
--- a/src/components/Asset/AssetActions/Download.tsx
+++ b/src/components/Asset/AssetActions/Download.tsx
@@ -60,7 +60,8 @@ export default function Download({
       if (
         asset?.accessDetails?.addressOrId === ZERO_ADDRESS ||
         asset?.accessDetails?.type === 'free' ||
-        (!poolData && asset?.accessDetails?.type === 'dynamic')
+        (!poolData && asset?.accessDetails?.type === 'dynamic') ||
+        !poolData
       )
         return
       setIsLoading(true)
@@ -82,7 +83,6 @@ export default function Download({
       )
 
       setOrderPriceAndFees(orderPriceAndFees)
-
       setIsLoading(false)
     }
 
--- a/src/components/Asset/AssetActions/Download.tsx
+++ b/src/components/Asset/AssetActions/Download.tsx
@@ -71,11 +71,17 @@ export default function Download({
         tokenInLiquidity: poolData?.baseTokenLiquidity,
         tokenOutLiquidity: poolData?.datatokenLiquidity,
         tokenOutAmount: '1',
-        opcFee: getOpcFeeForToken(poolData.baseToken.address, asset?.chainId),
+        opcFee: getOpcFeeForToken(
+          poolData?.baseToken?.address
+            ? poolData?.baseToken?.address
+            : asset?.accessDetails?.addressOrId,
+          asset?.chainId
+        ),
         lpSwapFee: poolData?.liquidityProviderSwapFee,
         publishMarketSwapFee: poolData?.publishMarketSwapFee,
         consumeMarketSwapFee: '0'
       }
+
       const orderPriceAndFees = await getOrderPriceAndFees(
         asset,
         ZERO_ADDRESS,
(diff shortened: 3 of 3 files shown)


kremalicious (approved):
good find. I see more weird things in that effect but will handle in #1357 like the loading spinner always being restarted whenever any of the dependency like `poolData` changes

EnzoVezzaro:
> #1357

yes, most probably for #1357 it's enough checking [these deps](https://github.com/oceanprotocol/market/blob/fix/issue-1356-loader-stuck-button-asset/src/components/Asset/AssetActions/Download.tsx#L90:L90) or just comment/remove these [2 lines](https://github.com/oceanprotocol/market/blob/fix/issue-1356-loader-stuck-button-asset/src/components/Asset/AssetActions/Download.tsx#L67-L68) - tried it, it works, but it's not the most clean solution.

I found another bug thought. Fixed assets are having issues when trying to buy. 

Screen Shot 2022-04-26 at 6 23 53 AM

kremalicious:
> but it's not the most clean solution.

yup, I would question putting a price loading spinner on the button in the first place, should be on the price as that's what's loading but not sure yet about best solution.

And that looks like a bug indeed, would be great to add this in a new issue unless you can quickly solve it in here with some optional chaining

EnzoVezzaro:
> 

Nope, this seems to be related to a query that is returning no information about the pool. Without this information the `order` call fails, it happens only for the `fixed` price asset. This bug actually made me go back to check my solution, and I found that this is what was actually causing the issue with the loader. I will continue solving the issue on this PR

alexcos20:
why are we buying from a pool when asset's price schema is fixed rate?

EnzoVezzaro:
> why are we buying from a pool when asset's price schema is fixed rate?

Yes. We had an issue with the logic, now should be fine.

jamiehewitt15 (approved):

src/components/Asset/AssetActions/Download.tsx
```diff
@@ -71,23 +71,29 @@ export default function Download({
...
+
       const orderPriceAndFees = await getOrderPriceAndFees(
         asset,
         ZERO_ADDRESS,
         params
       )
 
       setOrderPriceAndFees(orderPriceAndFees)
-      setIsDisabled(false)
+      if (isAssetNetwork) {
```
kremalicious, line 90: in theory, this makes `isAssetNetwork` a dependency of the effect, just like the removed `isLoading` so should be added to dependency array. That is unless something else in existing dependency would change on wallet network change too, in which case we would know better than ESLint and we should add an ESLint-disabling comment on the dependency array so the warning here doesn't appear:

Screen Shot 2022-04-27 at 21 48 29

src/components/Asset/AssetActions/Download.tsx
```diff
@@ -98,7 +104,9 @@ export default function Download({
 
     const isDisabled =
       !asset?.accessDetails.isPurchasable ||
+      !isAssetNetwork ||
```
kremalicious, line 107: this is such an obvious change and I we just missed that cause it's already in below crazy logic line. Which makes we wonder if we should rather split this up, as on quick look like user could have case where only `! isBalanceSufficient` happens which then would be also missed by below line. Can't remember what exactly the purpose of this line is so least we should do adding code comment describing intention.

src/components/Asset/AssetActions/Download.tsx
kremalicious: another thing, we might not need to call `setIsDisabled()` here based on `isAssetNetwork` as further down in the next effect we only deal with the disable state of the button.

src/components/Asset/AssetActions/Compute/index.tsx
```diff
@@ -420,7 +420,11 @@ export default function Compute({
             assetTimeout={datasetTimeout}
             hasPreviousOrderSelectedComputeAsset={hasPreviousAlgorithmOrder}
             hasDatatokenSelectedComputeAsset={hasAlgoAssetDatatoken}
-            oceanSymbol={accessDetails ? accessDetails.baseToken.symbol : ''}
+            oceanSymbol={
+              accessDetails?.baseToken?.symbol
```
kremalicious, line 424: `accessDetails?.baseToken?.symbol || ''` should be the same, just shorter?

src/@utils/accessDetailsAndPricing.ts
```diff
@@ -256,6 +256,7 @@ export async function getOrderPriceAndFees(
     asset?.services[0].serviceEndpoint
   )
   orderPriceAndFee.providerFee = initializeData.providerFee
+  if (!orderPriceAndFee.providerFee) return
```
kremalicious, line 259: double checking: under what circumstances could this `orderPriceAndFee.providerFee` be `null` or `undefined`? As this line stops all price fetching in case this is not defined. So question is if we should if maybe somewhere the provider owner would have not set it or something. Do we have a default set wherever we get this from which is not  `null` or `undefined`?

EnzoVezzaro: This is a edge case when the network is slow. It doesn't happen a lot, but basically yesterday I encounter an issue where the query failed (because infura was slow/down, today i haven't had this issue) and this failing call makes [this line](https://github.com/oceanprotocol/market/blob/fix/issue-1356-loader-stuck-button-asset/src/@utils/accessDetailsAndPricing.ts#L289) of code throw an error. To avoid this extremely limited edge case, I added this return when providerFee is null (aka there's an error retrieving the data). We can remove it, if we think this will not create a problem in the future.

src/components/Asset/AssetActions/Download.tsx
EnzoVezzaro: removed both as they're not necessary

src/components/Asset/AssetActions/Download.tsx
EnzoVezzaro: tested all possible combinations, didn't encounter errors. I have left a comment on it, but I'm not sure why [these conditions](https://github.com/oceanprotocol/market/blob/fix/issue-1356-loader-stuck-button-asset/src/components/Asset/AssetActions/Download.tsx#L108) are grouped though.

kremalicious: awesome, thanks for testing and those code comments are really great. We better not touch the line then if all works 🙃

PR #1368, opened 2022-04-27
Making error message more generic

### LINKED ISSUE #1367: 'Failed to buy datatoken from pool!' error message doesn't make sense for Free and Fixed priced assets
'Failed to buy datatoken from pool!' error message doesn't make sense for Free and Fixed priced assets.

Network error - 2022-04-27_15-15

jamiehewitt15:
Fixes #1367 

Changes proposed in this PR:

- The current error message doesn't really make any sense for free and fixed priced assets so I've replaced it with a more generic error message.


kremalicious (approved):
Ideally we would have different user-friendly messages based on different errors in background but this sentence is already way less confusing, great!

PR #1370, opened 2022-04-28
add google meta verification

mihaisc:


kremalicious:
What is the intention of this? We already have a `google-site-verification` txt entry on DNS level but I guess this only accounts for root domain:

Screen Shot 2022-04-28 at 15 15 37

PR #1371, opened 2022-04-28
Fix edit form validation

### LINKED ISSUE #1365: No validation error message displayed on edit form
No validation error message is displayed in edit form when changing input content, although the `Submit` button remains disabled:

Screenshot from 2022-04-27 13-50-24

jamiehewitt15:
Fixes #1365

Changes proposed in this PR:

- Fixing form validation messages and styles for the edit metadata form
- Removing some V3 code that wasn't used anymore

### DIFF
--- a/src/components/@shared/FormInput/index.tsx
+++ b/src/components/@shared/FormInput/index.tsx
@@ -96,14 +96,11 @@ export default function Input(props: Partial<InputProps>): ReactElement {
 
   const isFormikField = typeof field !== 'undefined'
   const isNestedField = field?.name?.includes('.')
-
   // TODO: this feels hacky as it assumes nested `values` store. But we can't use the
   // `useField()` hook in here to get `meta.error` so we have to match against form?.errors?
   // handling flat and nested data at same time.
-
   const parsedFieldName =
     isFormikField && (isNestedField ? field?.name.split('.') : [field?.name])
-
   const hasFormikError = checkError(form, parsedFieldName, field)
 
   const styleClasses = cx({
--- a/src/components/@shared/FormInput/index.tsx
+++ b/src/components/@shared/FormInput/index.tsx
@@ -94,33 +94,15 @@ export default function Input(props: Partial<InputProps>): ReactElement {
     disclaimerValues
   } = props
 
+  const isFormikField = typeof field !== 'undefined'
+  const isNestedField = field?.name?.includes('.')
+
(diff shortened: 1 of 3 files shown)


EnzoVezzaro:
Found some issue:

- validation on description doesn't have the correct border color (is not red when invalid as it should)
- validating Sample File breaks the app

https://user-images.githubusercontent.com/13741335/165929718-debc314c-e08b-44a9-aee0-3d9d47b903be.mp4

jamiehewitt15:
@EnzoVezzaro that's interesting, the border colour is red for me:

edit form validation - 2022-04-29_13-48

The file validation works if you use a valid asset URL but seems to break if it's not valid. I'll fix that.

EnzoVezzaro:
> @EnzoVezzaro that's interesting, the border colour is red for me:
> 
> edit form validation - 2022-04-29_13-48
> 
> The file validation works if you use a valid asset URL but seems to break if it's not valid. I'll fix that.

Yeah, sorry. I wasn't clear enough. Basically, on the title input the red border stays invalid even if you're in focus. The description field instead has the border invalid only if you're not in focus. If the description field is invalid and you're in focus on it, it shows the normal color border instead of the red one like in the title.

jamiehewitt15:
> The description field instead has the border invalid only if you're not in focus. If the description field is invalid and you're in focus on it, it shows the normal color border instead of the red one like in the title.

Ok yeah, I see what you mean now. I just checked the publish page on the v4 branch and the same thing is happening there - the border turns black when in focus. Seems that there must be some other issue causing that.

red title black border - 2022-04-29_15-03

claudiaHash (approved):
lgtm!

EnzoVezzaro:
> > The description field instead has the border invalid only if you're not in focus. If the description field is invalid and you're in focus on it, it shows the normal color border instead of the red one like in the title.
> 
> Ok yeah, I see what you mean now. I just checked the publish page on the v4 branch and the same thing is happening there - the border turns black when in focus. Seems that there must be some other issue causing that.
> 
> red title black border - 2022-04-29_15-03

It's a CSS related issue. It's caused by [this line](https://github.com/oceanprotocol/market/blob/feature/alert-story/src/components/@shared/FormInput/InputElement.module.css#L21:L21). The correct `InputElement_input` class get's applied, but the CSS overrides it.

EnzoVezzaro (approved):

PR #1375, opened 2022-04-29
Fix order and price

mihaisc:


claudiaHash (approved):

EnzoVezzaro:
Loading issue is back. The button keeps updating it's state (this was disabled as it was prompt to issues). When you start a transaction and wait until the timer kicks in, the button gets restored even though the transaction has not finished yet

https://user-images.githubusercontent.com/13741335/165952230-5f00f370-7725-4622-b60a-6c51286d2a7f.mp4

mihaisc:
yep, i'll check it out

EnzoVezzaro (approved):

kremalicious (approved):
awesome, good changes, could not break it. I still think the button should not get a loading state as it is the price which gets fetched so this should get loading state. Now I get a weird UX where buy button appears disabled, then gets replaced with spinner, then gets back in disabled state, then becomes active. But all working is more important right now

kremalicious:
price UX moved to #1378

PR #1377, opened 2022-04-29
Converting accountId and owner to lowercase for the comparison to avoid problems caused by inconsistent cases

jamiehewitt15:
Fixes #1377

Changes proposed in this PR:

- Converting accountId and owner to lowercase for the comparison to avoid problems caused by inconsistent cases

### DIFF
--- a/src/components/Asset/Edit/index.tsx
+++ b/src/components/Asset/Edit/index.tsx
@@ -39,7 +39,7 @@ export default function Edit({ uri }: { uri: string }): ReactElement {
 
   return asset &&
     asset?.accessDetails &&
-    accountId.toLowerCase() === owner.toLowerCase() ? (
+    accountId?.toLowerCase() === owner?.toLowerCase() ? (
     <Page title={pageTitle} noPageHeader uri={uri}>
       <div className={styles.container}>
         <Tabs items={tabs} defaultIndex={0} className={styles.edit} />
@@ -47,9 +47,7 @@ export default function Edit({ uri }: { uri: string }): ReactElement {
     </Page>
   ) : asset &&
     asset?.accessDetails &&
-    accountId &&
-    owner &&
-    accountId.toLowerCase() !== owner.toLowerCase() ? (
+    accountId?.toLowerCase() !== owner?.toLowerCase() ? (
     <Page title={pageTitle} noPageHeader uri={uri}>
       <Alert
         title="Edit action available only to asset owner"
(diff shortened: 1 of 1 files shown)


src/components/Asset/Edit/index.tsx
```diff
@@ -37,23 +37,23 @@ export default function Edit({ uri }: { uri: string }): ReactElement {
     }
   ].filter((tab) => tab !== undefined)
 
-  return asset && asset?.accessDetails && accountId === owner ? (
+  return !asset || !accountId || !owner || !asset?.accessDetails ? (
```
kremalicious, line 40: we should not add another set of crazy long clauses here, and continue to use the loader as the fallback only so we don't have to add every possible condition here. All those conditions should live on the next clause to define if edit should be shown or not.

jamiehewitt15:
Actually, after a bit more testing it looks like this issue is caused by inconsistent capitalisation between the accountId / owner account

Account Id - 2022-05-02_13-01

claudiaHash (commented):
After entering on an asset which I own page without a wallet connected, and then I connect my wallet, I got this error:

Screenshot from 2022-05-02 14-40-23

jamiehewitt15:
> After entering on an asset which I own page without a wallet connected, and then I connect my wallet, I got this error:

@claudiaHash thanks for pointing that out. Should be working ok now. If you're on the edit page and disconnect your account you should see the "Edit action available only to asset owner" error message

claudiaHash (approved):
LGTM

src/components/Asset/Edit/index.tsx
jamiehewitt15: I've removed this

src/components/Asset/Edit/index.tsx
```diff
@@ -37,13 +37,17 @@ export default function Edit({ uri }: { uri: string }): ReactElement {
     }
   ].filter((tab) => tab !== undefined)
 
-  return asset && asset?.accessDetails && accountId === owner ? (
+  return asset &&
+    asset?.accessDetails &&
+    accountId?.toLowerCase() === owner?.toLowerCase() ? (
```
EnzoVezzaro, line 42: I guess is fine, but why are we getting this ID formatted differently?

EnzoVezzaro (approved):

src/components/Asset/Edit/index.tsx
jamiehewitt15: I'm not really sure to be honest

PR #1382, opened 2022-05-02
Removing "no price set message"

jamiehewitt15:
### Issue
For a brief moment before the prices have loaded the assets show "No price set" and then it's replaced by the actual price. This is distracting and confusing. 

Screenshot from 2022-05-02 13-41-20

The same thing also happens on the asset detail page. Showing "No price set" when it's actually just loading. This creates a poor user experience and the initial impression is that it looks like it's not working.   

Screenshot from 2022-05-02 13-47-21

### Changes proposed in this  PR

- Removing the "No price set"  message. Given the new publish flow we shouldn't really have any assets without a price.

### DIFF
--- a/src/components/@shared/Price/index.tsx
+++ b/src/components/@shared/Price/index.tsx
@@ -18,7 +18,6 @@ export default function Price({
   conversion?: boolean
   size?: 'small' | 'mini' | 'large'
 }): ReactElement {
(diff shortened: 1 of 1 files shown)


jamiehewitt15:
This is being delt with in https://github.com/oceanprotocol/market/pull/1386

Issue #1378, opened 2022-04-29
Better price loading UX


kremalicious:
We have this weird UX where buy button appears disabled, then gets replaced with spinner, then gets back in disabled state, then becomes active. The loading state should be only used for the actual order flow, not for data fetching.

If we fetch the price then logically the price should get a loading state, albeit a spinner is not the best solution here, maybe dimmed price which flickers until final price is retrieved gives better impression for users to what exactly is loading. And during this price loading time, the buy button stays disabled.

Another UX problem is the difference between spot & consume price which becomes obvious to users when browsing between asset teaser and asset details. From user perspective would be better to always show the final consume price, but that's tricky technically. To combat that we should display tooltip and list fees on all prices in asset details view.

mihaisc:
> Another UX problem is the difference between spot & consume price which becomes obvious to users when browsing between asset teaser and asset details. From user perspective would be better to always show the final consume price, but that's tricky technically. To combat that we should display tooltip and list fees on all prices in asset details view.

We discussed this in the past. It is not really accurate. On the teasers it's ok to show the spot price because that's a more relevant information for both the person that wants to buy the asset and the person that wants to add liquidity or do pool related stuff. The buy price is not just the spot price + fees, it's the price to buy 1 dt  (not necessarily  equal to spot price)  + fees. 

There is already a fee related issue https://github.com/oceanprotocol/market/issues/1082 .

Agree we need to rethink the button interactions. Maybe always show the button and under it or somewhere else show what is happening (spinner + message)

jamiehewitt15:
This also comes under the topic of better price loading UX... For a brief moment before the prices have loaded the assets show "No price set" and then it's replaced by the actual price. I think a better placeholder would be for it to say "Loading price" or even nothing. There shouldn't be any assets without prices in the V4 market

![Screenshot from 2022-05-02 13-41-20](https://user-images.githubusercontent.com/20739535/166222499-98383fea-5542-4a9e-9cf3-e93fb2b5188c.png)

jamiehewitt15:
Same thing also happens on the asset detail page. Showing "No price set" when it's actually just loading  

![Screenshot from 2022-05-02 13-47-21](https://user-images.githubusercontent.com/20739535/166222746-8a8ab488-858d-48ae-92ea-51acedee3746.png)

jamiehewitt15:
Another improvement would be to have just a single loading icon and message. Here we have two messages saying "Retrieving price" and "Refreshing price":

[...]

PR #1386, opened 2022-05-03
fix loader and "Not price set" message

### LINKED ISSUE #1384: Stucked loader spinner on assets with no price
**Describe the bug**
I found that the assets with no price are not loading correctly in the homepage, as well as presenting some issues on the details page.

**To Reproduce**
1. Go to Homepage
2. Click on any asset with the stucked loader spinner
3. Open console

**Screenshots**
Screen Shot 2022-05-03 at 6 39 43 AM

Screen Shot 2022-05-03 at 7 07 48 AM

EnzoVezzaro:
- Fixes #1384 

Changes proposed in this PR:

- stop loader spinner on assets without prices
- UI fixes when loading/downloading assets


src/components/@shared/Price/index.tsx
```diff
@@ -27,7 +27,7 @@ export default function Price({
       conversion={conversion}
       type={accessDetails.type}
     />
-  ) : !accessDetails || accessDetails?.type === '' ? (
+  ) : !accessDetails || !accessDetails?.price || accessDetails?.type === '' ? (
```
jamiehewitt15, line 30: The logic here needs a rethink. It's already creating a poor user experience where the "No pricing mechanism has been set on this asset yet." message is temporarily shown for all assets until the price has loaded. It's distracting and creates a poor first impression. As described [here](https://github.com/oceanprotocol/market/issues/1378#issuecomment-1114716729)

EnzoVezzaro:
@jamiehewitt15 I just pushed some changes to address some of the concern. Now, you won't see the "No price set" label while loading. Hopefully I didn't messed up something else 😬 please take some time to test this PR carefully as it touches some critical components, and it might required some more work.

src/components/@shared/AssetList/index.tsx
```diff
@@ -52,14 +52,15 @@ export default function AssetList({
...
     if (!assets) return
     setAssetsWithPrices(assets as AssetExtended[])
-    setLoading(false)
+    setLoading(true)
     async function fetchPrices() {
       const assetsWithPrices = await getAccessDetailsForAssets(
         assets,
         accountId || ''
       )
       if (!isMounted()) return
```
jamiehewitt15, line 61: If it returns here, it will be permanently loading

src/components/@shared/AssetTeaser/AssetTeaser.tsx
```diff
@@ -51,7 +54,11 @@ export default function AssetTeaser({
           </div>
 
           <footer className={styles.foot}>
-            <Price accessDetails={asset.accessDetails} size="small" />
+            <Price
+              accessDetails={asset.accessDetails}
+              loading={loading}
```
jamiehewitt15, line 59: If we take this approach here, then we may as well do the same on the asset detail page so that the "No price set" message doesn't pop up there

jamiehewitt15 (commented):
Looks good, seems to work well on the preview link. I've just left a few comments

src/components/Asset/AssetActions/Consume.tsx
```diff
@@ -20,7 +20,7 @@ import { useIsMounted } from '@hooks/useIsMounted'
 import { usePool } from '@context/Pool'
 import { useMarketMetadata } from '@context/MarketMetadata'
 
-export default function Download({
+export default function Consume({
```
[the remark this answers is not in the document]
mihaisc: don't change it :)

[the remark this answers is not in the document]
EnzoVezzaro: I changed it because I found that in v4 we call this component's [context](https://github.com/oceanprotocol/market/tree/v4/src/components/Asset/AssetActions/index.tsx#L123:L123) as "Consume" instead of download. I just thought "Consume" makes more sense than download, but if not, we can just change all to "Download".

src/components/@shared/AssetTeaser/AssetTeaser.tsx
EnzoVezzaro: mmmm.... what do you mean? Do you see any "No price set" on the detail page? You shouldn't 

https://user-images.githubusercontent.com/13741335/166895382-c7064545-8dfd-4814-800e-c8ddb072be7c.mp4

src/components/@shared/AssetList/index.tsx
EnzoVezzaro: We can actually remove this line as it doesn't really affects anything, but I'm not sure what you're referring to. How do we replicate this case?

src/components/Asset/AssetActions/Consume.tsx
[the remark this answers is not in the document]
mihaisc: That is just an old reference, it should have been Download there as well.

src/components/@shared/AssetTeaser/AssetTeaser.tsx
jamiehewitt15: Yeah, I'm seeing it show "No price set" for a second before the price loads

https://user-images.githubusercontent.com/20739535/166912417-1c868cdb-185e-4da9-8ace-d2367ec2478a.mp4

EnzoVezzaro: > Yeah, I'm seeing it show "No price set" for a second before the price loads
> 
>  Berlin.Fish.Fauna.2013.-.Ocean.Market.-.5.May.2022.mp4

Interesting. I'll take a look at it as you shouldn't see it.

src/components/@shared/AssetList/index.tsx
jamiehewitt15: Yeah, it's not something easy to replicate. I was just thinking about the implications of setLoading being set to true before this while previously it was false, potentially that could mean people see the loading spinner for longer in some situations

src/components/@shared/AssetTeaser/AssetTeaser.tsx
jamiehewitt15: Ok cool, thanks

EnzoVezzaro: pushed fix now

jamiehewitt15 (approved):
Looks good to me

src/components/@shared/AssetList/index.tsx
```diff
@@ -52,14 +52,15 @@ export default function AssetList({
   useEffect(() => {
     if (!assets) return
     setAssetsWithPrices(assets as AssetExtended[])
-    setLoading(false)
+    // setLoading(true)
```
kremalicious, line 55: this makes all AssetList show the loader way longer to user than it should be, as by removing this assets will only be visible after the price has been fetched. Everything else to display AssetTeaser we have before fetching the prices

src/components/@shared/AssetList/index.tsx
kremalicious: This line's goal is to prevent setting state on a component which is not rendered on screen anymore, leading to memory leaks. This can happen when users load front page and while app is fetching data, user browses to another route, unmounting the component. Without this line, React will then try to update state on the unmounted component

kremalicious:
in general I would suggest taking a step back from this as by trying to fix things this PR makes some things worse, like more than doubling the time it takes before users see assets.

Also can't see how plastering loading spinners everywhere is any better than before, if you have 3 loading spinners in this small space, then something is wrong conceptionally, this is what I get with this PR:

test

Often it is better to show nothing in place of the thing you are fetching, and just make sure that the empty space takes same space as the filled element. Which could be used for the price alone, just like we have right now on `v4` on all AssetTeaser. More ideas are on #1378

So would suggest on #1378 communicate visually first those states

1. AssetTeaser without price loaded 
2. AssetDetail without price loaded
3. AssetDetail with spot price loaded
4. AssetDetail with final consume price loaded

And then start implementing which in turn should fix multiple issues instead of just adding more spinners

EnzoVezzaro:
> Better price loading UX #1378

Agree, we should closed this issue and continue in #1378

PR #1389, opened 2022-05-05
Atomic components stories

### LINKED ISSUE #1399: Basic Components: Atoms
Add stories for the following components 

- [x] Button (case example)
- [x] Alert
- [x] Badge
- [x] Blockies 
- [ ] Box
- [x] Container
- [x] Copy
- [x] List
- [x] Loader
- [x] Logo
- [x] Modal
- [x] Status
- [ ] Table
- [x] Tabs
- [x] Tags
- [x] Time
- [x] Tooltip

claudiaHash:
Changes proposed in this PR:

- stories added for the components under `/atoms` folder
- make storybook use webpack 5, mimic Next.js webpack config rules
- closes #1399

##### Atoms
- [x] Button (case example)
- [ ] Alert
- [x] Badge
- [x] Blockies 
- [ ] Box
- [x] Container
- [ ] Copy
- [x] List
- [ ] Loader
- [X] Logo
- [ ] Modal
- [x] Status
- [ ] Table
- [ ] Tabs
- [x] Tags
- [x] Time
- [ ] Tooltip


EnzoVezzaro:
We need to change the `state` input for a dropdown with options and remove the class field. Please also check if there's any text on this component.

Screen Shot 2022-05-09 at 6 37 25 AM

claudiaHash:
> @claudiaHash We need to change the `state` input for a dropdown with options and remove the class field. Please also check if there's any text on this component.

I just added a fix, please check this change when you can. No, this component has no text, it simply displays a circle/ square/ triangle associated with the state

EnzoVezzaro:
> > @claudiaHash We need to change the `state` input for a dropdown with options and remove the class field. Please also check if there's any text on this component.
> 
> I just added a fix, please check this change when you can. No, this component has no text, it simply displays a circle/ square/ triangle associated with the state

Looks good ✌🏽 great!

EnzoVezzaro:
I think we can add a Button here which onClick it triggers the Alert. What do you think?

Screen Shot 2022-05-09 at 6 59 08 AM

EnzoVezzaro:
> I think we can add a Button here which onClick it triggers the Alert. What do you think?
> 
> Screen Shot 2022-05-09 at 6 59 08 AM

or wait. It should be render as the Market's.

claudiaHash:
> > I think we can add a Button here which onClick it triggers the Alert. What do you think?
I was just Working on this :) Sure is the best way to show the component's behaviour

EnzoVezzaro:
> > > I think we can add a Button here which onClick it triggers the Alert. What do you think?
> 
> I was just working on this :) Sure is the best way to show the component's behaviour. What do you mean by rendering as Market's?

what I meant is that we need to make it renders like in the market. This component Alert should be rendered like this in Storybook:

Screen Shot 2022-05-09 at 7 43 14 AM

This for example is what we use [here](https://github.com/oceanprotocol/market/blob/feature/alert-story/src/components/Asset/Edit/index.tsx#L52:L52) in edit. Take this as reference.

kremalicious:
pushing merge conflict resolution in a sec., to then have a look at this Alert test we talked about

src/@utils/accessDetailsAndPricing.ts
```diff
@@ -16,7 +16,7 @@ import { AccessDetails, OrderPriceAndFees } from 'src/@types/Price'
 import Decimal from 'decimal.js'
 import { consumeMarketOrderFee } from '../../app.config'
 
-const TokensPriceQuery = gql`
+const TokensPriceQueryOp = gql`
```
kremalicious, line 19: what was the reason for renaming those?

src/components/@shared/ButtonBuy/index.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { FormEvent, ReactElement } from 'react'
-import Button from '../atoms/Button'
+import Button from '../atoms/Button/index'
```
kremalicious, line 2: do we really have to do this with the `/index` at the end? Cause then we have to touch almost every file in the codebase

src/components/@shared/FormFields/FilesInput/Info.tsx
```diff
@@ -3,6 +3,7 @@ import { prettySize } from './utils'
 import cleanupContentType from '@utils/cleanupContentType'
 import styles from './Info.module.css'
 import { FileMetadata } from '@oceanprotocol/lib'
+import Alert from '@shared/atoms/Alert/index'
```
kremalicious, line 6: why add this import in this file?

src/@utils/accessDetailsAndPricing.ts
EnzoVezzaro: at same point we have an error when compiling because it was named the same as the import in [line 8](https://github.com/oceanprotocol/market/blob/da7ccf392add5ad996e4dba16873e2f409215bdb/src/%40utils/accessDetailsAndPricing.ts#L8)

src/components/@shared/FormFields/FilesInput/Info.tsx
EnzoVezzaro: yeah, not need to. Issue already raised, we have fix some of them, but there's a still a couple to remove.

kremalicious:
As I'm looking at Alert component, noticed we are using for multiple components a test file and a story. Think that's overkill, goal of running stories as unit tests was to remove need to write a unit test file when a component already has a story, as the story is also rendering all the states.

Then noticed we maintain a webpack 4 config for storybook, and webpack 5 for Next.js. Unified this so all is webpack 5 now, as with this most image asset issues should be gone without downgrading anything

kremalicious:
Did we find a way to actually tell which story produces an error? As this is quite unhelpful and the more stories we have, the harder it would be to figure out which story failed. Ideally we get every story as individual test displayed, instead of 1 test for all stories and when then one of them fails we have no clue which one

Screen Shot 2022-05-10 at 12 15 13

src/@utils/accessDetailsAndPricing.ts
kremalicious: alright, makes sense, wanted to find a way to not touch this file to get the CodeClimate errors away which are all reported cause we touched this file here

kremalicious:
I move to a new PR only dealing with test setup changes as this PR is already weirdly big although it should only touch things in `atoms/` so don't want to add more to the confusion. Definetly our story shots are not working as intended

Like, when just [following instructions I find here](https://storybook.js.org/addons/@storybook/addon-storyshots) for using with react-testing-library, I get proper split up test file reporting right away:

Screen Shot 2022-05-10 at 12 37 34
 
But doing this setup-only thing in a new branch so would suggest holding off on this PR for now as it tries to solve too many things at once and then we end up with more test files we actually need

kremalicious:
so my solution to a big PR was just another big PR #1415 🙃

After that one is merged, this PR here should be greatly reduced and can concentrate on only adding tests for the atoms, for most part only touching files in that folder. It might be easier to branch off fresh from `v4` after #1415 is merged, and copy over selected changes.

claudiaHash:
Working right now on a creating a new branch from v4 ( where Matthias merged the #1415 ), I'll move there the stories for components which are checked in the description and then we'll continue on that branch. Is this approach ok , @kremalicious @EnzoVezzaro ?

kremalicious:
yup, sounds great! Those merge conflicts on here do not look fun

claudiaHash:
[Here](https://github.com/oceanprotocol/market/pull/1422) is the new PR for adding stories/ tests for atomic components. I think we can close this one and continue there.

PR #1393, opened 2022-05-05
add decimals to pool calls

### LINKED ISSUE #1390: pass decimals to all methods with excessive contract calls
Following [this change over in ocean.js](https://github.com/oceanprotocol/ocean.js/pull/1424) we should be able to dramatically limit all the excessive conrtract calls the app is doing by passing the exact decimal values from the app instead of having ocean.js fetch them all over again.

Requires ocean.js `v1.0.0-next.39`

mihaisc:
fixes #1390 

For pool tokens i hardcoded 18, no need to bother with it.


mihaisc:
I don't get it. I cleared everything locally, reinstalled, it's always building ok.

kremalicious:
I can replicate the build error locally after checking out this PRs branch:

1. `nvm use`
2. `npm i`
3. `npm run build`

Not sure what's going on though, maybe some build tool changed over on ocean.js between the versions?

kremalicious:
So I would assume Next.js tries to use `./dist/lib.modern.js` as we instruct it to do in `package.json`, but there's now only `./dist/lib.modern.mjs` over in ocean.js:

Screen Shot 2022-05-05 at 16 21 39

mihaisc:
good catch... the wonders of minor changes

kremalicious:
Not sure this works though, just wanted to test quick fix first. But locally I now get new import errors during build process which might require more tweaks in ocean.js. If in doubt, we can always downgrade `microbundle` for now but should be last resort

mihaisc:
no dice

mihaisc:
I think the problem is still web3 and its deps

kremalicious:
nope, it's all js ecosystem transitioning to ESM stuff. Came very close to solution in combination of ocean.js & market locally but would now have to use Node.js 17.5+ (which our deployment tools don't support) or run all our scripts with `--experimental-json-modules` to move forward. So locking microbundle in ocean.js to earlier version now seems like most feasible solution

kremalicious:
think I got it for the remove screen, some issue with that `useRef`/`debounce` not having access to outside state variables for whatever reason, so just passing them all as a fix. Needs more testing

kremalicious (approved):

mihaisc:
Looks good

EnzoVezzaro:
I notice that "Your Liquidity" is not working. It stays always at 0.
https://user-images.githubusercontent.com/13741335/167401904-6ce55d41-3577-4174-b504-69c6d65aa804.mp4

In v4 is fine
Screen Shot 2022-05-09 at 7 36 52 AM

Minor fix - not really related but nice to have:
when you clear the input field, this label should indicate the min amount required, not the maximun as now (`Must have maximum 6 decimal digits`)
Screen Shot 2022-05-09 at 7 24 56 AM

claudiaHash:
I got some errors when accessing data sets  with fixed price:
Screenshot from 2022-05-09 14-39-01

mihaisc:
looking into it

claudiaHash (approved):
LGTM! couldn't reproduce the former issues

mihaisc:
hope for the best :)

PR #1410, opened 2022-05-09
ENS integration, the right way

### LINKED ISSUE #1388: wrong ENS name if account owns multiple ENS names
When using an account which has multiple ENS names registered under its account, we show one of them, but not the actual reverse record.

- account 0xf50f267b5689b005fe107cfdb34619f24c014457
- has `kremalicious.eth` set as reverse record
  Screen Shot 2022-05-05 at 12 06 19

- some other ENS domain shows up instead, one which is owned by this account, but is not set as reverse record:
  Screen Shot 2022-05-05 at 12 00 01

### Expectation

Only the ENS name which is set as reverse record for an address is displayed.

### Possible solution

For performance reasons we query an ENS subgraph, this might only hold infos about registered domains for an account, but not for its actual reverse record. Look if this is actually possible with this implementation.

Alternatively, replace this mechanism with the commented out code which does actual blockchain query to find and verify reverse record.

### LINKED ISSUE #1396: ENS for all identity metadata
- kick out 3box
- replace all fields we use from 3box with respective fields from ENS
- most likely fix #1388 along the way

### LINKED ISSUE #1623: Profile images in publisher teasers not woking for some users
image

in my network tab there are some infura ipfs calls that are failing 
image
ex: https://infura-ipfs.io/ipfs/QmWHkFZvQwGgw2WMwfpC91ofG8qdvTsUYbh8BX8dhA1Bp2
Not sure if they are relevant

### LINKED ISSUE #1662: Integrate ENS Proxy
https://github.com/oceanprotocol/ens-proxy

kremalicious:
- closes #1396
- fixes #1388
- fixes #1623
- closes #1662
- uses https://github.com/oceanprotocol/ens-proxy

### Pretext

Ideal case would be to use `wagmi` hooks + ethers.js and out of the box we would not have any problems. Web3.js does not support reverse lookups yet (address → ENS Name) so have to use ens.js library, which they suggest to replace with ethers.js in future.

In any case, using the ENS subgraph is not the proper way for getting ENS names, as we can't reliably get reverse records with it, and we can't get any content or text record entries.

### Solution

- most of the things described below have been moved to an [ENS proxy](https://github.com/oceanprotocol/ens-proxy), and this PR fetches data from there
- kick out all 3Box and custom profile stuff, recreate `Profile` schema based on what we get from ENS
- use ens.js in one generalized method `getEnsProfile()`
- always uses ETH mainnet (`1`)
- support all kinds of avatars, including those set from NFTs, by hitting https://metadata.ens.domains/docs
- always double check name reverse resolution with forward resolution [as suggested by ENS](https://docs.ens.domains/dapp-developer-guide/resolving-names#reverse-resolution)
- fetch set text records before doing contract calls for them, filtered by hardcoded key list
- ...

### Fixes

- refactor `Blockies` component into generic presentational `Avatar` component, which uses Blockies as fallback, 100% test coverage for it
- `npm audit fix` run, as ens.js introduces some vulnerable packages
- Rename and move to folder `PublishersWithMostSales.tsx → TopSales/index.tsx`, also consolidate components and typings for it (no more `viewModels/AccountTeaserVM.d.ts`)


kremalicious:
Functionality as POC done to see what's possible.

ENS records:
Screen Shot 2022-05-09 at 18 14 09

Our profile page in this PR created from them:
Screen Shot 2022-05-09 at 18 14 16

Also used in wallet:

Screen Shot 2022-05-09 at 18 15 32

This is all fine for asset details & profile page, as we only ever fetch one profile on there. Tricky part are the asset teasers, making loading our frontpage hit our Infura endpoint pretty hard, options would be:

1. run with it and hope our Infura endpoint can handle it
2. remove publisher line from asset teasers
3. create proxy with cached responses as suggested in #1388, Vercel Serverless Function should be enough for this
4. batch call, collecting all accountIds on one screen and query in one request

Option 4. is not possible yet with ens.js, ethers.js might support it. If we go with option 3., then option 4 becomes possible too for limiting requests to that endpoint.

EnzoVezzaro:
I've been having mix results. Most of the times works, but on some refresh, it doesn't. 

- Here worked on the account information, but it didn't on the profile page
Screenshot 2022-07-05 at 10 26 58

- Here didn't work for both
Screenshot 2022-07-05 at 10 30 07

- Here worked on the account information, but it didn't on teaser list
Screenshot 2022-07-05 at 10 41 39

Most of the time works, you need to keep refreshing to break it.

kremalicious:
Something weird with profile page, this works when just browsing to that page, but full reload breaks it

mihaisc:
Don't do a full reload :D , already a bug for this #1572

AnaLoznianu:
@kremalicious On the main [readme](https://github.com/oceanprotocol/market/blob/main/README.md) we have references for the 3box support. Let's cleanup this one as well

kremalicious:
pushed some profile loading tweaks. Turned out that `ens.name(ensName).getAddress()` will return `0x0000000000000000000000000000000000000000` initially for whatever reason, leading to cool info like this in console:

Screen Shot 2022-07-06 at 14 16 08

We were also fetching everything else on profile page before getting the correct address so now this all works as expected, called once and in the right order:

Screen Shot 2022-07-06 at 14 16 37

This then solves Trent's profile page for me. And should also solve some issues when page is reloaded, or the initial landing page.

> the avatar image fallback doesn't seem to work for me

this looks more like the image attached to your ENS domain is not reachable. What does `https://metadata.ens.domains/mainnet/avatar/YOUR_ENS_NAME` return for you?

EnzoVezzaro:
> this looks more like the image attached to your ENS domain is not reachable. What does `https://metadata.ens.domains/mainnet/avatar/YOUR_ENS_NAME` return for you?

Exactly. More than not reachable, has [not been set](https://metadata.ens.domains/mainnet/avatar/vezzaro.eth). This is a newly created account, so it's empty for now. 

`
{
"message": "There is no avatar set under given address"
}
`

kremalicious:
Thanks, got it! So then indeed the blockies fallback is messed up in wallet alone for whatever reason.

As for Trent's profile, all correct, there is no primary ENS name set for that account. But it doesn't crash anymore so that's the cool new feature 😀

EnzoVezzaro:
On my end the image in the top bar still is missing. Other than that, the rest seems fine 👍🏽

Screenshot 2022-07-28 at 05 43 50

[ENS account](https://metadata.ens.domains/mainnet/avatar/vezzaro.eth)

EnzoVezzaro:
It's working as expected. Just a small fix on the profile image, if missing we might want to keep the avatar instead. 

Screenshot 2022-09-22 at 10 54 04

EnzoVezzaro (approved):

Issue #1409, opened 2022-05-09
coverage reporting not working


kremalicious:
Our `coverage` job runs successfully but with weird values in the end (we don't use `yarn`):

<img width="437" alt="Screen Shot 2022-05-09 at 14 47 57" src="https://user-images.githubusercontent.com/90316/167425306-e86f2db7-494d-420e-a41b-14d0fcc18250.png">

In any case, no coverage is reported anywhere, neither on CodeClimate or in console or wherever, and goal is to have something show up, which then also gives us coverage diff comments on all PRs.

<img width="482" alt="Screen Shot 2022-05-09 at 14 54 52" src="https://user-images.githubusercontent.com/90316/167425700-05014710-32da-4d63-b696-1c9e36c2868b.png">

Another ting worth trying is to generate coverage in the `coverage` job, as the multiple overwrites of our coverage artifacts (because `npm test` runs as matrix multiple times, also uploading artifacts multiple times) might lead to problem here.

kremalicious:
this is how it should look like, from [one of my own projects](https://github.com/kremalicious/portfolio/runs/6341173734?check_suite_focus=true).

<img width="714" alt="Screen Shot 2022-05-09 at 18 48 05" src="https://user-images.githubusercontent.com/90316/167467762-fc78d4bb-3b81-4ae7-bca0-816dbaca22ff.png">

Also seems we are using slightly older version of that action, maybe what we encounter was a bug in older version of action

EnzoVezzaro:
> bumped to v2.7.5 over in #1415 and right away getting actual error:
> 
> <img alt="Screen Shot 2022-05-10 at 14 26 28" width="1117" src="https://user-images.githubusercontent.com/90316/167639344-3f5552f7-dd3b-4e76-9b02-4ec05bc93a40.png">
> 
> Update: never mind, we ran into this already as it looks like

Yes, exactly. This is the reason why we decided to downgrade it, but it might be actually related somehow. It would be great anyway to move this to the lastest (as today [v3.0.0](https://github.com/paambaati/codeclimate-action/releases))

kremalicious:
first step, we now have correct coverage reporting in the console with #1415:

<img width="577" alt="Screen Shot 2022-05-10 at 15 41 01" src="https://user-images.githubusercontent.com/90316/167655546-329c4a00-dd94-454b-a5af-4d07b21615e8.png">

[...]

PR #1415, opened 2022-05-10
Test setup tweaks

kremalicious:
- all about making `npm test` make sense and giving us metrics to follow
- storybook add-on cleanup, remove everything we don't use (only add new add-ons & tools when actually using them)
- make storybook use webpack 5, just like main app
- mimic Next.js custom webpack config to support everything main app supports (did not quickly find a way to just import what we want directly from `next.config.js` as to not duplicate things)
- rewrite Button stories, add Alert stories
- rewrite Button test to get full coverage
- bump React to v18, cause most recent versions of `react-testing-library` & `jest` require that. There should be no noticeable changes as we are not using any React v18 feature yet
- bump `react-testing-library` & `jest` to most recent major new versions
- remove all the props interface duplications we had in all stories, just import from component
- split up build & test into different CI jobs, so we don't have to wait for build to finish to get test results as everything runs in parallel now. `test` is also always quicker than `build`, in turn making `coverage` run earlier than before
- make coverage reporting work in console, have it collect from the correct files so we finally get a total coverage number, partially fixing #1409
- move all Jest config files under `.jest/`
- bump all ESLint packages, adapt to new rules like [`no-unsafe-optional-chaining`](https://eslint.org/docs/rules/no-unsafe-optional-chaining)
- add eslint-plugin-testing-library & eslint-plugin-jest-dom, helping in local editor when writing tests
- start storybook in quiet mode
- update docs
- test storybook build as part of CI, cause we want to make sure no updates break it assuming that when storybook build works, `npm run storybook` works too

All in all we now know this PR brings total code coverage to a whopping 0.63% 🎉 

Screen Shot 2022-05-10 at 19 46 08


kremalicious:
alright, as for visual developer tool changes, this is how I split up Button & Alert stories as a first quick test as the current labeling of `Primary` & `Secondary` is super confusing:

Screen Shot 2022-05-10 at 13 35 21

While we have those great _Controls_ to fiddle with, we need to have the automatic test runs in mind for which we need to render multiple states. I think a good rule of thumb would be to render a default state by default, and then see if component renders more elements based on props, then also render this. Like you see in above screenshot for the Alert component.

This then has the advantage of each story being also run as a single test:

Screen Shot 2022-05-10 at 13 35 36

Also, as you see in file changes, I removed the dedicated test file for the button, as we instruct with our `.storybook/storyshots.test.tsx` file to run all stories as unit tests through react-testing-library. Would suggest to only start creating dedicated test files when testing interactions and such. Storybook also has some interactions testing but not sure we can actually automate this, which is what we want.

So Storybook stories are for rendering visual states alone, with the bonus of giving us cool development tool, all the rest can be dealt with in a unit test with react-testing-library

kremalicious:
Contrary to my former comment, added back a dedicated test file for the button but completely different than before, before it was simply rerendering the button as defined in the stories file. That's basically testing nothing, as the story covers most things already.

After this PR's refactoring, using only the story, you can see in last column that some lines are not covered:

Screen Shot 2022-05-10 at 18 42 44

Those lines concern an important thing, they switch the output markup of the button to be either an `<a>` or a `<button>` depending on passing `href`/`to` prop. So this is definitely something we want to test so now a perfect candidate to test actual concerns in addition to Storybook. React testing library makes this simple and does not require any further stories important as before, as the test is an addition:

```tsx
import React from 'react'
import { render, screen } from '@testing-library/react'
import Button from './'
test('returns correct markup when href or to is passed', async () => {
  const { rerender } = render(
    <Button href="https://oceanprotocol.com">Hello Button</Button>
  )

  let button = screen.getByText('Hello Button')
  expect(button).toHaveAttribute('href', 'https://oceanprotocol.com')
  expect(button).toContainHTML('<a')

  rerender(<Button to="/publish">Hello Button</Button>)
  button = screen.getByText('Hello Button')
  expect(button).toHaveAttribute('href', '/publish')
  expect(button).toContainHTML('<a')
})

test('returns correct markup when no href or to is passed', async () => {
  render(<Button>Hello Button</Button>)

  const button = screen.getByText('Hello Button')
  expect(button).toContainHTML('<button')
})
```

Which then gives us full coverage, and a test making us actually confident it is testing real things:

Screen Shot 2022-05-10 at 19 06 18

So perfect example for not just blindly creating test files along the stories files which only duplicate things and test nothing of importance in the end.

EnzoVezzaro:
the coverage report is incredible! thank you so much for looking into this. I got what you done, now it's actually useful. 

I found just one problem so far. The app is crashing on profile.

https://user-images.githubusercontent.com/13741335/167817516-7c726853-ef76-4dca-aa21-fd70df7dfc4f.mp4

other than that, the rest looks good so far.

kremalicious:
great, the only `atom` where I had to change something beyond just code style. But honestly don't know what it wants from me here, this is what I got after TypeScript update in here:

Screen Shot 2022-05-11 at 10 40 06

So I made it like so:

Screen Shot 2022-05-11 at 10 41 39

But that now produces the error you found. Fun stuff. We are major versions behind with that `react-data-table-component` dependency used in this component but did not want to touch that as it affects too many components. So ideally we just figure out how exactly we need to pass our Pagination component to it so Typescript and dependency is happy

kremalicious:
ok, should be fixed, YOLOing my way out of it with this:

Screen Shot 2022-05-11 at 10 46 53

EnzoVezzaro:
> great, the only `atom` where I had to change something beyond just code style. But honestly don't know what it wants from me here, this is what I got after TypeScript update in here:
> 
> Screen Shot 2022-05-11 at 10 40 06
> 
> So I made it like so:
> 
> Screen Shot 2022-05-11 at 10 41 39
> 
> But that now produces the error you found. Fun stuff. We are major versions behind with that `react-data-table-component` dependency used in this component but did not want to touch that as it affects too many components. So ideally we just figure out how exactly we need to pass our Pagination component to it so Typescript and dependency is happy

Great! it's working 👍🏽 I'm curious though, why didnt it get caught on the tests? 🤔 but YOLOing solution works 😬

EnzoVezzaro (approved):

kremalicious:
well, in here there are only tests written for the Button & Alert atoms as I only wanted to deal with the setup part in here so #1389 can be reduced to only adding new tests. While those 2 tests alone cover more components than just those 2, nothing in them affects the Table atom so it stays untested. That error would have shown up through any render test which we should have after #1389

.github/workflows/ci.yml
```diff
@@ -56,6 +82,35 @@ jobs:
         with:
           name: coverage
 
-      - uses: paambaati/codeclimate-action@v2.7.1 # using 2.7.1 for issue: https://github.com/paambaati/codeclimate-action/issues/316
+      - uses: paambaati/codeclimate-action@v2.7.3 # using 2.7.3 for issue: https://github.com/paambaati/codeclimate-action/issues/235
         env:
           CC_TEST_REPORTER_ID: ${{ secrets.CC_TEST_REPORTER_ID }}
+        with:
+          debug: true
```
kremalicious, line 89: right! Should have set that already considering how much I tried with this damn action. Like the idea of trying to figure it out in here. Even CodeClimate documentation suggests using this action, but I was already thinking of coming up with quick shell or js script just using that CodeClimate binary directly. But that action just working would be preferred of course

kremalicious: I think I know what's wrong! Please update the action to most recent version but then we have to do a bit more.

kremalicious: Here's what happens right now:

- all test runs will upload and overwrite the shared artifacts folder
- those coverage reports include file paths
- in our case, the test runs on Windows always finish last, so their uploaded artifact is the final one
- our `coverage` job then runs on Ubuntu and will download that coverage artifact folder, which then when trying to convert into CodeClimate format with the binary, will fail because the file system paths are different

That's why we see this:

Screen Shot 2022-05-11 at 13 49 50

As job runs on a unix file system it can't handle windows paths like `D:\a\market`

Solution: make artifact uploads system agnostic and then download the one generated with ubuntu

EnzoVezzaro: Great! Hopefully this solves it, I'm kind of swimming in their github repo. I keep jumping from issue to issue, but without luck so far. At least now we'll have some more logs in debug, hopefully we see something useful.

kremalicious: it looks like we now finally arrived at a proper error message not caused by our test setup:

Screen Shot 2022-05-11 at 14 10 54

Indeed when browsing `coverage/` folder, the reports reference a `src/@types/subgraph/globalTypes.d.ts` but we actually don't have that. And also never access this from our source code...

EnzoVezzaro: Yeah, but this is getting longer and longer. Maybe we should take this to it's actual [issue](https://github.com/oceanprotocol/market/issues/1409) to avoid making this longer by our tests.

kremalicious: yup, just pushed last try. If that doesn't work I just roll back to get CI green without that coverage job for now

kremalicious: trying to upload/download them as artifacts too didn't work so most likely the coverage job needs a full `npm ci` + `npm run codegen:apollo`

claudiaHash (approved):

PR #1416, opened 2022-05-10
fix pool context

### LINKED ISSUE #1397: Url caching issues
The library still sometimes caches the data

mihaisc:
Fixes #1397
Guess it was our fault not the library .


kremalicious (approved):
good find, something with that `prevState` seem to not have worked as I expected it to work when splitting up everything into multiple effects. Important thing is that it works reliably

mihaisc:
It wasn't the `prevState` it was the `useEffect` that wasn't triggering, it was a bad design from the start, we were trying to update the state of an object when a prop from that same object changed.

PR #1417, opened 2022-05-10
fix community fee displayed value

### LINKED ISSUE #1414: Convert Community Fee from decimal to percentage
We need to convert the community Fee from decimal to percentage. It should show 0.1% instead of 0.001%

image

claudiaHash:
Fixes #1414  .

### DIFF
--- a/src/components/Publish/Pricing/Fees.tsx
+++ b/src/components/Publish/Pricing/Fees.tsx
@@ -44,13 +44,13 @@ export default function Fees({
   pricingType: 'dynamic' | 'fixed'
 }): ReactElement {
   const [field, meta] = useField('pricing.swapFee')
-  const [communityFee, setCommunityFee] = useState<string>('')
+  const [oceanCommunitySwapFee, setOceanCommunitySwapFee] = useState<string>('')
   const { chainId } = useWeb3()
   const { appConfig } = useMarketMetadata()
 
   useEffect(() => {
     getOpcFees(chainId || 1).then((response: OpcFeesData) => {
-      setCommunityFee(
+      setOceanCommunitySwapFee(
         response?.swapOceanFee
           ? new Decimal(response.swapOceanFee).mul(100).toString()
           : '0'
@@ -81,10 +81,10 @@ export default function Fees({
         )}
 
         <Default
-          title="Community Fee"
+          title="Community Swap Fee"
           name="communityFee"
           tooltip={tooltips.communityFee}
-          value={communityFee}
+          value={oceanCommunitySwapFee}
         />
 
         <Default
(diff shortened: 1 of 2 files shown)


src/components/Publish/Pricing/Fees.tsx
```diff
@@ -43,13 +44,17 @@ export default function Fees({
   pricingType: 'dynamic' | 'fixed'
 }): ReactElement {
   const [field, meta] = useField('pricing.swapFee')
-  const [opcFees, setOpcFees] = useState<OpcFeesData>(undefined)
+  const [communityFee, setCommunityFee] = useState<string>('')
```
kremalicious, line 47: good thinking making variable consistent to what we display as this stays confusing in code (like, `response.swapOceanFee` is the "community fee" coming from `opcFees` query). But this raises the question what exact "community fee" this actually is and f we should update the UI label too.

Based on [our docs](https://docs.oceanprotocol.com/concepts/fees/#ocean-community-fee), there is a "Ocean Community swap fee" and a "Ocean Community order fee". If we use those 2 distinctions in the docs then we need to have this distinction in our UI too

EnzoVezzaro:
found a little thing. Maybe it's even worth doing anything as it doesn't impact anything, it's just a weird little thing 🙃 

if you're not connected, the community fee is '0'

Screen Shot 2022-05-11 at 9 42 57 AM

EnzoVezzaro:
found a little thing. Maybe it's not even worth doing anything as it doesn't impact anything at all, it's just a weird thing to see 🙃

if you are not connected, the amount shown is '0'

Screen Shot 2022-05-11 at 9 42 57 AM

claudiaHash:
> if you are not connected, the amount shown is '0'

hmmm, don't think it's introduced by the changes done here since I can reproduce it on main v4, but looking into fixing this

src/components/Publish/Pricing/Fees.tsx
mihaisc: yep , we need both swap and consume (this is how they are called in the sub query)

mihaisc: aaaand i see a redesign incoming 😆

kremalicious:
> if you are not connected, the amount shown is '0'

first guess is that we get those fees with web3 call now, which is fine but we do not fall back to our "dummy Web3", meaning Infura. Noticed this in the ENS PR and best solution would be to adapt our Web3 context to do that automatically. Otherwise we end up having to handle the web3 fallback in multiple places. So would say this should not be done in this PR to keep it light

kremalicious:
And if we want to get this PR with original number-only fix through, we should at least update the label we show too, then handle adding the "Ocean Community order fee" in another PR, as this most likely requires layout or design changes

kremalicious (approved):

PR #1422, opened 2022-05-12
Atomic components stories

### LINKED ISSUE #1399: Basic Components: Atoms

claudiaHash:
Closes #1399  .

Changes proposed in this PR:

- stories & tests added for the components under `/atoms` folder
- refactor multiple components to split up presentation & logic
- move out radio/checkbox component from `InputElement` mega component
- reduced passing around of props in input components
- setup fixes for Storybook & Jest

##### Atoms
- [x] Button (case example)
- [x] Alert
- [x] Badge
- [x] Blockies 
- [ ] Box
- [x] Container
- [x] Copy
- [x] List
- [x] Loader
- [X] Logo
- [x] Modal
- [x] Status
- [x] Table
- [x] Tabs
- [x] Tags
- [x] Time
- [x] Tooltip


src/components/@shared/atoms/Container/index.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+    narrow: boolean
+    children: ReactNode
+    className: string
+  }
+}
+
+export const Primary: Props = Template.bind({})
+Primary.args = {
+  narrow: true,
+  children: (
```
kremalicious, line 27: What is going on here, as this is nesting under `Container` a new div which then has the same styles again as the `Container` we are already testing? `children` should just be the actual content we put in the `Container`, so something like this:

```js
children:  <>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Mauris aliquam facilisis molestie. Integer eget congue turpis, in pharetra lectus. Sed urna dolor, porttitor luctus mauris eget, lacinia consectetur eros. Duis consequat, turpis et porttitor cursus, ante lacus placerat arcu, vel pellentesque enim orci ac sem.</>
```

src/components/@shared/atoms/Container/index.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+  component: Container
+} as ComponentMeta<typeof Container>
+
+const Template: ComponentStory<typeof Container> = (args) => (
+  <Container {...args} />
+)
+
+interface Props {
+  args: {
+    narrow: boolean
```
kremalicious, line 18: this is always unnecessary duplication, just import the props interface from the component. If props aren't abstracted into exportable interface in the component, then refactor them. See how the `Button` or `Alert` stories do that

src/components/@shared/atoms/Container/index.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+      <div
+        className={styles.content}
+        dangerouslySetInnerHTML={{
+          __html:
+            'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Mauris aliquam facilisis molestie. Integer eget congue turpis, in pharetra lectus. Sed urna dolor, porttitor luctus mauris eget, lacinia consectetur eros. Duis consequat, turpis et porttitor cursus, ante lacus placerat arcu, vel pellentesque enim orci ac sem.'
+        }}
+      />
+    </>
+  ),
+  className: styles.container || 'className'
```
kremalicious, line 38: render a default state by default, meaning not passing e.g. additional `className`

An additional story could then include additional `className` with which we could test if that overwrites the existing CSS correctly but rarely we would have to output custom class names in any story

src/components/@shared/atoms/Loader/index.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+const Template: ComponentStory<typeof Loader> = (args) => <Loader {...args} />
+
+interface Props {
+  args: {
+    message: string
+  }
+}
+
+export const Primary: Props = Template.bind({})
+Primary.args = {
```
kremalicious, line 20: please not use hierarchical terms for something which has no hierarchy. We do not have a primary or secondary loader. We always have a default state of a component which is a component with only the required props passed. Then we call that `Default`. Every additional story then should get meaningful names, so in this case we could have `Default` and `WithMessage`

kremalicious:
@EnzoVezzaro could you please get [4003b25](https://github.com/oceanprotocol/market/pull/1422/commits/4003b257f459e1a15a45b640dd17ca2f71ef8454) cherry-picked into a new PR? Thanks! 

Otherwise we would have to wait for that fix until this whole PR here is finished and I would rather spend more time to really nail those atoms tests so we do not have to think about them ever again

EnzoVezzaro:
> @EnzoVezzaro could you please get [4003b25](https://github.com/oceanprotocol/market/pull/1422/commits/4003b257f459e1a15a45b640dd17ca2f71ef8454) cherry-picked into a new PR? Thanks!
> 
> Otherwise we would have to wait for that fix until this whole PR here is finished and I would rather spend more time to really nail those atoms tests so we do not have to think about them ever again

yeah, sure. [PR](https://github.com/oceanprotocol/market/pull/1428) created

src/components/@shared/atoms/Status/index.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+interface Props {
+  args: StatusProps
+}
+
+export const Default: Props = Template.bind({})
+Default.args = {}
+
+export const Warning: Props = Template.bind({})
+Warning.args = {
+  state: 'warining'
```
kremalicious, line 22: small typo:

```suggestion
  state: 'warning'
```

kremalicious, line 22: small typo:

```suggestion
  state: 'warning'
```

claudiaHash: Busted :) thanks!

kremalicious:
overall, this is already looking great!

Screen Shot 2022-05-13 at 14 46 01

`Blockies` does not ned to be touched further, that's good enough in here as #1410 refactors this into a new `Avatar` component, which is fully tested over there

claudiaHash:
> What's strange is that I actually saw the `tab` component working before, as well as the `table`'s. We might have changed something in the config that is breaking them.

yes, exactly. I'm on them right now, but I pushed the changes regarding stories because I don't think the problem comes from those files

EnzoVezzaro:
> > What's strange is that I actually saw the `tab` component working before, as well as the `table`'s. We might have changed something in the config that is breaking them.
> 
> yes, exactly. I'm on them right now, but I pushed the changes regarding stories because I don't think the problem comes from those files

Yes, I think our basic story should work (at least tabs and table). I'm on this too as well as fighting with svg testing still.

src/components/@shared/atoms/Modal/index.tsx
```diff
@@ -1,8 +1,8 @@
 import React, { ReactElement, ReactNode } from 'react'
 import ReactModal from 'react-modal'
-import styles from './Modal.module.css'
+import styles from './index.module.css'
 
-if (process.env.NODE_ENV !== 'test') ReactModal.setAppElement('#__next')
+if (process.env.NODE_ENV !== 'test') ReactModal.setAppElement('#root')
```
kremalicious, line 5: are you sure about this not breaking the modal in the app? Cause I don't see any `root` id on any markup in the production build.

src/components/@shared/atoms/Table/index.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+    {
+      name: 'Column 2',
+      maxWidth: '10rem'
+    },
+    {
+      name: 'Column 3',
+      right: true
+    }
+  ],
+  data: []
```
kremalicious, line 34: would be super handy as a bonus to fill it with some dummy data which we could store in `.jest/__fixtures__/`. Cause the table dependency needs some major version jump and having this isolated in the stories is amazing to make sure to get the update right

src/components/@shared/atoms/Time/index.stories.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react'
+import { ComponentStory, ComponentMeta } from '@storybook/react'
+
+import Time, { TimeProps } from '@shared/atoms/Time'
+import { relative } from 'path'
```
kremalicious, line 5: this import shouldn't be here

src/components/@shared/atoms/Modal/index.tsx
claudiaHash: it actually worked using the old test file, thanks for the tip!

src/components/@shared/atoms/Table/index.stories.tsx
claudiaHash: yes, sure. left it like that because I have to add somehow the `useUserPreferences` hook in here, to get rid of that type error given by the `chainIds` absence, so still on this story

src/components/@shared/atoms/Modal/index.text.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+import Modal from '@shared/atoms/Modal'
+import ReactModal from 'react-modal'
+
+describe('Modal', () => {
```
kremalicious, line 6: sure, why not, but:

1. file is named wrong so test is not run
2. we already have a story, if that covers the component already, then goal was to use `ReactModal.setAppElement` on the story, not in an extra test. But if only this test can give full coverage then keep it

src/components/@shared/atoms/Lists/index.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+interface Props {
+  args: ListItemProps
+}
+
+export const Default: Props = Template.bind({})
+Default.args = {
+  children: (
+    <>
+      <a>List Item</a>
```
kremalicious, line 23: all technically correct as we have a really weird list component here. Let's see if we can get this more useful for real life. So one of the things we usually want with lists, is to make sure the spacing between the list items stays as we want it to be. And for this we simply need multiple list items. So from pure markup perspective we should have something like this output in the stories here:


```html
<ul>
  <li>List item short</li>
  <li>List item long ipsum dolor sit amet, consectetur adipiscing elit. Mauris aliquam facilisis molestie</li>
  <li>List item long ipsum dolor sit amet, <a href="#">consectetur adipiscing elit</a></li>
</ul>
```

And then the same again for a `ol`. That somehow should be possible with storybook instead of just working with those `args`

src/components/@shared/atoms/Table/index.stories.tsx
kremalicious: right, mocking hooks in storybook... kinda the limit of storybook, it really is meant for pure UI components, could not find any builtin way it has to mock hooks or function calls. Mocking those things for Jest is quite easy though. 

So there are then only 2 possibilities:

1. don't use stories for component, but only test file and either globally or within the test mock `useUserPreferences()` returns
2. refactor component to split-up UI & data fetching in 2 components: one hook-accessing wrapper and one pure markup component which gets all data via props passed down. The UI component can then be tested with Storybook, and the wrapper component tested with Testing Library.

kremalicious: quickly tested that last thing mentioned, and wrapped our preview with almost all context providers we have so that should help in some cases to at least get the default returned values. Pushing in a sec

src/components/@shared/atoms/Lists/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+    : `${styles.item} ${styles.ulItem}`
+
+  return (
+    <li className={classes}>
+      <span>{children}</span>
+    </li>
+  )
+}
+
+export function List({ items, ordered }: ListProps): ReactElement {
```
kremalicious, line 26: mmmh, would not refactor this component just for storybook and this looks like you do just that which is a bad practice as now we are testing something which is actually not in use. If you want to create markup only for storybook, then please use their decorators https://storybook.js.org/docs/react/writing-stories/decorators

src/components/@shared/atoms/Table/index.stories.tsx
kremalicious: not sure but in last CI run, tests succeed. But the last errored storybook build simply says this:

```
ERR! Module not found: Error: Can't resolve '../../../content/networks-metadata.json' in '/home/runner/work/market/market/src/@hooks/useNetworkMetadata'
```

So, you have to have this file generated before running `storybook:build`

kremalicious: so you could have `npm run pregenerate` although we only need the last line from the script so far to get this network metadata written out. So maybe `node ./scripts/write-networks-metadata > content/networks-metadata.json` is enough

kremalicious: and for the build error you have to merge in the latest `v4` changes as we had contract & subgraph updates in the background

kremalicious: > I end up just adding the missing data into the story to be able to render the component

yes, this is the point of mocking. In storybook we never want to have our components try to load actual real data, that would never work

kremalicious: had a look into that Table component and basically the `Empty` component it uses should not have any `useUserPreferences()` access. Somebody just hacked this in there instead of constructing the message outside of the component and pass `emptyMessage` to the table component instead. None of our atoms should have any hook access, they are supposed to be pure UI components

claudiaHash: In this case, should we do the fix in the Table component first?

kremalicious: Table component is almost fully covered, only that Empty component lines are missing. Proper way would be to move out any logic around that `useUserPreferences()` access to wherever we actually want this 'No network selected' message, and then from there pass it as `emptyMessage` to component

src/components/@shared/atoms/Modal/index.tsx
```diff
@@ -1,8 +1,6 @@
 import React, { ReactElement, ReactNode } from 'react'
 import ReactModal from 'react-modal'
-import styles from './Modal.module.css'
-
-if (process.env.NODE_ENV !== 'test') ReactModal.setAppElement('#__next')
```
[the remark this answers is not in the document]
kremalicious: FYI, the solution here was not to remove this line but to prevent it from firing when run in storybook, as that would make everything fail as there is no `#__next` id. Storybook does not set `NODE_ENV=test` like a Jest run does automatically, so I added this manually to the npm command

src/components/@shared/atoms/Table/index.tsx
```diff
@@ -3,18 +3,23 @@ import DataTable, { IDataTableProps } from 'react-data-table-component'
...
   sortAsc?: boolean
   className?: string
+  chainIds: number[]
 }
 
-function Empty({ message }: { message?: string }): ReactElement {
-  const { chainIds } = useUserPreferences()
+function Empty({
+  message,
+  chainIds
```
kremalicious, line 18: we should not deal with any `chainIds` in here, this component should not be concerned about any of that and it is only used to construct a different `emptyMessage`. so all this logic should be moved over to wherever we need that 'No network selected' message

kremalicious:
So much green already!

Screen Shot 2022-05-19 at 12 51 37

Let's recap to see which ones we can still get to 100%:

- `Blockies` is fine as is, cause #1410 refactors this completely and has 100% coverage over there
- `Copy` needs a dedicated test, which just clicks on it. That then should cover the remaining lines
- `Modal` seems fine as is, not really easy way to test this `ReactModal.setAppElement` line
- `Table` could be investigated a bit more to see if we get more lines covered, with more logic removed from it, coverage should become better right away
- `Tabs` leave me clueless as to why they are not more covered, our stories render all the lines the report says are not covered. Might be we also need a dedicated test here, which then clicks around and switches tabs
- `Tooltip` would either need a story which renders the tooltip-visible-state, or dedicated test firing hover event on it

src/components/@shared/atoms/Modal/index.tsx
```diff
@@ -1,8 +1,9 @@
 import React, { ReactElement, ReactNode } from 'react'
 import ReactModal from 'react-modal'
-import styles from './Modal.module.css'
+import styles from './index.module.css'
 
-if (process.env.NODE_ENV !== 'test') ReactModal.setAppElement('#__next')
+if (process.env.NODE_ENV !== 'development' && process.env.NODE_ENV !== 'test')
```
kremalicious, line 5: which tool sets `development` to justify this change? Otherwise no reason to touch this line at all

kremalicious:
Those last bits were not easy to do without some substantial refactoring of the tested components but here we are:

Screen Shot 2022-05-19 at 20 33 59

As said above, Blockies & Modal are fine like that.

Bonus now is that we already cover more components, although this shows that our smallest components here are not fully isolated making automated testing always a challenge. By moving out the radio/checkbox component we already have full coverage of this as it's rendered in our `Tabs` stories. Same goes for `Pagination`, which I made show up in a new `Table` story, or the `Markdown` component.

We will continue running into issues when trying to write tests for our components but that's all expected, but most likely we have to continue to be creative and possibly refactor some components to split up presentation & logic

EnzoVezzaro:
Found few issues on the inputs:

https://user-images.githubusercontent.com/13741335/169501398-79161521-c77d-4b19-a600-0bcfa8a56100.mp4

EnzoVezzaro:
Tooltips too seem to have issues, they are in weird positions and became super small:

Screen Shot 2022-05-20 at 5 43 58 AM
Screen Shot 2022-05-20 at 5 43 45 AM

claudiaHash:
And checkbox as well:

Screenshot from 2022-05-20 12-46-32

kremalicious:
cool, thanks for testing!

for tooltips, I'm inclined to just roll back to what we had, but this will spill console errors in tests and on production as we are using a deprecated way of creating those animations. Rebuilt to use newer way of using `useSpring()` but the docs look good but are actually not that extensive once you want to add a custom animations. https://react-spring.io/hooks/use-spring#usespring Maybe somebody else has a more luck, if not, we can roll back and live with console errors & not fully covered component.

As for radio/checkbox, I'll try to fix them all as splitting up our mega input component is the best way forward so would rather not roll this back. But that was another fun thing, we kinda setup input elements in multiple different ways

src/components/@shared/FormInput/InputElement.tsx
```diff
@@ -106,17 +106,17 @@ export default function InputElement({
         />
       )
     case 'files':
-      return <FilesInput name={name} {...field} {...props} />
+      return <FilesInput name={props.name} {...field} {...props} />
```
kremalicious, line 109: did you try just removing `name`? Cause it's part of `props` which we spread on component anyway

EnzoVezzaro:
This seems to be fixed:

Screen Shot 2022-05-20 at 8 00 08 AM

Working on 

> And checkboxes as well edit view:
> 
> Screenshot from 2022-05-20 12-46-32

kremalicious:
boxes in edit view I made work visually in [f7ee471](https://github.com/oceanprotocol/market/pull/1422/commits/f7ee471398d3fdbc056b95c67e1b7898f7e91ff3), but there's an issue with toggling `checked` attribute. Also have this in publish form for terms checkbox

Screen Shot 2022-05-20 at 13 11 43

kremalicious:
got it with the checked state in a way, it's weird. Theoretically we should always have a `checked` attribute when radio/checkbox is selected but for whatever reason this doesn't work in all places. I changed the CSS selector for this to always deal with the `checked` attribute alone:

```css
.checkbox[checked] {}
```

Now, if I use the pseudo element selector from before:

```css
.checkbox:checked {}
```

Then it works again where it doesn't right now. 

BUT, then our tab radios behave like this where for whatever reason the `checked` attribute changes, but `:checked` pseudo class does not:

Screen Shot 2022-05-20 at 13 21 27

Always having `checked` attribute should be the way to go but not clear why Formik only handles this in some places

kremalicious:
Tooltip (and with it all our popovers like in menubar) is so fun, their `api.stop()` apparently stops animation after 1st frame so just removed that. Still, only getting our "wobbly" animation to fire on first tooltip open, consecutive opens seem to ignore our `tension` & `friction` config. But that would be already good enough for now.

Also, what a joy to use Storybook to figure this out! Amazing help for fixing UI things so far

Checkboxes/radio `checked` attribute basically does not get updated in DOM although the prop changes although we do everything right, double checked prop changes through our input component chain with React dev tools.

`props.checked` says false in virtual DOM:

Screen Shot 2022-05-20 at 14 22 26

but `checked` stays true in DOM:

Screen Shot 2022-05-20 at 14 22 16

EnzoVezzaro:
> Tooltip (and with it all our popovers like in menubar) is so fun, their `api.stop()` apparently stops animation after 1st frame so just removed that. Still, only getting our "wobbly" animation to fire on first tooltip open, consecutive opens seem to ignore our `tension` & `friction` config. But that would be already good enough for now.
> 
> Also, what a joy to use Storybook to figure this out! Amazing help for fixing UI things so far
> 
> Checkboxes/radio `checked` attribute basically does not get updated in DOM although the prop changes although we do everything right, double checked prop changes through our input component chain with React dev tools.
> 
> `props.checked` says false in virtual DOM:
> 
> Screen Shot 2022-05-20 at 14 22 26
> 
> but `checked` stays true in DOM:
> 
> Screen Shot 2022-05-20 at 14 22 16

Sorry, I was in a meeting. But true, it's so weird. It's just the DOM that is not reflecting what's in state. 🤷🏽‍♂️

EnzoVezzaro:
Tests are failing. To fix these error we need to remove one of these [parameters](https://github.com/oceanprotocol/market/blob/feature/issue1399-atoms-stories/src/components/@shared/atoms/Tabs/index.tsx#L32:L32). If we do so though, the tabs don't work as intended anymore.

Screen Shot 2022-05-23 at 5 38 19 AM

Also, using the navigation at the top of the publish widzard after changing the price crashes the app.

https://user-images.githubusercontent.com/13741335/169792511-e0f37764-bd34-4ded-a7b0-8619462d5b73.mp4

kremalicious:
Console warning ≠ test failing. Still agree we should deal with it as fix is simple.

Looking into that tab change issue

EnzoVezzaro:
> Console warning ≠ test failing. Still agree we should deal with it as fix is simple.
> 
> Looking into that tab change issue

just pushed a change to restore it. Not sure why the latest changes though.

kremalicious:
So basically on Friday I switched around the tactic in [b2d71a7](https://github.com/oceanprotocol/market/pull/1422/commits/b2d71a7cb1ccf1e93b8607abc4e6a03e731395c3) and instead of figuring out why `checked` attribute appears only sometimes, I switched back to using `:checked` selector, and rather fixed those radios on the Tab component with a lot of trial and error and pushed what I had as tests were green with it. Looks like added a bit too much new attributes then.

Can't hunk any errors anymore regarding radio usage with the latest changes you pushed 🦑

EnzoVezzaro:
Beautiful 🙌🏽 it's almost all `100%`.

Screen Shot 2022-05-23 at 6 33 18 AM

kremalicious (approved):

PR #1426, opened 2022-05-13
disable loader on compute job tab (Profile)

### LINKED ISSUE #1425: Profile: Loader in compute job tab doesn't stops
**Describe the bug**
Loader in compute job tab (profile screen) doesn't stops loading.

**To Reproduce**
1. Go to Profile
2. Click on compute job tab
3. See loader

**Expected behavior**
To finish loading at some point

**Screenshots**
Screen Shot 2022-05-13 at 7 20 09 AM

EnzoVezzaro:
- Fixes #1425 

Changes proposed in this PR:

- disable loader on get compute job

### DIFF
--- a/src/components/Profile/History/ComputeJobs/index.tsx
+++ b/src/components/Profile/History/ComputeJobs/index.tsx
@@ -86,16 +86,18 @@ export default function ComputeJobs({
     if (!chainIds || chainIds.length === 0 || !accountId) {
       setJobs([])
       setIsLoading(false)
-      return
     }
+    /*
+    // TODO: restore after fixes on compute job
     try {
       setIsLoading(true)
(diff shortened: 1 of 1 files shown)


kremalicious:
is this really necessary to do right now against `v4`? Cause compute is not supposed to work yet in `v4` branch and on top of that, this PR would create a merge conflict for #1069. If this is an actual problem in `feature/v4-c2d` branch, then this PR here should go against that branch

EnzoVezzaro:
> is this really necessary to do right now against `v4`? Cause compute is not supposed to work yet in `v4` branch so not sure for what the loader should stop. And on top of that, this PR would create a merge conflict for #1069. If this is an actual problem still present in `feature/v4-c2d` branch, then this PR here should go against that branch

This has been solved in #1069 

closing issue

PR #1428, opened 2022-05-13
Fix svgr in storybook & jest

EnzoVezzaro:
Fixes #

Changes proposed in this PR:

- fix svg loader in storybook


EnzoVezzaro:
Still not luck. Being swimming in more issue related to this. Mosthly: 

Tried documente solution: 
https://react-svgr.com/docs/jest/#2-configure-jest

Try some suggested fixes, but nothing seems to work:
- https://github.com/nrwl/nx/issues/4850
- https://github.com/gregberge/svgr/issues/83 ([here](https://github.com/gregberge/svgr/issues/83#issuecomment-1121826489))
- https://github.com/vercel/next.js/discussions/31152 

Still working on it

kremalicious:
also tinkering around on this but no luck so far. What I don't like is that the tests don't fail in the end although they clearly should. It looks like all the storyshots will succeed even if they spit out console errors.

Quickly tried minimal Logo test, and this fails successfully with the same error where the storyshots don't fail:

```tsx
import React from 'react'
import { render } from '@testing-library/react'
import Logo from './'

test('Logo', async () => {
  render(<Logo />)
})
```

So getting the test run to fail in this current state would be another goal

EnzoVezzaro:
> So getting the test run to fail in this current state would be another goal

yeah, this is cra! 😬 ok, Let's added it to list 👍🏽

EnzoVezzaro:
Still no luck on this. I tried so many way to do the same thing, but nothing seems to be working. I'm sure at the end it'll be a trivial setup issue (hopefully), but all of those issue related to the same thing is kind of discouraging though. 😪

EnzoVezzaro:
oh boy, finally. I confirm the test is passing now. Let's see what we can do about the console erros.

kremalicious (approved):
awesome! And as usual lots of research needed to get to a code change which is quite small in the end

claudiaHash (approved):
👏

PR #1464, opened 2022-05-24
Fixes

kremalicious:
- success state for lock alert
- info state for pool alert
- making sure effect works with all dependents

### DIFF
--- a/src/components/molecules/Migration/index.tsx
+++ b/src/components/molecules/Migration/index.tsx
@@ -100,7 +100,8 @@ export default function Migration(): ReactElement {
   }
 
   useEffect(() => {
-    if (!accountId) return
+    if (!accountId || !poolShares || !lockedSharesV3) return
+
     const poolSharesNumber = isNaN(Number(poolShares)) ? 0 : Number(poolShares)
     const lockedSharesNumber = parseInt(
       Web3.utils.fromWei(lockedSharesV3 || '0')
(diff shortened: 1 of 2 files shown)


src/components/molecules/Migration/index.tsx
```diff
@@ -129,7 +137,7 @@ export default function Migration(): ReactElement {
             <Alert
               title={title}
               text={message}
-              state="info"
+              state={Number(lockedSharesV3) > 0 ? 'success' : 'info'}
```
kremalicious, line 140: do we have any other way to check if lock process has been successfully done? Cause this might be problematic with decimals between 0 and 1

mihaisc: i think we just need to check if there is a pool transaction 
https://v4.subgraph.rinkeby.oceanprotocol.com/subgraphs/name/oceanprotocol/ocean-migration/graphql?query=%7B%0A%20%20poolTransactions(where%3A%20%7BuserAddressStr%3A%20%220x968d4670c346275edd5cc1a7535d14b631ffc49c%22%2C%0A%20%20%20%20poolAddress%3A%20%220xb582c3a37d09f21cf95715f759086fa4475111c6%22%7D)%20%7B%0A%20%20%20%20id%0A%20%20%20%20timestamp%0A%20%20%20%20sharesTransferAmount%0A%20%20%7D%0A%7D%0A

mihaisc (approved):
should be ok

PR #1471, opened 2022-05-24
Feat: autocomplete tags functionality

### LINKED ISSUE #664: Autocomplete tags in form
There are 2 steps here :
- create an autocomplete component 
- use the component for tags in the publish form by passing to it a list of tags. I will provide a query in this issue to fetch all the tags , but it's currently not possible in aquarius

### LINKED ISSUE #1264: Edit tags functionality
When editing the metadata of an asset, the user should be able to edit the tags as well

LucaMilanese90:
- Fixes #664 
- Fixes #1264

Changes proposed in this PR:

- get list of existing tags from aquarius
- add autocomplete component for tags field in form publish
- add edit tags functionality


EnzoVezzaro (approved):

mihaisc:
we will merge this after the launch. Until then just bugfixes

LucaMilanese90:
As mentioned in the frontend planning, the tag aggregation query is not working as expected: the result is always the same list and it doesn't seem to be aware of new tags after the publication of a new asset.

EnzoVezzaro:
> As mentioned in the frontend planning, the tag aggregation query is not working as expected: the result is always the same list and it doesn't seem to be aware of new tags after the publication of a new asset.

Yeah, new tags are not being pull correctly. We are able to create new tags ([they show](https://market-kuqvznmnu-oceanprotocol.vercel.app/asset/did:op:72468f062a26f9eb668b6392a0bbdeebd33f889b2c31e2c91768eba35a31cd94) correctly also), but their are not available in the list of tags (edit/publish). I'm sure they exists as they get pulled on the view page, but they might be getting ignore when fetching the tags.

Screen Shot 2022-05-31 at 7 17 42 AM

On edit I'm able to use it as expected:
Screen Shot 2022-05-31 at 7 18 10 AM

LucaMilanese90:
@EnzoVezzaro don't think the problem is that we need to re-fetch after publishing since the new tags don't show up even after page refresh. Most likely the problem is in the query or in some ElasticSearch-related unexpected behavior.

LucaMilanese90:
Update: increased the size of the aggregation tags to 1000 (should still be fine performance-wise) and updated the code to follow the new edit refactor from #1546

mihaisc:
This seems to be working ok with the latest change. If in time we find that 1000 is too much we can reduce it so no problems there.

EnzoVezzaro:
Tested. LGTM.

mihaisc:
We will test more and merge after the H20 release. Do not merge for now

mihaisc:
Functionally this looks fine. @kremalicious ?

EnzoVezzaro:
This PR looks ok. @jamiehewitt15 @kremalicious can you guys please take a look? I think this PR can be merged.

jamiehewitt15:
To me it fees quite weird to be presented with this random list of suggestions before I've typed anything. Maybe we could only show the suggestions after the user has typed at least one or two characters?



We also need to do some more work on how the suggested tags are filtered and ordered. When I start typing `b` I would expect to see suggestions that begin with `b` at the top, but that's not the case.

jamiehewitt15:
At the moment, it seems that I have to type 3 characters before I get a useful suggestion. 

For example, when typing `re` I don't see `restaurant` in the list of suggestions.




I have to type `res` before I get what I'm looking for.

LucaMilanese90:
Looks like I'm not authorized to commit to this repo anymore, I forked the repo and opened a PR with the requested fixes #1679

mihaisc:
Ready for review

EnzoVezzaro (approved):
made a small change 👍🏽

src/components/@shared/FormInput/TagsAutoComplete.tsx
```diff
@@ -0,0 +1,92 @@
+import React, { ReactElement, useEffect, useState } from 'react'
+import CreatableSelect from 'react-select/creatable'
+import { OnChangeValue } from 'react-select'
+import { useField } from 'formik'
+import { InputProps } from '.'
+import { getTagsList } from '@utils/aquarius'
+import { chainIds } from 'app.config'
```
kremalicious, line 7: is this intended to get the tags of all initially defined `chainIds`? Cause usually we base everything on the `chainIds` the user has selected, which this here would not get

LucaMilanese90: Yes, it is intended behaviour since the request was to generate the list of existing tags coming from mainnets. Using the user selection would require an additional filter to exclude the testnet chainIds.

mihaisc: yep, as intended. No need to fetch for each chain, and no need to fetch tags from test chains.

src/components/@shared/FormInput/TagsAutoComplete.tsx
```diff
@@ -74,10 +74,12 @@ export default function TagsAutoComplete({
       defaultValue={defaultTags}
       hideSelectedOptions
       isMulti
-      noOptionsMessage={() => 'Start typing to get suggestions'}
+      noOptionsMessage={() =>
+        'Start typing to get suggestions based on tags from all published assets.'
+      }
       onChange={(value: AutoCompleteOption[]) => handleChange(value)}
       onInputChange={(value) => handleOptionsFilter(tagsList, value)}
-      openMenuOnClick={false}
```
LucaMilanese90: @kremalicious just to give you some context: this was set to false because it was suggested in the comments not to show the complete list before the user has started to type something, to always show only relevant suggestions.

kremalicious: interesting, when I tried it was just the difference on first click inside between this:

Screen Shot 2022-09-27 at 14 18 49

and this

Screen Shot 2022-09-27 at 14 17 30

As soon as I type 1 letter I get popup suggestion right away in both cases:

Screen Shot 2022-09-27 at 14 19 29

So to make it more clear how the tags field can be used right away, opted to show and modify that help text to see expectations right away. Otherwise it is immediately confusing why I should `Create "d"`. That's for the UI, basically a suggestion of what the UI does and introduce all elements already to reduce surprises.

But agree that it stays a bit confusing as it looks like the tool in background needs at least 3 characters to produce anything meaningful:

Screen Shot 2022-09-27 at 14 24 59

Screen Shot 2022-09-27 at 14 25 09

Both cases should give me suggestions already, as "data" contains "da". So this sounds more like we should prevent tag suggestions up until the 3rd character is typed, or modify how we use that `match-sorter`

LucaMilanese90: Yes, you are correct, it was to avoid opening an empty list when clicking on the input. Personally, I would prefer to show matches right after the first character has been typed (especially now that `match-sorter` has been implemented), but of course, that's up to you to decide 😄

kremalicious:
If I type in the placeholder:

Screen Shot 2022-09-27 at 14 37 07

With that we currently make it impossible for people to paste in a list they have created somewhere else, which I'm sure is quite common. We need to add auto-splitting on commas and transform them all into tags in the UI too.

src/components/@shared/FormInput/TagsAutoComplete.tsx
mihaisc: we can just change it to 2 chars, instead of 3, or 1 . change this `options={!input || input?.length < 3 ? [] : matchedTagsList}` to whatever we want. We can launch it with whatever and then collect feedback.

mihaisc:
I mean, true , but how do we know this is a common use-case? Were are they copying these tags from? We can just launch it and then add features based on feedback.

src/components/@shared/FormInput/TagsAutoComplete.tsx
kremalicious: works like every autosuggestion out there now with suggesting upon first key stroke. That `Create "xxx"` stays a bit confusing as it implies they are "created" upon click but in reality they are only "created" once publish finishes. But good enough

kremalicious:
when no tags are set and I end up on edit asset screen, I get this:

Screen Shot 2022-10-05 at 14 47 39

And without any interaction on that field I then get empty tag saved in DDO.

kremalicious (approved):

PR #1478, opened 2022-05-26
Only showing pool share locking message on eligible assets

jamiehewitt15:
Changes proposed in this PR:

- Only showing pool share locking message on eligible assets


kremalicious:
I learned today, that the pool stats regarding liquidity stay intact if pool shares are being sent around into "migration" contract, so we might want to start showing the charts and all again, and show the lock message somewhere else. But should be tested on some eligible pools to see how this behaves in current state

DMats (approved):

claudiaHash (approved):

PR #1480, opened 2022-05-30
added @shared components stories

### LINKED ISSUE #1400: Basic Components: Shared

EnzoVezzaro:
- Fixes #1400 

Changes proposed in this PR:

- added ...


EnzoVezzaro:
Please @claudiaHash take some time to review the market as this last commit touches few files. Also, this last commit will break the CI, but I'd prefer to bring some of the work to review while I'm trying to fix the CI issue.

src/components/@shared/AssetList/AssetListTitle/index.tsx
```diff
@@ -1,25 +1,27 @@
...
-  did?: string
-  title?: string
-}): ReactElement {
+}: AssetListTitleProps): ReactElement {
   const { appConfig } = useMarketMetadata()
   const [assetTitle, setAssetTitle] = useState<string>(title)
 
   useEffect(() => {
-    if (title || !appConfig.metadataCacheUri) return
+    if (title || !appConfig || !appConfig.metadataCacheUri) return
```
claudiaHash, line 24: think here we can have !appConfig?.metadataCacheUri

EnzoVezzaro:
> > Please @claudiaHash take some time to review the market as this last commit touches few files. Also, it will break the CI, but I'd prefer to bring some of the work to review while I'm trying to fix the CI issues.
> 
> Even though the `locale` change brings some modifications in a lot of files, from a writing stories perspective it makes our job easier

Yeah. I couldn't find a way to work around all the issues that our hooks creates. 

I found also that our script to automate the tests is actually testing running some tests twice or is ignoring our customs `.test` files. 

For example: I'm working on AssetList. This components needs a custom `.test` file to run correctly. Our script runs correctly our custom test but also it runs another test on the component which fails because it's missing what on our custom `.test` file.

This is our custom `.test`
Screen Shot 2022-06-03 at 6 40 45 AM

This is the automated test
Screen Shot 2022-06-03 at 6 42 07 AM

EnzoVezzaro:
Greener and greener 💚

Screen Shot 2022-06-10 at 7 36 44 AM

Tests coverage is on the rise. We still need to deal with a couple of issues:

- Automated test fails: `@shared/FormFields/AssetSelection/index.stories.tsx`

Screen Shot 2022-06-10 at 7 42 48 AM

- Automated test fails: `@shared/AssetList/index.tsx` (manual test passes)

Screen Shot 2022-06-10 at 7 42 37 AM

package.json
```diff
@@ -17,7 +17,7 @@
     "type-check": "tsc --noEmit",
     "deploy:s3": "bash scripts/deploy-s3.sh",
     "postinstall": "husky install",
-    "codegen:apollo": "apollo client:codegen --endpoint=https://v4.subgraph.rinkeby.oceanprotocol.com/subgraphs/name/oceanprotocol/ocean-subgraph --target typescript --tsFileExtension=d.ts --outputFlat src/@types/subgraph/",
+    "codegen:apollo": "apollo client:codegen --endpoint=https://v4.subgraph.ropsten.oceanprotocol.com/subgraphs/name/oceanprotocol/ocean-subgraph --target typescript --tsFileExtension=d.ts --outputFlat src/@types/subgraph/",
```
claudiaHash, line 20: why changing the endpoint here?

EnzoVezzaro: because this morning rinkeby wasn't working, so we ended up with some errors during build. This should really impact anything, and also rinkeby will be deprecated at some point. But after the fixes on the rinkeby subgraph we should be good to go back to it if we want to.

kremalicious:
just quickly looking over this, refactoring all components to not use any hooks is not feasible. That's the point of hooks, to not using a gazillion props on everything, which for some components this PR now does.

You need to mock hooks like `useWeb3()` or `useUserPreferences()` so that whenever Jest runs, those hook calls are returning your mocked values.

EnzoVezzaro:
Let's recap the current state of this PR before moving forward.

Current State of the Art: 

Screen Shot 2022-06-23 at 9 13 36 AM

Issues unsolved: 

- [ ] Automated test should stops when encountered `console.error`
- [ ] Automated test should ignore components with custom `.test.tsx` file passed (see [here](https://github.com/oceanprotocol/market/pull/1480#issuecomment-1145838401))
- [ ] Test failing on package `is-url-superb`
Screen Shot 2022-06-23 at 9 21 28 AM
- [ ] `Error fetchData:  Cannot read properties of undefined (reading 'query')` - errors with contexts (we need to mock correctly the context). Example: `MarketMetadataContext`
- [ ] `An update to AssetList inside a test was not wrapped in act` - see [here](https://github.com/oceanprotocol/market/pull/1480#issuecomment-1145838401)

As for right now, We should stop moving forward as these are critical issue we need to solve to be able to keep going effectively. As stated previously, we need to mock our context providers. At that point we'll probably endup with custom `.test.tsx` files to maintain. At the moment we don't see feasible decoupling logic from UI, and also doesn't make much sense to keep going on storybook if we'll need to do custom work on the testing side of things. On that regard, we wanted to ask you @kremalicious, what you think is the best way to move forward here?

PR #1482, opened 2022-06-02
disable job selection during actions execution

EnzoVezzaro:
Fixes [comment](https://github.com/oceanprotocol/market/pull/1069#issuecomment-1143647095).

Changes proposed in this PR:

- disabled the selection of a job during the execution of an action

### DIFF
--- a/src/components/Asset/AssetActions/Compute/FormComputeDataset.tsx
+++ b/src/components/Asset/AssetActions/Compute/FormComputeDataset.tsx
@@ -169,7 +169,7 @@ export default function FormStartCompute({
           {...field}
           options={algorithms}
           component={Input}
-          disabled={isLoading || !isAssetNetwork}
+          disabled={isLoading}
         />
       ))}
 
--- a/src/components/Asset/AssetActions/Compute/FormComputeDataset.tsx
+++ b/src/components/Asset/AssetActions/Compute/FormComputeDataset.tsx
@@ -105,7 +105,7 @@ export default function FormStartCompute({
       )
       const extendedAlgoAsset: AssetExtended = {
         ...algorithmAsset,
-        accessDetails: accessDetails
+        accessDetails
       }
       setSelectedAlgorithm(extendedAlgoAsset)
     }
@@ -169,6 +169,7 @@ export default function FormStartCompute({
           {...field}
           options={algorithms}
           component={Input}
+          disabled={isLoading || !isAssetNetwork}
         />
       ))}
 


src/components/Asset/AssetActions/Compute/FormComputeDataset.tsx
```diff
@@ -169,6 +169,7 @@ export default function FormStartCompute({
           {...field}
           options={algorithms}
           component={Input}
+          disabled={isLoading || !isAssetNetwork}
```
kremalicious, line 172: not saying we shouldn't block actions when user wallet is not connected to asset network, but which actions in the job selection actually require this? It would be great to be able to get a final price even without a wallet connected, which this `!isAssetNetwork` here would prevent

EnzoVezzaro: I don't fully agree, I think we shouldn't calculate the price on a network that is not supported. We can't even tell if the price is correct since we don't actually support that network (so it's really a reliable information to have). In my opinion blocking the user from selecting anything is a better approach to avoid issues like the one reported [here](https://github.com/oceanprotocol/market/pull/1069#issuecomment-1144586188). Otherwise, we can just fix the issue, and be done with it.

EnzoVezzaro: Like one of the problems is we're unable to fetch correctly the [user's oceans' balance](https://github.com/oceanprotocol/market/blob/feature/disable-selection-when-job-starts/src/@context/Web3.tsx#L160:L160) which is needed to [calculate the balance](https://github.com/oceanprotocol/market/blob/feature/disable-selection-when-job-starts/src/components/Asset/AssetActions/Compute/FormComputeDataset.tsx#L160:L160). When trying to `getOceanBalance` we get `No config found for given network '5'` from the lib. I'm trying to workaround this issue. Not completely sure it's the only error here as I'm getting weird numbers when trying to pass manually these values. 

Screen Shot 2022-06-06 at 10 25 00 AM

bogdanfazakas: Me and @kremalicious were talking about displaying the prices for the networks we already suport without a wallet connected, and isAssetNetwork is telling us just if the asset's network is the same as the user's wallet connection.
Definitely we should take care of the unsupported networks as well, but not sure if we should do it this PR

EnzoVezzaro: ok, 👍🏽 I'll remove `isAssetNetwork` to close this PR, and let's keep this discussion on a separate PR

EnzoVezzaro: PR open #1487

bogdanfazakas (approved):

PR #1487, opened 2022-06-06
allow price calculation when users are on unsupported network / not connected

EnzoVezzaro:
Fixes [comment](https://github.com/oceanprotocol/market/pull/1069#issuecomment-1144586188).

Continuing discussion from [comment](https://github.com/oceanprotocol/market/pull/1482#discussion_r890212775)

Changes proposed in this PR:

- block users from selection when on unsupported asset


kremalicious:
but why? We should be able to show a price for any asset, no matter if user is connected with wallet or not. This here simply circumvents actually fixing the reason the app crashes

mihaisc:
If network not supported just treat it as if the user is not connected at all.

EnzoVezzaro:
> if the user is not connected at all.

that's actually another issue. Right now, if you're not connected it doesn't calculate the price. I'll take at look at it. 

https://user-images.githubusercontent.com/13741335/173661943-9679770e-46b8-4c25-abf9-0aa1a03ac8fd.mp4

mihaisc:
that is not an issue, you don't have an infura id in .env

EnzoVezzaro:
> that is not an issue, you don't have an infura id in .env

Mmmm... I do have this NEXT_PUBLIC_INFURA_PROJECT_ID on my .env file. I'll check again to see if it's still valid.

mihaisc:
sorry, i just saw that we are on c2d branch , might be a bug , my bad.

EnzoVezzaro:
> sorry, i just saw that we are on c2d branch , might be a bug , my bad.

No worries 😊 Yes, it's not really a bug per se', it's actually missing the logic for users not connected 😅. I'm working on it right now. Once I have in place the `dummyWeb3` and changed a couple of this, we should be good to go.

kremalicious:
please update PR titles as the scope of a PR changes. I only see that this PR is supposed to block price calculation on unsupported networks, but the latest code changes seem to do the opposite. So PR title needs to be updated

kremalicious:
I just ran into this when testing compute. Super confusing with no wallet connected, as simply nothing happens in the UI when I click on an algo

mihaisc:
please merge main

mihaisc:
Didn't really check this, but not sure what it fixes. 
image
image
The price is calculated ok on both download and compute.

mihaisc:
it seems to not work as expected 
image

EnzoVezzaro:
> it seems to not work as expected image

Got it, the last merge broke the unsupported networks. Working on it 👍🏽

EnzoVezzaro:
> it seems to not work as expected image

Haven't been able to find this asset 😅 Can you share the link or ddo of this asset? On other assets I see all working normally.


https://user-images.githubusercontent.com/13741335/195615032-7db4cd89-ca17-497c-8325-930a01de4f58.mp4

EnzoVezzaro:
> did:op:cad4a81c9a8e1c1071ccf3e9dea6f8f42d58e100fa3ddf2950c8f0da9e0dda46

Mmmm.... 🤔 can't replicate this:

Screenshot 2022-10-13 at 11 47 39

can you re-test please, and let me know what exactly you did to break it since I can't find the bug on my end.

mihaisc:
Just go to url , connected on goerli it work
image
i swith network to kovan, refresh page, select random algo , the price is stuck in loop
image
console: 
image

EnzoVezzaro:
🤔 this is interesting and annoying at the same time. For some reason on my end the app doesn't goes into infinite loop, I just get a RPC error:

https://user-images.githubusercontent.com/13741335/195830555-5c6f8828-c58d-4b0a-adeb-283bcf6a338b.mp4

I added a catch anyway to the query context since it doesn't make sense to query to an unexisting / wrong subgraph.

mihaisc:
Test on the preview. You locally probably don't have proper rpc configured in env and it probably crashed before the loop

EnzoVezzaro:
> You locally probably don't have proper rpc configured in env and it probably crashed before the loop

I see. Yah, probably there's that. I'll check infura. It happens in the `preview` too. My txs trow an error and the calc get stopped, but I never get to the loop. In the meanwhile, are you still experiencing this issue after the latest changes?

mihaisc:
same 
image

EnzoVezzaro:
> Screenshot 2022-10-19 at 14 59 29

Strange, not sure why your algo is 0 in this case, can you please give me the DDO.

On the asset we're testing this is what I see when the wallet is not connected:

Screenshot 2022-10-19 at 08 26 18

All fields are valued except 'Dataset' which is always 0. I'll investigate as we have it in prod too, probably we're dragging it for some time.

bogdanfazakas:
> Strange, not sure why your algo is 0 in this case, can you please give me the DDO.

In this case it's fine for the algo to be 0, because I have a valid order for it, i had the wallet connected in that example 
but this is the did : did:op:9c3102ef636d92ba265dd3d21e42b0774813aac7bf87d94db3d82781e2382e84

src/@context/MarketMetadata/index.tsx
```diff
@@ -42,7 +42,7 @@ function MarketMetadataProvider({
           ),
           swapApprovedFee: response.data?.opc.swapOceanFee,
           swapNotApprovedFee: response.data?.opc.swapNonOceanFee
-        } as OpcFee)
+        } as unknown as OpcFee)
```
bogdanfazakas, line 45: do we really need this `unknown` ?

src/@utils/subgraph.ts
```diff
@@ -76,6 +77,12 @@ export function getSubgraphUri(chainId: number): string {
 
 export function getQueryContext(chainId: number): OperationContext {
   try {
+    const isNetworkSupported = chainIdsSupported.includes(chainId)
+    if (!isNetworkSupported)
```
bogdanfazakas, line 81:
```suggestion
    if (!appConfig.chainIdsSupported.includes(chainId))
```

src/@utils/subgraph.ts
```diff
@@ -5,6 +5,7 @@ import { getOceanConfig } from './ocean'
 import { AssetPreviousOrder } from '../@types/subgraph/AssetPreviousOrder'
 import { OrdersData_orders as OrdersData } from '../@types/subgraph/OrdersData'
 import { OpcFeesQuery as OpcFeesData } from '../@types/subgraph/OpcFeesQuery'
+import { chainIdsSupported } from 'app.config'
```
bogdanfazakas, line 8: to keep imports constant through the app we should use relative paths here also
```suggestion
import appConfig from '../../app.config'
```

src/components/@shared/ButtonBuy/index.tsx
```diff
@@ -42,12 +43,14 @@ function getConsumeHelpText(
   assetType: string,
   isConsumable: boolean,
   isBalanceSufficient: boolean,
-  consumableFeedback: string
+  consumableFeedback: string,
+  isSupportedOceanNetwork: boolean,
+  web3: any
```
bogdanfazakas, line 48: we should use the proper type here instead of any
```suggestion
  web3: Web3
```

src/components/@shared/ButtonBuy/index.tsx
```diff
@@ -57,6 +60,38 @@ function getConsumeHelpText(
   return text
 }
 
+function getAlgoHelpText(
```
bogdanfazakas, line 63: is the method used ?, could not find the call for it, if used sorry 😄

src/components/Asset/AssetActions/Compute/FormComputeDataset.tsx
```diff
@@ -122,23 +122,25 @@ export default function FormStartCompute({
...
-        : new Decimal(
-            datasetOrderPriceAndFees?.price || asset.accessDetails.price
-          ).toDecimalPlaces(MAX_DECIMALS)
+    const priceDataset = new Decimal(
+      hasPreviousOrder
+        ? 0
+        : datasetOrderPriceAndFees?.price || asset.accessDetails.price
+    ).toDecimalPlaces(MAX_DECIMALS)
+
     const priceAlgo =
```
bogdanfazakas, line 136: we should have similar conditions for both dataset and algorithm

src/components/Asset/AssetActions/Compute/FormComputeDataset.tsx
```diff
@@ -147,10 +149,10 @@ export default function FormStartCompute({
       .plus(providerFees)
       .toDecimalPlaces(MAX_DECIMALS)
       .toString()
+
     setTotalPrice(totalPrice)
   }, [
-    asset?.accessDetails,
-    selectedAlgorithmAsset?.accessDetails,
+    asset,
```
bogdanfazakas, line 155: we should keep the algorithm here also

bogdanfazakas:
tested the various scenarios on this asset from Mumbai: did:op:85d859a9a5e307b4ca001dc81ea35d9acc9fc4b4e8d2d223f1cefaab15282eb4
works as expected in the following scenario:
 - no wallet connected
 - wallet connected to an unsupported network
 - wallet connected to asset network (Mumbai in this case)
 - wallet connected to Goerli network

fails with the out of gas issue:
- when connected to other supported networks rather than the assets network or Goerli (Eth main, polygon, EWC, BSC, Moonbase, etc) 

https://user-images.githubusercontent.com/13447142/196925215-77766d68-d169-424a-b96b-b27c088e4896.mov

src/components/@shared/ButtonBuy/index.tsx
EnzoVezzaro: yeah, probably something got missed up when merging 😅

bogdanfazakas:
The issue mentioned above still persists on my end when connected to the mentioned networks
Screenshot 2022-10-25 at 10 58 07

bogdanfazakas (approved):

bogdanfazakas:
This needs a rebase

EnzoVezzaro:
> This needs a rebase, and there are also some conflicts that popped up after the merge of the compute fixes PR that need to be taken care off

ready for review 👍🏽

bogdanfazakas:
> > This needs a rebase, and there are also some conflicts that popped up after the merge of the compute fixes PR that need to be taken care off
> 
> ready for review 👍🏽

think new conflicts appeared, we should fix them and merge this 😃

jamiehewitt15 (approved):
Yeah, this works well and solves the initial problem. 

Possible extension, it would be great if you wrote at least one test for one of the files within this, `PriceOutput` would probably be a good one. I realise this PR dates back from June though so would be good to get it merged sooner rather than later.

PR #1495, opened 2022-06-08
update ui components for mobile

### LINKED ISSUE #1250: V4: Update UI components for Mobile
Some UX tweaks/polishing needed over time, like buttons running out of the screen or some default-zoom action upon tapping “search"

https://drive.google.com/file/d/1q0lxISUHisUuYejK4MvVA0GXIrnIbWnR/view?usp=sharing

The V4 market works pretty well on mobile, but there are some UI glitches we need to fix.

Abrom8:
Fixes #1250

Changes proposed in this PR:
Kept changes to a minimum.

- fix mobile view of filters (global search + profile assets view)
- fix  mobile view of tabs on profile page
- fix to prevent zoom on input focus for the search bar


src/components/Header/SearchBar.module.css
```diff
@@ -26,6 +26,9 @@
 }
 
 .input {
+  font-size: var(
+    --font-size-large
```
kremalicious, line 30: good point, and we should revisit this for all text form fields actually. But in end we just need to get to computed font-size of 16px, while this here leads to a size of around 18px which appears way too big up there. Sao to move forward quickly with this PR, and as there is a code comment, we can get away with hardcoding to `font-size` to `16px` here

Abrom8: `font-size` is now hardcoded to `16px`.

kremalicious:
sorry, old PR, think you have to merge or rebase against `main`, as some subgraph structure had changed and that might be the issue leading Vercel to failed deployments

Abrom8:
@kremalicious looks like Vercel deployment works now

kremalicious (approved):
cool, nice surgical fixes

PR #1498, opened 2022-06-08
fix ci

mihaisc:


kremalicious:
in `ci.yml` at very top we say to run it for pushes against `main` & `v4` branch so we should change this to `v3` branch too

kremalicious (approved):

PR #1507, opened 2022-06-09
NFT display tweaks

### LINKED ISSUE #1494: OpenSea redirect not working for assets on polygon
The correct url for polygon is : https://opensea.io/assets/matic/....

kremalicious:
* fixes #1494
* also support `nft.image` key, as `nft.image_data` is meant for raw SVG image data
* change no image found messaging
* show OpenSea link even if no image has been set
* typo fixes

This asset has empty `tokenURI` (for whatever crazy reason) and we were showing "This NFT was not created in Ocean Market" which was simply wrong. So replaced with this more generic message as we are checking exactly for this:

Screen Shot 2022-06-09 at 11 05 32


kremalicious:
More important to get the fixes in here out, than figuring out just for this fix how to mock all our hooks for the tests

mihaisc:
Issues in the publish preview:
- owned by
- accessed by
- it is shown as algorithm instead of data set
image

EnzoVezzaro (approved):

kremalicious:
> tokenURI is empty so where does opensea get the image? 

good find for #1515. OpenSea image is not the completely correct one either as it's scaled down. Would assume OpenSea picks it up from some other related asset or so

mihaisc (approved):

PR #1508, opened 2022-06-09
Temp form

mihaisc:
This is for people to "resume" publishing the asset if the pool was created but the metadata update failed. The nft and datatoken addresses are hardcoded so those need to be changed from person to person.


kremalicious:
should not be needed anymore after merge of #1511. Can be reopened for eventual edge cases we might encounter

PR #1511, opened 2022-06-09
Retry failed transactions during publish

### LINKED ISSUE #1509: submission: if last setMetadata tx fails, present action to trigger that same tx again
Action: "Try failed steps again"

_Developing..._

kremalicious:
Quickest thing with most impact I could think of without much refactoring of that `handleSubmit`:

- split up `handleSubmit` into multiple methods
- collect everything we need to define success for each step in local state in addition to method-scoped variables
- when `handleSubmit` runs, only run each step if result is not yet present in local state
- when anything fails, _Submit_ button switches to _Retry_
- when anything fails, show error message encouraging users to hit _Retry_
- fixes #1509

user flows for testing:

- publish as before, verify all 3 steps can succeed
- reject first approve tx in wallet, verify all steps fail and publish stops
- hit Retry, verify all steps now succeed
- repeat for step 2 & step 3, somehow making them fail (like, running locally and sending bogus parameters, for step 1 & 3 rejecting in MetaMask works too), then hitting _Retry_ again and verify only relevant steps will run
- buy/download the final asset file, to make sure created tokens fit the created pricing and such

### DIFF
--- a/src/components/@shared/atoms/Button/index.module.css
+++ b/src/components/@shared/atoms/Button/index.module.css
@@ -20,15 +20,6 @@
   text-align: center;
 }
 
-.button:first-child {
-  margin-left: 0;
-}
-
-.button:last-child {
-  margin-right: 0;
-  min-width: auto;
-}
-
 .button:hover,
 .button:focus {
   color: var(--brand-white);
--- a/src/components/@shared/atoms/Loader/index.module.css
+++ b/src/components/@shared/atoms/Loader/index.module.css
@@ -23,6 +23,11 @@
   margin-left: calc(var(--spacer) / 4);
 }
 
+.loader.white {
+  border-color: rgba(255 255 255 / 0.3);
+  border-top-color: var(--brand-white);
+}
+
 @keyframes loader {
   to {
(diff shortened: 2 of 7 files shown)


mihaisc:
break it i shall

mihaisc:
1st try (fixed price)
image

mihaisc:
It seems to be working, the error states are all over the place but i guess that is to be expected.

kremalicious:
this needs a rethink as of course the `handleSubmit` method is frozen once called so won't receive value changes in any React state. Could now mash nit all into 1 try/catch but also exploring to split up into multiple methods called conditionally

mihaisc:
Just slap that try catch and continue with the proper idea later 😄

claudiaHash:
With/ without a wallet connected I keep getting this when accessing the Publish form:

Screenshot from 2022-06-14 14-17-41

EnzoVezzaro (approved):
Great! 👍🏽

claudiaHash (approved):

mihaisc (approved):

PR #1512, opened 2022-06-10
fix global stats

### LINKED ISSUE #1506: Incorrect global stats
image
EWB shouldn't be 0

mihaisc:
Fixes #1506

Not a good idea to update a state in a loop and have an effect that monitors that state.


EnzoVezzaro (approved):
LGTM 👍🏽

Issue #1483, opened 2022-06-03
Data Locker using Estuary


calebtuttle:
**Is your feature request related to a problem? Please describe.**
Ocean frontend currently does not provide an interface where users can easily upload their data.

**Describe the solution you'd like**
Add a Data Locker page (as described in [this proposal](https://port.oceanprotocol.com/t/round-14-ocean-market-data-locker-automated-filecoin-archival-ipfs-data-services/1414)) that provides a way for users to easily upload data to IPFS + Filecoin and then publish the data on Ocean.

**Implementation**. The Data Locker consists, at present, of a page on the frontend where whitelisted users can upload their data, through a drag and drop interface, to IPFS and Filecoin, and where they can view their uploaded data. Uploads to IPFS and Filecoin are done using Estuary.

**Flow**. After a user drops data in the frontend and clicks “Submit,” the user must sign a message with their web3 wallet, and the data and signature are sent to a proxy server. The **proxy server** runs validation checks, including a check that the address that signed the message is on a whitelist maintained by the proxy server.

If the checks pass, the server submits the data to Estuary, which then pins the data to IPFS and establishes deals with Filecoin storage miners. On the frontend, the uploaded file(s) are displayed to the user. An additional modification was made to the Publish form: Under every FileInput, there is a dropdown listing the files a user has uploaded, and the user can select the filename to have the field populated with the file's CID.

**Describe alternatives you've considered**
One alternative we considered is building (or requiring users to use) a generic frontend for storing data on IPFS + Filecoin, but this does not provide an easy way for the user to go from uploading data to publishing it on Ocean.

**Additional context**
I have been working on this feature for OpSci Commons, and it is still a work in progress. At the Ocean core tech meeting yesterday (June 1, 2022), a desire for a pull request for this feature was expressed. I'm hoping this issue starts a conversation about which features of a Data Locker would benefit Ocean and about the best way to implement these features.

The Data Locker feature, as described above, is implemented within [my fork of Ocean](https://github.com/calebtuttle/ocean-market). The feature includes a proxy/access control server, which can be found [here](https://github.com/opscientia/opsci-commons-rbac).

Open questions/issues:

- Should the proxy server be folded into Ocean Provider? If so, should it be rewritten in Python, wrapped by Provider, or something else?
- How should directories be handled? Uploaded directories are currently stored as (and would be published as) CAR files, but there might be a better approach.
[...]

Issue #446, opened 2020-08-10
People can upload data easily (and we don't store it)


trentmc:
**Summary**
Provide publishers with an easy solution to drag & drop some files during publish flow and in background they are added to decentralized storage. But where _we_ don't store it or pay for storage.

**Motivation**
Status quo: publishers must a provide url for their data, which takes technical savvy. Removing this friction will help adoption.

In Ocean Commons, we had where people could "upload to IPFS" for a smoother flow. They would drag and drop the file, and it would get uploaded. However since IPFS does not do storage, we were actually doing the storage on Pacific nodes. This is not a good idea because then we inadvertently start hosting a ton of files for free. 

**Target outcome**
* have a GUI / UX similar to what we had in Ocean Commons
* but store somewhere that we don't have to pay for. 

**Where to store**
* Recommended: [Arweave](arweave.org) - pay once, store forever
* (Backup: Filecoin - pay once, store for a long time)
* IPFS could still be used as the intermediary ("IPFS pinning"). Data would be encrypted.

seldamat:
Our team is working on this! As mentioned in our recent issue/PR, overhead is still a challenge. There may be some ways to decrease overhead but will require a bit of engineering + experimentation https://github.com/oceanprotocol/market/issues/1483


https://port.oceanprotocol.com/t/project-coral-core-tech-improvement-desci-oceandao-berlin-hackathon-round-17/1824

trentmc:
> our recent issue/PR

To be clear, that's an issue. And it's fine as an issue right now. A PR will be needed at some point though.

kremalicious:
As a reminder cause it's mentioned in this issue's description as a goal. This is how the file form field looked in Commons:

<img width="569" alt="Screen Shot 2022-06-06 at 12 49 43" src="https://user-images.githubusercontent.com/90316/172147080-3cabc1d2-ac32-43da-b19d-01577d050fae.png">

"From URL" revealed the URL text input, "Add to IPFS" opened up a drag & drop field, dropping a file was sending this to our own IPFS node's API and we pinned it there. App then receives back the new IPFS URL and uses that as the final file URL.

mihaisc:
We have other issues that handle this in a more direct way

PR #1514, opened 2022-06-11
Add option to upload on publish

calebtuttle:
This is the frontend part of the implementation for the spec described in #1483 (and #446).

Changes proposed in this PR:

- Add option (on Step 2 of Publish form) to upload file through drag and drop interface, instead of requiring user to provide file URL.
- Add additional feedback (on Step 5 of Publish form) for file uploads. This feedback only displays if the user is uploading a file.
- When uploading file on publish, user must pay fee charged by Provider.
- Use the initializeUpload/ and upload/ endpoints proposed in [PR #482 for Provider](https://github.com/oceanprotocol/provider/pull/482)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,6 +43,7 @@
         "react-data-table-component": "^6.11.7",
         "react-dom": "^18.1.0",
         "react-dotdotdot": "^1.3.1",
+        "react-dropzone": "^14.2.1",
         "react-modal": "^3.15.1",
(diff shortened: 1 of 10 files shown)


kremalicious:
closing as out of scope as the solution presented here has a lot of requirements which are unlikely to be included across the whole stack:

- new fee in provider
- provider as an upload tool
- provider needs to work together with an Estuary node, which we most likely have to run on our own

PR #1516, opened 2022-06-13
Dynamic token list support for publishing

### LINKED ISSUE #1335: Dynamic tokens support for publishing
Right now, we hardcode OCEAN token address in config.
Also, the publish form only supports one token.

Proposal:
- on publish, fetch approved tokens from subgraph:
```
query {
 opcs{
  approvedTokens
}
 
}
```

result:
```
{
  "data": {
    "opcs": [
      {
        "approvedTokens": [
          "0x8967bcf84170c91b0d24d4302c2376283b0b3a07"
        ]
      }
    ]
  }
}
```

- then allow publishers to use any of those approved tokens when publishing assets

LucaMilanese90:
Fixes #1335

Changes proposed in this PR:

- Add support for approved baseTokens


mihaisc:
i'll do a quick pr to solve this https://github.com/oceanprotocol/ocean-subgraph/issues/355 so you can display token name, not address

LucaMilanese90:
That would be great actually, can you also make the decimals available? This address has 6 digits which creates a lot of problems during publish

mihaisc:
yes, it will be a 'Token' object

mihaisc:
fixed and deployed 
image

LucaMilanese90:
@mihaisc I updated the query to display the symbols instead of the addresses and I'm updating the labels accordingly during the publish flow. I'm able to publish with fix price (pool still gives errors) but trying to buy does not work. Is there something else I have to change other than baseTokenAddress and decimals?

mihaisc:
hmm, it should just work, i published some time ago with tusd and it just worked. I'll do some tests.

LucaMilanese90:
Have you published with pool from the preview link of from some local branch? I keep getting the same error
Screenshot 2022-06-14 at 11 53 43

mihaisc:
directly from preview, do you have tusdt in your wallet? not sure

LucaMilanese90:
Yep, I have 10, very strange🤔

Screenshot 2022-06-14 at 13 52 35

mihaisc:
but how to you want to create a pool with only 10?

LucaMilanese90:
Yes, I  realized right after I wrote the comment, I thought I had 100 🤦‍♂️

LucaMilanese90:
That's because the balance in the Web3 context provider is only checking for eth and ocean, I'm changing the Web3 context provider to also include the other tokens

LucaMilanese90:
When publishing with pool price the weight of the baseToken is '5000000000000' instead of '5' when using TUSDT (the 12 decimals of difference between TUSDT and OCEAN), but the decimals seem to be set correctly in the publish phase, any ideas?

mihaisc:
I don't think i understand.

mihaisc:
The base token weight will be fixed here https://github.com/oceanprotocol/ocean-subgraph/pull/468 . Is there anything else left to do here? Seems good

LucaMilanese90:
@mihaisc I just pushed the last changes to show the balance for the correct token in pool add/remove and trade tabs. Add/remove and trade seems to work correctly, only buy (for fixed and dynamic) is not currently working correctly. 
How do we deal with the text on the publish flow? Do we want to replace OCEAN with a more neutral text or is it ok to keep it like this?
Screenshot 2022-06-23 at 17 43 23

And what about the Ocean logo next to the basetoken? Is that ok or should we use a different image for the alternatives?
Screenshot 2022-06-23 at 17 43 36

mihaisc:
>  buy (for fixed and dynamic) is not currently working correctly.

I'll do some debugging as well, we also had some issues with the provider at the time you were testing , maybe that was the problem.

We should replace `OCEAN` with a more neutral text like `selected base token` 

We are using H2O so you can leave the same image. H2O doesn't have an actual image

mihaisc:
Please update ocean.js to 1.1.3 , it should fix ordering

LucaMilanese90:
Still not working, now I get an error as soon as I click "buy" both on fixed and dynamic price assets. 
Screenshot 2022-06-25 at 10 26 47
 
It looks like the price is not calculated correctly, for example on this asset did:op:286af94120a77fb3cfae56cbe37789a44c40d86d9ec90b2c711ffc7dd16bdfe9 I see the tokenAmount is "0.000000000001012125", which is not correct.

mihaisc:
Ok, i'll take a look.

mihaisc:
All good now. We were setting the wrong decimals for the datatoken in publish. So you need to publish new assets to test.

LucaMilanese90:
@mihaisc amazing! Tried with fixed and pool and the buying process is working correctly. Same for adding/removing from the pool. The only thing that I can see it's still failing is trading datatokens for basetokens, which is strange because the opposite seems to be working correctly (not sure if it's related to the fact the basetokenWeight is still "50000000000000") on an [asset](https://market-x31mi2462-oceanprotocol.vercel.app/asset/did:op:ead984089c868121b22d4a5e2c9ba53664b2f0eff5cfcd2bea3c262e77e456e3) published from the new deployed preview 🤔

mihaisc:
That's probably normal since i didn't deploy the subgraph 😁. I'll do all the updates tomorrow

LucaMilanese90:
Ok, then I guess we are ready for review?

mihaisc:
All good now.

EnzoVezzaro (approved):
LGTM

EnzoVezzaro (dismissed):
Wait. Not really.

Screenshot 2022-06-29 at 11 16 57

The message at the bottom should have TUSDT instead of OCEAN in [this case](https://market-1zxmteee6-oceanprotocol.vercel.app/asset/did:op:ead984089c868121b22d4a5e2c9ba53664b2f0eff5cfcd2bea3c262e77e456e3).

LucaMilanese90:
> Wait. Not really.
> 
> Screenshot 2022-06-29 at 11 16 57
> 
> The message at the bottom should have TUSDT instead of OCEAN in [this case](https://market-1zxmteee6-oceanprotocol.vercel.app/asset/did:op:ead984089c868121b22d4a5e2c9ba53664b2f0eff5cfcd2bea3c262e77e456e3).

Should be fixed now

kremalicious:
Why is BaseToken this huge new single `select` form field? Where we then only have 2 options?

This should be just a `select` on the price input field prefix, where instead of OCEAN I can select the other base tokens. Selecting this then switches base token across whole pricing form

Select is just empty without a wallet, and tokens are wrong too, showing "DT" and "Data Token" on OCEAN field

Screen Shot 2022-06-30 at 11 12 55

After connecting wallet on this screen as I am prompted there, I get this, with something selected in that `select` at top, which then is not reflected in the form:

Screen Shot 2022-06-30 at 11 15 09

I first have too select OCEAN in that `select` again, then switch back to H2O. Only then changes are reflected

mihaisc:
There is no h20 logo

kremalicious:
How to test H2O publishing? Only appears for me on main networks, but not on any test network. If there is only 1 approved token, the UI becomes even more redundant, where in case of making the input prefix into a `select` we can simply hide the caret in this case to remove confusion. In this current UI iteration this is not possible

kremalicious:
> There is no h20 logo

well, but we can agree that the Ocean logo is not the H2O logo? And Coingecko shows something for me: https://www.coingecko.com/en/coins/h2o

kremalicious (changes requested):
remove redundant base token `select` field and deal with some visual and interaction weirdnesses described in former comments

LucaMilanese90:
Went for the big select because the only other select implementation on the publish flow is the "timeout" on step 2, so I thought you wanted to keep the same style for this one as well. But no problems, I'll move it to the price label.
As for the different logo based on the token, how can I query for the appropriate images?

kremalicious:
> As for the different logo based on the token, how can I query for the appropriate images?

That would need to be figured out. If the goal here is to launch as quickly as possible then would suggest dealing with this manually for now and copy over some SVG assets for the first approved tokens only.

For full automation, there are a bunch of token lists out there where projects self-maintained them and apps then use those token lists to get metadata about tokens. Like this one https://github.com/trustwallet/assets

This would then also require to deal with multiple asset types then, as images could be bitmap or vector assets so that all becomes a bit more tricky to implement. Would say 3-5 approved tokens is a good baseline after which we should implement more automatic way. Until then copying in assets should be enough.

kremalicious:
code archaeology from coin selection in v3:

https://github.com/oceanprotocol/market/blob/9f2cbaafc1fbae74303ef8b478677a9a46d0d385/src/components/organisms/AssetActions/Pool/Add/FormAdd.tsx#L137-L156

https://github.com/oceanprotocol/market/blob/9f2cbaafc1fbae74303ef8b478677a9a46d0d385/src/components/organisms/AssetActions/Pool/CoinSelect.tsx

https://github.com/oceanprotocol/market/blob/9f2cbaafc1fbae74303ef8b478677a9a46d0d385/src/components/organisms/AssetActions/Pool/CoinSelect.module.css

mihaisc:
My bad on the image i found another H2O and when i asked i was  told there is none and dropped it.
Agree with everything. If the image implementation is too complicated just stick with a hardcoded array/list and add the required images in whatever format we desire. We are the ones who add approved tokens, and we don't plan more for now. So doing a very complicated implementation would not really make sense

kremalicious:
some more minor UI thingies:

found another mention of OCEAN hardcoded, also not sure if it's of any concern because of test token, but the conversion into user currency would be wrong for a supposed stable coin. Although stable coins loosing their peg are hip right now

Screen Shot 2022-06-30 at 21 05 07

The error validation message overlapping the input bug is back:

Screen Shot 2022-06-30 at 21 07 20

This is only in this PR, not present on live where it behaves as intended:
Screen Shot 2022-06-30 at 21 09 03

That balance check message in last screenshot appears and reappears sporadically for me when I have none of that token in my wallet. But we have already WIP PR going on for it with #1435 so might be rather solved there.

EnzoVezzaro:
If I'm not connected I just see an empty input. It's kind of weird to see. Maybe we add a placeholder instead.

Screenshot 2022-07-04 at 07 59 09

src/@context/Web3.tsx
```diff
@@ -155,16 +157,33 @@ function Web3Provider({ children }: { children: ReactNode }): ReactElement {
     if (!accountId || !networkId || !web3) return
 
     try {
+      const { oceanTokenAddress } = getOceanConfig(networkId)
+      const approvedTokensBalance: { [key: string]: string } = {}
+      const approvedTokens = (await getOpcsApprovedTokens(chainId)).filter(
```
kremalicious, line 162: I see we call `getOpcsApprovedTokens` multiple times in various places, again on the pricing page. Would it be more effective if we fetch and expose approved tokens from `MarketMetadata`? Cause this Web3 context provider here can then just grab it from there, as `MarketMetadata` is our highest context provider. Cause in end we should only fetch the approved tokens once, when user initially enters app

LucaMilanese90: Yes, that is a good idea actually! I'll modify it as suggested

LucaMilanese90: In that provider I'm not going to be able to check for the selected chainId though, right? Should I expose the list for all the supported chainIds then?

kremalicious: right, didn't think of that and if I count the requests current implementation could be actually most effective already. Think we're good here, sorry!

Cause if we always fetch for all chains on app start we force at least 5 or 9 requests on all users although so far they are only relevant in the publish form. But if e.g. user is connected to Polygon and publishes on Polygon we end up with 2 requests (1 in Web3 context, 1 in pricing form). We could reduce this to only 1 by storing the fetched tokens in some context but too much refactor for this PR

kremalicious:
awesome, thanks for all the latest changes, much better UI and interactions now. Could only find this on the first time user enters this screen, where it says "Data Token" where it should say "Ocean Token":

Screen Shot 2022-07-05 at 12 57 18

Works correctly when switching to H2O, then back to OCEAN.

Then was already trying to ditch heading font for token names but then realized there is actually a real space in code and not just some bad kerning:

Screen Shot 2022-07-05 at 12 51 13

As for the overlapping error state, this is because when H2O is selected the conversion div under the input field is gone, moving the absolute positioned error up. So if we can get the conversion to show up on H2O, we don't need to do anything regarding error placement, if we want to leave it out for now, guess we have to do some CSS tweaks for that error alone.

As for conversion in general I can look into teaching our price context provider more tokens but would do that in another PR as most likely will change lots of files. We can start with hardcoded list as we would have to map approved tokens to Coingecko token ids but as long as token is listed there this should work out

LucaMilanese90:
> awesome, thanks for all the latest changes, much better UI and interactions now. Could only find this on the first time user enters this screen, where it says "Data Token" where it should say "Ocean Token":

Yep, I'm still working on it, I'll fix this😉

src/components/Publish/Pricing/index.tsx
```diff
@@ -11,75 +11,123 @@ import Free from './Free'
...
 
   // Connect with main publish form
   const { values, setFieldValue } = useFormikContext<FormPublishData>()
   const { pricing } = values
-  const { price, amountOcean, weightOnOcean, weightOnDataToken, type } = pricing
+  const { price, amountBaseToken, weightOnBaseToken, weightOnDataToken, type } =
+    pricing
+
+  useEffect(() => {
+    if (!chainId) return
```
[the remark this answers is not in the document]
LucaMilanese90: Not 100% sure why but with `(chainId || 1)` everything breaks and I always get back the results from chainId = 1 🤔

LucaMilanese90:
> Then was already trying to ditch heading font for token names but then realized there is actually a real space in code and not just some bad kerning:

Can I simply use the string in `token.name` then? The problem is that while for ETH is fine and the name is 'Ocean Token' 
Screenshot 2022-07-05 at 14 54 01

for the testnets it's not, on Rynkeby for example it is 'oceanToken'
Screenshot 2022-07-05 at 14 54 45

I tried to convert the camelCase string with regex but there are too many edge cases to handle.

Also, would it be safe for me to assume that the default token is always the one in position `0` from the approved tokens array? Or are they ordered alphabetically?

kremalicious:
> Can I simply use the string in token.name then?

yes! No need to complicate this, wasn't sure where exactly `h2 o` was coming from so that's the only concern regarding spacing. Your screenshot is technically correct, as on Rinkeby the token name in the deployed contract is indeed `OceanToken` as we never deployed the token update there we had on mainnet, where we also changed the name to `Ocean Token`. In any case, think it's fine just passing through `token.name`

Over in #1573 made the prices context more token-agnostic so we can deal with that token conversion topic later over there.

LucaMilanese90:
@kremalicious update:

> As for the overlapping error state, this is because when H2O is selected the conversion div under the input field is gone, moving the absolute positioned error up. So if we can get the conversion to show up on H2O, we don't need to do anything regarding error placement, if we want to leave it out for now, guess we have to do some CSS tweaks for that error alone.

The error positioning should be fixed now even if the conversion is not shown.

> in this case, we should have a `chainId` fallback so that for demonstration reasons we also fetch when no wallet is connected. In this case the whole publish form defaults to `1` so if we remove this here and further down do `await getOpcsApprovedTokens(chainId || 1)`, then we have this functionality for the no wallet case too

This is now implemented, I've also added a check to verify if the chainId is supported, if not it's reverting back to the default 1

As for the token logos, should the color scheme remain the same (black for basetoken and pink for datatoken)?
Screenshot 2022-07-06 at 10 39 34

These logos are updated as well (don't mind the wrong token name, I couldn't find a published asset with H20 so I just switched the logos for the screenshot)

Screenshot 2022-07-06 at 10 46 14


Is the logo in the wallet supposed to stay ocean or should that change too? 
Screenshot 2022-07-06 at 10 50 58

mihaisc:
> Is the logo in the wallet supposed to stay ocean or should that change too?

Leave it as it is for now, we need to rethink that a bit.

src/@context/Web3.tsx
```diff
@@ -155,16 +157,33 @@ function Web3Provider({ children }: { children: ReactNode }): ReactElement {
...
+          address,
+          web3
+        )
+        approvedTokensBalance[symbol.toLocaleLowerCase()] = tokenBalance
+      })
+
       const balance = {
         eth: web3.utils.fromWei(await web3.eth.getBalance(accountId, 'latest')),
-        ocean: await getOceanBalance(accountId, networkId, web3)
+        ocean: await getOceanBalance(accountId, networkId, web3),
```
kremalicious, line 178: bonus points for not making duplication in codebase worse: `getOceanBalance()` from `@utils/ocean` is redundant and unnecessarily specific, so can be replaced completely with the newly created `getTokenBalance()`.

src/components/Header/Wallet/Details.tsx
```diff
@@ -67,7 +67,7 @@ export default function Details(): ReactElement {
               <AddToken
                 address={oceanTokenMetadata?.address}
                 symbol={oceanTokenMetadata?.symbol}
-                logo="https://raw.githubusercontent.com/oceanprotocol/art/main/logo/token.png"
+                logo="ocean"
```
kremalicious, line 70: up here, the logo is supposed to be for the datatoken, as this action adds the datatoken to the wallet. Just like with pricing screen, this raises question what exact logo datatokens should have, as this was never consciously decided. I would see 3 options:

1. Giving datatokens the ocean logo, but in violet, was a quick solution when building v3, but it also reflects that they are still OCEAN-branded tokens in a way. From that perspective we could go with violet OCEAN logo for all datatokens, regardless of base token.

2. Current implementation where base token logo is made violet works also well if we are fine with them being branded differently. But mechanics-wise the datatoken as access token is not bound or derived from the base tokens so argument could be made this could be confusing to users.

4. We could come up with a new logo, for all datatokens. Of course would take longest as design work has to be done first.

mihaisc: 3 would be the best , but i would go with 1. for now.

src/@context/Web3.tsx
LucaMilanese90: I refactored the user balance fetch and took the occasion to move the approvedBaseTokens list into the web3 provider to reduce the number of calls of the `getOpcsApprovedTokens()` function.

src/components/Header/Wallet/Details.tsx
LucaMilanese90: So, are we going with solution 1 for the moment? Can I go ahead and change it?

mihaisc: go for it

LucaMilanese90: Done

mihaisc:
On polygon where do we get that name from ? `H20data` . And i assume the icon doesn't change because of the name

image

LucaMilanese90:
> On polygon we display `H20data` . I assume the icon doesn't change because of the name image I think we should use the symbol. Looks on all networks

Looks lithe the problem is that on Ethereum the symbol is H2O (with the letter "o" at the end), while on polygon it's spelled with the number 0.

mihaisc:
Ok , i think everything is covered. @kremalicious see any major things that are missing?

EnzoVezzaro:
Just [published](https://market-5sqxkn7eu-oceanprotocol.vercel.app/asset/did:op:950293cefa3fe53cd15040564ceb0a6e8fcd40a84f77172ce3a042c41010f8e0) on polygon, I was able to publish with h2o, but the details page breaks because some data is missing



On the response I see missing some info in baseToken

Screenshot 2022-07-18 at 09 55 08

- on edit I see `Ocean` instead of `h20`

Screenshot 2022-07-18 at 09 54 06

- on the publish form found an issue on the checkbox of the pricing tab

`Fixed` is fine
Screenshot 2022-07-18 at 09 48 34

`Free` is missing filled bullet
Screenshot 2022-07-18 at 09 48 39

mihaisc:
The problem is that on mumbai we have only 1 token [query](https://v4.subgraph.mumbai.oceanprotocol.com/subgraphs/name/oceanprotocol/ocean-subgraph/graphql) so don't understand why there are two in the publish form, will check it out

LucaMilanese90:
I only see 1 token when connected to Mumbai (don't see the dropdown at all). I think the problem is if you compile the form without being connected with your wallet and then connect after the price step: since it was decided to default to `chainId` 1 to fetch the approved baseTokens when no wallet is connected, you'll always be able to see H20 and Ocean when disconnected. We could reset the baseToken value when the `chainId` changes at any point during the publish flow or check during the creation of the nft if the baseToken address is coherent with the selected network. Which option would you prefer?

mihaisc:
>  We could reset the baseToken value when the chainId changes at any point during the publish flow or check during the creation of the nft if the baseToken address is coherent with the selected network.

this sounds good

LucaMilanese90:
Which one? Or do you mean both?

mihaisc:
omg... sorry 

> We could reset the baseToken value when the chainId changes at any point during the publish flow

AnaLoznianu:
Please keep in mind that the AMM pool removal will be merged in the main branch on the 1st of August and we will merge this PR soon after.

kremalicious (approved):

kremalicious:
Super big bug, on Polygon the symbol uses a `0` instead of a big `O`

Screen Shot 2022-07-27 at 16 46 05

Another weird but small interaction bug: select `H2O`, then go to next screen. Then go back, and the form data will be reset to the initial default base token.

Apart from that, all seems to be working well, great job!

Remaining conversion issue then can be integrated and finalized in #1573

LucaMilanese90:
> > On polygon we display `H20data` . I assume the icon doesn't change because of the name image I think we should use the symbol. Looks on all networks
> 
> Looks lithe the problem is that on Ethereum the symbol is H2O (with the letter "o" at the end), while on polygon it's spelled with the number 0. I added these options to the TokenLogo switch case.

@kremalicious yes, I mentioned it in an old comment: the H2O written in two different ways is the result of the approved token query, we can manually override it in the frontend but it would be better to have the correct data from the query itself. 

On the default token, up until now, it was decided to use the first token in the list as default for each network. We can change that and use a `.find()` or something else to catch the ocean token from the list.

You are right on the weird reset of the default token when going back and forth between steps, that should be an easy fix.

mihaisc:
> Super big bug, on Polygon the symbol uses a 0 instead of a big O

Not a bug, that is how it's on the chain https://polygonscan.com/address/0xc5248aa0629c0b2d6a02834a5f172937ac83cbd3

LucaMilanese90:
Update:
- the default baseToken is set to look for the ocean option
- the token reset bug when switching tabs back and forth should now be fixed (it will still reset if the chainId changes during the publish process)

src/components/Publish/Pricing/Price.tsx
```diff
@@ -59,12 +58,17 @@ export default function Price({
                 <Conversion price={field.value} className={styles.conversion} />
               </h4>
             </div>
-
-            <h4 className={styles.datatoken}>
-              = <strong>1</strong> {dataTokenOptions.symbol}{' '}
-              <Conversion price={field.value} className={styles.conversion} />
-            </h4>
           </div>
+          {firstPrice && (
```
kremalicious, line 62: there is no `firstPrice` anymore so this can be removed again, same for respective CSS rules.

mihaisc: cool ,didn't know what was that for

jamiehewitt15:
In terms of the visual aspect, the grey background makes the "OCEAN" look like something that can't be changed (like the fee fields beneath it)

H2O - 2022-08-02_16-08

By contrast, in other drop down boxes that we allow people to change the background is white. 
Timeout - 2022-08-02_16-08

I would suggest changing the background colour to white to bring more attention to it.

jamiehewitt15:
Another potential improvement would be to put a tool tip next to pricing so that we can explain briefly what H2O is. As it stands, there is no explanation about H2O or any links to where users can get more info. At the moment we don't have anything in the docs about it either (I'll create an issue for that). 

The functionally works well though, I haven't been able to break it or find any bugs.

mihaisc:
The docs are enough, no more tooltips required.

bogdanfazakas:
Should small issues on c2d be fixed before merging or we can fix them later on since C2D  is still in a testing phase ?
In this case there is a problem at display when we mix dataset and algorithm with diferent tokens (dataset is TUSDT but algo is OCEAN in the case below)

Screenshot 2022-08-03 at 09 50 20

mihaisc:
we can fix them after we merge in a separate pr

kremalicious:
> the grey background makes the "OCEAN" look like something that can't be changed

code to make its background light is actually there but it won't kick in as we are using a `background` later in the code to create the caret without any images. On top of that this is our input prefix component so we should not change them all to white, as we only want more affordance in it if it is an actual `select`. Which also won't be all the time in that price field, like on network where only OCEAN is supported.

So as compromise so we don't have to refactor all input stuff just for this:

Only 1 option:

Screen Shot 2022-08-03 at 12 43 57

With multiple options:

Screen Shot 2022-08-03 at 12 45 09

kremalicious:
In more surprising problems, I am the one with this last remaining conversation but I can't get to it as clicking the View link takes me to long loading changeset view but this conversation to resolve is nowhere to be found.

Screen Shot 2022-08-03 at 12 46 59


So we might have to quickly change merge permissions and back to get this merged, also because @EnzoVezzaro has an ongoing change requests which also prevents merge unless resolved.

mihaisc (approved):

Issue #1244, opened 2022-03-23
Edit UI


kremalicious:
It works, but it could be better. Biggest issues to solve:

- mix of form elements where some are prefilled and some are not making it hard to know which fields will be edited or cleared
- How to clear fields
- ...

EnzoVezzaro:
> * mix of form elements where some are prefilled and some are not making it hard to know which fields will be edited or cleared

Here the difficulty is getting the File and Sample file url. What we can do is placing a "placeholder" like this: 

![Screenshot 2022-08-09 at 10 08 02](https://user-images.githubusercontent.com/13741335/183671536-bd980a78-a9a3-4d1e-89e5-4b4e06cc96ed.png)

if the user clicks on "Download" then we start a tx to download the file or sample file (after the download we should display it correctly). 

![Screenshot 2022-08-09 at 10 07 48](https://user-images.githubusercontent.com/13741335/183671883-fc2aafcc-e190-47b9-9378-d93ea06b23fa.png)

Otherwise the user can remove the placeholder with the "x" to add another link. 

> * Remove/Hide tabs during editing

If we remove the tabs, we won't be able to access the "EDIT COMPUTE SETTINGS":
![Screenshot 2022-08-09 at 10 00 37](https://user-images.githubusercontent.com/13741335/183668815-7c979e6c-9fd2-4257-924d-5214f8a0fb66.png)

> * How to clear fields

Not sure if a clean button like an styled "x" on the single field would do or do you a mean a generic button that clears the whole form? 🤔

EnzoVezzaro:
On the other hand, we can implement an "edit" CTA on each field to let the user know exactly which field he's editing. Any change will trigger a tx to change this specific field. What I proposed is something like this: 

Edit Metadata:
<img width="1001" alt="Screenshot 2022-08-17 at 09 48 54" src="https://user-images.githubusercontent.com/13741335/185150963-89971a0c-1595-4466-a73f-d8805054b846.png">

Compute Settings: 
<img width="996" alt="Screenshot 2022-08-17 at 09 49 00" src="https://user-images.githubusercontent.com/13741335/185151036-a229fabb-ec1a-49ae-abfd-11c73d2955a3.png">

EnzoVezzaro:
> On the other hand, we can implement an "edit" CTA on each field to let the user know exactly which field he's editing. Any change will trigger a tx to change this specific field. What I proposed is something like this:
> 
[...]

PR #1518, opened 2022-06-14
Update file

mihaisc:
to test this : 
- use rinkeby
- publish a file in market (not this branch) or just take one older published asset (version 4.0.0)
- edit asset and change file url
- check if asset is downloadable and that version is 4.1.0

### DIFF
--- a/src/components/Asset/Edit/EditMetadata.tsx
+++ b/src/components/Asset/Edit/EditMetadata.tsx
@@ -107,12 +107,16 @@ export default function Edit({
 
       // TODO: remove version update at a later time
       const updatedAsset: Asset = {
-        ...asset,
+        ...(asset as Asset),
         version: '4.1.0',
         metadata: updatedMetadata,
         services: [updatedService]
       }
 
+      // delete custom helper properties injected in the market so we don't write them on chain
+      delete (updatedAsset as AssetExtended).accessDetails
+      delete (updatedAsset as AssetExtended).datatokens
+      delete (updatedAsset as AssetExtended).stats
       const setMetadataTx = await setNftMetadata(
         updatedAsset,
         accountId,
--- a/src/components/Asset/Edit/EditMetadata.tsx
+++ b/src/components/Asset/Edit/EditMetadata.tsx
@@ -24,6 +24,7 @@ import EditFeedback from './EditFeedback'
 import { useAsset } from '@context/Asset'
 import { setNftMetadata } from '@utils/nft'
 import { sanitizeUrl } from '@utils/url'
+import { getEncryptedFiles } from '@utils/provider'
 
 export default function Edit({
   asset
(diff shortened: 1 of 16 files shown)


kremalicious:
this asset worked on 2nd try again after failing at final download with that known provider issue
https://market-git-fix-updatefilestructure-oceanprotocol.vercel.app/asset/did:op:4a89a8cbd80d08080924ed85b3751c16b63085166c78e2f2a9a3a8ea0944263a

Tried publish new one, worked well, just slow to index in end, was only showing up after like 1-2 min.
https://market-git-fix-updatefilestructure-oceanprotocol.vercel.app/asset/did:op:ff882dc0147a180d2d6ca1a9bc61fe6fa9779af2517924afafc1d7e3a3f1e53b

Download shows same `sender of order transaction is not the consumer/payer` provider issue. I have to do full browser page refresh to get Download to work in end. But as far as this PR goes, seems to be all alright

kremalicious:
that edit form is a UX nightmare but guess we keep it for #1244 where I added that issue which is made worse through adding the file input. Makes it hard to know what field gets edited or cleared. So both sample & file should start out in "added" mode which has to be dismissed to change. Current implementation will make people think the file & sample gets cleared based on the other form elements. Quick fix is to be consistent with the labels and also use "New ..." on file & sample

bogdanfazakas:
> Works flawlessly on Rinkeby, but got some issues on Ropsten & Polygon: the file url it's not updated, on download I still get the old file

I think the provider version needed for this was just deployed Rinkeby, also but for aquarius it takes some time to index the updated metadata so you probably can encounter this if the update was not indexed when the asset was reloaded

kremalicious:
Should only work on Rinkeby right now, as only there new provider is deployed so seems all good. Also tested again on Rinkeby, verified updated file is actually updated and can be downloaded. All flawless

EnzoVezzaro:
I also confirm it's all good on Rinkeby, just tested.

EnzoVezzaro (approved):

claudiaHash (approved):

mihaisc:
Not too fond about deleting props but... it works 😁

src/components/Asset/Edit/EditMetadata.tsx
```diff
@@ -78,17 +80,43 @@ export default function Edit({
...
+      }
       const updatedService: Service = {
         ...asset.services[0],
-        timeout: mapTimeoutStringToSeconds(values.timeout)
+        timeout: mapTimeoutStringToSeconds(values.timeout),
+        files: updatedFiles
       }
 
+      // TODO: remove version update at a later time
       const updatedAsset: Asset = {
```
kremalicious, line 109: on first look seems we could have spotted this issue earlier if we would use the `DDO` typing here instead?

```ts
const updatedAsset: DDO
```

mihaisc: nope, typescript does not remove the props. `Asset`  doesn't have accessDetails and yet it's still there.

PR #1527, opened 2022-06-20
Arweave Integration Milestone 1: Publish using transaction ID

MantisClone:
Towards https://github.com/oceanprotocol/multi-repo-issue/issues/151.

Changes proposed in this PR:

- Add transactionId field
- 

> **Note**
> This does not include the ability to upload to Arweave. That is left for a future PR.


MantisClone:
In https://github.com/oceanprotocol/multi-repo-issue/issues/151 I proposed a Publish UI like the image below.  But when I dug into the code, I couldn't find an existing component like this, so I used a `boxSelection` instead.

Proposed:
image

Approximate Implementation (still work in progress):
Screenshot from 2022-06-21 08-47-48

kremalicious:
this needs design work first. Existing components are not fit for any of this.

kremalicious:
sorry, wasn't clear enough. I meant multiple types of file storages for the publish form, I did not mean to imply that one asset can have multiple files which are then from multiple file locations. We are talking UI design here, and just listing all possible options with all their attached input fields at once is the worst option. In the commons screenshot as an example all actions were opening different sets of input fields, as different file storage types require different user input.

So we need a concept what gets clicked, which fields get revealed and so on. And on top of that make it look good.

MantisClone:
@kremalicious @mihaisc 
I have prepared a rough Figma mockup of how I think the publish screen would look when publishing an asset that is stored in Arweave. Can you please take a look at let me know what you think?

https://www.figma.com/file/HT1h5yNiQs8yVTGv7TAu8j/Figma-to-HTML%2C-CSS%2C-React-%26-more!-(Community)?node-id=0%3A1

Basically, when the "Arweave" storage type is selected, the "File" field turns into the "Transaction ID" field.

MantisClone:
Here's a quick video of what the latest changes look like: https://user-images.githubusercontent.com/2530913/181637257-2f134e62-490e-486c-910f-3c69fc2a0ce5.mp4

It shows me switching between the URL and Arweave storage types. It also shows that switching between Download and Compute doesn't affect the storage type selection.

MantisClone:
After the most recent merge, `FileInfo` moved inside ocean.js, so this PR not relies on https://github.com/oceanprotocol/ocean.js/pull/1618

MantisClone:
@alexcos20 told me yesterday at the Core Tech WG meeting that there was little to no appetite from the core team to merge this PR. Therefore, I am abandoning it.

PR #1541, opened 2022-06-23
Correctly display trusted algorithms in compute asset

### LINKED ISSUE #1538: Datasets have all algos allowed by default
When publishing a compute data set, by default all published algorithms are allowed to run on it:

Screen Shot 2022-06-23 at 12 24 50

Probably related with #1537 as this is the reason all the algos show up in preview too.

Expectation is that, as in v3, compute datasets are published with nothing allowed to run on them.

mihaisc:
Fixes #1538 
By default no algos are allowed now.
Also added filtering for trusted algo publishers. We can't add those in the market yet in edit, but they can be added from external sources.
Related to https://github.com/oceanprotocol/docs/pull/1032

### DIFF
--- a/src/@utils/compute.ts
+++ b/src/@utils/compute.ts
@@ -151,19 +151,20 @@ export async function getComputeEnviroment(
 
 export function getQueryString(
   trustedAlgorithmList: PublisherTrustedAlgorithm[],
+  trustedPublishersList: string[],
   chainId?: number
 ): SearchQuery {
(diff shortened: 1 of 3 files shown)


kremalicious:
mostly fixed over in compute PR in edd80eed7719f779a044e71639edbf97d3f04822 & ec34dc8f537b5c3e957661833f12567bf7d6403d.

But still missing that empty array is correctly read out in edit form

kremalicious (approved):
perfect, `trustedPublishersList` stays untested for now but most likely will work

PR #1546, opened 2022-06-24
Edit screen fixes

### LINKED ISSUE #1539: Can't remove access to compute dataset
After publishing, asset was wrongly set to allow all algos. Going into edit and unticking everything like so:

Screen Shot 2022-06-23 at 12 28 50

Will simply do nothing after successful transaction. Going then back into edit mode shows same settings as before:

Screen Shot 2022-06-23 at 12 30 25

Same goes for other user flow, where e.g. 1 algo is set:
Screen Shot 2022-06-23 at 12 32 41

If I untick that, after successful transaction, now all published algorithms are allowed on the data set(!)

Screen Shot 2022-06-23 at 12 34 14

kremalicious:
* fixes #1539
* addresses #1244
* refactor compute edit form into declarative pattern as we have on publish form for easier maintenance and additions
* lots of unused or duplicated css cleaned up
* disable submit button until user does something with forms
* rudimentarily deal with `publisherTrustedAlgorithmPublishers`, making sure we always keep it at `[]` in all flows

Lots of new styling with more helpful messaging on all screens:

New general layout with more clear content boxes and proper page header:

Screen Shot 2022-06-24 at 14 04 29

By default, after compute dataset is published, this is the fixed "nothing allowed" state. Kinda fine for now and will add some copy to form description to explain "none" state. But actually we would need 3 options on this screen for users to start from, which then reveals various fields (none, selected, all). But didn't want to go there now

Screen Shot 2022-06-24 at 14 14 56

New loading layout:

Screen Shot 2022-06-24 at 14 06 03

New success with updated messaging:

Screen Shot 2022-06-24 at 14 18 51


claudiaHash (commented):
Acessing 'Edit compute settings' for this asset did:op:900ca24c835430fd5f374cca4cc7c52f40796910c146693fe98a39507a8e55f2/edit I get: 

Screenshot from 2022-06-24 16-31-37


Also, maybe it's a good idea to disable the submit button in edit form until the user makes an actual change for its data

kremalicious:
that app crash on that asset should be fixed now.

> Also, maybe it's a good idea to disable the submit button in edit form until the user makes an actual change for its data

yeah, that's good point, I'll try to quickly add this, if not we leave it for #1244

EnzoVezzaro (approved):

PR #1551, opened 2022-06-27
put back Remove Liquidity components

### LINKED ISSUE #1550: V3: Enable remove liquidity
Re-enable remove liquidity on the V3 market. The 'Add liquidity' functionality remains disabled.

kremalicious:
* closes #1550

Changes from former v3 state:
- show Remove button at all times, disable with conditions which were hiding it before
- longer label "Remove Liquidity" on pool tab
- remove migration status alert
- remove Pool Share Lock message above pool tabs

### DIFF
--- a/src/components/organisms/AssetActions/Pool/Actions.module.css
+++ b/src/components/organisms/AssetActions/Pool/Actions.module.css
@@ -1,0 +1,25 @@
+.actions {
+  margin-left: -2rem;
+  margin-right: -2rem;
+  padding-left: var(--spacer);
+  padding-right: var(--spacer);
+  padding-top: calc(var(--spacer) / 1.5);
+  border-top: 1px solid var(--border-color);
+  text-align: center;
+  display: flex;
+  justify-content: center;
+}
+
+.success {
+  margin-top: calc(var(--spacer) / 2);
+  margin-bottom: calc(var(--spacer) / 2);
+}
+
+.actions button {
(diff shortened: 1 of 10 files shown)


claudiaHash (approved):
After removing max amount of pull shares from [this pool ](https://deploy-preview-1551--v3-market.netlify.app/asset/did:op:aA16f337604cb5716995d30c23E92057ef2826Ef) I still can access _Remove Liquidity_ and remove more liquidity from the pool

kremalicious:
> After removing max amount of pull shares from [this pool ](https://deploy-preview-1551--v3-market.netlify.app/asset/did:op:aA16f337604cb5716995d30c23E92057ef2826Ef)I still can access Remove Liquidity

not so sure we can do anything about it as in v3 the calculations always led to not being able to remove really 100%, always some tiny amounts of pool shares were left in wallet. 

> and remove more liquidity from the pool

but how much is shown to you in terms of pool shares and expected output? Is it that remaining tiny amount? Or you mean UI didn't update and still shows same amount in your wallet after removal?

claudiaHash:
> not so sure we can do anything about it as in v3 the calculations always led to not being able to remove really 100%, always some tiny amounts of pool shares were left in wallet.
> 

tested again this morning and you're right, don't think we can do anything about it and it's not worth to do it since we have a perfected remove liquidity implementation in v4.

claudiaHash (approved):

PR #1553, opened 2022-06-27
Fix compute fre order and bump ocean lib

bogdanfazakas:
Fixes # .

Changes proposed in this PR:

- bump ocean lib to 1.0.13
- fix fre approve before calling buyFromFreAndOrder


src/@utils/order.ts
```diff
@@ -83,6 +88,8 @@ export async function order(
         exchangeContract: config.fixedRateExchangeAddress,
         exchangeId: asset.accessDetails.addressOrId,
         maxBaseTokenAmount: orderPriceAndFees.price,
+        baseTokenAddress: asset?.accessDetails?.baseToken?.address,
+        baseTokenDecimals: asset?.accessDetails?.baseToken?.decimals || 18,
```
EnzoVezzaro, line 92: I think we should move '18' to its own variable. Not sure how many 18's we have in the code, but I've seen it a couple of times, we should consider move it to a variable.

EnzoVezzaro (approved):
LGTM! just what suggested in comment

kremalicious (approved):
Confirmed on fixed price compute dataset it can be started with this PR

src/@utils/order.ts
kremalicious: it's not about being optional, it's about the app not breaking when it is already rendering but that specific data is not there yet. For hardcoded `18` we should have some const somewhere, but also through #1516 might be not needed anymore

PR #1555, opened 2022-06-29
remove onChange handlers on edit

kremalicious:
Hotfix. Violated my own rule of never trying to be smarter than Formik and overwriting any form field's `onChange` handler.

Currently editing the list of trusted algos is not working as expected so this PR restores that functionality by removing this `onChange` handler completely instead trying to hack it more.

Downside: Submit button gets activated although user hasn't done anything on the metadata & compute form yet. Solving this requires different approach, where we would have to teach `AssetSelection` component to validate itself with `validateField` on each click on the `input`, which requires a lot of refactoring in that component and almost all `Input...` components.


claudiaHash (approved):
Seems a fair compromise, works well on my end

PR #1556, opened 2022-06-29
added total orders to asset details

### LINKED ISSUE #1510: Add number of order on asset details
We have the number of orders on datatoken in the subgraph. It would be easy to display this value somewhere in Use or in details.

EnzoVezzaro:
- Fixes #1510 

Changes proposed in this PR:

- added total orders value to asset details

Screenshot 2022-06-29 at 09 43 07


src/components/Asset/AssetActions/Download.tsx
```diff
@@ -206,6 +206,10 @@ export default function Download({
       <div className={styles.info}>
         <div className={styles.filewrapper}>
           <FileIcon file={file} isLoading={fileIsLoading} />
+          <div className={styles.totalOrders}>
+            Total Orders:{' '}
+            <span className={styles.totalAmount}>{asset.stats.orders}</span>
```
claudiaHash, line 211: Think we should use asset.stats.consume here, in order to get rid of the build errors

EnzoVezzaro:
waiting for [updates](https://github.com/oceanprotocol/ocean.js/pull/1540 ) in ocean.js

kremalicious:
will work on some other placement somewhere as current implementation breaks easily with bigger numbers:

Screen Shot 2022-06-30 at 12 26 02

Also tending towards calling it "sales" in the UI which could still be explained further in info tooltip where we could explain what "order" means.

kremalicious:
Maybe too early but this gives us some nice space to summarize in future whatever more key metrics about an asset we come up with, like "xxx sales by xxx different accounts." or so

Screen Shot 2022-06-30 at 13 04 24

kremalicious (approved):

bogdanfazakas (approved):

PR #1568, opened 2022-07-05
Fix/change c2d orders order

bogdanfazakas:
Fixes # .

Changes proposed in this PR:

- order asset with provider fees last so that transaction is the last and we can use the block to validate c2d resource
- add proper digest values to our default algorithm container images


src/components/Publish/_constants.tsx
```diff
@@ -109,13 +109,15 @@ export const algorithmContainerPresets: MetadataAlgorithmContainer[] = [
     image: 'node',
     tag: 'latest',
     entrypoint: 'node $ALGO',
-    checksum: '' // TODO: how to get? Most likely needs to be fetched from DockerHub.
+    checksum:
+      'sha256:c60692d492dda8a432e2bb423a700fd470768eb394d182e76b5fe5f058f98cbc' // TODO: how to get? Most likely needs to be fetched from DockerHub.
```
kremalicious, line 113: so whenever there's a new Node.js version, this checksum changes? If so, would suggest switching to some fixed version image instead of `latest`

bogdanfazakas: yep that's an issue I wanted to discus in 10 min 😄

bogdanfazakas: we need to pick a specific version since this property can't be updated and stick with that for a published asset

bogdanfazakas:
closing this PR since it's handled in #1615

PR #1573, opened 2022-07-05
refactor price context to fetch multiple tokens

### LINKED ISSUE #1632: Make currency conversion work for h2o
started in #1573

kremalicious:
- Closes #1632

Putting foundation in place so we can make the conversion component work for any token which is listed on Coingecko. Starting with H2O. Still requires some hardcoded symbols as they are simply different in different network token contracts.

### Principle

There's a hardcoded const in `app.config.js`:

```js
coingeckoTokenIds: ['ocean-protocol', 'h2o']
```

Based on that, the Coingecko API URL is formed and fetched data is returned as is. So final response object is based on those Coingecko tokenIds:

```json
{
  "h2o": {
    "eur": 0.00
  },
  "ocean-protocol": {
    "eur": 0.00
  }
}
```

Then the `Conversion` component requires a new prop `symbol` so it can map passed `price` to above data structure.


src/components/@shared/FormFields/Nft/TxFee.tsx
```diff
@@ -55,7 +55,7 @@ export default function TxFee({
   return gasFee ? (
     <p>
       Gas fee estimation for this artwork
-      <Conversion price={gasFee} />
+      <Conversion price={gasFee} symbol={'OCEAN'} />
```
mihaisc, line 58: this doesn't seem ok, what's the relation between gas fee and ocean?

kremalicious: this is how we do currency calculations in codebase, always converting from OCEAN value, which the `gasFee` is in. The `Conversion` component needs to know the value (`price`) and the denomination of the value (`symbol`).

Agree this isn't super clear and was thinking to use different named props like `price` & `priceSymbol` to make connection more clear.

But in this specific case, for the gas fee, we should denominate in ETH, so could refactor this file a bit where we convert fetched gas fee to OCEAN somewhere above

mihaisc: oh, didn't imagine  it was this complicated. Maybe leave it for another pr since it's not really related to this one.

mihaisc:
while we're at it should we add conversions for the main token as well?
image

EnzoVezzaro:
Price Conversion is working fine. I have a question though, I don't see "H2o" on matic, is it normal? probably not related. 

Screenshot 2022-08-10 at 06 20 43

src/components/@shared/FormFields/Nft/TxFee.tsx
kremalicious: alse realized I closed the other gas fee estimation PRs so even tending towards removing estimation from here too, as the same argument applies here: nobody uses ETH mainnet and estimations are only useful for this network

kremalicious:
> I don't see "H2o" on matic testnet, is it normal?

I think so but I'll double check. As far as I know H2O token contract is only deployed on ETH & Polygon Mainnets. For testing other tokens we can only use Rinkeby where we we use the TUSDT token, which can be obtained for free

kremalicious:
> while we're at it should we add conversions for the main token as well?

total, did that, and more subtle tweaks:

- styling tweaks to put more focus on actual main number value
- tweak how we construct the user balance object in `@context/Web3` so this works for all tokens now (we hardcoded `eth` key before)
- hide conversion if value is 0

src/components/@shared/FormFields/Nft/TxFee.tsx
kremalicious: just removed it for now as same applies as for other gas fee estimation implementations: we can go back to it once we have ETH mainnet usage

kremalicious:
> I think so but I'll double check.

Now that I had a deep dive into what's going on in that wallet popover, this is how it is constructed:

1. we fetch the network's native token balance from user wallet, and add it to the balance object
2. we fetch the approved tokens for the network
3. with that, we fetch those token balances from user wallet, and add them to the balance object

So that wallet popover will always reflect the native token & approved tokens for each network

EnzoVezzaro:
if you connect to an empty wallet, we get NaN.

Screenshot 2022-08-12 at 05 21 34

kremalicious:
> if you connect to an empty wallet, we get NaN.

There has to be more to it, this is what I get when connecting empty wallet. On ETH Mainnet:

Screen Shot 2022-08-12 at 10 41 13

Polygon:

Screen Shot 2022-08-12 at 10 42 46

Only able to reproduce this on unsupported network, where we should simply prevent fetching approved tokens and clear the tokens up there:

Screen Shot 2022-08-12 at 10 43 11

mihaisc:
I get the NaN message for just a moment if i quickly switch wallets and open the popup , it shows proper values in less than 1 sec. Not sure if relevant , wouldn't focus on this. This is not necessarily related to what @EnzoVezzaro  said

mihaisc (approved):
i give this my `good enough` stamp of approval 🤣

EnzoVezzaro (approved):

Issue #1534, opened 2022-06-22
Show TVL in asset teaser


mihaisc:
I often am interested in the liquidity of an asset and this info feel kinda far sometimes .  Alternatives:
a. Just find a place to put it in the teaser (disclaimer: the following image is just a quick edit , it does not reflect a good UI or placement for this information !!)
![image](https://user-images.githubusercontent.com/26000280/175053078-8570a3f8-9b86-4a81-9112-4e26167ba1bd.png)
b. Create a larger version of the asset teaser that can be toggled in the settings ( simple teaser , extended teaser ) . Similar to how in other places you can pick between grid and list. Or maybe experiment with lists? Not to fond of lists.
![image](https://user-images.githubusercontent.com/26000280/175055096-4b174bd5-3b5f-4b3d-8226-6a5c573aacc9.png)

Also we already have all the data we need, so it's just a visual thing.

jamiehewitt15:
Seems like a good idea to me, particularly as the TVL is one of the main ways we have of curating good quality assets. The first visual seems like the most straight forward implementation

PR #1577, opened 2022-07-06
adding TVL to asset teaser

EnzoVezzaro:
- Feature #1534 .

Changes proposed in this PR:

- show TVL in asset teaser (option a)

Final result: 

Screenshot 2022-07-06 at 07 25 04


EnzoVezzaro:
> * show currency
> * the currency for TVL is a user setting so this number needs to change upon changing currency too

🤔  maybe i misunderstood the [assignment](https://github.com/oceanprotocol/market/issues/1534#issue-1280248477). I though we're showing the base token liquidity at first (basically option a). I think the currency thing goes more into [option b](https://github.com/oceanprotocol/market/issues/1534#issue-1280248477). Should we start working on it ?

EnzoVezzaro:
> image So it doesn't need to be currency, it's in base token

Do we need changes on this PR?

PR #1594, opened 2022-07-13
add MDEditor in Publish / Edit

### LINKED ISSUE #1496: Add markdown editor in publish form
We need a markdown editor for the description.
We could check this out https://uiwjs.github.io/react-md-editor/

EnzoVezzaro:
- Fixes #1496 .

Changes proposed in this PR:

- Added MDEditor in Description field

Screenshot 2022-08-22 at 07 19 47

WIP:

- [X] finish logic & styling
- [X] make sure "all markdown previews we have in our app renders exactly like we render the final asset details."
- [X] add tabs for preview
- [X] light / dark mode
- [x] bug: in long texts - after 500 words or so, it splits into a weird transparent editor (if you select you see the text) - WIP

For testing: 

- Markdown Editor:
-- Publish Form
-- Edit Form
- Mardown Preview
-- Publish (preview step)
-- Edit Form (preview tab)
-- Asset Details
-- Asset Teaser


next.config.js
```diff
@@ -1,3 +1,5 @@
+const removeImports = require('next-remove-imports')()
```
kremalicious, line 1: I see the dependency says this is a requirement for tit to work with Next.js, which is not good. Basically this implies we have to double check every dependency we use which might import their own CSS. Data table component, toast, etc.

Adding this `next-remove-imports` means we remove all those imports from all dependencies which seems excessive to get 1 dependency to work

EnzoVezzaro: Ok, I will check what are the consequences and see if it can be removed (or some other tweak).

kremalicious:
As we basically just want the formatting buttons, maybe https://github.com/github/markdown-toolbar-element is a better choice for solving this. Then we don’t need to deal with multiple ways of rendering markdown

EnzoVezzaro:
> As we basically just want the formatting buttons, maybe https://github.com/github/markdown-toolbar-element is a better choice for solving this. Then we don’t need to deal with multiple ways of rendering markdown or the potential overhead currently choosen dependency might bring

- move to #1599

jamiehewitt15:
If we want to make this even more friendly for non-developers we could user a wysiwyg editor that can export the content as markdown. Something like [this](https://www.npmjs.com/package//react-draft-wysiwyg) looks like it could do the trick

EnzoVezzaro:
> If we want to make this even more friendly for non-developers we could user a wysiwyg editor that can export the content as markdown. Something like [this](https://www.npmjs.com/package//react-draft-wysiwyg) looks like it could do the trick

Thanks @jamiehewitt15. I wouldn't introduce another editor as for now 😅 Both editors (MDEditor / Markdown-toolbar) actually do the trick, but they don't solve all the UX problems. Here, for example, seems to be appreciated the [split screen](https://github.com/oceanprotocol/market/pull/1594#issuecomment-1183192612) functionalities, so makes more sense for us to keep working with MDEditor. What we discussed earlier in standup should be enough for our use case.

next.config.js
EnzoVezzaro: without `next-remove-imports`, this doesn't work at all. 😞

EnzoVezzaro:
Being struggling to style this component, to style whatever I need to find some tricky CSS way to do it. Also, it should handle dark-mode, but doesn't seem to work at all. I've spent too much time trying to make work, but I haven't being able to. This affects not dark mode, but also ignores light theme's color scheme.

Screenshot 2022-08-19 at 07 41 54
Screenshot 2022-08-19 at 07 46 10

Considering also the problem with `[next-remove-imports](https://github.com/oceanprotocol/market/pull/1594#discussion_r920019139)`, I would be more inclined to change to @jamiehewitt15 's [suggestion](https://github.com/oceanprotocol/market/pull/1594#issuecomment-1216693600) or similar.

kremalicious:
Agree with Enzo, no new editor trying to solve all use cases. To me, this is what we want to solve here, and nothing more:

1. previewing description is the most important preview, yet the preview step is not easy to reach
2. some people are not used to Markdown but still want to do small formatting

For solving 1, it absolutely requires that we render the markdown the same way we render the big preview/final description. We have methods and components for that. This can‘t be solved with another rendering of some plugin. In theory we do not even need a dedicated markdown preview plugin here, we have that already with our own components.

Then, UI-wise, not sure we want to go into creating own custom UI by just copying the layout of GitHub comments authoring. As long as we manage to separate write & preview and not have it side-by-side, it should be good

EnzoVezzaro:
haven't find a way to fix this: 

- bug: in long texts - after 500 words or so, it splits into a weird transparent editor (if you select the text you'll see it)

Screenshot 2022-08-22 at 07 22 41

It should work as expected by default (in the [docs](https://uiwjs.github.io/react-md-editor/#support-nextjs) works), but our project seems to have something that the component doesn't like.

jamiehewitt15 (approved):
This works well. Have tested in both light mode, dark mode and on both the publish and edit pages

kremalicious:
quick review so it doesn't get merged with all those problems:

- preview is the most important thing we want to solve here, yet the button for it is hidden and not clear at all. And impossible to hit on mobile, which holds true for all buttons in that toolbar, unusable on mobile.
- editor uses different typography
- editor uses different colors, like suddenly there's a lot of blue text which we never use in the UI.
- editor preview uses its own markdown rendering, independent from how we render the final description, which is exactly what we want to prevent. The preview here, the big publish preview, and the final page all need to have the same rendering, otherwise it's not a preview. Markdown is not Markdown, there are many fine differences in different libraries, we try to always use [GitHub flavored Markdown](https://github.github.com/gfm/) which all our existing Markdown components do. Stylistically the current preview renders our markdown completely different than our actual rendering, completely different styles like quote style and so on. Again, theoretically we do not need a plugin for the preview, we can do it on our own as this PR already creates a custom component with full control over markup.
- button starts titles at h1. Look at the final page and see which headings are already taken by other elements like the main title, then adapt editor buttons to only include titles starting from that hierarchy. Most likely h2 or h3. We can't have multiple h1 on the final asset details page. Would opt for having only 1 level of title hierarchy as GitHub does (and they make it an h3) as those buttons take away too much space but that's debatable

mihaisc:
we will stop working on this. Too much effort. Wil not delete the branch, maybe we will continue at a later time.

PR #1596, opened 2022-07-13
remove tvl from footr

### LINKED ISSUE #1595: Remove tvl from footer
Remove tvl info
image

mihaisc:
Fixes #1595


kremalicious:
We should keep the tooltip, but only the part that it’s counted on-chain. Otherwise we create this confusion again where asset numbers in footer do not match the search results

mihaisc:
good catch

EnzoVezzaro (approved):

PR #1601, opened 2022-07-14
catch nft state on asset detail

### LINKED ISSUE #1600: Proper state check for assets
We need to have a proper message based on state. For example https://market.oceanprotocol.com/asset/did:op:73fcf31760687e5deb8e75688e3ae090adc29c8699e3e74d23bbac9a28627f7d has state 3 , revoked.

EnzoVezzaro:
- Fixes #1600 .

Changes proposed in this PR:

- added catch on nft state (solves app crashing)
- handled nft states

Result:

Screenshot 2022-07-15 at 06 41 37

Current states according to [docs](https://docs.oceanprotocol.com/concepts/did-ddo/#state)

0 | Active.
1 | End-of-life.
2 | Deprecated (by another asset).
3 | Revoked by publisher.
4 | Ordering is temporary disabled.

Messages: 

0: no message
1: 'This asset is in End-of-life.'
2. 'This asset has been deprecated.'
3: 'This asset has been revoked.'
4. no message

### DIFF
--- a/src/@context/Asset.tsx
+++ b/src/@context/Asset.tsx
@@ -100,7 +100,7 @@ function AssetProvider({
             break
           case 4:
             message =
-              'This asset has been disabled. If you just published this asset, please contact support.'
+              'This asset has been temporary disabled. If you just published this asset, please contact support.'
             break
         }
 
--- a/src/@context/Asset.tsx
+++ b/src/@context/Asset.tsx
@@ -100,7 +100,7 @@ function AssetProvider({
 
         setTitle(`This asset has been flagged as "${state}" by the publisher`)
         setIsV3Asset(await checkV3Asset(did, token))
-        setWarning(`\`${did}\`` + `\n\nPublisher Address: ${asset.nft.owner}`)
+        setError(`\`${did}\`` + `\n\nPublisher Address: ${asset.nft.owner}`)
         LoggerInstance.error(`[asset] Failed getting asset for ${did}`, asset)
         return
       }
--- a/src/@context/Asset.tsx
+++ b/src/@context/Asset.tsx
@@ -78,6 +78,13 @@ function AssetProvider({
             '\n\nWe could not find an asset for this DID in the cache. If you just published a new asset, wait some seconds and refresh this page.'
         )
         LoggerInstance.error(`[asset] Failed getting asset for ${did}`, asset)
+      } else if (asset.nft.state === 3) {
+        setIsV3Asset(await checkV3Asset(did, token))
+        setError(
+          `\`${did}\`` +
+            '\n\nThis asset has been reboked. If you just published a new asset, please contact support.'
+        )
+        LoggerInstance.error(`[asset] Failed getting asset for ${did}`, asset)
       } else {
         setError(undefined)
         setAsset((prevState) => ({
--- a/src/@context/Asset.tsx
+++ b/src/@context/Asset.tsx
@@ -25,6 +25,7 @@ export interface AssetProviderValue {
   title: string
   owner: string
   error?: string
+  warning?: string
   isAssetNetwork: boolean
   isV3Asset: boolean
   isOwner: boolean
@@ -52,6 +53,7 @@ function AssetProvider({
   const [owner, setOwner] = useState<string>()
   const [isOwner, setIsOwner] = useState<boolean>()
   const [error, setError] = useState<string>()
+  const [warning, setWarning] = useState<string>()
   const [loading, setLoading] = useState(false)
   const [isAssetNetwork, setIsAssetNetwork] = useState<boolean>()
   const [isV3Asset, setIsV3Asset] = useState<boolean>()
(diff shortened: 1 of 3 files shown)


mihaisc:
we need to handle all the states https://docs.oceanprotocol.com/concepts/did-ddo/#state

mihaisc:
Just put in place some generic messages and we can work on those as we go. I'm sure there will be suggestions 😁

mihaisc:
Why did you put `If you just published this asset, please contact support` at the end of all messages?

EnzoVezzaro:
> Why did you put `If you just published this asset, please contact support` at the end of all messages?

There's no really a reason 😅 it's just something I made up, we can remove it. The rest of the messages is fine?

mihaisc:
`1: 'This asset is End-of-life.'` 
Also don't cover state 4, because the asset is indexed in that case, we just need to disable the buy button

kremalicious:
Please use the title of the error alert to communicate the actual error, so users actually get that. So in the shown case the title should say "This asset is revoked", or "Revoked asset" or so.

The text under the alert title it is meant to inform users on what to do next, not hide the actual reason somewhere in between.

EnzoVezzaro:
> * Proper state check for assets #1600

Currently we're showing a [generic text "Could not retrieve asset"](https://github.com/oceanprotocol/market/blob/fix/issue-1600-proper-state-check-asset/src/components/Asset/index.tsx#L21) in the title if there's no asset or there's an error or warning. Show we change this logic to handle the title if passed?

kremalicious:
I would say yes, cause so far we used a generic title as we used for all „not found“ cases without knowing the exact reasons why it can‘t be retrieved. Now that we add cases where we exactly know the reason the asset can’t be retrieved we need to communicate it. And more than kinda hidden in the longer text

EnzoVezzaro:
ok, sounds good. Do we have any additional documentation regarding this states? in the [docs](https://docs.oceanprotocol.com/concepts/did-ddo/#state ) we explain the possible states, but we're not actually saying what those states actually mean or how they are raised. Even though it's pretty self-explanatory :-p, but it would be helpful to know how exactly these states changes.

kremalicious:
If we only repeat in the message what‘s in the title we don‘t need the message. Ideally, message in this context tells user how to resolve the error but not sure what to do for depreceated/eol assets. „Contact the publisher…“ would be maybe a bit useful if we also output the publisher but not sure we want to go there. The info that all those states are done by the publisher is useful though, so if we use message this should be on all states.

mihaisc:
We are overcomplicating thing. It's enough to just write a warning title with  "Asset flagged as {state text} by publisher" . And maybe show publisher address. That's it . We can improve this in time if we see fit.

EnzoVezzaro:
Something like this could work?

Screenshot 2022-07-25 at 09 27 06

mihaisc:
Perfect from my point of view.

src/@context/Asset.tsx
```diff
@@ -25,6 +25,7 @@ export interface AssetProviderValue {
   title: string
   owner: string
   error?: string
+  warning?: string
```
kremalicious, line 28: It’s a nice touch to differentiate between warning & error in UI, but in this case the asset the user looked for and not being shown is always an error. So fine if we handle all this in existing `error` state

src/@context/Asset.tsx
```diff
@@ -78,7 +80,32 @@ function AssetProvider({
...
+          case 3:
+            state = 'revoked'
+            break
+        }
+
+        setTitle(`This asset has been flagged as "${state}" by the publisher`)
+        setIsV3Asset(await checkV3Asset(did, token))
+        setWarning(`\`${did}\`` + `\n\nPublisher Address: ${asset.nft.owner}`)
+        LoggerInstance.error(`[asset] Failed getting asset for ${did}`, asset)
+        return
```
kremalicious, line 105: Just double checking, are `end-of-life` assets indexed or not? Cause in that casde returning here will not show the asset. Do we even have docs on what exactly the states should imply?

EnzoVezzaro: not sure what implies, I couldn't find anything useful on the docs, but maybe I'm just not looking in the correct place. @mihaisc do you know where we can find this info?

mihaisc: to quote myself `1,2,3 - same action just different message` . The metadata is not indexed.

src/@context/Asset.tsx
kremalicious: Some small leftover here, and further down `warning` state is still defined

kremalicious (approved):

src/@context/Asset.tsx
EnzoVezzaro: > Some small leftover here, and further down `warning` state is still defined

I've removed the unused states, but I can't remove line 28 because if I remove `warning` from the AssetProviderValues then the [`AssetDetails`](https://github.com/oceanprotocol/market/blob/fix/issue-1600-proper-state-check-asset/src/components/Asset/index.tsx#L12) breaks. If we remove it from the provider, we'll need to remove it completely.

kremalicious: Ah yes, also remove from `components/Asset/index.tsx`, then nothing should break as we are not using this newly introduced `warning` state

PR #1606, opened 2022-07-18
Fix long title

mihaisc:
`Dotdotdot` not working as expected so we need to slice for bigger titles. Why 200? Totally random, more than enough for the size of the teaser.
We need to find a good approach for the title in the asset page. Just a simple slice is not optimal , it loads faster but you still get this beauty
image

[test asset](https://market-git-fix-fixlongtitle-oceanprotocol.vercel.app/asset/did:op:dd5f3be4dd045027c8aa4e2fba935def2e071dbce1e8e5349672becd8462f861)

### DIFF
--- a/src/components/@shared/Page/index.tsx
+++ b/src/components/@shared/Page/index.tsx
@@ -2,7 +2,6 @@ import React, { ReactNode, ReactElement } from 'react'
 import PageHeader from './PageHeader'
 import Seo from './Seo'
 import Container from '@shared/atoms/Container'
-import Dotdotdot from 'react-dotdotdot'
 
 export interface PageProps {
   children: ReactNode
@@ -27,7 +26,7 @@ export default function Page({
       <Container>
         {title && !noPageHeader && (
           <PageHeader
-            title={<Dotdotdot clamp={3}>{title.slice(0, 300)}</Dotdotdot>}
+            title={<>{title.slice(0, 400)}</>}
             description={description}
             center={headerCenter}
           />


EnzoVezzaro:
Here you could try these CSS properties in [module.css](https://github.com/oceanprotocol/market/blob/main/src/components/@shared/Page/PageHeader.module.css#L10) and see how come up:

`
   white-space: nowrap;
    overflow: hidden;
    text-overflow: ellipsis;
    display: inherit;
`

kremalicious:
> Here you could try these CSS properties in [module.css](https://github.com/oceanprotocol/market/blob/main/src/components/@shared/Page/PageHeader.module.css#L10) and see how come up

This will cut off the title with `…` in the first line in all viewports.

In the shown screenshot in PR description, the problem is this long string is one word for the browser. The title will line break if it is too long, but only between words. So if we want to have long words break into new lines too, then `overflow-wrap: break-word;` would be the solution.

For more playing around: https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/

mihaisc:
Sorry, not sure i understand , what would be the desired way to go for this case?

kremalicious:
```css
.title {
  overflow-wrap: break-word;
  word-wrap: break-word;
  word-break: break-word;
  hyphens: auto;
}
```

resulting in

Screen Shot 2022-07-27 at 13 13 54

mihaisc:
Added a test asset in the main comment

mihaisc:
I would stick with 3 rows. If in the future it makes sense to show more, we can easily increase the number.
image

kremalicious:
we shouldn't use DotDotDot for the asset details, we still can `slice` it there though to prevent those extreme edge cases locking the browser.

Just cuts off way too many valid use cases as number of lines depends on way too many factors, like on [this asset](https://market-git-fix-fixlongtitle-oceanprotocol.vercel.app/asset/did:op:fc0ffb0059c36894631a93118b55126fc12e7368d0e70ffda23d2f9bf2c36344):

Screen Shot 2022-07-28 at 15 03 03

Also doesn't correctly work when window is resized and such, and on top of that user has no way to see the full title

mihaisc:
True, didn't check that. I increased the size to 400. I think it's more than enough
image

EnzoVezzaro (approved):
LGTM

kremalicious (approved):

Issue #1567, opened 2022-07-04
Market and Barge Script results in TypeError


Simon-Pokorny:
Hi everyone,

I would like to setup a local development environment and want to use the barge script to have all depencencies running locally. Unfortunately I am confronted with a TypeError as soon as I switch the network to localhost in Metamask. You can find a detailed bug description below. Any help with this issue is much appreciated.  
Thank you in advance!  

**Describe the bug**
As soon as I switch the network in metamask to localhost 8545 I see the following issue:
<img width="716" alt="image" src="https://user-images.githubusercontent.com/54535910/177133799-d213ab4d-91cd-4ff7-b52f-93f8ce6a4c7e.png">

**To Reproduce**
1. Started Barge script with `./start_ocean.sh --with-thegraph`
2. Build a docker image with market with the following dockerfile:

        FROM node:16.14  
        COPY . /market   
        RUN chown -R node /market  
        WORKDIR /market  
        USER node  
        RUN npm install
      Where the copy comand in line 2 simply copies the content from this repository into the folder /market
3. Set NEXT_PUBLIC_METADATACACHE_URI="http://127.0.0.1:5000" in .env file
4. Adjusted subgraphUri to http://127.0.0.1:9000 in the function getDevelopmentConfig in /src/@utils/ocean.ts
5. Copy contact addresses  

          ./scripts/copy-contracts.sh
6.   Start market container with artifacts

          docker run -it --rm --network host -v $(pwd)/node_modules/@oceanprotocol/:/market/node_modules/@oceanprotocol/ontracts/artifacts <IMAGENAME> npm start
        
7. Added private key to metamask wallet and confirmed, that the expected balance is shown. 
8. Switched network in metamask to localhost and the error above occures.
**Expected behavior**
I would expect, that the market works in the same way as with other networks

**Desktop (please complete the following information):**
Browser: Chrome
OS:  Ubuntu 16.04.7 LTS

kremalicious:
looks like an oversight in code as local network can't be found in the network metadata. So in the error you could try adding a `!network` to the condition at the beginning.

Developing market against barge has always been a bit tricky so we always suggest to use a remote test network, which the app is setup to do by default.

htcarr3:
@kremalicious, can you explain what you mean by "test network"? Do you mean running all of those containers in the cloud as opposed to running them locally?

Simon-Pokorny:
[...]

PR #1608, opened 2022-07-19
V4 barge

rogargon:
This PR tries to address some issues when trying to use the market just over locally deployed infrastructure (aquarius, ganache, contracts, thegraph, provider,...), similar to those described in #1567.

Changes proposed in this PR:
- Example .env to use with barge
- Add local barge to networks-metadata.json
- Use dev config with locally deployed contracts for network 8996
- start-local script triggering local ocean-subgraph codegen

Altogether, they should make it easier to use the market locally. Tested and able to publish both a dataset and an algorithm. I also bought and downloaded the dataset.


kremalicious:
this affects too much running the app against remote networks, like always adding dev network to network selection at all times. And unnecessary duplication with added scrips which only change 1 line.

So this would need some more thinking and a better strategy, like:

- read out the `chainIdsSupported` & `chainIds` arrays from `.env` vars
- read out subgraph URI for pre-generation from `.env` vars
- then pretty sure we need some signal from user that they want to run against barge, as we need to conditionally load or not load contract artifacts. Ideas welcome

htcarr3:
I'm glad to see someone trying to get some traction on this. The docs don't paint an accurate picture on how to make barge/market stack locally, and has made onboarding our team slow and developing on ocean difficult. There are errors if you don't include an infura project id, errors when trying to connect to local ganache, and various development configs (ocean.ts) are commented out without any mention in the docs. We would be happy to help develop a friendlier local dev environment. Can you help clear up what are the known local development issues?

shadmoney:
> I'm glad to see someone trying to get some traction on this. The docs don't paint an accurate picture on how to make barge/market stack work locally, and has made onboarding our team slow and developing on ocean difficult. There are errors if you don't include an infura project id, errors when trying to connect to local ganache, and various development configs (ocean.ts) are commented out without any mention in the docs. We would be happy to help develop a friendlier local dev environment. Can you help clear up what are the known local development issues?

Agreed on this sentiment, we would also like to contribute back to the documentation on the missing gaps. This will help more teams build on ocean quickly and efficiently.

rogargon:
The same for us, not being able to run the Market on a local Barge was complicating our development efforts. These changes are just a quick and dirty successful test as we have been able to make it work. We will continue elaborating on them, so the tests pass and it integrates smoothly into the main branch. Additionally, we can add a section to the README about the steps to get it working

htcarr3:
> 

Thank you, yes at the very least the README should make it clear that, while barge can be run locally, market does not work with it out of the box. The docs currently read as if it should work out of the box.

kremalicious:
This will be taken over by #1729 and all solutions to this incorporated there

PR #1609, opened 2022-07-19
fix download

mihaisc:
Will move this to ocean.js after testing. 

for this: did:op:2b9f0d0438eb603b30b4ac99a701f52d48a9b8d15e2254ea8b5e7f816f587987 it does nothing because it wants to redirect
for normal asset should just download.
Need to fix the name.

### DIFF
--- a/src/@utils/provider.ts
+++ b/src/@utils/provider.ts
@@ -115,7 +115,7 @@ export async function downloadFile(
   xhr.responseType = 'blob'
   xhr.open('GET', downloadUrl)
   xhr.onload = () => {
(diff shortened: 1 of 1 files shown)


EnzoVezzaro:
Tested, LGTM. It's working as described. 👍🏽

As for the title, you could add these properties in the [module.css](https://github.com/oceanprotocol/market/blob/main/src/components/@shared/Page/PageHeader.module.css#L10) and see how come up:

`
   white-space: nowrap;
    overflow: hidden;
    text-overflow: ellipsis;
    display: inherit;
`

mihaisc:
I think you have the wrong pr 😁  , it's another one with the title, but good tips

PR #1614, opened 2022-07-21
Remove AMM Pools

### LINKED ISSUE #1607: Remove AMM pools functionality
As we are moving away from the dynamic pricing schema, we need to remove everything POOL related.

- [x] #1614 Remove AMM related code (UI)
- [x] #1614 Remove AMM related code (cleanup)
- [x] #1614 Remove tests
- [x] #1610 Update READMEs (docs)

EnzoVezzaro:
- Fixes #1607 .

Changes proposed in this PR:

- refactor PriceUnit
- fix fees component after rework
- removed pool and trade tabs completely
- removed pool shares and pool transactions in profile
- change main query in homepage
- rework approve button removal
- remove price display from asset teasers
- rework on various components to remove amm references
- simplify Price component
- remove unnecessary packages
- refactor pricing step on publish 
- typings and publish data cleanup
- content cleanup

Deprecated "dynamic" pricing assets:

- Dataset/Algo with Dynamic Pricing

- Removed dynamic pricing algo from list selection:

Free and Fixed pricing assets should display normally (no change).


src/@utils/aquarius.ts
```diff
@@ -47,6 +47,7 @@ export function generateBaseQuery(
           ...(baseQueryParams.filters || []),
           getFilterTerm('chainId', baseQueryParams.chainIds),
           getFilterTerm('_index', 'aquarius'),
+          getFilterTerm('accessDetails.type', ['free', 'fixed']),
```
mihaisc, line 50: `accessDetails` is not on the ddo, we accidentally added it for some assets in the edit form. So doesn't have any purpose here

EnzoVezzaro: 🤔 mmm... this is a problem then. I'm looking at the [DDO docs](https://docs.oceanprotocol.com/concepts/did-ddo/#full-enhanced-ddo-example) now, but I'm not sure how to structure the query to get only "fixed" and "free" assets. What do you think is the best approach to filter these results?

EnzoVezzaro: We don't need to filter these assets. Instead, we just need to alert the user on the Button. Let's just show "price not available" on these assets.

EnzoVezzaro:
😅 I'm not really happy with the result. @kremalicious can you help me out here please, I don't think we're communicating correctly what's the problem is here. Maybe we can add a note bellow the button to let the user know what the problem is. Something like: "This asset's pricing schema is not valid anymore. Contract the publisher if you need this asset." or something similar. 

Also the label "Price is not available" is kind of ugly 🥲 do you have any suggestion?

Screenshot 2022-07-26 at 06 45 28

kremalicious:
> Also the label "Price is not available" is kind of ugly 🥲 do you have any suggestion?

It's worse that the label is simply lying to the user, as I can clearly see a price. So this screenshot does not make sense.

So replace everything which is bound to price:

- price
- action
- action help text

And replace it with an `Alert` of type `warning` or `info` explaining what's going on, also linking two some blog post further explaining what happened to pools.

EnzoVezzaro:
> > Also the label "Price is not available" is kind of ugly 🥲 do you have any suggestion?
> 
> It's worse that the label is simply lying to the user, as I can clearly see a price. And if you disable a button to the user they will try to figure out how to enable it again but it's an impossible task. So this screenshot does not make sense.
> 
> So replace everything which is bound to price:
> 
> * price
> * action
> * action help text
> 
> And replace it with an `Alert` of type `warning` or `info` explaining what's going on, also linking to some blog post further explaining what happened to pools.

Mmmm.... 🤔  Thanks for that, I guest we can do that, but on our last discussion we talked about just disabling the button and changing the label. Not sure how to proceed.

kremalicious:
we can get around all this by hiding dynamic priced algos in that algo selection list. There we always query subgraph too so we should have the price type on them somewhere.

For dynamic compute dataset: hide everything (CtD warning, algo selection, action button, action help text) and replace with alert described above

EnzoVezzaro:
> we can get around all this by hiding dynamic priced algos in that algo selection list. There we always query subgraph too so we should have the price type on them somewhere.
> 
> For dynamic compute dataset on its asset details page: replace everything (CtD warning, algo selection, action button, action help text) with alert described above

As for datasets with dynamic pricing, this is result after removing all the actions.

Screenshot 2022-07-27 at 09 32 37

is the copy correct?

kremalicious:
Maybe short and to the point, as "disabled" implies they might be activated again:

`Dynamic pricing with pools [is deprecated](https://linkme).`

EnzoVezzaro:
> we can get around all this by hiding dynamic priced algos in that algo selection list.

this has some implications related to the `accessDetails` property that makes it difficult to do without some hacky stuff in the middle. I think showing the alert in this case would be fine, what you think?

Screenshot 2022-07-27 at 12 47 02

kremalicious:
thanks! Pushed too early as I was finding my way through the code removing things where I started with the typings so I get errors I have to deal with. Removing everything and rather hiding prices seems like better approach, and with it we do not have to keep some parts in place to show useless prices on those dynamic assets. Pushing more removal in a sec where I just removed the price from the asset teasers. So if we also hide price on asset details we do not need to fetch anything for the dynamic assets anymore

kremalicious:
dynamic compute asset, no further problems anymore:

Screen Shot 2022-07-27 at 18 51 51

src/components/Asset/AssetActions/Compute/index.tsx
```diff
@@ -406,21 +369,32 @@ export default function Compute({
...
+          />
+        )}
       </div>
 
       {asset.metadata.type === 'algorithm' ? (
         <>
           <Alert
-            text="This algorithm has been set to private by the publisher and can't be downloaded. You can run it against any allowed data sets though!"
+            text={
+              "This algorithm has been set to private by the publisher and can't be downloaded. You can run it against any allowed data sets though!"
```
kremalicious, line 386: this is something we need to solve as this goes into the compute to data design. We need a new mechanism to signal algos to be downloadable or not.

kremalicious: whoops, ignore, mixed something up so there's no issue here, push incoming

kremalicious:
fun fact: we were using pool swap fee config values when calculating final price of fixed rate assets. Did not lead to problems as most fees are 0 in our case. Fixed in f934e8ccc37b6d26870c169807871b1d22d29ffd

mihaisc:
on pr: 
image
on main :
image

the last profile on the first row is this https://market.oceanprotocol.com/profile/jellymcjellyfish.eth

EnzoVezzaro:
> on pr: image on main : image
> 
> the last profile on the first row is this https://market.oceanprotocol.com/profile/jellymcjellyfish.eth

this has been solved in #1410 

Screenshot 2022-07-28 at 08 41 29

kremalicious:
> the last profile on the first row is this

I can't reproduce this on my side, I do not get double-trentmc0:

Screen Shot 2022-07-28 at 13 59 50

mihaisc:
Not sure what to think about this message `Dynamic pricing with pools [is deprecated](https://blog.oceanprotocol.com/ocean-market-changes-3384fd7e113c).` . It is very misleading and says nothing about the asset. Is it supposed to be temporary?
For example  [this asset](https://market-git-fix-issue-1607-remove-amm-code-oceanprotocol.vercel.app/asset/did:op:02b51f91cff9eecf8d444f2e9a8c7509927202d076f0ae026dbfaec437737f99) never had a pool

kremalicious:
> For example [this asset](https://market-git-fix-issue-1607-remove-amm-code-oceanprotocol.vercel.app/asset/did:op:02b51f91cff9eecf8d444f2e9a8c7509927202d076f0ae026dbfaec437737f99) never had a pool

looking into this, but as said we might have to live with edge cases like this with utility methods without any loading state management

mihaisc:
My point is we never should bother with this. Just say "No pricing schema available for this asset" and that's it, you cover all cases.

EnzoVezzaro:
> Not sure what to think about this message `Dynamic pricing with pools [is deprecated](https://blog.oceanprotocol.com/ocean-market-changes-3384fd7e113c).` . It is very misleading and says nothing about the asset. Is it supposed to be temporary? For example [this asset](https://market-git-fix-issue-1607-remove-amm-code-oceanprotocol.vercel.app/asset/did:op:02b51f91cff9eecf8d444f2e9a8c7509927202d076f0ae026dbfaec437737f99) never had a pool

Agree, we have some files without `accessDetails` which are not related to the pool stuff. What about these messages: 

- Asset with pools: "Dynamic pricing with pools is deprecated"
- Asset with no accessDetails: "Pricing schema has not been defined"
- Free / Fixed Asset: no alert

is there any other state? for example assets with `accessDetails`, but without a pricing schema defined (thinking about ocean.js)

mihaisc:
image
Total sales seems to have some issues

EnzoVezzaro:
> image Total sales seems to have some issues

looking into it

kremalicious:
So basically, without querying for `pools` in the subgraph there is no way to reliably differentiate between dynamic price and no price in our `getAccessDetailsFromTokenPrice()` method:

```
if (
  tokenPrice?.dispensers?.length === 0 &&
  tokenPrice?.fixedRateExchanges?.length === 0
) {
  accessDetails.type = 'dynamic_DEPRECATED'
  return accessDetails
}
```

This would be only way as far as I see, but would require keeping pools in the subgraph:

```
if (tokenPrice?.pools?.length > 0) {
  accessDetails.type = 'dynamic_DEPRECATED'
  return accessDetails
}
```

mihaisc:
at this point (or next week :)  ) dynamic price  = no price, so let's keep it simple

EnzoVezzaro:
> at this point (or next week :) ) dynamic price = no price, so let's keep it simple

Great, I'll change it to always display: "No pricing schema available for this asset"

Screenshot 2022-07-28 at 10 59 31

kremalicious (approved):

PR #1651, opened 2022-08-12
fix tag search

### LINKED ISSUE #1650: Tag search returns all assets
If you click on a tag on an asset page, it starts a search for that tag, but the results are all results.

E.g.
https://market.oceanprotocol.com/search?tags=agriculture&sort=metadata.created&sortOrder=desc

mihaisc:
This is a good enough fix for this bug. A more thorough refactor should be done in the bigger search issue to figure out what exact searches we want.
Fixes #1650


EnzoVezzaro (approved):
LGTM

kremalicious (approved):
works

PR #1655, opened 2022-08-15
adding UI for field files / links in edit form

### LINKED ISSUE #1244: Edit UI

EnzoVezzaro:
- Fixes #1244 .

Changes proposed in this PR:

- load 'sample links' from subgraph and validate

Screenshot 2022-08-15 at 09 29 33

- add placeholder for 'new url' (as we can't access the asset's url)

Screenshot 2022-08-15 at 09 29 39

The blur effect shows up only when you first access the edit form. If you edit the "new link" input, the blur is not apply to the new url.


bogdanfazakas:
found one issue, if the asset you want to edit already has sample files, the submit button is always disabled even if you update the sample field or any of them, on the other side the submit button is always enabled on assets without sample files set, even if you have even touched or updated any form input fields.


https://user-images.githubusercontent.com/13447142/195508812-78725c38-68b8-4476-8ffd-dccd22db4ecb.mov

EnzoVezzaro:
> found one issue, if the asset you want to edit already has sample files, the submit button is always disabled even if you update the sample field or any of them, on the other side the submit button is always enabled on assets without sample files set, even if you have even touched or updated any form input fields.
> 
>  Screen.Recording.2022-10-13.at.08.13.21.mov

oh boy 😅 getting that validation to work was challenging 😂 please let me know if it needs more changes 👍🏽

bogdanfazakas (approved):

mihaisc:
i updated this https://market-git-feature-issue-1244-edit-ui-oceanprotocol.vercel.app/asset/did:op:50c60ce0810bf8f8b1c6ca421e5231b106cf4de784945037498c84147e11dbc5 and when i download i still get the old file

mihaisc:
Updating just the sample file didn't work

mihaisc (approved):
lgtm

src/components/@shared/FormFields/FilesInput/Info.module.css
```diff
@@ -31,6 +31,10 @@
   padding-right: calc(var(--spacer) / 2);
 }
 
+.hideUrl {
+  filter: blur(0.2rem);
```
kremalicious, line 35: What is the purpose of this blurring?

Screen Shot 2022-10-14 at 13 44 01

Either we prevent access to that URL completely (I can just select and copy it, I can see it in the source code, etc.) or we show it. This in between does not solve anything I think. And on top of that, every new file I add gets blurred too. It is always fine showing that URL to the actual publisher.

kremalicious: aaah sorry, I get it now, was already wondering and started inspecting how the heck we get that existing URL decrypted :-) 

Still, when adding a new file, we have to show the just added URL

mihaisc: yea, i was a bit confused about the blur, not sure what the purpose it.

EnzoVezzaro: the blur is used to hide the fact that we can't access the url of the asset. @kremalicious is right, we should display the blur only when the user access the edit form, but if he edits the fields, the new url should be displayed without blur.

bogdanfazakas:
think we can merge this, just double checked latest changes and all seems to be in place

bogdanfazakas (approved):

PR #1656, opened 2022-08-15
change label 'data set' with 'dataset'

### LINKED ISSUE #1648: Consistent spelling of dataset
It's only a small thing but I noticed that we are not consistent with how we spell dataset (or data set). I think neither is technically wrong but it would be nice if we could be consistent.

data set vs dataset - 2022-08-11_08-56

EnzoVezzaro:
- Fixes #1648 

Changes proposed in this PR:

- change label "data set" with "dataset"

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,12 +103,12 @@ cp .env.example .env
 
 ## 🦀 Data Sources
 
-All displayed data in the app is presented around the concept of one data set, which is a combination of:
+All displayed data in the app is presented around the concept of one dataset, which is a combination of:
 
-- metadata about a data set
-- the actual data set files
-- the NFT which represents the data set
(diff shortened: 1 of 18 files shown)


jamiehewitt15 (approved):
Looks good to me

kremalicious:
just a sec., specially in readme this needs more thought:

- changing to "dataset" should be fine, also consistent how we do it in the docs
- we used "data set", especially in readme, before we introduced algorithms. After that we started using the term "asset" as it can either be data set or algorithm, which is then also the term used in the docs as far as I can see

So we should do:
- replace some occurrences depending on context with "asset" instead
- "dataset" can still be used when talked in context about algorithms too

kremalicious (approved):

PR #1660, opened 2022-08-23
fix publishing when connecting wallet on publish form

### LINKED ISSUE #1659: Publishing fails if user connects its wallet on the publish form
**Describe the bug**
Publishing fails with error `Cannot read properties of null (reading 'address')` if the user connects its wallet on the publish form.

**To Reproduce**
1. Go to Publish
2. Complete form
3. On the last step, connect wallet
4. CLick submit
5. See error

**Expected behavior**
It should run the operation

**Screenshots**

https://user-images.githubusercontent.com/13741335/185946943-d5517d44-e670-4624-b437-6d09b2aca458.mp4

EnzoVezzaro:
- Fixes #1659 .

Changes proposed in this PR:

- fix connection issue when the user connects the wallet on the publish form

### DIFF
--- a/src/components/Publish/Steps.tsx
+++ b/src/components/Publish/Steps.tsx
@@ -1,9 +1,14 @@
 import { ReactElement, useEffect } from 'react'
 import { useFormikContext } from 'formik'
-import { wizardSteps, initialPublishFeedback } from './_constants'
+import {
+  wizardSteps,
+  initialPublishFeedback,
+  initialValues
+} from './_constants'
 import { useWeb3 } from '@context/Web3'
 import { FormPublishData, PublishFeedback } from './_types'
 import { getOceanConfig } from '@utils/ocean'
+import router from 'next/router'
 
 export function Steps({
   feedback
@@ -22,11 +27,15 @@ export function Steps({
     setFieldValue('user.accountId', accountId)
   }, [chainId, accountId, setFieldValue])
 
-  // Reset the selected baseToken on chainId change
   useEffect(() => {
     if (!chainId) return
 
-    setFieldValue('pricing.baseToken', null)
+    // Reset the pricing values on chainId change
+    // the user needs to update the pricing schema on network changes
+    if (values.pricing.price) {
+      setFieldValue('pricing', initialValues.pricing)
+      router.push(`/publish/3`)
+    }
   }, [chainId, setFieldValue])
 
   // auto-sync publish feedback into form data values
--- a/src/components/Publish/Steps.tsx
+++ b/src/components/Publish/Steps.tsx
@@ -30,9 +30,10 @@ export function Steps({
   useEffect(() => {
     if (!chainId) return
 
-    // Reset the pricing values on chainId change
-    // the user needs to update the pricing schema on network changes
-    if (values.pricing.price) {
+    // Reset the pricing values on chainId change:
+    // - the user needs to update the pricing schema on network changes
+    // - if the tx has been started, don't restore pricing values
+    if (values.pricing.price && values.feedback[1].status === 'pending') {
       setFieldValue('pricing', initialValues.pricing)
       router.push(`/publish/3`)
     }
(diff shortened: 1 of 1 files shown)


jamiehewitt15:
Something odd happens. After connecting my account on the submit tab, it sends me back to the price tab. It's a bit disorienting. Publishing seems to work well though.

Screen recording:
https://www.loom.com/share/0afa79aa70bd40dab3189e645fa22b42

EnzoVezzaro:
> Something odd happens. After connecting my account on the submit tab, it sends me back to the price tab. It's a bit disorienting. Publishing seems to work well though.
> 
> Screen recording: https://www.loom.com/share/0afa79aa70bd40dab3189e645fa22b42

yeah, it shouldn't happen. It seems from the video that the form lost track of the price. I'll check 👍🏽

EnzoVezzaro:
> Something odd happens. After connecting my account on the submit tab, it sends me back to the price tab. It's a bit disorienting. Publishing seems to work well though.

Hi @jamiehewitt15 , please test it again now.

EnzoVezzaro:
So what we need to test is as follow: 

Issue:
- Go to [market](https://market.oceanprotocol.com/) and fill out the form without wallet. Connect the wallet ONLY at the end. You should get an error `Cannot read properties of null (reading 'address')`. This because the object `pricing.baseToken` is not form correctly: 

Screenshot 2022-08-24 at 06 52 04

PR solution: 
- Repeat the process above in this branch, this time you should be able to run the tx. The pricing object should be filled correctly. For some reason before we're emptying the pricing baseToken address. 

Screenshot 2022-08-24 at 06 51 54

jamiehewitt15:
> * Go to [market](https://market.oceanprotocol.com/) and fill out the form without wallet. Connect the wallet ONLY at the end. You should get an error `Cannot read properties of null (reading 'address')`. This because the object `pricing.baseToken` is not form correctly:

Yep, managed to recreate that error on the market. 

> * Repeat the process above in this branch, this time you should be able to run the tx. The pricing object should be filled correctly. For some reason we were emptying the pricing baseToken address when changing network.

Yeah it worked, looks like this solved now

src/components/Publish/Steps.tsx
```diff
@@ -1,9 +1,14 @@
 import { ReactElement, useEffect } from 'react'
 import { useFormikContext } from 'formik'
-import { wizardSteps, initialPublishFeedback } from './_constants'
+import {
+  wizardSteps,
+  initialPublishFeedback,
+  initialValues
```
jamiehewitt15, line 6: You've imported `initialValues` but aren't using it

src/components/Publish/Steps.tsx
```diff
@@ -1,9 +1,14 @@
...
+} from './_constants'
 import { useWeb3 } from '@context/Web3'
 import { FormPublishData, PublishFeedback } from './_types'
 import { getOceanConfig } from '@utils/ocean'
+import router from 'next/router'
```
jamiehewitt15, line 11: You've imported `router` but it isn't used

jamiehewitt15 (approved):
I did another test and it's working well. Thanks for fixing this

kremalicious:
Just a note: please ALWAYS test the code you are changing against what the code was actually doing. Not just for the issue you are trying to solve. 

Just by looking at this changeset, so is the basetoken correctly reset when I change the network? How if that code is fully removed? Cause every network has different approved basetokens.

EnzoVezzaro:
> Just a note: please ALWAYS test the code you are changing/removing against what the code was actually doing. Not just for the issue you are trying to solve.
> 
> Just by looking at this changeset, so is the basetoken correctly reset when I change the network? How exactly would that work if that code is fully removed? Cause every network has different approved basetokens.

Yes, I spent some time trying to understand why this [baseToken 'reset'](https://github.com/oceanprotocol/market/blob/main/src/components/Publish/Steps.tsx#L29) was put in place, as the baseToken actually updates correctly when we change network: 

Rinbeby: 
Screenshot 2022-08-25 at 06 44 45

Matic: 
Screenshot 2022-08-25 at 06 44 57

I'm still not sure why this was added to the codebase in PR #1516

bogdanfazakas (approved):
seems to work fine, also the switch baseToken looks ok

PR #1679, opened 2022-09-09
enhancement: autocomplete tag component config

LucaMilanese90:
Changes proposed in this PR:

- only include tags from mainnet in the autocomplete suggestions
- start showing matching results after at least 3 characters are typed

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,0 +1,11 @@
+[[headers]]
+  for = "/_next/image/*"
+
+  [headers.values]
+    Content-Security-Policy= "upgrade-insecure-requests"
+    Strict-Transport-Security = "max-age=63072000; includeSubDomains; preload"
+    X-XSS-Protection = "1; mode=block"
+    X-Frame-Options = "DENY"
+    X-Content-Type-Options = "nosniff"
+    Referrer-Policy = "strict-origin-when-cross-origin"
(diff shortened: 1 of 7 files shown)


EnzoVezzaro:
Found a couple of small things:

1) Styling:
- [ ] change font color to `var(--font-color-heading)` (in dark mode is not very visible)
- [ ] add `cursor: text;` to input

2) Results: I'm not completely sure it's fine. In the example down below I'd expect to see `opensea` as first option instead of second. 

Screenshot 2022-09-12 at 07 28 31

EnzoVezzaro:
> Not sure why but I already see the correct color for the input, this is from localhost though since I don't have access to the Vercel preview but the color is already set to be `--font-color-heading`. The search can also force to show results that only start with the typed input, not sure if this is the behaviour you expect, in this case typing `ope` would only show `opensea`. Screen Shot 2022-09-12 at 14 21 03

Yeah, it might be then. But why you can't access the preview? I'm not completely aware of the vercel setup, but I think you can access this [preview build](https://market-fbnspp9ys-oceanprotocol.vercel.app/publish/1).

> not sure if this is the behaviour you expect

Not really, the problem that I was referring to is related to the order itself. Basically if I have `ope` as input, I'd expect to have "opensea" as first option since it start with `ope` instead of "Europe" that ends with `ope`. I wonder what would happen if there's a lot of options (will words with the match at the beginning endup at the end of the list?)

LucaMilanese90:
I can see the link but the Vercel page was showing a 404 error for me before. Anyway, I tweaked the order/filter for the matched options, lets see if it works for you this time

jamiehewitt15:
I had a look at this but I think the vercel deployment is out of date, the matches are the same as before (Europe at the top). Probably something to do with the vercel settings

LucaMilanese90:
Yes, it says 'authorization required to deplyo'. I confirmed a couple of times but it's not working. Maybe one of you has to do it. This fork is a pain, was much easier to work on the main repo😅

LucaMilanese90:
@jamiehewitt15 can you please authorize the deployment? The sort order should be fine now and hopefully also the input font color

LucaMilanese90:
@EnzoVezzaro the requested changes should be hopefully addressed, please have a look when you have a moment.

PR #1685, opened 2022-09-12
integrate shacl schema into market's validation

### LINKED ISSUE #1657: Integrate shacl schema on market publish process
Aqua issues:
* https://github.com/oceanprotocol/aquarius/pull/867
* https://github.com/oceanprotocol/aquarius/issues/871

EnzoVezzaro:
- Fixes #1657 .

Changes proposed in this PR:

- [X] integrate Shacl schema from acquarius into the market
- [x] add Shacl validation to publish form
- [x] ~~add Shacl validation to edit form~~ moved to #1739


src/@utils/aquarius.ts
```diff
@@ -164,6 +165,31 @@ export async function retrieveAsset(
     }
   }
 }
+let schema: any
+export function retrieveShaclSchema(): Promise<ShaclSchema> {
+  console.log(schema)
```
jamiehewitt15, line 170: You can remove this log

src/@utils/schaclSchema.ts
```diff
@@ -0,0 +1,55 @@
+import { retrieveShaclSchema } from '@utils/aquarius'
+import { ShaclSchemaField } from '@context/MarketMetadata/_shaclType'
+
+import { capitalizeFirstLetter } from '@utils/textTransform'
+
+function getMinMax(valueField: string, field: ShaclSchemaField) {
+  let min = field.minLength
+  const max = field.maxLength
+
+  // TODO: remove once all fields in schema has minLength property
```
jamiehewitt15, line 10: I don't see an issue in aquarius for doing this? If it's not done yet we should add an issue. It would also be good to note the issue in the comment so we can track when it is done

src/@utils/schaclSchema.ts
```diff
@@ -0,0 +1,55 @@
...
+export async function validateFieldSchaclSchema(
+  keyField: string,
+  valueField: string,
+  value: any,
+  createError: any
+): Promise<any> {
+  const schemaFields: any = await retrieveShaclSchema()
+  const fieldSchema: ShaclSchemaField = schemaFields[keyField][valueField]
+  const { minLength, maxLength } = getMinMax(valueField, fieldSchema)
+  console.log(schemaFields)
```
jamiehewitt15, line 37: this log can be removed

src/@utils/schaclSchema.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  // TODO: add minLength when integrated in endpoint
```
jamiehewitt15, line 39: Same applies as above, lets make sure we are tracking when the change has been made

src/components/Publish/index.tsx
```diff
@@ -19,6 +25,7 @@ import { getOceanConfig } from '@utils/ocean'
 import { validationSchema } from './_validation'
 import { useAbortController } from '@hooks/useAbortController'
 import { setNFTMetadataAndTokenURI } from '@utils/nft'
+import { useMarketMetadata } from '@context/MarketMetadata'
```
jamiehewitt15, line 28: `useMarketMetadata` is another unused dependency

app.config.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
 
   // List of chainIds which metadata cache queries will return by default.
   // This preselects the Chains user preferences.
-  chainIds: [1, 137, 56, 246, 1285],
+  chainIds: [1, 137, 56, 246, 1285, 8996],
```
jamiehewitt15, line 14: What network is it that you're adding here? It's not in the list of supported chainIds so I'm not sure why you are adding it here in the pre-selected list?

bogdanfazakas: that is barge, definitely can be removed

src/@utils/aquarius.ts
EnzoVezzaro: yeah, some how I didn't push my last commit with these cleanup changes. But thanks for pointing this out.

src/@utils/schaclSchema.ts
EnzoVezzaro: yeah, this too is old comment. It has been removed, as this was already fixed on aqua.

jamiehewitt15 (approved):
Looks good to me

src/@utils/aquarius.ts
```diff
@@ -164,6 +165,30 @@ export async function retrieveAsset(
...
+let schema: Promise<ShaclSchema>
+export function retrieveShaclSchema(): Promise<ShaclSchema> {
+  if (schema) return schema
+
+  try {
+    const getSchema = axios.get(
+      `${metadataCacheUri}/api/aquarius/validation/schema`
+    )
+
+    return getSchema.then((res: any) => {
```
bogdanfazakas, line 177: can't we use here this type `ShaclSchema` ?

src/components/Publish/_validation.ts
```diff
@@ -1,31 +1,73 @@
...
 const validationMetadata = {
   type: Yup.string()
     .matches(/dataset|algorithm/g, { excludeEmptyString: true })
     .required('Required'),
   name: Yup.string()
-    .min(4, (param) => `Title must be at least ${param.min} characters`)
-    .required('Required'),
+    .required('Required')
+    .test(async (value, { path, createError }): Promise<any> => {
+      if (!value) return
```
bogdanfazakas, line 17: think we can move this logic in a separate method and just call it in the` test()` . we have the same logic repeating in quite a few places here.

bogdanfazakas:
could not validate (Continue button disabled) first publish step in the preview (Metadata) even tho there are no validation or console errors
Screenshot 2022-11-22 at 11 49 40


also there seems to be an issue with the tags:
Screenshot 2022-11-22 at 11 49 46

EnzoVezzaro:
> also there seems to be an issue with the tags:

I see 🙄 probably the last chances broke those fields, I'll take a look.

EnzoVezzaro:
ready for review @bogdanfazakas

src/@context/MarketMetadata/_shaclType.ts
```diff
@@ -0,0 +1,96 @@
...
+      values: ShaclSchemaField
+    }
+    deny: {
+      type: ShaclSchemaField
+      values: ShaclSchemaField
+    }
+  }
+  id: ShaclSchemaField
+  metadata: {
+    additionalInformation: any
```
kremalicious, line 23: be slightly more precise as this really shouldn't be e.g. a top-level string. So:

```
addionalInformation: {
  [key: string]: any
}
```

src/@context/MarketMetadata/_shaclType.ts
```diff
@@ -0,0 +1,96 @@
...
+      description: ShaclSchemaField
+      label: ShaclSchemaField
+      name: ShaclSchemaField
+      options: ShaclSchemaField
+      required: ShaclSchemaField
+      type: ShaclSchemaField
+    }
+    datatokenAddress: ShaclSchemaField
+    description: {
+      maxCount: number
```
kremalicious, line 78: why is this typed out when it's the same as `ShaclSchemaField`?

src/@utils/aquarius.ts
```diff
@@ -164,6 +170,30 @@ export async function retrieveAsset(
     }
   }
 }
+let schema: Promise<ShaclSchema>
+export function retrieveShaclSchema(): Promise<ShaclSchema> {
```
kremalicious, line 174: this should be an async function in which you then `await` things

src/@context/MarketMetadata/_shaclType.ts
```diff
@@ -0,0 +1,96 @@
...
+      maxLength: number
+      minLength: number
+      pattern: string
+    }
+    serviceEndpoint: ShaclSchemaField
+    timeout: ShaclSchemaField
+    type: ShaclSchemaField
+  }
+  version: ShaclSchemaField
```
kremalicious, line 95: this file seems unrelated to the MarketMetadata context provider so why is it here? This should be a global typing in `src/@types`

src/@utils/schaclSchema.ts
```diff
@@ -0,0 +1,42 @@
+import { retrieveShaclSchema } from '@utils/aquarius'
+import { ShaclSchemaField } from '@context/MarketMetadata/_shaclType'
+
+import { capitalizeFirstLetter } from '@utils/textTransform'
+
+function getMinMax(valueField: string, field: ShaclSchemaField) {
```
kremalicious, line 6: what is the point of this method? Seems over engineered as the only usage of this method further down in this file can be replaced with:

```
const { minLength, maxLength } = field
```

src/@utils/schaclSchema.ts
```diff
@@ -0,0 +1,42 @@
...
+  return {
+    minLength: min,
+    maxLength: max
+  }
+}
+
+export async function validateFieldSchaclSchema(
+  path: string,
+  value: string,
+  createError: any
```
kremalicious, line 19: this `createError` should be typed as its always same usage so pretty sure we can figure out its typing

src/@utils/aquarius.ts
```diff
@@ -22,6 +23,11 @@ export function escapeEsReservedCharacters(value: string): string {
   return value.replace(pattern, '\\$1')
 }
 
+export interface ShaclSchemaRes {
```
kremalicious, line 26: and in general do not type out things already in axios. As with all our other axios usage, this is how you type their responses:

```
type Asset = { example: string }
const response: AxiosResponse<Asset> = await axios.get()
```

kremalicious:
conceptionally, di I get this right, that you are fetching the schema on every user keystroke? If so, this needs to be changed to fetch the schema only e.g. when user arrives on the publish form

mihaisc:
taking too much time, we will maybe think of a lighter implementation to do this validation

PR #1691, opened 2022-09-14
added N/A for negative values as well as missing data

### LINKED ISSUE #1684: Correctly handle number of sales
When aquarius returns `-1` it means it couldn't fetch the number of orders.
In this case we should display `n/a`

image

EnzoVezzaro:
- Fixes #1684 .

Changes proposed in this PR:

- added N/A for negative values as well as missing data

### DIFF
--- a/src/components/@shared/Price/Conversion.tsx
+++ b/src/components/@shared/Price/Conversion.tsx
@@ -28,13 +28,7 @@ export default function Conversion({
   const priceTokenId = getCoingeckoTokenId(symbol)
 
   useEffect(() => {
-    if (
-      !prices ||
-      !price ||
-      price === 0 ||
-      !priceTokenId ||
-      !prices[priceTokenId]
-    ) {
+    if (!prices || !price || !priceTokenId || !prices[priceTokenId]) {
       return
     }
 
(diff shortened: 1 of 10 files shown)


mihaisc:
add the fix on the teaser as well and check if we display it in other places , like profile page

image

bogdanfazakas:
> add the fix on the teaser as well and check if we display it in other places , like profile page

Yes we have the issue in the profile too
Screenshot 2022-09-14 at 15 30 59

EnzoVezzaro:
made changes when there's sales numbers. Hopefully I covered all components with these numbers.

src/components/@shared/Price/Conversion.tsx
```diff
@@ -27,19 +27,21 @@ export default function Conversion({
   // referring to Coingecko tokenId in Prices context provider
   const priceTokenId = getCoingeckoTokenId(symbol)
 
+  console.log(price)
```
bogdanfazakas, line 30: missed this 😄

bogdanfazakas (approved):
LGTM, i remembered that we have decided to have prices as strings and send them between various services like that to avoid further issues, but as far i can see this update to Number affects just UI elements so we should be fine

src/components/@shared/Price/Conversion.tsx
```diff
@@ -31,15 +31,15 @@ export default function Conversion({
     if (
       !prices ||
       !price ||
-      price === '0' ||
+      price === 0 ||
```
catalin8v, line 34: Screenshot 2022-09-17 at 10 49 47

this line is not necessary anymore :)

EnzoVezzaro: good catch, thanks 😊

PR #1696, opened 2022-09-19
Enforce docker container checksum in publish form

### LINKED ISSUE #1672: Enforce docker digest on publish
at publish, docker digest is required. Dapps may try to use a proxy to fetch the digest from dockerhub, but if the image is served by another container service, it will fail. Then user must fill in that value

(see https://github.com/oceanprotocol/pm/issues/175 for Epic details)

bogdanfazakas:
Fixes #1672  .

Changes proposed in this PR:

- get container checksum from docker hub api using the existing proxy
- allow users add manually the container checksum for images not hosted on docker hub and for private images where we can't fetch the value
- fetch the checksum for our default container images


kremalicious:
fetches everything correctly in some quick tests. Some findings:

1. When switching in UI between node:latest & python:latest, 4 of the same requests with the same response are happening. So let's double check if we can reduce those requests.

2. Tried with the example in the placeholder and got:
Screen Shot 2022-09-22 at 18 45 14

As `latest` does not exist for this, let's use `oceanprotocol/algo_dockers:node-vibrant` as placeholder example.

Because then all is good:

Screen Shot 2022-09-22 at 18 48 54

UI for success path is good enough considering this is advanced feature, one small addition could be that in above success case, the filled _Docker Image Checksum_ becomes `readOnly`.

3. The always fun error states. This is really confusing for an image which does not exist, I typed in `oceanprotocol/hello`:

Screen Shot 2022-09-22 at 18 51 16

So in this case I would expect UI telling me this image does not exist. And no empty _Image:_ & _Tag:_ But not sure if we can reliably detect if there's a Docker image behind a URL? Easy for shortcut notation `oceanprotocol/hello` but not with a full URL I suppose.

So we would need some error state for cases where we can detect that it's not a Docker image

bogdanfazakas:
> fetches everything correctly in some quick tests. Some findings:
> 
> 1. When switching in UI between node:latest & python:latest, 4 of the same requests with the same response are happening. So let's double check if we can reduce those requests.
> 2. Tried with the example in the placeholder and got:
> 
> As `latest` does not exist for this, let's use `oceanprotocol/algo_dockers:node-vibrant` as placeholder example.
> 
> Because then all is good:
> 
> UI for success path is good enough considering this is advanced feature, one small addition could be that in above success case, the filled _Docker Image Checksum_ becomes `readOnly`.
> 
> 3. The always fun error states. This is really confusing for an image which does not exist, I typed in `oceanprotocol/hello`:
> 
> 
> So in this case I would expect UI telling me this image does not exist. And no empty _Image:_ & _Tag:_ But not sure if we can reliably detect if there's a Docker image behind a URL? Easy for shortcut notation `oceanprotocol/hello` but not with a full URL I suppose.
> 
> So we would need some error state for cases where we can detect that it's not a Docker image

Think i cover most or all of the things mentioned above 😃

mihaisc:
The only issue that i found is that we don't specify that the validation works only for docker hub. So based on the placeholder i write my custom url and then click 'Use' . I get an error where it tells me tot check my url
image

bogdanfazakas:
> The only issue that i found is that we don't specify that the validation works only for docker hub. So based on the placeholder i write my custom url and then click 'Use' . I get an error where it tells me tot check my url

this should have covered the following scenarios:

1. if docker image inserted is a url : we we do a head query on that url to see if it is a valid url (think this is the most we can do in this case), if that succeeds we inform the user to add manually the container checksum, otherwise we display the error you've got
2. if not an url, we hit docker-hub via our docker-hub-proxy, if image-tag not found we display that error, if found we add the checksum and disable that input field

mihaisc:
Ok, after further discussion there will be no checks/validation. For example : `docker pull quay.io/startx/mariadb`  in the field you put just `quay.io/startx/mariadb` a url doesn't work and you can't even check it. If it's custom we can't really validate, the user just need to input the right values for name and checksum.

Docker Image URL - > Custom Docker Image ( there is no url involved, never)

PR #1704, opened 2022-09-22
add Görli, remove Rinkeby & Ropsten

kremalicious:
- add Goerli to network selector
- switch subgraph typings generation to goerli subgraph
-  for https://github.com/oceanprotocol/pm/issues/178

resulting in new network defaults:

Screen Shot 2022-09-23 at 11 21 44


mihaisc:
Wait, don't kick out rinkeby and ropten.

mihaisc:
incorrect ocean value in wallet
image
image
console error: `RROR: Failed to get the balance: Returned values aren't valid, did it run Out of Gas? You might also see this error if you are not using the correct ABI for the contract you are retrieving data from, requesting data from a block number that does not exist, or querying a node which is not fully synced.` 
A first guess would be incorrect ocean token address

kremalicious:
> console error: 

Getting this error only when switching networks, no matter if Görli or not. But on other networks the balance then is correct despite this error message so this might be caused by some too early call before all values are present for it.

Contract address is correct, was able to add Görli OCEAN to MetaMask with that address

kremalicious:
Think I got it, the OCEAN contract addresses we get from contracts and from the approved token query are different.

Contracts say `0xCfDdA22C9837aE76E0faA845354f33C62E03653a` but approved token query says `0xd8992ed72c445c35cb4a2be468568ed1079357c8`.

mihaisc:
@alexcos20 can you please update the approved token so it's the correct ocean token address.

mihaisc:
Also blocked by https://github.com/oceanprotocol/aquarius/issues/893

EnzoVezzaro:
the market's `OCEAN` balance is not updating. I have some on my wallet, but the market is not reflecting that:

Screenshot 2022-09-26 at 07 00 25

kremalicious:
> the market's OCEAN balance is not updating. I have some on my wallet, but the market is not reflecting that:

Anything in the browser console?

EnzoVezzaro:
> Anything in the browser console?

Nope. Unfortunately, there's no error thrown.

Screenshot 2022-09-26 at 09 52 39

kremalicious:
> the market's OCEAN balance is not updating.

got it, and it's [still this issue](https://github.com/oceanprotocol/market/pull/1704#issuecomment-1256035590) so nothing we can do on market

alexcos20:
added 0xCfDdA22C9837aE76E0faA845354f33C62E03653a
removed 0xd8992Ed72C445c35Cb4A2be468568Ed1079357c8

it should be fine now

kremalicious:
getting balance on Görli seems really flaky, only successful every 2nd or 3rd reload for me

And the approved token query seems to still return `0xd8992Ed72C445c35Cb4A2be468568Ed1079357c8` for me

But sometimes balance actually works for whatever reason:

Screen Shot 2022-09-27 at 10 21 33

alexcos20:
that's definitely a bug on the subgraph, I can see TokenRemoved event on router contract:

https://goerli.etherscan.io/address/0x159924ca0f47d6f704b97e29099b89e518a17b5e#events

kremalicious:
> that's definitely a bug on the subgraph

cool, thanks for figuring out. For now added in a hacky filter in [4914314](https://github.com/oceanprotocol/market/pull/1704/commits/4914314288b09b2844bb1baa18212c49f41cb8ff) so at least it works now and we are not forced to rush that subgraph fix.

mihaisc:
An aquarius update was made and it seems to be working now. So, i guess we can merge this.

alexcos20:
bug was fixed on subgraph as well

mihaisc (approved):

Issue #1644, opened 2022-08-09
Integrate new error status codes from Provider


AnaLoznianu:
Blocked by: https://github.com/oceanprotocol/provider/issues/523

Display user-friendly error messages for C2D failed jobs

AnaLoznianu:
The provider issue is merged. We can start working on this one

AnaLoznianu:
Waiting for the documentation of the provider errors before merging this

bogdanfazakas:
fixed with #1940

PR #1715, opened 2022-09-30
C2D fixes

### LINKED ISSUE #1631: Compute fixes for approved tokens
over from https://github.com/oceanprotocol/market/pull/1516#issuecomment-1203562974:

1. there is a problem at display when we mix dataset and algorithm with diferent tokens (dataset is TUSDT but algo is OCEAN in the case below)

Screenshot 2022-08-03 at 09 50 20

2. h2o algorithms do not show up in edit compute settings, the query returns them but we filter them out at a point

bogdanfazakas:
Fixes #1631  .

Changes proposed in this PR:

- correctly display tokens in tooltip
- display all approve tokens in c2d total price display
- fix edit compute settings for free assets

Pretty hard to test now on test networks, after the drops of  rinkeby and ropsten, since we have just one approve token (OCEAN) on the remaining test networks

We can't really do more on the error display or logs thingy since we need to make some changes to the ocean lib also. Will do this in #1644 and in a separate PR on oceanJS


EnzoVezzaro (approved):
👍🏽

PR #1716, opened 2022-09-30
Fix/issue 1533 fix loading results compute

### LINKED ISSUE #1533: Listing of Compute jobs on asset details and profile constantly gets into loading mode
And whenever its loading state kicks in, the compute job details modal closes automatically.

1. Go to an asset you have started a compute job on
2. Open up _Your Compute Jobs_
3. Click _Show Details_
4. Wait like 10 sec.
5. See the modal close cause list refreshes from very beginning with empty compute jobs incl. spinner

Screen Shot 2022-06-22 at 14 36 23

Sounds like bad data refresh management. A loading spinner should only appear at very beginning when no data is available. Every further refresh of data needs to update existing data on screen, instead of replacing them all with empty list. But in end most likely component shouldn't refresh at all, as nothing changed.

EnzoVezzaro:
- Fixes #1533 .

Changes proposed in this PR:

- loading spinner gets activated only in certain cases (initial load, open/close accordion)
- fix modal closing automatically when timer kicks in

### DIFF
--- a/src/components/Asset/AssetActions/Compute/History.tsx
+++ b/src/components/Asset/AssetActions/Compute/History.tsx
@@ -5,14 +5,17 @@ import Caret from '@images/caret.svg'
 
 export default function ComputeHistory({
   title,
-  children
+  children,
+  refetchJobs
 }: {
   title: string
   children: ReactNode
+  refetchJobs?: any
 }): ReactElement {
   const [open, setOpen] = useState(false)
 
-  function handleClick() {
+  async function handleClick() {
+    await refetchJobs(true)
     setOpen(!open)
   }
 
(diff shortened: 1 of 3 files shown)


mihaisc:
jobs_refresh
This is what i was supposed to test right? If so seems to be working ok now. Doesn't start loading randomly.

mihaisc (dismissed):

EnzoVezzaro:
@mihaisc what did you find?

bogdanfazakas:
> @mihaisc what did you find?

for eg, the compute jobs section seems to be broken, we have compute jobs displayed if we go to an asset for eg, but the list is empty on the profile->compute jobs

EnzoVezzaro:
> > @mihaisc what did you find?
> 
> the profile compute jobs section seems to be broken, we have compute jobs displayed if we go to an asset for eg, but the list is empty on the profile->compute jobs

👍🏽 great catch, thanks!

PR #1722, opened 2022-10-03
Add top allocations

mihaisc:
Adds top 6 assets based on ve allocation. 
Added allocation in asset teaser and in asset details.
Reduced all home components to 6 assets.
Added a warning on asset teaser when no price available
image


kremalicious:
Something like this for asset teasers makes more sense, where we put everything which holds content which never grows into top line. Which gives nice space in bottom line to put allocation & sales numbers:

Screen Shot 2022-10-03 at 15 16 51

Anybody has a shorter synonym for "allocations"?

And then next step would be to get allocations for each asset, which ideally we would get delivered with the DDO from Aquarius like the sales numbers. Do we have this prepared or planned? Otherwise this could be an expensive query

mihaisc:
The first one looks great. for now display this in the top component. will create issue for aqua

mihaisc:
good to go

EnzoVezzaro (approved):
👍🏽

app.config.js
```diff
@@ -11,10 +14,12 @@ module.exports = {
 
   // List of chainIds which metadata cache queries will return by default.
   // This preselects the Chains user preferences.
-  chainIds: [1, 137, 56, 246, 1285],
+  chainIdsProduction: chainIdsProduction,
```
kremalicious, line 17: do we really need all this split-up introduced here? On first look, seems this is only used in `getVeChainIds()` which has more hardcoded chainIds. So would suggest keeping app config as before, and rather use new setting dedicated for ve stuff, like `veChainId` or so.

src/@utils/subgraph.ts
```diff
@@ -71,6 +71,8 @@ const OpcsApprovedTokensQuery = gql`
 
 export function getSubgraphUri(chainId: number): string {
   const config = getOceanConfig(chainId)
+  if (chainId === 1)
+    config.subgraphUri = `https://v4.subgraph2.mainnet.oceanprotocol.com`
```
kremalicious, line 75: hardcoded different subgraph URI for all mainnet requests?

src/components/Asset/AssetActions/AssetStats/index.tsx
```diff
@@ -5,17 +5,25 @@ import styles from './index.module.css'
 export default function AssetStats() {
   const { asset } = useAsset()
 
+  // TODO: get this number for each asset
```
kremalicious, line 8: another hardcoded value we need to fetch, to keep stats on teaser and asset details consistent

src/@utils/subgraph.ts
mihaisc: good catch, missed this one ;D

src/components/Asset/AssetActions/AssetStats/index.tsx
mihaisc: didn't see you added it here as well, will implement it here

src/components/@shared/AssetTeaser/index.tsx
```diff
@@ -23,37 +23,52 @@ export default function AssetTeaser({
...
 
-          <footer className={styles.foot}>
-            <Price accessDetails={asset.accessDetails} size="small" />
-            <NetworkName networkId={asset.chainId} className={styles.network} />
+          <Price accessDetails={asset.accessDetails} size="small" />
+
+          <footer className={styles.footer}>
+            {allocated && allocated > 0 ? (
+              <span className={styles.typeLabel}>
+                {allocated < 0 ? '' : `${allocated} veOCEAN`}
```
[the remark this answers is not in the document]
mihaisc: hmm, true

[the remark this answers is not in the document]
mihaisc: ok, added our formatting component, not sure it's working as expected.

mihaisc:
Added totals in the footer as well

mihaisc:
Time out until we have https://github.com/oceanprotocol/aquarius/issues/908. Most of this pr will be refactored.

jamiehewitt15:
> Anybody has a shorter synonym for "allocations"?

We could say "6,000 veOcean votes"

EnzoVezzaro:
> oceanprotocol/aquarius#908

For me it's a bit weird to see some numbers without thousands separators. I think we should have all numbers formatted the same to avoid confusion. 

> Added totals in the footer as well image

I know these represent different values (units doesn't have ',' while currency has), but I think we should have the same format for all numbers displayed.

kremalicious:
Slight differences in all numbers between this list and df-web:

Screen Shot 2022-10-07 at 10 56 43

Screen Shot 2022-10-07 at 10 56 37

Regardless, didn't we say that we need to have only percentage representations of allocations as those absolute veOCEAN numbers are not correct?

mihaisc:
The slight different is because of the sync difference. No, you can't have percetages. The total values are calculated every 5 minutes in a backed. Then we sync that with aquarius.

kremalicious (approved):

PR #1728, opened 2022-10-05
Change datatoken template and ensure compatibility with both

### LINKED ISSUE #1652: Support both datatoken templates when ordering an asset
We need to support the other template as well for assets created outside the market. 
In order, instead of calling  `buyFromFreAndOrder` or `buyFromDispenserAndOrder` , we need to buy 1 dt from the fre/dispenser and the do the order. We can get the templateId from the subgraph.

### LINKED ISSUE #1725: Asset published in Market fails to get dispensed
### The error / how to reproduce

What I did:

1. Published CEXA price feed via Ocean Market, onto Polygon. (Published [here](https://market.oceanprotocol.com/asset/did:op:0dac5eb4965fb2b485181671adbf3a23b0133abf71d2775eda8043e8efc92d19)).

2. Then, tried to consume it via Ocean Market, with a different address. It failed. [This comment has details.](https://github.com/oceanprotocol/market/issues/1725#issuecomment-1268215847).

I also tried to consume it via [this ocean.py README](https://github.com/oceanprotocol/ocean.py/blob/main/READMEs/predict-eth.md), with a different address than the publishing one. It failed too. Digging into it...

It appears that by default, the dispenser contract, [`allowedSwapper`](https://github.com/oceanprotocol/contracts/blob/722f6cb91ec2a3c5966b919888b2633601e6ecab/contracts/pools/dispenser/Dispenser.sol#L25) gets set to a nonzero address (not 0x00...). 

Which meant that *only* the publishing account can dispense tokens, unless it's changed via [`setAllowedSwapper()`](https://github.com/oceanprotocol/contracts/blob/722f6cb91ec2a3c5966b919888b2633601e6ecab/contracts/pools/dispenser/Dispenser.sol#L205).

I looked around in Ocean Market to try to change it, such that I could allow anyone to request tokens. Eg next to "change ddo". Such an option did not exist.

But I did find a workaround, in code: via Python, I updated the allowedSwapper to zero address (0x00..) manually. In that setting, it means that anyone can request tokens ([ref](https://github.com/oceanprotocol/contracts/blob/722f6cb91ec2a3c5966b919888b2633601e6ecab/contracts/pools/dispenser/Dispenser.sol#L238)).

### The error, seen again

A user in a data challenge ran into the same issue. See [ocean.py#1161](https://github.com/oceanprotocol/ocean.py/issues/1161)

### Summary

We published a free asset in Ocean Market

A user tried to call dispense(), it didn't work. (Implicitly via a call to ocean_assets::download_file()). It would also fail in any dispense() call from ocean.js, etherscan, or elsewhere.

"Free" assets should be dispense-able. 

The root problem is that: template 2's allowed_swapper != ZERO, which means that no one can call dispense(). 

The challenge is: using template 1 solves the problem, but incurs a second tx and more gas fees (either on publish, or on consume). How do we crack this?

### Candidate approaches, recommendation

**[This comment](https://github.com/oceanprotocol/market/issues/1725#issuecomment-1341398180)** lists several candidate approaches, analyzes them, and makes recommendation on how to proceed. (Hint: it's not "use template 1", we can do better!).

mihaisc:
Changed default datatoken template to 1
Consume is compatible with both templates.

- fixes #1725 
- fixes #1652


trentmc:
I just tested this for the problem reported in #1725, where the python README couldn't dispense. It works. 

Details:
1. I Published CEXA price feed via Ocean Market, onto Polygon. (Published [here](https://market.oceanprotocol.com/asset/did:op:0dac5eb4965fb2b485181671adbf3a23b0133abf71d2775eda8043e8efc92d19)).

Then, tried to consume it via Ocean Market, with a different address. It failed. https://github.com/oceanprotocol/market/issues/1725#issuecomment-1268215847.

I also tried to consume it via [this ocean.py README](https://github.com/oceanprotocol/ocean.py/blob/main/READMEs/predict-eth.md), with a different address than the publishing one. It failed too.

kremalicious:
So `predict-eth.md` in ocean.py does not use dispense method anymore, so the only known case of this being used is now gone. Why and for whom exactly are we doing this PR here then?

trentmc:
> So predict-eth.md in ocean.py does not use dispense method anymore

Not true. Look closely. It does. Via a new top-level call to simplify the interface

> so the only known case of this being used is now gone

Not true. We can expect it to be a common flow: publish url via webapp, then consume via python (ocean.py).

alexcos20:
Let's keep template 1 as default.
Can we have a config variable as well? In case someone needs to change this

mihaisc:
added `defaultDatatokenTemplateIndex` in `app.config.js`

bogdanfazakas (approved):
lgtm

EnzoVezzaro (approved):

src/components/Publish/_utils.ts
```diff
@@ -203,7 +204,7 @@ export async function createTokensAndPricing(
 
   // TODO: cap is hardcoded for now to 1000, this needs to be discussed at some point
   const ercParams: DatatokenCreateParams = {
-    templateIndex: 2,
+    templateIndex: defaultDatatokenTemplateIndex,
```
kremalicious, line 207: this is all a bit confusing so let's double check. If I get the thread in here right, we want to continue using the so called "datatoken enterprise template".

We usually refer to this with template index `2`, like we did here before. But this change now sets this to `1` based on the config value in `app.config.js`.

This change here will make new assets published use the default template, `1`. Is this all intended?

mihaisc: @alexcos20 ?

kremalicious: bump @alexcos20

kremalicious:
we need to get to a decision in here, supporting both templates during order flow is great and we should have it. But switching the default template still has [outstanding decision](https://github.com/oceanprotocol/market/pull/1728#discussion_r1004243454) as we have conflicting direction instructions.

The reluctance to continue speedily on this PR also tells me that the original intention is not important at all anymore?

src/components/Publish/_utils.ts
alexcos20: my mistake for spelling error.
We should publish using template 2 (Enterprise Template), because it has a better UX  (buy / order in a single TX)  compared to template 1 (where you need to buy DT first (one tx) + order (2nd tx)

But we need to support both templates when consuming.

app.config.js
```diff
@@ -18,6 +18,7 @@ module.exports = {
 
   infuraProjectId: process.env.NEXT_PUBLIC_INFURA_PROJECT_ID || 'xxx',
 
+  defaultDatatokenTemplateIndex: 1,
```
alexcos20, line 21:
```suggestion
  defaultDatatokenTemplateIndex: 2,
```

trentmc:
> [alex] Let's keep template 1 as default.

Note: this PR, despite being merged, kept template 2 as a default. (2 = Enterprise template)

Therefore issue [market#1725](https://github.com/oceanprotocol/market/issues/1725) was not completed. I'm re-opening it.

PR #1732, opened 2022-10-07
Pass UrlFile object to Provider.checkFileUrl instead of string

MantisClone:
Changes proposed in this PR:

- Pass `UrlFile` object to `Provider.checkFileUrl` instead of a string

Depends on https://github.com/oceanprotocol/ocean.js/pull/1627

Reasoning:

This is an incremental step towards supporting other file storage types like Arweave, Filecoin, etc...

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "@coingecko/cryptoformat": "^0.5.4",
     "@loadable/component": "^5.15.2",
     "@oceanprotocol/art": "^3.2.0",
-    "@oceanprotocol/lib": "^2.2.3",
+    "@oceanprotocol/lib": "^2.3.0",
     "@oceanprotocol/typographies": "^0.1.0",
     "@oceanprotocol/use-dark-mode": "^2.4.3",
     "@tippyjs/react": "^4.2.6",
--- a/src/@utils/provider.ts
+++ b/src/@utils/provider.ts
@@ -83,7 +83,7 @@ export async function getFileDidInfo(
   }
 }
 
-export async function getFileUrlInfo(
+export async function getFileInfo(
   file: UrlFile,
   providerUrl: string
 ): Promise<FileInfo[]> {
--- a/src/@utils/provider.ts
+++ b/src/@utils/provider.ts
@@ -1,5 +1,4 @@
 import {
-  Arweave,
   ComputeAlgorithm,
   ComputeAsset,
   ComputeEnvironment,
@@ -85,7 +84,7 @@ export async function getFileDidInfo(
 }
 
 export async function getFileUrlInfo(
-  file: UrlFile | Arweave,
+  file: UrlFile,
   providerUrl: string
 ): Promise<FileInfo[]> {
   try {
--- a/src/@utils/provider.ts
+++ b/src/@utils/provider.ts
@@ -1,4 +1,5 @@
 import {
+  Arweave,
   ComputeAlgorithm,
   ComputeAsset,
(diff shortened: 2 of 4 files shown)


MantisClone:
Hello, @kremalicious Thanks for taking a look at this and approving the CI checks to run.  I believe the CI checks are failing because this PR depends on https://github.com/oceanprotocol/ocean.js/pull/1627 which has not yet been merged/released.

MantisClone:
Hello @alexcos20 @kremalicious @mihaisc @bogdanfazakas

This PR depends on https://github.com/oceanprotocol/ocean.js/pull/1627, but other than that, I think it is ready for review.

src/@utils/provider.ts
```diff
@@ -83,11 +84,11 @@ export async function getFileDidInfo(
 }
 
 export async function getFileUrlInfo(
```
kremalicious, line 86: Same as over on ocean.js, if it does more than checking URLs, method needs a new name. Here we can simply remove `URL` and end up with `getFileInfo()`

src/components/@shared/FormFields/FilesInput/index.tsx
```diff
@@ -23,7 +23,12 @@ export default function FilesInput(props: InputProps): ReactElement {
         ? values?.services[0].providerUrl.url
         : asset.services[0].serviceEndpoint
       setIsLoading(true)
-      const checkedFile = await getFileUrlInfo(url, providerUrl)
+      const urlFile: UrlFile = {
+        type: 'url',
+        url: url,
```
kremalicious, line 28: shorthand notation please, which then most likely allows us to have this short line in one line, like:

```js
const urlFile: UrlFile = { type: 'url', url, method: 'GET' }
```

kremalicious:
Misses one more usage of that file check:
https://github.com/oceanprotocol/market/blob/83d83447b6f241740d3fdd1c3b78f99a01b3f5c9/src/components/Asset/AssetActions/index.tsx#L44-L82

kremalicious:
ocean.js v2.3.0 has the changes required for this PR

src/@utils/provider.ts
MantisClone: Done.

MantisClone:
> Misses one more usage of that file check, be sure to handle this for publish preview & published asset which is what the condition for `fileInfoResponse` in here is for:
> 
> https://github.com/oceanprotocol/market/blob/83d83447b6f241740d3fdd1c3b78f99a01b3f5c9/src/components/Asset/AssetActions/index.tsx#L44-L82

I looked at this briefly, but I don't know how to fix it.

package.json
```diff
@@ -26,7 +26,7 @@
     "@coingecko/cryptoformat": "^0.5.4",
     "@loadable/component": "^5.15.2",
     "@oceanprotocol/art": "^3.2.0",
-    "@oceanprotocol/lib": "^2.2.3",
+    "@oceanprotocol/lib": "^2.3.0",
```
kremalicious, line 29: need to run `npm install` so `package-lock.json` updates, then commit that file too for actual upgrade

kremalicious:
This needs merge or rebase against `main` and then continue. No need to update ocean.js in here then.

> I looked at this briefly, but I don't know how to fix it.

ok, what did you try to solve this? What exactly is unclear?

kremalicious:
closing as abandoned, and #1765 is picking things up from here.

PR #1737, opened 2022-10-12
Tests for @shared components

### LINKED ISSUE #1400: Basic Components: Shared

kremalicious:
This picks up from #1480, addressing #1400, but with different tactics as already prepared in earlier PRs:

- unit tests are the default now, where every component has to have one
- use a `testRender` method which is the bare minimum for each component
- storybook stories are optional and only used for UI development, but not for automatic test runs
- all the new stories in here were copied over from #1480 (thanks @EnzoVezzaro!), but source all their dummy data & dummy props from the respective component test, or from `__fixtures__`
- mock most important hooks usage with dummy data for all tests
  - these hooks are working across all tests now: `useWeb3()`, `useUserPreferences()`, `useMarketMetadata()`, defined in `.jest/jest.setup.js` with dummy data defined in `.jest/__fixtures__/`
  - downside: redefining their return data in individual tests does not really work so careful which hooks to mock globally. Alternatively, things can be always mocked at beginning of individual tests.
- provide pattern for mocking our `@utils` methods
  - Jest-native mocking, e.g. using `__mocks__/@utils/accessDetailsAndPricing.ts` makes Jest pick this up automatically.
  - Also works for mocking all dependencies, e.g. to create mocks for anything in ocean.js, create file structure `__mocks__/@oceanprotocol/lib/index.ts`
- better separation between mocks & fixtures
- NO extensive refactoring of all our components
- closes #1400
- closes #1480

_Developing..._


kremalicious:
We should get this through already as it sets the base for writing tests and provides many patterns which can be followed already. Main reason to merge already is that I end up writing tests for everybody’s changes we ship onto `main` in last days/weeks to keep my PR here green. So we all need to get into the habit of writing and/or modifying tests if we add new stuff

mihaisc (approved):

PR #1743, opened 2022-10-14
Various veOCEAN features

mihaisc:
Closes #nothing
To test this go to https://test-df.oceandao.org/data (on goerli) lock ocean and then allocate. All the feature respect the network selection.

- Update top message to `[Lock your $OCEAN](https://df.oceandao.org/) to get veOCEAN, earn rewards and curate data.` 
- Added `Assets with allocations` in home page (if no allocations exist then that section is not displayed ) . This could have been in the profile page, but I intentionally added it on the homepage to increase the importance of df.
image

- Added locked ocean in profile
image

- View own allocation on an asset (if available)
image

- Added new sort options : `Sales`, `Price`, `Total Allocation`
image

ToDo (@kremalicious  assistance required 😀 ) : 
- when you have no locked ocean i intentionally added a tooltip to "encourage" the user to go lock 😁 , but it looks .....
image
- For locked ocean should we display also the actual ocean amount not just the conversion?
- the sort tab looks kinda off now
image

### DIFF
--- a/content/site.json
+++ b/content/site.json
@@ -14,7 +14,7 @@
       "link": "/profile"
     }
   ],
-  "announcement": "Explore [OceanONDA V4](https://blog.oceanprotocol.com/how-to-publish-a-data-nft-f58ad2a622a9).",
+  "announcement": "[Lock your $OCEAN](https://df.oceandao.org/) to get veOCEAN, earn rewards and curate data.",
   "warning": {
     "ctd": "Compute-to-Data is still in a testing phase, please use it only on test networks."
   }
--- a/src/@types/aquarius/SearchQuery.ts
+++ b/src/@types/aquarius/SearchQuery.ts
@@ -7,7 +7,8 @@ export enum SortTermOptions {
   Created = 'nft.created',
   Relevance = '_score',
   Orders = 'stats.orders',
-  Allocated = 'stats.allocated'
+  Allocated = 'stats.allocated',
+  Price = 'stats.price.value'
 }
 
 // Note: could not figure out how to get `enum` to be ambiant
--- a/src/@utils/aquarius.ts
+++ b/src/@utils/aquarius.ts
@@ -215,6 +215,28 @@ export async function getAssetsFromDtList(
   }
 }
 
+export async function getAssetsFromNftList(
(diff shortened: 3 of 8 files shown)


kremalicious:
Started some refactor on front-page but not sure we want to go there. We might want to just leave out that front-page allocated asset section.

Problem is that if we show assets based on a metric, we also have to show that metric and order the list by it. This is not happening right now. On top of that, all those asset teasers on front-page are quite distracting if we use them for each and every section. Then again, asset teasers include metric numbers for all other sections, but not for the own allocation.

To combat both, started rendering this in a list:

Screen Shot 2022-10-14 at 12 20 17

But then still the issue remains to get allocations for each asset rendered. Will push what I have. We can still decide to kick it out, or figure out how to add own allocations for each asset on there

mihaisc:
Ok, leave that home component as it is, I'll add the allocation, btw they are percentages

src/components/Home/Allocations/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+        const allocations = await getOwnAllocations(chainIds, accountId)
+        setHasAllocations(allocations && allocations.length > 0)
+
+        const baseParams = {
+          chainIds,
+          filters: [
+            getFilterTerm(
+              'nftAddress',
+              allocations.map((x) => x.nftAddress)
```
mihaisc, line 55: we already have the allocation % here, i just didn't propagate it because i didn't have any idea how to display it.

kremalicious:
Regarding profile stats, all debatable which currency to show. Tinkered around with hasAllocation/has no allocations states and switched to showing the OCEAN value on there to stay consistent with what df-web shows.

Something is locked:

Screen Shot 2022-10-14 at 13 06 57

Nothing is locked:

Screen Shot 2022-10-14 at 13 17 08

Another variant is this one when something is locked, combines both but still think a bit unclear what that is exactly:

Screen Shot 2022-10-14 at 13 17 30

mihaisc:
Percentages are now displayed in the table 
image

mihaisc:
Should be ready for testing.

mihaisc:
Fixed

PR #1746, opened 2022-10-18
Disable ordering when state is 4

mihaisc:
When the state of an nft is 4 ordering is imposible. Thest asset : did:op:6098e339b178d4d9273cd77e0f5d22394fef09fba3498f71abff5cabec306e37
image

### DIFF
--- a/src/components/Asset/AssetActions/Download.tsx
+++ b/src/components/Asset/AssetActions/Download.tsx
@@ -45,11 +45,15 @@ export default function Download({
   const [isPriceLoading, setIsPriceLoading] = useState(false)
   const [isOwned, setIsOwned] = useState(false)
   const [validOrderTx, setValidOrderTx] = useState('')
+  const [isTemporaryDisabled, setIsTemporaryDisabled] = useState(false)
   const [orderPriceAndFees, setOrderPriceAndFees] =
     useState<OrderPriceAndFees>()
 
   const isUnsupportedPricing = asset?.accessDetails?.type === 'NOT_SUPPORTED'
 
+  useEffect(() => {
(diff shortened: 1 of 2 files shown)


src/components/Asset/AssetActions/Download.tsx
```diff
@@ -183,26 +187,36 @@ export default function Download({
   const AssetAction = ({ asset }: { asset: AssetExtended }) => {
     return (
       <div>
-        {isUnsupportedPricing ? (
+        {isTemporaryDisabled ? (
           <Alert
             className={styles.fieldWarning}
             state="info"
-            text={`No pricing schema available for this asset.`}
+            text={`The publisher temporary disabled ordering for this asset`}
```
kremalicious, line 194: Typo, same for variable name:

```suggestion
            text={`The publisher temporarily disabled ordering for this asset`}
```

Maybe call the variable 'isOrderDisabled' or something?

mihaisc: the suggestion is very confusing for me. Sure i can change the variable to `isOrderDisabled`. Also the text is ok
image

kremalicious: yes, but in this context you are looking for an adverb, not an adjective as you are not modifying a noun. So with adjective it would only work like so, but it sounds weird: "Temporary disability of ordering by the publisher."

As "disabled" is not a noun we have to use an adverb.

mihaisc: touché

kremalicious (approved):
grammatically correct PR now 😄

PR #1748, opened 2022-10-19
Showing related assets

### LINKED ISSUE #1710: Related assets
From https://github.com/oceanprotocol/market/issues/1635#issuecomment-1257800448

> Below each data asset add a section to showcase all data assets that are somehow related to the ones into the profile.
> Reason: similarly to how Google handles backlinks, or Amazon handles suggestions, it would be interesting to find a way to showcase datasets that are similar to the one searched for (an initial suggestion could come from filtering per tags and sales). As the market grows, this could help users find datasets relevant to their case.

Not sure exactly where , we should have a list of teasers with related assets. At first glance under the actions component might be nice. 
List contains: 
- first 2-3 assets that have tags in common, ordered by sales.
- assets from the same publisher 

And with the power of paint...
Screenshot 2022-09-27 161327

jamiehewitt15:
Fixes #1710 

Changes proposed in this PR:

- Showing related assets on asset detail page


EnzoVezzaro (approved):
Great job 👍🏽

jamiehewitt15:
> Great job 👍🏽

Thank you

mihaisc:
i'm looking at[ this asset ](https://market-git-issue-1710-related-assets-oceanprotocol.vercel.app/asset/did:op:81a2c9959bf2a03e7c9b21f95e24b457cb28c880e28e3670cb0a736372d0ad05) . The first related asset is [this](https://market-git-issue-1710-related-assets-oceanprotocol.vercel.app/asset/did:op:ba8753de3a2715d05960092f35bc2c74a0d63bb4998a4fcb583bae60e335cf89) they have no tags in common and are from separate owners

jamiehewitt15:
> i'm looking at[ this asset ](https://market-git-issue-1710-related-assets-oceanprotocol.vercel.app/asset/did:op:81a2c9959bf2a03e7c9b21f95e24b457cb28c880e28e3670cb0a736372d0ad05) . The first related asset is [this](https://market-git-issue-1710-related-assets-oceanprotocol.vercel.app/asset/did:op:ba8753de3a2715d05960092f35bc2c74a0d63bb4998a4fcb583bae60e335cf89) they have no tags in common and are from separate owners

The problem is caused by the hyphen that's in the tag, so elastic search returns `data` etc as a match for `data-farming`. I've tried escaping the hyphen with `replace(/-/g, '\\-')` (as mentioned in the [docs](replace(/-/g, '\\-')) but that's not making any difference. Let me know if you have any other ideas?

mihaisc:
The search seems to work ok https://market.oceanprotocol.com/search?tags=data-finance&sort=_score&sortOrder=desc

jamiehewitt15:
> The search seems to work ok https://market.oceanprotocol.com/search?tags=data-finance&sort=_score&sortOrder=desc

Have a look at the [search results](https://market.oceanprotocol.com/search?sort=_score&sortOrder=desc&text=ocean-market) for `ocean-market`. It returns lots of results that don't have the exact phrase or tag "ocean-market": [1](https://market.oceanprotocol.com/asset/did:op:8a3739a8b3eb494cdcaf463a1d78e9c3f6dfedb214bcee65e394a03d060fe70f) [2](https://market.oceanprotocol.com/asset/did:op:6637c63a7be53c4d7c6204b92e1508c928f9090ca822cec42782c8b1ec33bb2f) [3](https://market.oceanprotocol.com/asset/did:op:aee900df7379cda6a5aa1b87bd77e053906002058f649825df0bffe5d8cf17dc) [4](https://market.oceanprotocol.com/asset/did:op:cd086344c275bc7c560e91d472be069a24921e73a2c3798fb2b8caadf8d245d6) [5](https://market.oceanprotocol.com/asset/did:op:fc153fad971db5de39e307e239eedf05cf407e6218ae70fc43a96ddd8ca590fe)   [6](https://market.oceanprotocol.com/asset/did:op:0d3c8845df538dff98a08e0cc9d572cf4703d9324eb5887e91e9c8d97474e8ae)

The reason is that elastic search treats `ocean-market` as two separate words and looks for a match for each one. For the main search that's fine as it's not a bad thing to show a wider selection of results. We need to do something differently though if we want exact matches for tags that have hyphens.

You can see this more dramatically if you search for `ocean-market-data-finance-test-123` as one term, there isn't any exact matches but you'll get lots of results.

mihaisc:
Yes, that is intentional, it's more "fuzzy". But not related to my suggestion

jamiehewitt15:
> Yes, that is intentional, it's more "fuzzy". But not related to my suggestion

What do you suggest we do to get exact matches when there is a hyphen in the tag? I already tried escaping the hyphen with `replace(/-/g, '\\-')` (as mentioned in the [docs](https://www.elastic.co/guide/en/elasticsearch/reference/8.4/query-dsl-query-string-query.html#_reserved_characters)) but that didn't work. There is a suggestion on the [ES discussion board](https://discuss.elastic.co/t/es-not-search-with-hyphen/300769) that it will work if we change the data type to keyword (currently tags are text) but that would require a change in Aquarius.

mihaisc:
Query for [this asset](https://market-git-issue-1710-related-assets-oceanprotocol.vercel.app/asset/did:op:81a2c9959bf2a03e7c9b21f95e24b457cb28c880e28e3670cb0a736372d0ad05):
1. Get x (we will start with 3, we'll change depending on the ui):

```
{
    "from": 0,
    "size": 3,
    "query": {
        "bool": {
            "must_not": {
                "term": {
                    "nftAddress.keyword": "0x08d2eF7c255834dBB7513e8b3E744D05748a7079"
                }
            },
            "filter": [
                {
                    "terms": {
                        "chainId": [
                            1,
                            137,
                            56,
                            246,
                            1285
                        ]
                    }
                },
                {
                    "terms": {
                        "metadata.tags.keyword": [
                            "data-farming",
                            "ocean-market"
                        ]
                    }
                },
                {
                    "term": {
                        "_index": "aquarius"
                    }
                },
                {
                    "term": {
                        "purgatory.state": false
                    }
                }
            ]
        }
    },
    "sort": {
        "stats.orders": "desc"
    }
}
```

2. Get y asset from the same owner

```
{
    "from": 0,
    "size": 3,
    "query": {
        "bool": {
            "filter": [
                {
                    "terms": {
                        "chainId": [
                            1,
                            137,
                            56,
                            246,
                            1285
                        ]
                    }
                },
                {
                    "term": {
                        "nft.owner.keyword": "0xEB18BaD7365a40E36A41fB8734eB0B855d13b74F"
                    }
                },
                {
                    "term": {
                        "_index": "aquarius"
                    }
                },
                {
                    "term": {
                        "purgatory.state": false
                    }
                }
            ]
        }
    },
    "sort": {
        "stats.orders": "desc"
    }
}
```
So, let's say that we decide we display only 3 related assets, then x+y =3. If we do the first query and we get 3 assets, then we don't do the second query. If we get only 2 assets from the first one, then we do the second one and we fetch only 1 asset from there.

src/components/Asset/RelatedAssets/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+      },
+      sortOptions: {
+        sortBy: SortTermOptions.Orders
+      } as SortOptions
+    } as BaseQueryParams
+    setQueryRelatedAssets(generateBaseQuery(baseParamsSales))
+  }, [chainIds, dtAddress, modifiedSearchTerm])
+
+  return (
+    <SectionQueryResult title="Related Assets:" query={queryRelatedAssets} />
```
kremalicious, line 53: We never do `:` in any headings, so remove `:`

src/components/Asset/AssetContent/index.tsx
```diff
@@ -78,6 +79,11 @@ export default function AssetContent({
               </Button>
             </div>
           )}
+          <RelatedAssets
+            tags={asset?.metadata?.tags}
+            dtAddress={asset?.datatokens[0]?.address}
```
kremalicious, line 84: if you actually intended to make sure this does not break even if `asset.datatokens` does not exist, then it needs to be `asset?.datatokens?.[0]?.address`

jamiehewitt15:
> `{
    "from": 0,
    "size": 3,
    "query": {
        "bool": {
            "must_not": {
                "term": {
                    "nftAddress.keyword": "0x08d2eF7c255834dBB7513e8b3E744D05748a7079"
                }
            },
            "filter": [
                {
                    "terms": {
                        "chainId": [
                            1,
                            137,
                            56,
                            246,
                            1285
                        ]
                    }
                },
                {
                    "terms": {
                        "metadata.tags.keyword": [
                            "data-farming",
                            "ocean-market"
                        ]
                    }
                },
                {
                    "term": {
                        "_index": "aquarius"
                    }
                },
                {
                    "term": {
                        "purgatory.state": false
                    }
                }
            ]
        }
    },
    "sort": {
        "stats.orders": "desc"
    }
}`

I tested this query out on the same asset and the results aren't any better: 


Screenshot from 2022-10-25 12-33-39

mihaisc:
I just ran that query in postman and i get different results. I get the following dids ( i pot just the did because the full response would be too long) :did:op:a3aa14de333ee1ecb3b8a842954033c1c0004f9e77020cacb861843478c1079c , did:op:dc1d8c161b641011614e4de03f5023bbba55fefcc91cc88d8074656ca91bf483, 
did:op:1d84394247d2a0ef463048b6a5aa9b580d9402bfd846550853f00079cd2fdc6c

All of them have relevant tags.

```
curl --location --request POST 'https://v4.aquarius.oceanprotocol.com/api/aquarius/assets/query' \
--header 'Content-Type: application/json' \
--data-raw '{ "from": 0, "size": 3, "query": { "bool": { "must_not": { "term": { "nftAddress.keyword": "0x08d2eF7c255834dBB7513e8b3E744D05748a7079" } }, "filter": [ { "terms": { "chainId": [ 1, 137, 56, 246, 1285 ] } }, { "terms": { "metadata.tags.keyword": [ "data-farming", "ocean-market" ] } }, { "term": { "_index": "aquarius" } }, { "term": { "purgatory.state": false } } ] } }, "sort": { "stats.orders": "desc" } }'
```

kremalicious:
Regarding UI: not a fan of putting them as big asset teasers on the side which puts an unnecessary priority on those assets. And it quickly takes up too much space when main asset's description is not that long. Then even 3 related assets push right side of screen down too much, as seen on Jamie's last screenshot.

My suggestion would be either:

1. put the related assets section below the main grid layout
2. use a minimal table to only list related asset names in the existing place

mihaisc:
> use a minimal table to only list related asset names in the existing place

Why not use the teasers? This is the pattern on most markets. Having just a minimal table might look weird

jamiehewitt15:
We could potentially do a reduced version of the asset teaser to save space? 

image

jamiehewitt15:
Another simple option for the UI would be a list of links on the side. The query has been updated including the feedback from @mihaisc so I think it's just the UI to sort out with this

image

EnzoVezzaro:
> Yup, the assets seem ok. Regarding the UI personally i would like something with teasers. The small ones look nice.

I do agree. 👍🏽

jamiehewitt15:
I've updated it with a minimal asset teaser

EnzoVezzaro (approved):

jamiehewitt15:
@mihaisc @kremalicious do you you have anymore feedback?

src/components/Asset/RelatedAssets/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+      chainIds,
+      esPaginationOptions: {
+        size
+      },
+      nestedQuery: {
+        must_not: {
+          term: { 'nftAddress.keyword': nftAddress }
+        }
+      },
+      filters: [
```
mihaisc, line 39: why are you adding all these filters? you are duplicating them (index, purgatory, chainid)

mihaisc:
```
{
    "from": 0,
    "size": 3,
    "query": {
        "bool": {
            "must_not": {
                "term": {
                    "nftAddress.keyword": "0x0E122f7d453e983eF0233AF4E2AF4Cb053e5F075"
                }
            },
            "filter": [
                {
                    "terms": {
                        "chainId": [
                            1,
                            137,
                            56,
                            246,
                            1285
                        ]
                    }
                },
                {
                    "terms": {
                        "metadata.tags.keyword": [
                            "crypto",
                            "cryptocurrency",
                            "web3",
                            "data",
                            "sentiment",
                            "frequency-analysis"
                        ]
                    }
                },
                false,
                {
                    "term": {
                        "_index": "aquarius"
                    }
                },
                {
                    "term": {
                        "purgatory.state": false
                    }
                },
                {
                    "terms": {
                        "chainId": [
                            1,
                            137,
                            56,
                            246,
                            1285
                        ]
                    }
                },
                {
                    "term": {
                        "_index": "aquarius"
                    }
                },
                {
                    "term": {
                        "purgatory.state": false
                    }
                }
            ]
        }
    },
    "sort": {
        "stats.orders": "desc"
    }
}
```

src/components/@shared/AssetTeaser/index.tsx
```diff
@@ -51,13 +53,15 @@ export default function AssetTeaser({
             <Dotdotdot tagName="h1" clamp={3} className={styles.title}>
               {name.slice(0, 200)}
             </Dotdotdot>
-            {!noPublisher && <Publisher account={owner} minimal />}
+            {!noPublisher && !noDescription && <Publisher account={owner} />}
```
kremalicious, line 56: Adding `noDescription` here is confusing and redundant. There is already a prop to hide the publisher, so use this by adding the `noPublisher` prop to the component

jamiehewitt15: Good point, I've removed it

src/components/@shared/AssetTeaser/index.tsx
```diff
@@ -51,13 +53,15 @@ export default function AssetTeaser({
             <Dotdotdot tagName="h1" clamp={3} className={styles.title}>
               {name.slice(0, 200)}
             </Dotdotdot>
-            {!noPublisher && <Publisher account={owner} minimal />}
+            {!noPublisher && <Publisher account={owner} />}
```
kremalicious, line 56: The `minimal` prop on the `Publisher` component needs to be kept in this context though. It is dimming the the publisher output so they are not all pink on all asset teasers, which this current change did

src/components/Asset/AssetContent/index.tsx
```diff
@@ -78,6 +79,11 @@ export default function AssetContent({
               </Button>
             </div>
           )}
+          <RelatedAssets
+            tags={asset?.metadata?.tags}
```
kremalicious, line 83: As this component conceptionally will always be used within the `Asset` context, there is no need to pass those props. Remove them and use the `useAsset()` hook to get them from inside the `RelatedAssets` component

src/components/Asset/RelatedAssets/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+    async function getAssets() {
+      setIsLoading(true)
+      const tagQuery = generateBaseQuery(
+        generateQuery(chainIds, asset.nftAddress, 4, asset.metadata.tags)
+      )
+      const tagResults = (await queryMetadata(tagQuery, newCancelToken()))
+        .results
+
+      if (tagResults.length === 4) {
```
kremalicious, line 31: I tried to follow in the comments what the intention of this check is and found this (replace 3 with 4 as I increased it):

>  If we do the first query and we get 3 assets, then we don't do the second query.

But pretty sure this wasn't meant as "we get EXACTLY 3 assets". Cause that's pretty much never going to happen in reality. So the me sounds like this check here should rather be representation of "if there are 4 or more results"?

So:

```suggestion
      if (tagResults.length >= 4) {
```

Which then needs another tweak to limit the `tagResults` to the first 4, and then set it into state.

jamiehewitt15: The original think was the following: If the `size` is set to 4 in `generateQuery` then the maximum results the query should return is 4, although in many cases it will be less and then in that case we also do the owner query

kremalicious (approved):

PR #1749, opened 2022-10-19
ignore assets with state 5

### LINKED ISSUE #1747: In Ocean Market, owner can list / unlist assets
**Summary**

- Problem: active assets are discoverable in Ocean Market, even if owner doesn't want to list them.
- Solution: In Ocean Market, owner can list / unlist assets. Eg via candidate solution (2) or (3) below.

**Motivation**

- Submissions from the first predict-ETH challenge show up in Ocean Market. [Here](https://market.oceanprotocol.com/asset/did:op:ec7fd49888469f10e234af498ec8e06a44881d48effdc6d284d6107586d6609d). Wed expected this; that's how Ocean is designed by default. However we don't want this to happen.
- To generalize, one can imagine many flows where someone publishes an asset in Ocean, but they don't want it to be listed / discoverable in Ocean Market or other public markets. 
- Or, they've published it in Ocean Market, and want to keep it active but don't want it discoverable anymore. 

**Candidate solutions**
1. In the predict-eth README, don't publish the ddo. Just the data NFT & datatoken (including url service), and not the asset. 
2. Leverage [erc721.state](https://docs.oceanprotocol.com/core-concepts/did-ddo#state). Any asset with state >= 1 isn't displayed in Ocean Market.
3. Have another parameter in [`ddo.metadata`](https://docs.oceanprotocol.com/core-concepts/did-ddo#metadata): `listInMarkets` (a simple boolean). Use this appropriately in Ocean Market.

**Analysis: candidate 1**

Candidate 1 solves the first issue listed above, not the other two. It _is_ useful, however, so it will be done via [ocean.py#1022](https://github.com/oceanprotocol/ocean.py/issues/1022).

**Analysis: candidate 2**

We already have [erc721.state](https://docs.oceanprotocol.com/core-concepts/did-ddo#state). Any asset with state >=1 will not be displayed in the market.

Here are the possible states:
states

Q: which state should the asset be, to not display? It's still **active**, after all (state 0). 

A: The question is N/A. "state" is orthogonal to "list y/n". So we need a different field.

**Analysis & details: candidate 3**

> 3. Have another parameter in [`ddo.metadata`](https://docs.oceanprotocol.com/core-concepts/did-ddo#metadata): `listInMarkets` (a simple boolean). Use this appropriately in Ocean Market.

This solves all three issues.

**Precise specs for solution:**

- Have another field in ddo.metadata: `"listInMarkets" : true | false`.
- In Market's browse/search results, it filters out assets having showInMarkets = False. Exception: when asset owner is logged in, the hidden assets will not be filtered, and will have an affordance so they know it's hidden.
- Market still shows the asset page to anyone, when direct url is given. (The asset isn't "listed" per se, but people can still look at it.)
- In Ocean Market, inside "edit ddo" page, there will be a new entry "show/hide" that can be toggled, which set the value for "listInMarkets".
- All of the behavior above is independent of "state" field

**TODOs:**
- [ ] In aquarius, support "precise specs" above as appropriate
- [ ] In market, ""
- [ ] In docs, document as appropriate
- [ ] Re-deploy components as needed

Background: [slack message](https://oceanprotocol.slack.com/archives/C047QMC1Z4G/p1666107788337639) that led to this.

mihaisc:
Closes #1747 

Assets with state 5 are ignored from all queries. 
They are shown in your own profile.

When viewing from another account: 
image

My account:
image

### DIFF
--- a/src/@context/Asset.tsx
+++ b/src/@context/Asset.tsx
@@ -77,7 +77,7 @@ function AssetProvider({
         return
       }
 
-      if (asset.nft.state) {
+      if ([1, 2, 3].includes(asset.nft.state)) {
         // handle nft states as documented in https://docs.oceanprotocol.com/concepts/did-ddo/#state
         let state
         switch (asset.nft.state) {
--- a/src/@context/Profile/index.tsx
+++ b/src/@context/Profile/index.tsx
@@ -46,10 +46,12 @@ const clearedProfile: Profile = {
 function ProfileProvider({
   accountId,
   accountEns,
+  ownAccount,
   children
 }: {
   accountId: string
   accountEns: string
+  ownAccount: boolean
   children: ReactNode
 }): ReactElement {
   const { chainIds } = useUserPreferences()
@@ -111,7 +113,8 @@ function ProfileProvider({
         const result = await getPublishedAssets(
           accountId,
           chainIds,
-          cancelTokenSource.token
+          cancelTokenSource.token,
+          ownAccount
         )
         setAssets(result.results)
         setAssetsTotal(result.totalResults)
@@ -134,7 +137,13 @@ function ProfileProvider({
     return () => {
       cancelTokenSource.cancel()
     }
-  }, [accountId, appConfig.metadataCacheUri, chainIds, isEthAddress])
+  }, [
+    accountId,
+    appConfig.metadataCacheUri,
+    chainIds,
+    isEthAddress,
+    ownAccount
+  ])
 
   //
   // DOWNLOADS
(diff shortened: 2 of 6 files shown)


src/@context/Asset.tsx
```diff
@@ -77,7 +77,7 @@ function AssetProvider({
         return
       }
 
-      if (asset.nft.state) {
+      if ([1, 2, 3].includes(asset.nft.state)) {
         // handle nft states as documented in https://docs.oceanprotocol.com/concepts/did-ddo/#state
```
kremalicious, line 81: while we are at it, this URL does not exist. It's `https://docs.oceanprotocol.com/core-concepts/did-ddo#state`

kremalicious (approved):

src/@context/Asset.tsx
mihaisc: that was an old comment,  good catch

PR #1751, opened 2022-10-24
update ocean to 2.4.0

mihaisc:


EnzoVezzaro (approved):

src/@utils/provider.ts
```diff
@@ -87,7 +88,13 @@ export async function getFileUrlInfo(
   providerUrl: string
 ): Promise<FileInfo[]> {
   try {
-    const response = await ProviderInstance.checkFileUrl(url, providerUrl)
+    const fileUrl: UrlFile = {
```
kremalicious, line 91: the method name `getFileUrlInfo()` does not fit anymore after this change so should be renamed to e.g. `getFileInfo()`

kremalicious: Forgot that comment, had still the changes from #1732 in mind. So are we going to leave the change to accept an object to cover all the other file types out of this PR?

kremalicious: alright, should be possible to defer the rest to #1732 as everything works with this PR and we need ocean.js v2.4.0 live, so good to go here

kremalicious (approved):

src/@utils/provider.ts
mihaisc: This is intended, this function `getFileUrlInfo` just returns file url info.

Issue #1531, opened 2022-06-22
replace `useCancelToken()` with `useAbortController()`


kremalicious:
axios `CancelToken` is deprecated and we already use a replacement for creating a standard `AbortSignal` with our `useAbortController()` hook.

So replace all usages of cancel token with abort signal.

PR #1754, opened 2022-10-27
Add most viewed

### LINKED ISSUE #895: Most popular assets
Top x assets based on netlify 'Top pages' . If i remember correctly we can fetch some analytics data from there. It would be cool to have a section based on user traffic.

inspiration: https://github.com/NiklasMerz/netlify-analytics-collector/blob/main/index.js . We need to create a small function to expose some metric. Tested the api described, it's working.

mihaisc:
Fixes #895


src/components/Home/index.tsx
```diff
@@ -66,6 +91,12 @@ export default function HomePage(): ReactElement {
       />
 
       <SectionQueryResult title="Most Sales" query={queryMostSales} />
+      <SectionQueryResult
+        title="Most Viewed"
+        query={queryMostViewed}
+        queryData={mostViewedDids}
+        tooltip="Assets from all supported chains. It is not influenced by selected networks"
```
EnzoVezzaro, line 98: Not sure why we're not filtering by selected networks. I think that's a valuable information to have 👍🏽

EnzoVezzaro (approved):
LGTM 👍🏽 left a comment, but it's not blocking

src/components/Home/index.tsx
mihaisc: Can't . This information is from netlify and it's limited. For most chains i would have 0 results.

mihaisc: Or maybe i can do something

mihaisc: nope way too complicated

jamiehewitt15, line 98:
```suggestion
        tooltip="Assets with the most traffic from all supported chains. It is not influenced by selected networks"
```

jamiehewitt15 (commented):
Looks good to me

kremalicious:
This breaks a bit the “most sections” conceptionally, to compare within the list we need to output the value we order by. Which is why we do this for veOCEAN allocation and sales. Can we find a way to output the views numbers then? Ideally in place of the veOCEAN number in this place

mihaisc:
I think we can

mihaisc:
I did good test? 🤣

src/components/Home/MostViews/index.test.tsx
```diff
@@ -0,0 +1,8 @@
+import testRender from '../../../../.jest/testRender'
+import { render } from '@testing-library/react'
+import React from 'react'
+import MostViews from '.'
+
+describe('components/Home/MostViews', () => {
+  testRender(<MostViews />)
```
kremalicious, line 7: works, but will lead to unreliable tests as by doing it this way, all the real requests are actually being fired, meaning the views endpoint and another `queryMetadata()` call. Ideally, both are mocked inside this file, also to test all possible return types. Could be good exercise for tomorrow's testing intro

src/components/Home/MostViews/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+import Tooltip from '@shared/atoms/Tooltip'
+import Markdown from '@shared/Markdown'
+import AssetList from '@shared/AssetList'
+import { LoggerInstance } from '@oceanprotocol/lib'
+import { sortAssets } from '@utils/index'
+
+export default function MostViews(): ReactElement {
+  const [loading, setLoading] = useState<boolean>()
+  const [mostViewed, setMostViewed] = useState<AssetExtended[]>([])
+  const newCancelToken = useCancelToken()
```
kremalicious, line 20: puh, this is fun and we should get https://github.com/oceanprotocol/market/issues/1531 done to prevent things like this, and always use `AbortController`

src/components/Home/MostViews/index.test.tsx
kremalicious: and also spills a console warning which should be addressed https://github.com/oceanprotocol/market/actions/runs/3413734266/jobs/5680799611#step:6:297

mihaisc: on myside there are multiple errors on the tests, didn't really understood the point. I'll look into mocking the call.

mihaisc:
👍 i approve 😆

PR #1757, opened 2022-10-31
Checking for google drive links

### LINKED ISSUE #1750: Add warning message when publishing assets with GDrive links
GDrive links are problematic and cause a lot of problems when consuming assets, let's add a warning message when users are publishing using these types of URLs.
Docs issue: https://github.com/oceanprotocol/docs/issues/1144

jamiehewitt15:
Fixes #1750 

Changes proposed in this PR:

- This prevents people from publishing assets that are hosted on google drive

### DIFF
--- a/src/components/Publish/_validation.ts
+++ b/src/components/Publish/_validation.ts
@@ -55,7 +55,7 @@ const validationService = {
           .url('Must be a valid URL.')
           .test(
             'GoogleNotSupported',
-            'Google drive is not a supported hosting service. Please use an alternative.',
+            'Google Drive is not a supported hosting service. Please use an alternative.',
             (value) => {
               return !value?.toString().includes('drive.google')
             }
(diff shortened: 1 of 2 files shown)


src/components/Publish/_validation.ts
```diff
@@ -32,7 +32,17 @@ const validationService = {
   files: Yup.array<FileInfo[]>()
     .of(
       Yup.object().shape({
-        url: Yup.string().url('Must be a valid URL.').required('Required'),
+        url: Yup.string()
+          .test(
+            'GoogleNotSupported',
+            'Google drive is not a supported hosting service. Please use an alternative.',
```
kremalicious, line 38: Good enough, we probably want this on Edit validation too?

```suggestion
            'Google Drive is not a supported hosting service. Please use an alternative.',
```

jamiehewitt15: Cool, yeah I'll add it on the edit form as well

kremalicious (approved):
Excellent minimal change. Good we didn’t end up with more warning banners in UI

PR #1765, opened 2022-11-01
adding ipfs / arweave support

EnzoVezzaro:
- Fixes #1711 .

Changes proposed in this PR:

- created storage type dropdown (Url / IPFS / Arweave)
- added storage type selection on publish
- added storage type selection on edit
- updated `@oceanprotocol/lib@2.5.2`

WIP: 

- [X] UI redesign fileInput
- [x] fix IPFS typing [issue](https://github.com/oceanprotocol/ocean.js/pull/1640)
- [x] wait for fix provider: [IPFS](https://github.com/oceanprotocol/provider/issues/558) and [Arwave](https://github.com/oceanprotocol/provider/issues/555)
- [x] validation and edit form
- [x] waiting for https://github.com/oceanprotocol/provider/issues/555 to test `ARWEAVE`
- [x] fix failing test after merged main
- [x] finish suggestions on review
  - [x] redesign UI uploader
  - [x] error messages and catch errors
  - [x] cleanup and functionality test
  - [x] tests

TO TEST:

- IPFS: you can upload a file in IPFS using [web3.storage](https://web3.storage) (to get CID)
- Arweave: you can use this[ little app](https://hv4gxzchk24cqfezebn3ujjz6oy2kbtztv5vghn6kpbkjc3vg4rq.arweave.net/PXhr5EdWuCgUmSBbuiU587GlBnmde1MdvlPCpIt1NyM/) to upload a file into Arweave (to get the transactionID)

## Publish

Before validation:
Screenshot 2022-11-11 at 10 44 47

After validation:
Screenshot 2022-11-14 at 11 16 57

## Edit
When entered in edit form:
Screenshot 2022-11-14 at 10 50 13

After clicking on "x":
Screenshot 2022-11-14 at 10 50 19

PR #1766, opened 2022-11-02
fix validation link field on edit view when the asset is already hosted on g drive

EnzoVezzaro:
Fixes # .

Changes proposed in this PR:

- fix links validation when url is on g drive

To test you need to be on a asset with sample file.


EnzoVezzaro:
I'm trying to show the validation error also, but I'm not having luck:

Basically if the user has already a sample file on g drive, when they arrive to the page they should see right away:

Screenshot 2022-11-02 at 06 36 39


Right now we're not showing the error which might confuse a bit:

Screenshot 2022-11-02 at 06 36 30

bogdanfazakas:
> I'm trying to show the validation error also, but I'm not having luck.
> 
> Basically if the user has already a sample file on g drive, when they arrive to the page they should see right away:
> 
> Screenshot 2022-11-02 at 06 36 39
> 
> Right now we're not showing the error which might confuse a bit:
> 
> Screenshot 2022-11-02 at 06 36 30
> 
> but you should not be able to validate a g drive url, making the submit disabled until you entered a valid url.
> 
> UPDATE: haven't been able to show the alert on page load, I will stop here to avoid spending too much time on this single thing.

I think if we find a way to kick the YUP validation here on files imput we will get that nice error displayed and we can get rid of the extra code added with this PR, will give a quick try

EnzoVezzaro:
@bogdanfazakas I merged main into this, but I'm not sure it's needed anymore. Right now It's an edge case I'm having trouble to replicate since we can't publish assets on g drive anymore. Also this logic has been heavily refactored in this [PR](https://github.com/oceanprotocol/market/pull/1765)

bogdanfazakas:
> @bogdanfazakas I merged main into this, but I'm not sure it's needed anymore. Right now It's an edge case I'm having trouble to replicate since we can't publish assets on g drive anymore. Also this logic has been heavily refactored in this [PR](https://github.com/oceanprotocol/market/pull/1765). Maybe we can keep it the way it is, to at least have the link field not validated anymore if the asset is on g drive.

Yep, I tried to add least add that error but no luck so far, so to avoid adding more complexity we can stick to this.
But I was wandering, should we invalidate existing sample links if they are on g drive even if the user has not touched the field? or we should somehow try to bypass the validation in this case rather than what we have build here, trying to invalidate and show errors in this case?

EnzoVezzaro:
> > @bogdanfazakas I merged main into this, but I'm not sure it's needed anymore. Right now It's an edge case I'm having trouble to replicate since we can't publish assets on g drive anymore. Also this logic has been heavily refactored in this [PR](https://github.com/oceanprotocol/market/pull/1765). Maybe we can keep it the way it is, to at least have the link field not validated anymore if the asset is on g drive.
> 
> Yep, I tried to add least add that error but no luck so far, so to avoid adding more complexity we can stick to this. But I was wandering, should we invalidate existing sample links if they are on g drive even if the user has not touched the field? or we should somehow try to bypass the validation in this case rather than what we have build here, trying to invalidate and show errors in this case?

Yeah, maybe this PR is not needed anymore. At first, this was high priority 'cause a user had an issue with it, but now I feel it's getting kind of worthless considering #1765 as I said before:

> Also this logic has been heavily refactored in this [PR](https://github.com/oceanprotocol/market/pull/1765). 

Not sure if we want to continue invalidating the sample file (on g drive) here or wait for #1765 (maybe it's best)

EnzoVezzaro:
we need to review this case after merging #1765 since this will be overritten.

jamiehewitt15:
This is a bit tricky to test as I don't have any pre-existing assets with google drive links as a sample file

PR #1768, opened 2022-11-02
Ensuring the market doesn't crash when asset has no services

jamiehewitt15:
Fixes #1768

Changes proposed in this PR:

- Ensuring the market doesn't crash when asset has no services

**To test:** view [this asset](https://market-r84wawq01-oceanprotocol.vercel.app/asset/did:op:984ae0bc96d95d0ac766f82f7449d897adfec916e586078f0e2ee3db394cb5b2)

### DIFF
--- a/src/@context/Asset.tsx
+++ b/src/@context/Asset.tsx
@@ -124,8 +124,8 @@ function AssetProvider({
 
     const accessDetails = await getAccessDetails(
       asset.chainId,
-      asset.services[0].datatokenAddress,
-      asset.services[0].timeout,
+      asset.services?.[0]?.datatokenAddress,
+      asset.services?.[0]?.timeout,
       accountId
     )
     setAsset((prevState) => ({
--- a/src/@context/Asset.tsx
+++ b/src/@context/Asset.tsx
@@ -120,12 +120,12 @@ function AssetProvider({
   // Helper: Get and set asset access details
   // -----------------------------------
   const fetchAccessDetails = useCallback(async (): Promise<void> => {
-    if (!asset?.chainId || !asset?.services) return
+    if (!asset?.chainId || !asset?.services?.length) return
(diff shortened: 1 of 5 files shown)


EnzoVezzaro:
Now the page is not crashing, but maybe we can do something for this two things:

1) show a message to the user instead of pricing calculation:
This:
Screenshot 2022-11-02 at 08 55 42

Instead of:
Screenshot 2022-11-02 at 08 54 47

2) let the user know the asset is missing its services (this might not be necessary, let's ask @mihaisc ) with a message somewhere in the UI

jamiehewitt15:
Yeah, good idea. I've updated the messaging so it shows the warning alert rather than infinite loading

src/components/Asset/AssetActions/Download.tsx
```diff
@@ -183,6 +183,7 @@ export default function Download({
       consumableFeedback={consumableFeedback}
     />
   )
+  console.log('asset.services', asset.services)
```
EnzoVezzaro, line 186: remove logs

EnzoVezzaro (commented):
found some logs to remove

src/@context/Asset.tsx
```diff
@@ -124,8 +124,8 @@ function AssetProvider({
 
     const accessDetails = await getAccessDetails(
       asset.chainId,
-      asset.services[0].datatokenAddress,
-      asset.services[0].timeout,
+      asset.services?.[0]?.datatokenAddress,
```
kremalicious, line 127: This method should not fire when there are no services. So instead of all this optional chaining on here, the first line should be modified to return early. 

Most likely:

```js

if (!asset?.chainId || !asset?.services?.length) return
```

kremalicious (approved):

PR #1770, opened 2022-11-03
Handle provider error & showing Retry button

### LINKED ISSUE #1646: Handle provider error on initialize
Sometimes the provider randomly fails. We need to catch that , display a proper error and prevent the user from ordering .
Replace `Buy` button with a `Retry` one

jamiehewitt15:
Fixes #1646

Changes proposed in this PR:

- In the case of a provider error, the error is logged to the console and a toast pop-up is shown asking the user to retry. The button text changes from "Buy" to "Retry"

### Screenshot

Screenshot from 2022-11-03 16-12-06

### To test:

In order to test it's necessary to run this locally and force the provider to have an error. This can be done by changing the arguments in: 

```Javascript
await ProviderInstance.initialize(
        asset.id,
        asset.services[0].id,
        0,
        accountId,
        asset.services[0].serviceEndpoint
      )
```

### DIFF
--- a/src/components/Asset/AssetActions/ButtonBuy/index.test.tsx
+++ b/src/components/Asset/AssetActions/ButtonBuy/index.test.tsx
@@ -49,96 +49,93 @@ const computeProps: ButtonBuyProps = {
 
 describe('Asset/AssetActions/ButtonBuy', () => {
   //  TESTS FOR LOADING
-  it('Renders Buy button without crashing', async () => {
+  it('Renders Buy button without crashing', () => {
     render(<ButtonBuy {...downloadProps} isLoading />)
     const button = screen.getByText('TEST')
     expect(button).toContainHTML('<Loader')
   })
 
   //  TESTS FOR DOWNLOAD
-  it('Renders Buy button without crashing', async () => {
+  it('Renders Buy button without crashing', () => {
     render(<ButtonBuy {...downloadProps} />)
     const button = screen.getByText('Buy for 1 day')
     expect(button).toContainHTML('<button')
   })
 
-  it('Renders Buy button without crashing when hasPreviousOrder=true', async () => {
-    downloadProps.hasPreviousOrder = true
-    render(<ButtonBuy {...downloadProps} />)
+  it('Renders Buy button without crashing when hasPreviousOrder=true', () => {
+    render(<ButtonBuy {...downloadProps} hasPreviousOrder />)
     const button = screen.getByText('Download')
     expect(button).toContainHTML('<button')
   })
 
-  it('Renders retry button for download without crashing', async () => {
+  it('Renders retry button for download without crashing', () => {
     render(<ButtonBuy {...downloadProps} retry />)
     const button = screen.getByText('Retry')
     expect(button).toContainHTML('<button')
   })
 
-  it('Renders get button for free download without crashing', async () => {
-    downloadProps.priceType = 'free'
-    downloadProps.hasPreviousOrder = true
-    render(<ButtonBuy {...downloadProps} />)
+  it('Renders get button for free download without crashing', () => {
+    render(<ButtonBuy {...downloadProps} priceType="free" hasPreviousOrder />)
     const button = screen.getByText('Download')
     expect(button).toContainHTML('<button')
   })
 
-  it('Renders "Get" button for free assets without crashing', async () => {
-    downloadProps.priceType = 'free'
-    downloadProps.hasPreviousOrder = false
-    render(<ButtonBuy {...downloadProps} />)
+  it('Renders "Get" button for free assets without crashing', () => {
+    render(<ButtonBuy {...downloadProps} priceType="free" />)
     const button = screen.getByText('Get')
     expect(button).toContainHTML('<button')
   })
 
-  it('Renders Buy button without crashing', async () => {
-    downloadProps.assetTimeout = 'Forever'
-    downloadProps.priceType = 'fixed'
-    downloadProps.isConsumable = false
-    render(<ButtonBuy {...downloadProps} />)
+  it('Renders Buy button without crashing', () => {
+    render(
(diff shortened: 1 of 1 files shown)


src/@utils/order.ts
```diff
@@ -20,6 +21,27 @@ import {
 } from '../../app.config'
 import { toast } from 'react-toastify'
 
+async function initializeProvider(
+  asset: AssetExtended,
+  accountId: string,
+  providerFees?: ProviderFees
+): Promise<ProviderInitialize> {
+  if (!providerFees) {
```
kremalicious, line 29: super small thing, but we mostly utilize return early pattern to prevent excessive nesting. So in this case just turn around logic to remove the first level nesting by doing `if (providerFees) return`.

For the rest, just noting here as I'm not sure we want to touch that mega `@utils/order` further, as changes are hard to test.

Another problem, but I get why it's here as it was just hacked into original code: accepting a parameter which then is only used to check if method should run seems unnecessary. Cause why is the method even called in the first place when existence of `providerFees` is known before calling it? Better pattern here would be to only call this method when necessary, which is only when `providerFees` are not present.

And finally, this method should always return something as it's kinda only luck leading to this not breaking as we have `providerFees || initializeData.providerFee` further down. Furthermore, if no `providerFees` are passed AND `initializeProvider()` returns empty `initializeData`, the app breaks. Solution could be optional chaining with `initializeData?.providerFee` and make `initializeProvider()` always return something

bogdanfazakas:
On the consume part looks good, but should we add some retry logic to compute as well ?

src/components/Asset/AssetActions/ButtonBuy/index.test.tsx
```diff
@@ -0,0 +1,291 @@
+import React from 'react'
+import { render, screen } from '@testing-library/react'
+import ButtonBuy from './'
+
+//  TESTS FOR DOWNLOAD
+test('Renders Buy button without crashing', async () => {
+  render(
+    <ButtonBuy
+      action="download"
```
kremalicious, line 9: lots of tests, nice!

They;'re just hard to follow with all those props which is of course caused by this mega component which should simply not have that many props. To keep at least the tests readable would suggest to make use of object spreading, define all the props you will not change as e.g. `defaultProps`, then spread them on component and after it redefine what you are testing, like so for the next `hasPreviousOrder=true` test:

```
<ButtonBuy {...defaultProps} hasPreviousOrder />
```

jamiehewitt15: Ok cool, yeah that makes sense. I was thinking the test file was getting a bit too big

kremalicious: also the structure should be this, just for consistency's sake, most likely won't change anything during runs:

```js
describe('Asset/AssetActions/ButtonBuy', () => {
  it('Renders Buy button without crashing', () => { ... })
  it('Renders Buy button without crashing when hasPreviousOrder=true', () => { ... })
})
```

jamiehewitt15: OK cool, looks good. I'll update it

bogdanfazakas:
can confirm that the _please retry_ message is displayed on compute as well, but the button doesn't turn into the **Retry** as it does on consume.
Screenshot 2022-11-10 at 08 34 44

jamiehewitt15:
> can confirm that the _please retry_ message is displayed on compute as well, but the buy button doesn't turn into the **Retry** as it does on consume. 

When I tested yesterday it seemed to be working but I'll have another look at it

src/components/Asset/AssetActions/ButtonBuy/index.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+  //  TESTS FOR DOWNLOAD
+  it('Renders Buy button without crashing', async () => {
+    render(<ButtonBuy {...downloadProps} />)
+    const button = screen.getByText('Buy for 1 day')
+    expect(button).toContainHTML('<button')
+  })
+
+  it('Renders Buy button without crashing when hasPreviousOrder=true', async () => {
+    downloadProps.hasPreviousOrder = true
```
kremalicious, line 66: Object reassignments here and in all the following tests lead to modifying the global object. You reassign e.g. here `downloadProps.hasPreviousOrder` so the next test running after it will have `downloadProps.hasPreviousOrder = true`.

You have to put the overwrite props after the object spread on the component as suggested before, like so:

```
<ButtonBuy {...downloadProps} hasPreviousOrder />
```

This needs to be changed for each and every test block in here. Another thing is that none of the tests should be async, as you are not `await` for anything, so the `async` can be removed from all tests too.

kremalicious: And despite the many tests in here and reporting 100% coverage, there are still large parts of the codebase untested. Because some messages are not triggered by any test:

Screenshot 2022-11-10 at 10 55 40

From what I'm seeing, another good test would be everything around the disabled button. Like, line 100 deals with `isBalanceSufficient` so testing that button stays disabled/gets enabled when we want it to would cover that and probably more

kremalicious (approved):
good enough regarding tests. As we quickly talked in standup, would only make sense going further in the tests by completely refactoring `ButtonBuy`, which is out of scope for this task. Thanks for pushing through all this @jamiehewitt15!


(did not recheck all functionality again so approval is mainly for the latest test-related changes)

jamiehewitt15:
I did another test and it's working alright, also working with compute:

image

bogdanfazakas:
> I did another test and it's working alright, also working with compute:

Just checked and i think there are some use cases that are not fully covered, then will check later what is missing then.

Screenshot 2022-11-11 at 12 04 32

PR #1771, opened 2022-11-04
package updates & technical debt cleanup

### LINKED ISSUE #1153: less duplication in getAssets methods in `@utils/aquarius`
With ... we have:

- `getAssetsFromDidList()`
- `getAssetsFromDtList()`
- `retrieveDDOListByDIDs()`

Where all of them have exact same fetch code with a slightly different `filter` being sent, and `retrieveDDOListByDIDs()` just returns results in different order.

This could be one method `getAssets()` or `getAssetsFromList()` or `getAssetsFromIds()`:

```ts
export async function getAssets(
  list: string[], // list of DIDs or Datatoken addresses
  filter: string, // 
  chainIds: number[],
  cancelToken: CancelToken
): Promise<Asset[]> {...}
```

But as utility methods they could be handy to keep separate, in that case we should at least have them all use a shared fetching method, and then only modify input/output differently in separate methods

kremalicious:
* bump to ocean.js v2.5.2, which requires web3.js v1.8.0
* bump to [Next.js v13](https://nextjs.org/blog/next-13)
* run CI against Node.js v18, switch to it in local development too (`nvm use`)
* yet another `lockfileVersion` update. Please stop pushing old `package-lock.json` into `main`. Always use the Node.js version defined in `.nvmrc`. And NEVER use `--legacy-peer-deps` when updating dependencies.
* run [codemod for new `next/link` behavior](https://nextjs.org/docs/advanced-features/codemods#new-link) 
* unused methods & GraphQL queries cleanup
* updated `tsconfig.json` for actual browser builds, this makes ESM modules not error builds anymore
* fix `tsc` not erroring on faulty import paths, which then lead to problems in test runs: fix many import paths like `src/components/` which should be either relative to the file (`../../src/` etc.) or use our aliases, so in this case `@components/`
* add mechanism for Jest to understand ESM modules, so those test runs with ESM modules now work too
* write tests for some things refactored by this PR
* final piece to close #1153
* update CODEOWNERS
* both remaining outside issues have been resolved:
  a) fix in new web3.js version for https://github.com/web3/web3.js/issues/5601
  b) Vercel ships Node.js v18 support https://github.com/vercel/community/discussions/986, which depends on https://github.com/aws/aws-lambda-base-images/issues/47


kremalicious:
issue for that web3.js + Next.js problem:
https://github.com/web3/web3.js/issues/5601

jamiehewitt15 (approved):

PR #1773, opened 2022-11-07
Fix compute asset selection lists

bogdanfazakas:
Fixes:
https://discord.com/channels/612953348487905282/720631837122363412/1039053158778941440

Changes proposed in this PR:

- fix algorithm asset selection list for algorithms with service of `acces` type 
- fix the list of datasets algorithm is allowed to run
- small UI fix for the datasets algorithm is allowed to run list

### DIFF
--- a/src/@utils/assetConvertor.ts
+++ b/src/@utils/assetConvertor.ts
@@ -19,7 +19,7 @@ export async function transformAssetToAssetSelection(
 
     if (
       asset?.accessDetails?.price &&
-      algoComputeService?.serviceEndpoint === datasetProviderEndpoint
+      algoService?.serviceEndpoint === datasetProviderEndpoint
     ) {
       let selected = false
       selectedAlgorithms?.forEach((algorithm: PublisherTrustedAlgorithm) => {


src/@utils/assetConvertor.ts
```diff
@@ -14,7 +14,8 @@ export async function transformAssetToAssetSelection(
   const algorithmList: AssetSelectionAsset[] = []
 
   for (const asset of extendedAssets) {
-    const algoComputeService = getServiceByName(asset, 'compute')
+    const algoComputeService =
```
kremalicious, line 17: based on what it is checking, the variable is not exclusive to the compute service anymore, so should be renamed to e.g.:

```suggestion
    const algoService =
```

kremalicious:
confirming that with this change, this asset is now showing an algo to run with https://market-git-fix-compute-selection-lists-oceanprotocol.vercel.app/asset/did:op:3b0cbedd39d7cd7b55fea3028e04fa71b354dc81f8bf4ebbffe1614d351b6a7d shows

src/@utils/assetConvertor.ts
bogdanfazakas: makes sense now 😃

mihaisc (approved):

PR #1774, opened 2022-11-07
Fetching prices from Aquarius in AssetList component

### LINKED ISSUE #1654: Fetch price from aquarius
Get price from aquarius so we don't have to do multiple sub queries. 

```
"price": {
      "value" : 33,
      "tokenAddress": "0x0....",
      "tokenSymbol": "ocean"
}
```
If value is 0 the it's a dispenser so it's free
if the price object is empty  (`{}`) then no valid pricing was found.

Requires new aquarius deployment( > 4.2.0)
Aqua issue: https://github.com/oceanprotocol/aquarius/issues/853

jamiehewitt15:
- Fixes #1654 

Changes proposed in this PR:

- Fetching prices from Aquarius in AssetList component


src/components/@shared/AssetList/index.tsx
```diff
@@ -41,22 +44,25 @@ export default function AssetList({
...
 
     setAssetsWithPrices(assets as AssetExtended[])
     setLoading(false)
-    async function fetchPrices() {
-      const assetsWithPrices = await getAccessDetailsForAssets(
-        assets,
-        accountId || ''
-      )
+
+    async function fetchPrices(token?: CancelToken) {
```
kremalicious, line 55: this seems like the wrong tactic. The prices coming directly from the Aquarius-delivered DDO implies we do not have to fetch them again. So this component here can finally be a pure visual component without any data fetching. That is the whole goal of getting prices from Aquarius.

Remove all fetching logic from in here and refactor to get the price from `assets`

jamiehewitt15:
Running into an issue with this where occasionally we don't receive the price from Aquarius https://github.com/oceanprotocol/aquarius/issues/853#issuecomment-1316802025

kremalicious:
Has this been abandoned?

kremalicious:
> Currently blocked by https://github.com/oceanprotocol/aquarius/pull/928 as we're not receiving the prices from Aquarius every time.

In frontend world, you can never assume all fetched data is always there so you have to find ways to deal with that. Especially with the things the publisher does not add, but Aquarius alone. So strongly suggest to actually put in logic dealing with a missing price object

jamiehewitt15:
> In frontend world, you can never assume all fetched data is always there so you have to find ways to deal with that. Especially with the things the publisher does not add, but Aquarius alone. So strongly suggest to actually put in logic dealing with a missing price object

Ok cool, will do  👍

kremalicious:
there is a final price difference on asset details between this PR and on live, please investigate and fix:

Live

Screenshot 2023-01-19 at 12 42 32

https://market.oceanprotocol.com/asset/did:op:3dd05b9f564b9f43609e03121351b83bb1fa0a64b94e5635a3990679318b5dc5

This PR's preview

Screenshot 2023-01-19 at 12 42 57

https://market-git-issue-1654-price-aquarius-oceanprotocol.vercel.app/asset/did:op:3dd05b9f564b9f43609e03121351b83bb1fa0a64b94e5635a3990679318b5dc5

Additionally there should be no `package-lock.json` changes in this PR as no dependencies were changed

jamiehewitt15:
> there is a final price difference on asset details between this PR and on live, please investigate and fix:

The issue is that Aquarius is sending the price without the fees. I'm going to update this PR so that the price from Aquarius is only used on the teasers

kremalicious:
> The issue is that Aquarius is sending the price without the fees. I'm going to update this PR so that the price from Aquarius is only used on the teasers

yes, that was the actual goal, only on the asset teasers. The asset details should always fetch the actual price with fees

bogdanfazakas (approved):
LGTM, before merging this pls copy the `package-lock.json` file content from `main` branch since we should not touch that if we have not updated the dependency

src/components/@shared/AssetList/index.tsx
```diff
@@ -1,19 +1,7 @@
 import AssetTeaser from '@shared/AssetTeaser'
-import React, { ReactElement, useEffect, useState } from 'react'
+import React, { ReactElement } from 'react'
 import Pagination from '@shared/Pagination'
 import styles from './index.module.css'
-import Loader from '@shared/atoms/Loader'
-import { useIsMounted } from '@hooks/useIsMounted'
-import { getAccessDetailsForAssets } from '@utils/accessDetailsAndPricing'
```
kremalicious, line 7: so this method `getAccessDetailsForAssets()` should in the end disappear completely, but seeing we are using it for bookmarks and stats on profile page. Can we also get rid of this? Or do in new PR?

src/components/@shared/AssetTeaser/index.tsx
```diff
@@ -60,15 +59,15 @@ export default function AssetTeaser({
...
-          <div className={styles.price}>
-            {isUnsupportedPricing || !asset.services.length ? (
-              <strong>No pricing schema available</strong>
-            ) : (
-              <Price accessDetails={asset.accessDetails} size="small" />
-            )}
-          </div>
-        )}
+        <div className={styles.price}>
+          {isUnsupportedPricing ||
```
kremalicious, line 63: all those clauses here should be part of `isUnsupportedPricing`

src/components/Asset/AssetActions/Download.tsx
```diff
@@ -199,7 +199,9 @@ export default function Download({
           />
         ) : (
           <>
-            {isUnsupportedPricing || !asset.services.length ? (
+            {isUnsupportedPricing ||
```
kremalicious, line 202: same here, put the following clauses behind the `isUnsupportedPricing` variable

kremalicious (changes requested):
seems to be all working as expected from functionality standpoint. Now please solve conflicts and address review comments

jamiehewitt15:
> seems to be all working as expected from functionality standpoint. Now please solve conflicts and address review comments

Cool, thanks for the review. I'll address the comments

src/@utils/accessDetailsAndPricing.ts
```diff
@@ -291,56 +291,3 @@ export async function getAccessDetails(
     LoggerInstance.error('Error getting access details: ', error.message)
   }
 }
-
-export async function getAccessDetailsForAssets(
```
kremalicious, line 295: cool, nice removal. This leaves many unused import at top of file, see the GitHub warnings further up

src/@utils/assetConvertor.ts
```diff
@@ -8,17 +7,14 @@ export async function transformAssetToAssetSelection(
...
   const algorithmList: AssetSelectionAsset[] = []
 
-  for (const asset of extendedAssets) {
+  for (const asset of assets) {
     const algoService =
       getServiceByName(asset, 'compute') || getServiceByName(asset, 'access')
 
     if (
-      asset?.accessDetails?.price &&
+      asset?.stats?.price?.value &&
```
kremalicious, line 17: just double checking: are we using what `transformAssetToAssetSelection()` returns on each asset to then start a compute job with it? If not, then all changes in this file are good. But if not, we would need to fetch actual `accessDetails` for respective assets

src/@utils/accessDetailsAndPricing.ts
jamiehewitt15: Yeah good point. I've removed the unused imports and also there was a query that we weren't using anymore that's also removed

src/@utils/assetConvertor.ts
jamiehewitt15: Yeah, we're not using it for anything in `src/components/Asset/AssetActions/Compute/index.tsx`

PR #1782, opened 2022-11-09
special veOCEAN number treatment

### LINKED ISSUE #1769: Round veOCEAN amounts
On the market homepage, at the bottom of each asset card, it is difficult to discern commas and decimal points in the veOCEAN value, especially with the grey colour.

For example, it is hard to tell that 403,832.438 veOCEAN is 403 thousand and not 403 million.

If this value was rounded (as is done in other places in the market), it would make comparing the values easier.

kremalicious:
* fixes #1769

Use a tweaked `formatNumber()` method for formatting all occurances of veOCEAN number in the UI, and remove the digits for better readability. This still keeps the formatting consistent with all our other numbers. Both, `formatNumber()` and our `PriceUnit` component now accept optional `decimals` param/prop.

This is not doing any rounding, as we only use this number for presentation and not doing any chain events with it so this should be fine.

Before:

Screenshot 2022-11-09 at 11 23 01

After:

Screenshot 2022-11-09 at 11 23 14

And on asset teasers:

Before:

Screenshot 2022-11-09 at 12 19 12

After:

Screenshot 2022-11-09 at 12 18 22

In footer:

After:

Screenshot 2022-11-09 at 13 02 05

### Bonus

There were a lot of duplicate declarations in `@utils/veAllocation`, meaning some import name which then is declared again as a `const` with same name. That really needs to be avoided as this will lead to all sorts of unintended consequences, like broken builds. We never catched this cause `@utils/veAllocation` was never part of any test run. Earlier iteration of this PR was touching that file and this had to be fixed then

### DIFF
--- a/src/@utils/numbers.ts
+++ b/src/@utils/numbers.ts
@@ -1,4 +1,19 @@
+import { formatCurrency } from '@coingecko/cryptoformat'
 import { Decimal } from 'decimal.js'
+
+export function formatNumber(
+  price: number,
+  locale: string,
+  decimals?: string
+): string {
+  return formatCurrency(price, '', locale, false, {
+    // Not exactly clear what `significant figures` are for this library,
+    // but setting this seems to give us the formatting we want.
+    // See https://github.com/oceanprotocol/market/issues/70
+    significantFigures: 4,
+    ...(decimals && { decimalPlaces: Number(decimals) })
+  })
+}
 
 // Run decimal.js comparison
 // http://mikemcl.github.io/decimal.js/#cmp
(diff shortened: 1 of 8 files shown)


EnzoVezzaro:
Maybe we can add this to the stats on the footer:

Screenshot 2022-11-09 at 08 35 38

kremalicious:
good point with the footer as this was exposing that having special methods only for veOCEAN numbers is not such a good idea, as what we actually want is that when certain numbers become very big, we don't want decimals to show.

So rather generalized a new `@utils/numbers/formatNumber` method and solving everything through there.

EnzoVezzaro (approved):

PR #1783, opened 2022-11-09
filter out `className` from `Tippy`

### LINKED ISSUE #1692: @tippyjs/react: Cannot use `className` prop in conjunction with `render` prop.
There is a warning in the console:
```
@tippyjs/react: Cannot use `className` prop in conjunction with `render` prop. Place the className on the element you are rendering. 
```

Exclude `className` from this object in order to fix it:

https://github.com/oceanprotocol/market/blob/4dca7d3442bbba51d98195150b78314967094234/src/components/%40shared/atoms/Tooltip/index.tsx#L63

____________

```
const { className, ...restProps } = props
const { content, children, trigger, disabled, placement } = props
```

and use `restProps` instead if `props`

kremalicious:
* fixes #1692

Was a little annoyance as this warning was showing up sometimes also in test runs. So just filter out `className` as suggested in #1692, change should not affect any behavior

### DIFF
--- a/src/components/@shared/atoms/Tooltip/index.tsx
+++ b/src/components/@shared/atoms/Tooltip/index.tsx
@@ -17,7 +17,8 @@ const DefaultTrigger = React.forwardRef((props, ref: any) => {
 })
 
 export default function Tooltip(props: TippyProps): ReactElement {
-  const { content, children, trigger, disabled, className, placement } = props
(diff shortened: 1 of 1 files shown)


EnzoVezzaro (approved):
👍🏽 LGTM. Didn't found any issue on the market.

PR #1779, opened 2022-11-08
NFT image & faulty metadata fixes

kremalicious:
* fixes #1778
* safeguard against faulty metadata.tags in `Asset` & `RelatedAssets` component
* properly handle image strings in tokenURI

Screenshot 2022-11-08 at 22 05 11

Click, click, test, test:

https://market-git-fix-1778-oceanprotocol.vercel.app/asset/did:op:4e95bb06d1342e027fcf3dea2a3030034b66fff4fa77c64369aeb99621e6b2f3

PR #1785, opened 2022-11-11
fix loading in preview

### LINKED ISSUE #1781: Related Assets try to load on publish preview
Just don't show them on the preview screen as it does not make sense there:

Screenshot 2022-11-09 at 09 35 47

mihaisc:
- closes #1781

Completely hiding it would involve a bit more complex logic and i think it's relevant to display it there in the preview.
image


kremalicious (approved):
this was already fixed as part of #1779, as the list is not loading on live. But since adding the `finally` clause, this `setIsLoading()` indeed is redundant

PR #1786, opened 2022-11-15
Allowing publisher to edit payment collector

### LINKED ISSUE #1775: Allow publishers to edit paymentCollector
Sometimes, a publisher will decide that entire revenue for a service should be sent to a different address (not the publisher's one).

Let's allow this in UI as well (maybe in Edit Metadata).

It's already in ocean.js (https://github.com/oceanprotocol/ocean.js/blob/main/src/contracts/Datatoken.ts#L418), so this is mostly design

Caller must be:  NFT.owner() or have feeManager rights

To get the current collector account, use https://github.com/oceanprotocol/ocean.js/blob/main/src/contracts/Datatoken.ts#L458

jamiehewitt15:
Fixes #1775 

Changes proposed in this PR:

- Allowing publisher to add a new payment collector in the edit form
- Adding a test to keep the test coverage going up.


content/pages/editMetadata.json
```diff
@@ -66,6 +66,13 @@
         "type": "tags",
         "placeholder": "e.g. logistics",
         "required": false
+      },
+      {
+        "name": "paymentCollector",
+        "label": "Payment Collector Address",
+        "placeholder": "e.g. 0X123ABC...",
+        "help": "This address will receive the revenue from all sales. More info available in our [docs](TODO: Update docs + Add link here).",
```
jamiehewitt15, line 74: Docs update is in review: https://github.com/oceanprotocol/docs/pull/1153

mihaisc: merged, you can update

bogdanfazakas:
one question regarding the functionality of this issue, should we display somehow the payment collector in the asset page? 
if no how would the users know that the payment collector was correctly updated ?

jamiehewitt15:
> one question regarding the functionality of this issue, should we display somehow the payment collector in the asset page? if no how would the users know that the payment collector was correctly updated just by going back to edit page ?

Good point, currently the only place where it is displayed is in the edit page. Maybe we could show it on the asset detail page only when it is different to the owner to avoid cluttering up the UI in the 99% of times when it's the same

bogdanfazakas:
> Good point, currently the only place where it is displayed is in the edit page. Maybe we could show it on the asset detail page only when it is different to the owner to avoid cluttering up the UI in the 99% of times when it's the same

Think that would do the job , also the preview suggested above is pretty clean 👍

src/components/Asset/Edit/EditMetadata.tsx
```diff
@@ -36,10 +37,27 @@ export default function Edit({
...
+    async function getInitialPaymentCollector() {
+      let paymentCollector
+      try {
+        const datatoken = new Datatoken(web3)
+        paymentCollector = await datatoken.getPaymentCollector(
+          asset?.datatokens[0].address
+        )
+        setPaymentCollector(paymentCollector)
+      } catch (error) {
+        console.error(error)
```
bogdanfazakas, line 55: think we can use LoggerInstance here and add also a description from where the errors comes from

src/components/Asset/Edit/EditMetadata.tsx
```diff
@@ -36,10 +37,27 @@ export default function Edit({
...
 
+  useEffect(() => {
+    async function getInitialPaymentCollector() {
+      let paymentCollector
+      try {
+        const datatoken = new Datatoken(web3)
+        paymentCollector = await datatoken.getPaymentCollector(
+          asset?.datatokens[0].address
+        )
+        setPaymentCollector(paymentCollector)
```
bogdanfazakas, line 53:
```suggestion
        setPaymentCollector(await datatoken.getPaymentCollector(
          asset?.datatokens[0].address
        ))
```
can't we do this inline this way ?

mihaisc:
But then you have cases like 
image
Revenue sent to ... nobody? 😁  Maybe in these cases put the owner there , or just hide the field.

jamiehewitt15:
> But then you have cases like image Revenue sent to ... nobody? grin Maybe in these cases put the owner there , or just hide the field.

Yeah that looks like a bug... I'll fix it

src/components/Asset/Edit/EditMetadata.tsx
jamiehewitt15: Yep, done

bogdanfazakas (approved):

jamiehewitt15:
> I have a question, why don't we add this in publish?

We discussed this briefly in the planning meeting and the decision was to leave it out for now and then possibly in the future we could introduce an "Advanced settings" part of the publish form that would include this and some other features

EnzoVezzaro (approved):

PR #1790, opened 2022-11-16
add calica integration

mihaisc:
Docs : https://github.com/oceanprotocol/docs/issues/1154

Currently works on goerli . Will be on mumbai at a later date

Draft (not sure what is possible yet, in talks with the developer) : 
- validate if we can detect if the contract is a calica contract so we can display something nice in  - confirmed
image

we can using :` https://calica.xyz/api/contract?address={{CONTRACT_ADDRESS}}&chain={{CHAIN}}`  , where chain is "goerli", "homestead", "matic", and "maticmum"
- if no split is set "split revenue" redirects to `https://calica.xyz/create?oceanAddress={dt_addr}`   - confirmed 
- if split is set redirects to `https://calica.xyz/goerli/{calical_contract_addr}`   - confirmed
- in profile add a small link to `https://calica.xyz/my-contracts/{wallet_addr}`   - confirmed


kremalicious:
Cleaning up some things, long story short:

Screenshot 2022-12-06 at 10 37 56

jamiehewitt15:
Couple of suggestions:

- We should use an arrow next to the "split revenue" link to indicate that it is an external link (same as we do in the footer)
- I'd suggest that there needs to be a tooltip with a link to either our docs or the calica docs. Otherwise people are just sent to calica.xyz and asked to connect their wallet, without any explanation of what calica is. 
- After implementing the revenue split, the "split revenue" button is still shown. It would be much better if this changed to a button that recognised a revenue split is already in place. For example either: "edit revenue split" or "view revenue split"
- After implementing a revenue split the metadata shows the revenue is sent to the calica contract, which is alright, but it would be much better if it showed the addresses that are actually receiving the revenue.  

image

jamiehewitt15:
Not sure how easy or possible it is but there is also potential for some improvements on  the calica side.

On the "My Ocean Contracts" page it would be better to show the token symbol rather than the name. We don't use the datatoken name on our asset detail page (only the datatoken symbol) so it's much easier and quicker to check that you're using the right token if it shows the symbol

image

It's weird how the dashboard for the calica contract doesn't show either the datatoken name, symbol, or address... There doesn't seem to be any easy way of seeing which asset this is for. That might be fine for someone with one asset but it's going to end up being very confusing for someone who has lots. 

image

jamiehewitt15:
Would actually be better if clicking "Split Revenue" took the user directly to the page for creating the revenue sharing agreement. This is easy to do on our side as we just need to change the URL and we know the address of the datatoken 

image

mihaisc:
Put a pin in this for now, i'll do some more research to spec it more.

mihaisc:
I'll start working on the changes described in the first comment

mihaisc:
havin some issues because of cors on their api

mihaisc:
I think it would be nice if we could add one or both these links in the profile on the market
manage revenue contracts, view created contracts (you can already get to this page by clicking `My contracts` in calica):  https://calica.xyz/my-contracts/0x3e81AA994f774eE914D57946DDF2486EA7d42D65 
image
create a new revenue split for ocean assets ( you can't get here from calica website), it's a list of all your assets that don't have a split created : https://calica.xyz/ocean/contracts/0x3e81AA994f774eE914D57946DDF2486EA7d42D65
image

@kremalicious any ideas how we can sneak one of there in the profile (or both, not sure)

PR #1795, opened 2022-11-24
Changing state of assets

### LINKED ISSUE #1776: Allow publishers to retire their assets
Sometimes, a publisher will decide that he/she wants to retire their asset.
We should allow this in UI as well.

It's already implemented in the backend (https://github.com/oceanprotocol/ocean.js/blob/main/src/contracts/NFT.ts#L639)

This requires UI work and some thinking about warning messages (are you sure, etc..)

Only nft.owner can call this

jamiehewitt15:
Fixes #1776
Fixes: #1798

Changes proposed in this PR:

- It's currently setup so that the user can update the state to 4 out of the [6 potential states](https://docs.oceanprotocol.com/core-concepts/did-ddo#state). If it's confusing having too many options I can remove some of them.
- The nft owner can now view their own retired assets, previously it was showing an error message for both the owner and everyone else.
- I've also added the asset state to the asset detail page so that the publisher can easily see the state of the assets they've retired. This is only shown when the asset isn't "Active" to avoid unnecessary clutter in the metadata.
- Tests have been added to keep the overall test coverage going up
- Adding the edit payment collector field back into the edit form

**Note:** It can take a few mins before the new state is reflected on the asset detail page.

image

If the same asset is viewed by someone who is not the nft owner they will see this:

image

**Possible extension:** We could potentially add something on the asset teaser about the state of the asset to make it easier to see which ones are retired on the publishers profile. Alternatively we could have them hidden and only shown if the user selects a filter for "Retired" assets.


kremalicious:
what's the (actual, practical) difference between:

- "End-of-life"
- "Deprecated (by another asset)"
- "Revoked by publisher"
- "Asset unlisted"

We always come back to this but those states are all overlapping so super confusing. Aren't they all leading to asset being "unlisted"? But then "unlisted" is its own state too?

If they all lead to to the same practical consequences, then there should be only one state for that, like "Unlist asset" and remove all the other states. Or "Asset unlisted" should not exist as an option.

jamiehewitt15:
> We always come back to this but those states are all overlapping so super confusing. Aren't they all leading to asset being "unlisted"? But then "unlisted" is its own state too?

Yeah, there is no practical difference. All it does is change the message that's displayed to the user e.g. "this asset has been flagged as End-of-life" vs "this asset has been flagged as Deprecated (by another asset)". 

It terms of making things simpler it seems reasonable to just have two options "Active" and "Asset unlisted"

EnzoVezzaro:
> > We always come back to this but those states are all overlapping so super confusing. Aren't they all leading to asset being "unlisted"? But then "unlisted" is its own state too?
> 
> Yeah, there is no practical difference. All it does is change the message that's displayed to the user e.g. "this asset has been flagged as End-of-life" vs "this asset has been flagged as Deprecated (by another asset)".
> 
> It terms of making things simpler it seems reasonable to just have two options "Active" and "Asset unlisted"

As for the functionality, it works fine. But I have few observations. 

First, having the info "Asset state" only when the asset is unlisted It's kind of weird to me. Would be better to have that display always?

Screenshot 2022-11-28 at 11 37 03

Also I think that would be helpful to add a tooltip saying something about the asset being unlisted. Is this asset available for computation after being unlisted? Maybe we should say something about that too.

jamiehewitt15:
> @jamiehewitt15 can we have the #1798 's fix on a separate PR please. To advance that one while we review this PR 👍🏽

Yeah, good idea. I've put it [here](https://github.com/oceanprotocol/market/pull/1799)

EnzoVezzaro (approved):

bogdanfazakas:
on the PR's preview there seems to be an issue with the published assets on any profile you try to check
PR preview:
Screenshot 2023-01-05 at 17 16 15

main ocean market:
Screenshot 2023-01-05 at 17 16 49

jamiehewitt15:
> on the PR's preview there seems to be an issue with the published assets on any profile you try to check

Hmm that's odd, I'll have a look into this

EnzoVezzaro:
Not sure why, but it doesn't seems to be working correctly. At first I though It was an issue with ES indexing slowly, but so far my assets haven't changed of state. 

On Polygon:

- https://market-htg0vyj0a-oceanprotocol.vercel.app/asset/did:op:e9831a65f2064b66a36b1f5e39353174975933819e081257e5a8fa0692c276c2
- https://market-htg0vyj0a-oceanprotocol.vercel.app/asset/did:op:138705d387c8f954b419bc5e477cd7349eeecedd93781ade3c0152677592a624


https://user-images.githubusercontent.com/13741335/212863748-f18b3154-ddba-450a-8664-0edfb0484b78.mp4


https://user-images.githubusercontent.com/13741335/212863753-db94511d-125e-4411-ab21-c7e87b7bf4bf.mp4



On Goerly it's even weirder 😅 it kind of works, but then it goes away (probably the content of the page gets overritten by some other call):


https://user-images.githubusercontent.com/13741335/212863912-9c2c790a-4bb1-4d7a-be12-8592edb91686.mp4

bogdanfazakas:
profile page looks good now, one more question though how much time should we wait for state to update  based on your previous experiences? on my end around 1h elapsed and the state has not updated yet

jamiehewitt15:
> profile page looks good now, one more question though in general how much time should it take for state to update based on your previous experiences? on my end around 1h elapsed and the state has not updated yet

Can you check again? It might have been an issue with the network or Aquarius. I just did a couple of tests on both mumbai and Goerli and they both updated instantly

EnzoVezzaro:
> > profile page looks good now, one more question though in general how much time should it take for state to update based on your previous experiences? on my end around 1h elapsed and the state has not updated yet
> 
> Can you check again? It might have been an issue with the network or Aquarius. I just did a couple of tests on both mumbai and Goerli and they both updated instantly

I retested too, On my assets it works fine for me on both mumbai and goerli 👍🏽

jamiehewitt15:
@bogdanfazakas Could you try testing this again?

bogdanfazakas:
> @bogdanfazakas Could you try testing this again?

Hi I've retested on friday but on my end i was still experiencing issues mentioned, will try again and update this comment in few min.

jamiehewitt15:
> > @bogdanfazakas Could you try testing this again?
> 
> Hi I've retested on friday but on my end i was still experiencing issues mentioned, will try again and update this comment in few min.
> 
> LE: retested , so I end up in this situation with the follwing steps
> 
> 1. go to edit page
> 2. unlist asset
> 3. on succes go to asset detail page from the succes alert
> 4. see asset updated
> 5. hit refresh
> 6. asset went back to initial state

Yeah, something weird is going on here, I'll investigate. Thanks for testing it again.

EnzoVezzaro:
> Yeah, something weird is going on here, I'll investigate. Thanks for testing it again.

I know if helps, but it's happening to me too now on the same assets I tested last time.

jamiehewitt15:
@bogdanfazakas @EnzoVezzaro I've fixed the bug where the state would revert back to it's original state. Can you have another look at this? Thanks a lot

bogdanfazakas:
> @bogdanfazakas @EnzoVezzaro I've fixed the bug where the state would revert back to it's original state. Can you have another look at this? Thanks a lot

Retested, the update actually works as expected now.

Not sure if we need to improve something for the 'revoked by publisher' state, the market throws errors in the browser if you try to acces assets using the did (see print screen), i know we can end up on this asset page just after updating the asset and hit 'back to asset' button or if you have the did saved since the assets with this state could not be discovered on the market and under the profile page. Maybe just promoting something like asset revoked by publisher error similar with the unlisted warning message.

Screenshot 2023-01-30 at 10 22 10

kremalicious:
> Not sure if we need to improve something for the 'revoked by publisher' state, the market throws errors in the browser if you try to acces assets using the did (see print screen)

yes, at no point in time user should see "application error". This needs an error screen telling user what's going on, and how to get out of it.

jamiehewitt15:
Yeah good spot @bogdanfazakas, I have changed that now so that you get this message:
 
image

bogdanfazakas (approved):
LGTM

jamiehewitt15:
@kremalicious @mihaisc any further feedback on this?

mihaisc (approved):

PR #1816, opened 2022-12-05
Avoiding getInitialPaymentCollector failure

### LINKED ISSUE #1813: getInitialPaymentCollector failure
Go to any asset and look into the browser console. All of them show:

> `[MetaFull: getInitialPaymentCollector] TypeError: undefined is not an object (evaluating 'new this.web3.eth')`

This was introduced in https://github.com/oceanprotocol/market/pull/1786, which assumes a web3 instance is always available AND which fires before any web3 instance is available even with wallet connected, leading to error.

jamiehewitt15:
Fixes #1813 

Changes proposed in this PR:

- Early return if there is no `web3` or `ddo` in `getInitialPaymentCollector` function
- Adding additional tests


EnzoVezzaro (dismissed):
LGTM! 👍🏽 couldn't replicate the issue after Fix

src/components/Asset/AssetContent/MetaFull.tsx
```diff
@@ -7,13 +7,14 @@ import { useWeb3 } from '@context/Web3'
 import { Asset, Datatoken, LoggerInstance } from '@oceanprotocol/lib'
 
 export default function MetaFull({ ddo }: { ddo: Asset }): ReactElement {
-  const [paymentCollector, setPaymentCollector] = useState<string>()
+  const [paymentCollector, setPaymentCollector] = React.useState<string>()
```
kremalicious, line 10: Our pattern is to always import React's methods directly, so in this case `useState` and not `React.useState`

src/components/Asset/AssetContent/MetaFull.tsx
```diff
@@ -7,13 +7,14 @@ import { useWeb3 } from '@context/Web3'
...
   const { isInPurgatory } = useAsset()
   const { web3 } = useWeb3()
 
   useEffect(() => {
     async function getInitialPaymentCollector() {
       try {
+        if (!web3 || !ddo) return
```
kremalicious, line 17: so without wallet I can't see the payment collector?

kremalicious:
Browser console with this PR, when I have a wallet connected:

Screenshot 2022-12-06 at 08 50 19

.jest/__fixtures__/algorithmAquarius.ts
```diff
@@ -0,0 +1,91 @@
+import { Asset } from '@oceanprotocol/lib'
+
+export const algorithmAquarius: Asset = {
```
kremalicious, line 3: if you introduce `algorithmAquarius` then the other ones can't be called `assetAquarius` anymore, as "asset" is the collection term for both, datasets & algorithms. So with a newly named `algorithmAquarius`, the other ones need to be renamed to `datasetAquarius`

src/components/Asset/AssetContent/MetaFull.tsx
jamiehewitt15: I used that to make the tests work, as `React.useState = jest.fn().mockReturnValue()` works but `useState = jest.fn().mockReturnValue()` doesn't. Do you have some advice on how to make the mock work without `React.useState`?

kremalicious: rather trigger whatever leads to updating that state instead of trying to mock it directly. So in this case you should not mock the state but the method which runs before state is updated, so `datatoken.getPaymentCollector()`

src/components/Asset/AssetContent/MetaFull.tsx
jamiehewitt15: Good point, I've removed this

EnzoVezzaro:
this PR is having the refresh issue (if you refresh the page, it breaks). You need to merge main, be aware that after merging the latest main on my local PR, I've been having issues when [running the app](https://github.com/oceanprotocol/market/issues/1823).

Screenshot 2022-12-07 at 10 29 27

jamiehewitt15:
> this PR is having the refresh issue (if you refresh the page, it breaks). 

hmm this is odd, it's actually already up to date with main. Not sure why that is still happening

EnzoVezzaro:
> > this PR is having the refresh issue (if you refresh the page, it breaks).
> 
> hmm this is odd, it's actually already up to date with main. Not sure why that is still happening

can it be an issue on the deployment side?

jamiehewitt15:
> can it be an issue on the deployment side?

Looks like I was wrong, it wasn't quite up to date. The reloading issue is fixed now

src/components/Asset/AssetContent/MetaFull.tsx
```diff
@@ -3,17 +3,20 @@ import MetaItem from './MetaItem'
...
   const [paymentCollector, setPaymentCollector] = useState<string>()
   const { isInPurgatory } = useAsset()
-  const { web3 } = useWeb3()
+  // const { web3 } = useWeb3()
 
   useEffect(() => {
     async function getInitialPaymentCollector() {
       try {
+        if (!ddo) return
+        const web3 = await getDummyWeb3(ddo.chainId)
```
kremalicious, line 19: this means we are always using the fallback web3, even if user has a wallet connected?

jamiehewitt15: Is there any disadvantage to using `getDummyWeb3` all the time to request the payment collector address?

jamiehewitt15: Not sure from you initial message if using `getDummyWeb3` causes any problems, seemed simple to use that all the time give this is just reading data from chain. Were you suggesting it needs to be changed?

EnzoVezzaro: I'm not sure if there's any disadvantages to always use ´getDummyWeb3´, but we usually use [this](https://github.com/oceanprotocol/market/blob/8d1782a800d03a194268671b03c39c21446f37ac/src/@utils/fixedRateExchange.ts#L23). Basically we check if there's a web3 connection available; if not, we use ´getDummyWeb3´.

EnzoVezzaro (approved):
LGTM!

bogdanfazakas (approved):

PR #1820, opened 2022-12-05
Fix #1809 - C2D text  update: proposal

### LINKED ISSUE #1809: C2D Warning may lead users to not using feature
User State: On mainnet. No wallet connected.

### Problem:
While working towards publishing an algo, the current messaging may be leading users to churn away from using system, and not publishing at all.

We want people to think they are able to publish algos.

### Candidate solution:
Change messaging to facilitate conversion while helping the user understand the risks associated with the sytem.

“Upload an algorithm using this feature. Please note that C2D is still in alpha phase.” 

### DoD:
- [ ] Ticket has been expanded to include any adjacent feature that can benefit from similar messaging improvement.
- [ ] Warning message/onboarding for algos, C2D, or any adjacent feature has been updated.

andreaarmanni:
As suggested by Veronica and Roberto, proposing to update the C2D alert in the publish flow to have a softer connotation. Requesting review/opinions to the OM team

Fixes #1809.

### DIFF
--- a/content/site.json
+++ b/content/site.json
@@ -16,6 +16,6 @@
   ],
   "announcement": "[Lock your OCEAN](https://df.oceandao.org/) to get veOCEAN, earn rewards and curate data.",
   "warning": {
-    "ctd": "Upload an algorithm using this feature. Please note that Compute-to-Data is still in alpha phase."
+    "ctd": "Please note that Compute-to-Data is still in alpha phase."
   }
 }
--- a/content/site.json
+++ b/content/site.json
@@ -16,6 +16,6 @@
   ],
   "announcement": "[Lock your OCEAN](https://df.oceandao.org/) to get veOCEAN, earn rewards and curate data.",
   "warning": {
-    "ctd": "Compute-to-Data is still in a testing phase, please use it only on test networks."
+    "ctd": "Upload an algorithm using this feature. Please note that Compute-to-Data is still in alpha phase."
   }
 }


kremalicious:
I don't see how replacing one warning with another one leads to more usage. Code clearly says it's a warning, which will also be shown on the asset details page. So just changing the wording of that makes even less sense.

If the features is ready to be used, there should be no warning at all so what exactly is preventing us from removing the warning?

graceful-coder:
Fixes: https://github.com/oceanprotocol/market/issues/1809

jamiehewitt15:
> If the feature is ready to be used, there should be no warning at all so what exactly is preventing us from removing the warning?

Yeah would be good to think about what needs to be done (if anything) for us to remove this warning. That's likely to have more of an impact that changing the wording.

graceful-coder:
@alexcos20 @jamiehewitt15 

Just got clarification that the READMEs for C2D x Predict-ETH will be optimized by Mid Jan, NOT C2D itself - sorry for the confusion. However, the proposed wording accurately depicts the state of C2D including the term "alpha" which indicates that it is out of testing. "Beta" also works.

jamiehewitt15:
I don't think the suggested message makes sense in the context where it's displayed. It's saying "Upload an algorithm using this feature." on the asset detail page, which isn't where users upload algorithms (that's on the publish page). If you want to remove the note about "still in testing" let's just remove that, no need to say  "Upload an algorithm using this feature." in the wrong place.

image

content/site.json
```diff
@@ -16,6 +16,6 @@
   ],
   "announcement": "[Lock your OCEAN](https://df.oceandao.org/) to get veOCEAN, earn rewards and curate data.",
   "warning": {
-    "ctd": "Compute-to-Data is still in a testing phase, please use it only on test networks."
+    "ctd": "Upload an algorithm using this feature. Please note that Compute-to-Data is still in alpha phase."
```
jamiehewitt15, line 19:
```suggestion
    "ctd": "Please note that Compute-to-Data is still in alpha phase."
```

graceful-coder:
@jamiehewitt15 I like your proposed change, and I also like the idea of removing the message completely. I am leaning toward the 2nd option because C2D has been in prod since 2019. Your thoughts?

content/site.json
graceful-coder: @jamiehewitt15 I like your proposed change, and I also like the idea of removing the message completely. I am leaning toward the 2nd option because C2D has been in prod since 2019. Your thoughts?

jamiehewitt15 (approved):

PR #1825, opened 2022-12-08
Multichain Provider

### LINKED ISSUE #1787: Multichain Provider
Provider issue: https://github.com/oceanprotocol/provider/issues/572
To support the multichain provider we need to remove a few filters and checks in the compute section. 
Do not limit assets to be on the same chain

jamiehewitt15:
Fixes #1787 

Changes proposed in this PR:

- Removing restriction of trusted algorithms needing to be on the same  chain
- 
-

### DIFF
--- a/src/@types/aquarius/BaseQueryParams.d.ts
+++ b/src/@types/aquarius/BaseQueryParams.d.ts
@@ -4,7 +4,7 @@ interface EsPaginationOptions {
 }
 
 interface BaseQueryParams {
-  chainIds: number[]
+  chainIds?: number[]
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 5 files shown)


jamiehewitt15:
So we need to think carefully through how this is going to work in the UI. The current implementation leads to a fairly poor user experience with a lot of metamask popups, we also can't tell if the user has enough tokens for the transactions to succeed as they will be on different networks. 

We also need progress on https://github.com/oceanprotocol/provider/issues/572 before this will actually work.

bogdanfazakas:
think we can close this PR since it is related to cross chain compute to data ,  which we are not going to do at the current stage.
We'll reopen it once we'll be working on integrating cross chain compute to data into the ocean-market

PR #1829, opened 2022-12-12
Fix compute basetoken issue

bogdanfazakas:
Fixes # .

Changes proposed in this PR:

- fix issue with base token on polygon main
- get rid of hardcoded provider fees token symbol and dynamically fetch the symbol 
- fix free assets symbol on polygon main assets

Note: builds are failing because of missing methods from Datatoken class in ocean.js that will be introduced by the following PR https://github.com/oceanprotocol/ocean.js/pull/1667


bogdanfazakas:
> What issue is this fixing?

main things we were hardcoding provider fees symbol as ocean token , this led to issues for polygon main, but now fetching them based on the token address received with the fees

jamiehewitt15 (approved):
Looks good to me

PR #1830, opened 2022-12-12
Fix filters

mihaisc:
Empty arrays `[]` were added in the filters at some point. These will break when we update aquarius to the new version with es 8.5

This needs to be merged before aqua update on live @alexcos20


jamiehewitt15 (approved):
Looks good to me

kremalicious (approved):

PR #1836, opened 2022-12-22
fix asset route

mihaisc:
Checks if route has a valid did


bogdanfazakas (approved):

jamiehewitt15 (approved):
This looks good. We should also add this into our tests at some point

EnzoVezzaro (approved):

PR #1845, opened 2023-01-04
share profile button

### LINKED ISSUE #1713: Enhance the “Sharing Profile” functionality
Currently someone’s own profile URL looks like this: `https://market.oceanprotocol.com/profile`

I propose to either: 

- [ ] Edit the URL to include the wallet address (or ENS, when present) at the end or;
- [ ] Add a “Share Profile” button in the Profile page as a visible CTA like in the following example: 
Screenshot 2022-09-29 at 14 08 46

Objective: Increase profile sharing amongst data publishers to drive more traffic to OM -> leading to better profile custom/optimisation -> more data on OM

jamiehewitt15:
Fixes #1713 

Changes proposed in this PR:

- Adding button on the profile so that users can easily share the link to the profile page with others

image


EnzoVezzaro:
Found a couple of issues:

1) when you hover over the share button, it moves a bit. I tried to inspect the elem to find the css property causing this, but I didn't have luck with it. It should be fairly easy to fix.

2) the actual link profile is broken on profiles with ENS accounts

On this [profile](https://market-li75x4hne-oceanprotocol.vercel.app/profile/autobotocean.eth)

Share link copied: https://market-li75x4hne-oceanprotocol.vercel.app/profile/autobotocean.eth/0xEB18BaD7365a40E36A41fB8734eB0B855d13b74F

But the actual account link shoud be (in this case):

oceanprotocol.vercel.app/profile/autobotocean.eth

https://user-images.githubusercontent.com/13741335/210804114-597293b2-76d9-46bb-8d98-d19ab2ed4110.mp4

PS: we found another bug here, don't we have a fallback page (404) in case we're on a non existing route?

jamiehewitt15:
> 1. when you hover over the share button, it moves a bit. I tried to inspect the elem to find the css property causing this, but I didn't have luck with it. It should be fairly easy to fix.

Not sure if this is an issue or not as all of our buttons have that movement. If you hover over the button at the bottom of the homepage that says "All datasets and Algorithms" you'll see the same movement. 

> 2. the actual link profile is broken on profiles with ENS accounts

Yeah good spot, they all should be working now.

> PS: we found another bug here, don't we have a fallback page (404) in case we're on a non existing route?

Yeah, not sure why that issue has come back again. Do you have any idea what's caused that?

EnzoVezzaro:
> Yeah, not sure why that issue has come back again. Do you have any idea what's caused that?

it seems related to the scheme color. The dark mode is not working on that page. Opened an issue for it https://github.com/oceanprotocol/market/issues/1847

jamiehewitt15:
> it seems related to the scheme color. The dark mode is not working on that page. Opened an issue for it #1847

Ok fair enough, that can be handled in a separate PR then

mihaisc:
Doesn't look good on mobile,  not working on your profile when no wallet connected, not implementing web share api.  Parking this for another time when we have time for it.

PR #1849, opened 2023-01-09
Check provider fees validity 

bogdanfazakas:
Fixes # .

Changes proposed in this PR:

- gets from subgraph existing provider fees and checks if still available before paying again


mihaisc:
is this in work?

mihaisc (approved):

PR #1852, opened 2023-01-12
Issue #1513 provider available

### LINKED ISSUE #1513: Check provider availability when publishing
During the publish process if a provider is down we still show it as "available" 
image
We should do a check. When the user tries to validate the data url it will fail and  will not understand why.

jamiehewitt15:
Fixes #1513

Changes proposed in this PR:

- 
- 
-


src/components/@shared/FormInput/InputElement/FilesInput/index.tsx
```diff
@@ -32,6 +32,13 @@ export default function FilesInput(props: InputProps): ReactElement {
           'Google Drive is not a supported hosting service. Please use an alternative.'
         )
       }
+      // Check if provider is a valid provider
+      const isValid = await ProviderInstance.isValidProvider(providerUrl)
+
+      if (!isValid)
+        throw Error(
+          '✗ Provider cannot be reached, please check status.oceanprotocol.com and try again later.'
```
bogdanfazakas, line 40: i think here we can add a reference to custom or self hosted provider as well, maybe adding also something like this at the end of the error message: `if you have your own hosted provider please check the status of the service`

mihaisc:
Please fix ci

src/components/@shared/FormInput/InputElement/FilesInput/index.tsx
jamiehewitt15: I'm not sure if that's necessary, as they should be stoped at the previous step if they have their own provider that doesn't work. If they enter there own provider then they have to click validate on the publish form, which shows them a separate error message:

image

kremalicious:
please fix CI. And why are there so many changes in this PR touching so many files? Like, almost all changes in here regarding test setups is also in #1774 making this all a huge hassle to review and later merge/rebase.

Please finalize PRs before opening up PRs with the same changes all over again. Or split out the shared changes into a new PR

jamiehewitt15:
> And why are there so many changes in this PR touching so many files? 

Those other changes were from on the release candidate branch. I was keeping this branch up to date with it as this was originally merging into that branch.

Easiest thing to do here is just to close this and open a new branch.

PR #1853, opened 2023-01-12
Checking if the Provider can be reached and throwwing an error

jamiehewitt15:
Fixes # .

Changes proposed in this PR:

- 
- 
-


mihaisc:
Seems to be a lot of code in this, what is goal ?

PR #1870, opened 2023-01-26
Encrypted Direct Messaging and Asset Commenting Modules

marcoelissa:
Changes proposed in this PR:

- Add encrypted direct messaging module using Orbis SDK
- Add commenting module on asset page using Orbis SDK (currently hidden/commented)


AnaLoznianu:
Hello, thanks for opening this PR. I tested it a bit, and the messaging part works like a charm on my end. 
I did though find a few things:

1. Can we have the send messages functionality triggered by the two operations?
- Via the send message button - this is already working
- Via the "Enter" keyboard button - can we add it back?

2. "Contact publisher" button from the asset page position.
The button is positioned in a place that is rendered dynamically based on the asset's metadata. Sometimes this positioning is good, other times the user needs to scroll a bit to find it. As this button's functionality is tightly related to the owner and not the asset itself, we should place it in the header or close to the header.  I am tagging my team members here as we should take care of this one. cc @matthias @bogdan

3. The "Contact publisher" button is available on the owner's published assets. We should verify who the owner of the asset is, and if it's the same as the connected wallet, hide the "Contact publisher" button(Like we do with the "Send Direct Message" button). Tapping on the button with the same account used to publish the asset, opens up a previous conversation.

4. For consistency, I suggest using the same label for the direct messaging functionality. Instead of "Contact publisher" and "Send Direct Message" labels, we can use something like "Send message".

src/@hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,42 @@
...
+      return item ? JSON.parse(item) : initialValue
+    } catch (error) {
+      // If error also return initialValue
+      console.log(error)
+      return initialValue
+    }
+  })
+  // Return a wrapped version of useState's setter function that ...
+  // ... persists the new value to localStorage.
+  const setValue = (value: T | ((val: T) => T)) => {
```
bogdanfazakas, line 23: not sure if we can reuse part of the work or move some into the  UserPreferencesProvider we have under @context/UserPreference.tsx

src/@types/Orbis.d.ts
```diff
@@ -0,0 +1,623 @@
+declare module '@orbisclub/orbis-sdk'
+
+declare interface IOrbisConstructor {
```
bogdanfazakas, line 3: as a general thing we tend to split type declaration files that apply to one component only into smaller files so would be easier to track and follow. as for eg we could have splitted this into _types.tsx files for the Orbis context types  and grouping them inside folders, and one new file under _types.tsx under the DirectMessages component. 
See as reference the `@context/Prices/` or `@context/MarketMetadata`

src/@utils/orbis.tsx
```diff
@@ -0,0 +1,96 @@
+import React, { ReactNode } from 'react'
+import Link from 'next/link'
+import reactStringReplace from 'react-string-replace'
+
+/** Regex patterns to use */
+const patternMentions = /\B@[a-z0-9_.⍙-]+/gi
+
+export function didToAddress(did: string) {
```
bogdanfazakas, line 8: we can apply the suggestion above with types to utils methods as well, and have more `_utils.tsx` where needed if the utils from the file can't be used in more than one component.

see '@context/Prices' for eg

bogdanfazakas:
seen there is an empty file in the PR as well `content/repo-metadata.json` that can be removed

src/@hooks/useLocalStorage.ts
marcoelissa: I don't quite understand what you mean here? Should I move the hook inside UserPreferencesProvider? But then again the stored data has nothing to do with user preferences.

bogdanfazakas: Yep, makes sense, I think this can stay as it is

marcoelissa:
> seen there is an empty file in the PR as well `content/repo-metadata.json` that can be removed

I have removed this file in my last commit

bogdanfazakas:
please rebase and fix the merge conflicts, also there are many failures in the CI ( build + tests) that need some attention 😃

mihaisc:
I can't send a message. I get the following

`need to create new conversation` 
` Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'session')`

image

baptistegreve:
Hey @mihaisc! Sorry about that this is actually my fault, I pushed by mistake a version of the SDK with a bug this morning, it was live for only an hour but it must be the version of the SDK this version is currently using.  If you install v`0.4.40` of the `@orbisclub/orbis-sdk` the bug should be fixed

bogdanfazakas (approved):

impactcapital:
Cool one!

Respectfully,
Silvio


* If urgent, pls call.

-------------------------------------

Silvio F. Pupo | Logos Capital

MIA•DC•SEU•TOK•ACC•LAG•BOG

360 NW 27th Street, Miami, FL 33127

Cel: +1.786.355.4261 | Tel: +1.818.564.6708

e: ***@***.******@***.***> w: www.LogosCapital.io <http://www.logoscapitalgroup.com/>

-------------------------------

All U.S. securities transactions are handled via SEC registered Broker Dealer(s), Member FINRA, SIPC, MSRB| This message may contain information that is confidential or privileged. If you are not the intended recipient, please advise the sender immediately and delete this message.

<https://aka.ms/o0ukef>
________________________________

PR #1899, opened 2023-03-21
add posthog

mihaisc:


jamiehewitt15:
This looks good, I think posthog is actually a lot better than plausible. Would be good to hear some feedback on it once it's merged, it might be worth us swapping the df site to use posthog

bogdanfazakas (approved):

PR #1912, opened 2023-04-21
Feature/wagmi

bogdanfazakas:
Fixes # .

Changes proposed in this PR:

- refactor wallet interactions and use wagmi
- integrate refactor version of oceanjs 3.0.0 using ethers


jamiehewitt15:
We could improve the styling of this box a bit as it blends into the background. Maybe add a border? 

image

jamiehewitt15:
Publishing and consuming worked fine for me, tested with a few assets (free and paid). Compute to data doesn't seem to be working though. 

Screenshot from 2023-05-10 12-42-20

Screenshot from 2023-05-10 12-42-50


image

jamiehewitt15:
> Compute to data doesn't seem to be working though.

Actually looks like it was a problem with the asset that was causing that issue

EnzoVezzaro:
As @jamiehewitt15, buying and consuming datasets work. For C2D I had mix-results. But the tx which failed, then worked if increased spending amount. In that regard I have a question, now are we gonna have this screen on every tx? how does it work? For me, it almost always shown when transacting. 

Screenshot 2023-05-10 at 08 01 31

For example: [here](https://market-cgpax5ye1-oceanprotocol.vercel.app/asset/did:op:1db94018e4f78cc59408310657a50a5039e75366eee97986843f4c949b98b50e)

First I tried with 10 ocean, it failed. Then I tried 20 ocean, and it passed.

EnzoVezzaro:
I guess is all fine, but I'm still think there's something off with the gas fees. 

Asset tested: https://market-frxu411ha-oceanprotocol.vercel.app/asset/did:op:1c53dff3c9031808f81c3330be0120f6a497388e726ef60cf014fa208868e461

When transacting with metamask I have to always use the aggressive mode instead of what suggested for "market" (or enter a value manually), otherwise the tx doesn't go through with a gas fee error. 

Screenshot 2023-05-11 at 11 08 15

Screenshot 2023-05-11 at 11 07 03

I also have to also enter an amount manually (sometimes, not always) in this new screen with a high number otherwise, again, you'd have the same gas fees issues as before. 

Screenshot 2023-05-11 at 11 08 53

This is not happening on production, I can transact just fine with what suggested for the site.

Screenshot 2023-05-11 at 11 16 08

Screenshot 2023-05-11 at 11 16 51

bogdanfazakas:
> I guess is all fine, but I'm still think there's something off with the gas fees.
> 
> Asset tested: https://market-frxu411ha-oceanprotocol.vercel.app/asset/did:op:1c53dff3c9031808f81c3330be0120f6a497388e726ef60cf014fa208868e461
> 
> When transacting with metamask I have to always use the aggressive mode instead of what suggested for "market" (or enter a value manually), otherwise the tx doesn't go through with a gas fee error.
> 
> I also have to enter an amount manually (sometimes, not always) in this new screen with a high number otherwise, again, you'd have the same gas fees issues as before.
> 
> This is not happening on production, I can transact just fine with what suggested for the site.
> 
> is there anything we can do about it?

This should be fine now, the site suggeste is back after the latest ocean js lib update (3.0.0-next.5)

bogdanfazakas:
> In that regard I have a question, now are we gonna have this screen on every tx? how does it work? For me, it almost always shown when transacting.
> 
> Screenshot 2023-05-10 at 08 01 31
> For example: [here](https://market-cgpax5ye1-oceanprotocol.vercel.app/asset/did:op:1db94018e4f78cc59408310657a50a5039e75366eee97986843f4c949b98b50e)
> 
> First I tried with 10 ocean, it failed. Then I tried 20 ocean, and it passed.
> 
> 
> can we improve this by asking just once, or give the user a message to guide them into the correct amount.

Unfortunately i was not able to find a work around for this yet, it looks it is been introduced for security reasons with latest metamask updates see:
_We designed this interface to give you more control and visibility over your token allowances, since you'll be prompted to always input a preferred limit, rather than default to the amount proposed by the dapp (though you can still select this default allowance if you choose)._ from this announcement [announcement](https://support.metamask.io/hc/en-us/articles/6055177143579-How-to-customize-token-approvals-with-a-spending-cap )

jamiehewitt15:
> > We could improve the styling of this box a bit as it blends into the background. Maybe add a border?
> 
> @jamiehewitt15 There are a limited options for customising the connect wallet modal, there is no border option unfortunately, I've added some shadows (see image below) but if you spark anything else we can use in the full list [here](https://docs.family.co/connectkit/theming) let me know smiley

That's fair enough. It looks better with the shadow, let's proceed with it looking like this.

mihaisc (approved):

PR #1992, opened 2023-11-02
remove provider url from check error method

### LINKED ISSUE #1991: broken custom provider validation
**Describe the bug**
Reported in #1989 

> I noticed this fix broke the custom provider validation (try to remove the default provider and add a new one and you'll see the app crash), you could probably just add !field.name.endsWith('.providerUrl') to the checkError function.

EnzoVezzaro:
- Fixes #1991 .

Changes proposed in this PR:

- fix provider url field

### DIFF
--- a/src/components/@shared/FormInput/index.tsx
+++ b/src/components/@shared/FormInput/index.tsx
@@ -72,11 +72,7 @@ export interface InputProps {
   disclaimerValues?: string[]
 }
 
-function checkError(
-  form: any,
-  parsedFieldName: string[],
-  field: FieldInputProps<any>
-) {
+function checkError(form: any, field: FieldInputProps<any>) {
   const touched = getObjectPropertyByPath(form?.touched, field?.name)
   const errors = getObjectPropertyByPath(form?.errors, field?.name)
 
@@ -85,7 +81,8 @@ function checkError(
     errors &&
     !field.name.endsWith('.files') &&
     !field.name.endsWith('.links') &&
-    !field.name.endsWith('consumerParameters')
+    !field.name.endsWith('consumerParameters') &&
+    !field.name.endsWith('.providerUrl')
   )
 }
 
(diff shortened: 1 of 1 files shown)


EnzoVezzaro:
@bogdanfazakas @mariacarmina can you please review this.

bogdanfazakas (changes requested):
this breaks the validation for Custom parameters if the users selects to publish and algo and use Custom parameters

EnzoVezzaro:
> this breaks the validation for Custom parameters if the users selects to publish and algo and use Custom parameters
> 
> LE: actualy not related to custom parameters, step 1 not validated at all for algorithms

Can you please provide the exact steps please. I haven't been able to reproduce it.

Screenshot 2023-11-08 at 17 26 41
Screenshot 2023-11-08 at 17 26 01


Validation Custom Parameters:

Screenshot 2023-11-08 at 17 26 10

bogdanfazakas:
sure, so on my end the issue the issue is that even if all fields are valid when trying to publish an algorithm the continue button is not active, so it should be something with the disable logic for this button

Screenshot 2023-11-09 at 11 24 35

bogdanfazakas:
when selecting dataset is all good with the disable logic and the validations

EnzoVezzaro:
> sure, so on my end the issue the issue is that even if all fields are valid when trying to publish an algorithm the continue button is not active, so it should be something with the disable logic for this button

Ok, got it. It looks like the custom params check is not optonal anymore. I'll take a look

EnzoVezzaro:
Found the issue, unfortunately is not related to this PR. I [created an issue](https://github.com/oceanprotocol/market/issues/1998) to keep track of it and fix it.

jamiehewitt15 (approved):
This looks good to me, it fixes the issue with the validation check on the custom provider URL. 

@bogdanfazakas can you also have another look at this? 

I have a separate PR for solving the issue with not being able to select the environment and publish an algorithm: https://github.com/oceanprotocol/market/pull/2004

bogdanfazakas (approved):
since we added a new pr for algorithm validation this looks good and we can move on and merge it

PR #1993, opened 2023-11-06
fix hidden field type on edit form load

### LINKED ISSUE #1949: File input not always populated in edit form
When using big files or api's behind the url used when creating the file, the file input doesn't get populated always which makes the whole validation form to fail and the submit button can't be used being disabled.
Eg in the image below, also in the debug the files type is no longer 'hidden' it defaults to 'ipfs' when this is reproduced.

Screenshot 2023-07-20 at 1 19 43 PM

EnzoVezzaro:
- Fixes #1949 .

Changes proposed in this PR:

- fix hidden field type on edit form when page is loading


jamiehewitt15:
I don't think I have ever experienced the hidden field type issue before, can you suggest some steps for replicating the issue and testing this?

EnzoVezzaro:
> I don't think I have ever experienced the hidden field type issue before, can you suggest some steps for replicating the issue and testing this?

Well, I don't think you'll be able to replicate it on this PR. But you can try what I commented [here](https://github.com/oceanprotocol/market/issues/1949#issuecomment-1786167931).

- Open a file you have in edit mode
- Open console and go on your nework tab
- Set "slow 3g"
- Refresh

You can try this on the main branch and compare the behaivior with this PR.

jamiehewitt15:
Something seems to have broken on this branch, when trying to open one of my assets the app crashes. I'm able to open the same asset on the main branch without any problems. 

image

jamiehewitt15:
> Something seems to have broken on this branch, when trying to open one of my assets the app crashes. I'm able to open the same asset on the main branch without any problems.

Looks like that issue has already been solved on main. It's not happening anymore after pulling the latest changes from main

jamiehewitt15 (approved):
This works well for me now

PR #2025, opened 2024-03-22
Added documentation 

andreaarmanni:
Added documentation only related to market deployment whilst kept documentation about components on main ocean docs

### DIFF
--- a/documentation/build-a-marketplace/README.md
+++ b/documentation/build-a-marketplace/README.md
@@ -1,0 +1,46 @@
+---
+title: Forking Ocean Market
+description: Forking and customizing Ocean Market (Frontend)
+---
+
+# Build a Marketplace
+
+## Outcome
+
+Your own fully functioning customized fork of Ocean Market is deployed and working.
+
+## Introduction
+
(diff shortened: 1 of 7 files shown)


andreaarmanni:
@mihaisc removed relative URLs and updated links

andreaarmanni:
> Can you also update the links on forking the market (in [this section](https://github.com/oceanprotocol/market?tab=readme-ov-file#-forking)) to link to these docs. The links currently go to gitbook, if the documentation is going to be in the repo then we should also update the links.

Thanks for the feedback Jamie, all links pointing to Gitbook are removed now

jamiehewitt15:
> all links pointing to Gitbook are removed now

Can you double-check that? There are no edits to the main readme in this PR. I'm talking about the links (in [this section](https://github.com/oceanprotocol/market?tab=readme-ov-file#-forking))

documentation/build-a-marketplace/README.md
```diff
@@ -19,9 +19,9 @@ Using Ocean Market is already a big improvement on the alternatives that are out
 
 The tutorial covers:
 
-* [Forking and running Ocean Market locally](forking-ocean-market.md)
-* [Customizing your fork of Ocean market](customising-your-market.md)
-* [Quick deployment of Ocean Market](deploying-market.md)
```
jamiehewitt15: These links to the other files are useful, I would suggest keeping them. Just double check they work

Issue #2027, opened 2024-05-08
Add GA4


mihaisc:
Add GA4 and proper cookie banner for gdpr

PR #2040, opened 2024-05-22
feat: add GA4

tom1145:
Fixes [#2027](https://github.com/oceanprotocol/market/issues/2027).

Changes proposed in this PR:

- add GA4 and enable cookie banner

image


jamiehewitt15 (commented):
You should also update the readme. This line is no longer true for example: 

> This feature is disabled per default since we do not use cookies requiring consent on our deployment of the market

mihaisc (approved):

Issue #2065, opened 2025-02-11
Fetch price schema from node instead of subgraph


mihaisc:

bogdanfazakas:
done

PR #2069, opened 2025-05-28
Feat/update v2

ahmedraza118:
Fixes # .

Changes proposed in this PR:

-  [Fetch price schema from node instead of subgraph](https://github.com/oceanprotocol/market/issues/2065) 
- 
- 
-


ahmedraza118:
> Based on my understanding, this PR seems to have been superseded by #2070 (Feat/update v3). Can you confirm if that’s the case, or if this PR is still relevant? or it can be closed ?

Yes, you are right, we can close this PR because feat/update v3 is updated and we don't need to merge the feat/update v2 anymore. we can only merge feat/update v3.

